PR #1, opened 2021-01-19
Stitches theme config

thomasdigby:
- Update theme with current values from main app
- Add breakpoints


src/stitches.ts
```diff
@@ -2,49 +2,99 @@ import { createStyled } from '@stitches/react'
 export * from '@stitches/react'
 
 export const tokens = {
+  colors: {
+    $tonal100: 'hsl(0, 0%, 15%)',
```
hollg, line 6: Do ths numbers in these names have a specific meaning?

thomasdigby: Hue, saturation, lightness. When we create our new theme I think `hsl` will be the right format to enable colour scales, we can tweak saturation and lightness easily, `hex` and `rgb` just don't work as well relative to other colours.

hollg: Haha sorry, no I understand what `hsl` is. I mean the `200` in `$tonal200` etc. Those numbers don't seem to relate to the `hsl` values in any consistent way.

thomasdigby: Ah sorry, yeah so it's a fairly standard naming convention for colour scales on the web, 100-900. https://eightshapes.com/articles/color-in-design-systems.html

hollg: Interesting read! I can't see anything about 100-900 in it though.

It says they should be named according to the lightness value, which would make total sense to me, but doesn't seem to be what's happening here. 
> We’ve all been there. A month into the system, the neutrals $color-gray-1, $color-gray-2, ... , $color-gray-7 — are stable. And then, in a stroke, you’ve got another tint to add stuck between -1 and -2. That numbering system stinks.

> Takeaway: Scale color names between 0 and 100 based on HSL’s lightness, such as $color-gray-05 and $color-gray-92. The scale reflects a familiar range from dark to light, allows for injecting new options between, and heck if I won’t remember $color-gray-93 until we retire it later.

I think those arguments are pretty compelling. My initial reasons for commenting were that `100`, `200` etc seem easy to forget because they're not linked to any other property of those colours and that because the scale doesn't match the scale of the colours themselves, adding to the palette over time could be tricky.

How do you feel about changing the names to match the lightness values?

thomasdigby: Can you give an example? e.g. in this case `$tonal15`?

thomasdigby: Sorry I thought he referenced the Material UI color palette https://www.materialui.co/colors

thomasdigby: Also, are you referencing this naming convention specific to `tonal` values, or across all color scales?

thomasdigby: Just as a final comment, I think the important thing that we need to confirm is the specification. I think a scale of `tonal100` to `tonal900` with 7 other values from light to dark (or `tonal10`, `tonal90`) should be enough to cover the specific colours we need in the UI.

I think naming based on lightness e.g. `tonal23` has the same issue to me as naming a color `dark-blue` when it's the main brand colour. We want to represent what the value means to the developer consuming it, `tonal10` implies lightness without having to define exactly what the lightness value is. Types will help when we want to autocomplete the tonal values available but I still think there's value in simplifying the interface to the colour scale, rather than representing it accurately.

>adding to the palette over time could be tricky.

On this point, the aim is to not add to it over time as we should have our scale in place. We may not use all the values but we have a system in place with which to use them if we need. We can always add extra color scales if we need breadth of colour instead of lightness variations

andoulla: I have used this convention in the past and it does make sense to me. It doesn't really need to reflex a value but it should be easy to understand the range of a colour ie: gray-100 is lighter from gray-110 but it will be very close to the first one.  does that make any sense?

src/stitches.ts
```diff
@@ -2,49 +2,99 @@ import { createStyled } from '@stitches/react'
...
+    $successDark: 'hsl(137, 60%, 31.5%)',
+
+    $warning: 'hsl(24 ,100%, 55%)',
+    $warningDark: 'hsl(24 ,100%, 41%)',
+
+    $dangerLight: 'hsl(354.5, 88%, 67%)',
+    $danger: 'hsl(354.5, 70.5%, 53.5%)',
+    $dangerDark: 'hsl(354.5, 70.5%, 40%)',
+
+    $english: '#b7373f',
```
andoulla, line 45: Are these always the same?  they don't have any variations?

src/stitches.ts
```diff
@@ -2,49 +2,99 @@ import { createStyled } from '@stitches/react'
...
+  },
+  shadows: {
+    $1:
+      '0px 2px 1px -1px rgba(0, 0, 0, 0.2), 0px 1px 1px 0px rgba(0, 0, 0, 0.14), 0px 1px 3px 0px rgba(0, 0, 0, 0.12)',
+    $2:
+      '0px 7px 8px -4px rgba(0, 0, 0, 0.2), 0px 12px 17px 2px rgba(0, 0, 0, 0.14), 0px 5px 22px 4px rgba(0, 0, 0, 0.12)',
+    $3:
+      '0px 11px 15px -7px rgba(0, 0, 0, 0.2), 0px 24px 38px 3px rgba(0, 0, 0, 0.14), 0px 9px 46px 8px rgba(0, 0, 0, 0.12)'
   },
   radii: {
```
andoulla, line 87: I didn't realise the plural for radious is radii

src/stitches.ts
```diff
@@ -2,49 +2,99 @@ import { createStyled } from '@stitches/react'
...
 
 export const utils = {
   p: (value: any) => ({
-    paddingTop: value,
-    paddingBottom: value,
-    paddingLeft: value,
-    paddingRight: value
+    padding: value
   }),
   pt: (value: any) => ({
```
andoulla, line 99: do we need to define the rest of the options? (pb, pl,pf)

Also, is the value typed as any? wouldn't it be number|string?

andoulla (commented):
It was very easy to read and understand.

src/stitches.ts
hollg: They actually definitely do have at least 2 variations. As well as the ones listed here, there's a lighter version of each color which is used as the background for icons that represent the subject.

thomasdigby: Ah that's interesting, I wonder how we can represent these, I kind of want to move away from light/dark variants but it doesn't make sense to have a scale

src/stitches.ts
thomasdigby: We have pl pt pb pr, as well as px for x axis padding and py for padding on the y axis. Are there any others that you think we'll need?

thomasdigby: Good point on the types, I copied this from the modulz design system but I'll update them as per your suggestion!

andoulla: No, I think these are more than enough. If we have one for the whole thing and one per side and px py that should be flexible enough.

src/stitches.ts
andoulla: Could we stick to the same convention? so we can set this to be the middle value $english500 and then this will allow us to add the lighter and darker values as we discover them

src/stitches.ts
hollg: I think I still don't fully understand the logic for the `100` - `900` scale but am content to go with the majority vote. If it's worked for you both in the past then I'm sure it will start making sense to me as we use it!

src/stitches.ts
```diff
@@ -93,62 +93,63 @@ export const tokens = {
 }
 
 export const utils = {
-  p: (value: any) => ({
+  p: (value: number | string) => ({
```
hollg, line 96: How would you feel about defining a type for this union and using it throughout? e.g.
```tsx
type CssValue = number | string

// ...

p: (value: CssValue) => ({ padding: value })
```

I think it removes a bit of noise from the function declarations.

src/stitches.ts
thomasdigby: I might leave this for now and look to add it in a future release

hollg (approved):

PR #2, opened 2021-01-20
Feat: Example component documentation

hollg:
Adds Storybook documentation for the `Box` component, including:
- [X] Into explaining what the box is / what it's for
- [X] Example of a card composed from `Box`es
- [X] A basic story with [controls](https://storybook.js.org/docs/react/essentials/controls)
- [X] Proposed directory pattern for documentation consisting of multiple files (in this case, the `CardExample` component is defined in another file)

Note: in the future we'll probably want to use Storybook controls to update the component's props, but `Box` only has `as`, `ref` and `css` so I just picked some specific `css` properties. The `ArgsTable` component will also be more useful in components where we've defined prop types and default values, as this information will be included.


src/primitives/box/stories/Box.stories.mdx
```diff
@@ -0,0 +1,79 @@
...
+# Box
+
+The `Box` is our most fundamental container component. Many other components are composed of `Box`es. For example, this card:
+
+<Card />
+
+`Box` falls back to rendering a `div`, but we can use it to render any other container element with e.g. `<Box as="section">`. In the card, `Box` is used to render a `section`,
+`header`, `figure` and `footer`.
+
+Avoid rendering any of these elements directly in our components — `Box` has all the same flexibility, but with the
```
hollg, line 38: I'm not sure whether this is the actual rule we want to follow. Perhaps instead we want to describe how to make the decision between using a `Box` and using a `section`, `footer` etc?

thomasdigby: I think it's worth having a blanket rule to use `<Box />` at all times, we gain the ability to easily add css and it keeps the view layer consistent. It would introduce a modicum of verbosity though, although hopefully going heavy on the promoters and components should help us avoid this affecting every view.

src/primitives/box/stories/CardExample.tsx
```diff
@@ -0,0 +1,63 @@
...
+      margin: 'auto',
+      height: '300px',
+      width: '250px',
+      display: 'flex',
+      flexDirection: 'column',
+      border: '1px solid #ebebeb',
+      borderRadius: '8px',
+      textAlign: 'center',
+      fontFamily: 'sans-serif',
+      boxShadow: 'rgba(100, 100, 111, 0.2) 0px 7px 29px 0px',
```
thomasdigby, line 18: Where we can, even in examples, I think we should use the theme values, in this case "$2" should work ok?

src/primitives/box/stories/Box.stories.mdx
```diff
@@ -0,0 +1,79 @@
...
+    background: { control: 'color' },
+    textAlign: {
+      control: { type: 'select', options: ['left', 'right', 'center'] }
+    },
+    border: { control: 'color' },
+    borderRadius: { control: 'number' },
+    padding: { control: 'number' },
+    ref: { table: { disable: true } },
+    as: { table: { disable: true } },
+    css: { table: { disable: true } }
```
thomasdigby, line 25: Might be cool to have a few storybook utils that we use to ensure that css, ref and as don't appear in any of our docs

src/primitives/box/Box.tsx
```diff
@@ -1,4 +1,3 @@
-import * as React from 'react'
-import { styled } from '../../stitches'
+import { css, styled } from '../../stitches'
```
thomasdigby, line 1: Do we need to import css here?

src/primitives/box/stories/Box.stories.mdx
hollg: OK cool! Do you have any strong feelings about how we define the list of elements that should only be rendered via `Box`? We _could_ use it for _everything_, but `h1` certainly feels very different to `section`.

src/primitives/box/stories/CardExample.tsx
hollg: The theme doesn't include `boxShadow`. Not until your PR is merged, at least. (I'm v happy to wait until it is, and then update this PR accordingly — I definitely agree that we should use the theme values as much as possible.)

src/primitives/box/stories/Box.stories.mdx
hollg: Oh yeah, nice. I'll look into that!

src/primitives/box/Box.tsx
hollg: Nope. Good catch — thanks!

src/primitives/box/stories/Box.stories.mdx
andoulla: If we are using it for everything it might be easier to list the exceptions.

src/primitives/box/stories/CardExample.tsx
hollg: Apparently I'm completely wrong about this, sorry! I didn't cotton onto the `radii` token!!

src/primitives/box/stories/CardExample.tsx
```diff
@@ -0,0 +1,63 @@
...
+    }}
+  >
+    <Box
```
andoulla, line 21: So the recommendation is that we wrap headings with a box element? I'm asking as I'm not used to having a Box component

hollg: I don't think every heading necessarily needs wrapping in anything. In this case, I needed a container to apply styling to, and `header` is the appropriate element so I used `Box` to render a `header`. So the choice to wrap  the `h2` in a container isn't really related to the fact that we have a `Box`. 

I hope that makes sense!

src/primitives/box/stories/Box.stories.mdx
thomasdigby: Yeah I think this will be easier to define when we have better idea of how to handle typography. I actually wonder whether it'll be better to use elements when they're missing class/attribute... might need to just see how the system evolves and revisit this in a few weeks

src/primitives/box/stories/Box.stories.mdx
```diff
@@ -0,0 +1,79 @@
...
+          width: '250px',
+          margin: 'auto',
+          display: 'flex',
+          flexDirection: 'column',
+          justifyContent: 'center',
+          backgroundColor: background,
+          border: `1px solid ${border}`,
+          borderRadius,
+          textAlign,
+          fontFamily: 'sans-serif',
```
thomasdigby, line 66: I think it's best to remove any text related css in our examples, `textAlign`, `fontFamily` et al should be controlled by the eventual `Text` component. I think if we're a bit strict about that now, it'll help reinforce it for others when reading the docs.

hollg: Oh yeah, I didn't think about this at all but that seems v sensible to me. Worth bearing in mind then that when we add new components, we should also come back to existing docs to update examples to use them.

thomasdigby (approved):

PR #4, opened 2021-01-20
Automated versioning

andoulla:
Documenting the approach that we will take in versioning and which tools are going to be used


andoulla:
@thomasdigby @hollg

hollg:
This looks great to me. I think strictly enforcing rules for versioning, commit messages etc will save us a lot of hassle in the long run and it sounds like this setup will do that!

thomasdigby (approved):
Awesome, I think this will be really great, looking forward to trying it out when we have our first alpha release ready 👍

.gitignore
```diff
@@ -1,3 +1,7 @@
 node_modules
 dist
 .cache
+.eslintcache
+yarn-error.log
+storybook-static
+storybook-static
```
andoulla, line 7:
```suggestion

```

PR #5, opened 2021-01-20
Feat: Sandbox setup

hollg:
# Sandbox setup

Adds a basic wrapper to the sandbox with a message about not commiting changes into `main` and a couple of links that I imagine will be useful while developing components. This will look much nicer as we update it to use typography components etc!

Also adds a global CSS reset to the main `index.ts` to help our components to render consistently without having to run the reset in each one separately.

- [X] Add a basic wrapper to the Sandbox app
- [X] Fix global CSS reset

## Screenshots
image


andoulla (approved):

src/index.ts
```diff
@@ -1,3 +1,8 @@
+import { reset } from 'stitches-reset'
 export { Box, Flex, Select } from './primitives'
 
-export { styled, css } from './stitches'
+import { css, styled } from './stitches'
+
+css.global(reset)()
```
thomasdigby, line 6: I think we should include this just within the sandbox, we'll want to include this in our main app but not as a side-effect from the components.

sandbox/Sandbox.tsx
```diff
@@ -0,0 +1,65 @@
...
+      backgroundColor: 'tonal',
+      p: '$3',
+      borderRadius: '$1',
+      mb: '$3'
+    }}
+  >
+    <h1 className={css({ margin: '$2 0' })}>Component lib sandbox</h1>
+    <p className={css({ margin: '0 0 $2 0' })}>
+      Use this app to quickly render components in an environment where you have
+      more control than in Storybook. But don't commit changes to this app into
```
thomasdigby, line 30: I wonder if we could include this in the sandbox JS file rather than the actual view? I'm kinda easy either way but think I'd prefer to keep the sandbox view clean.

src/index.ts
hollg: Yeah, I went back and forth on this quite a lot tbh. I eventually landed on leaving it here because
- It will help our components render as consistently as possible across, hypothetically, any app 
- If I've understood correctly, this will only affect our components

Am I wrong about one of the above? Or are there just other considerations you're thinking of?

sandbox/Sandbox.tsx
hollg: Yeah that makes sense, I'll change that now

src/index.ts
thomasdigby: The reset will just apply globally so It definitely needs to be opt-in. Theoretically if this package was consumed by another team/app then they're also accepting that a reset will be applied, which may not be wanted/needed. It also introduces a side-effect which may remove the ability to tree-shake our components.

To help render our components consistently we should try to apply resets at the component level or just accept browser defaults for elements.

hollg: Gotcha, thanks for explaining!

PR #6, opened 2021-01-21
Initial unit test set-up work

thomasdigby:
- Set up Jest and related services for unit tests
- Included `jest-stitches` to include the `css` contents in the snapshot
- Run tests on precommit


package.json
```diff
@@ -10,6 +10,7 @@
     "dev": "microbundle watch  --jsx React.createElement",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
+    "test": "jest",
```
andoulla, line 13: should we add a test:watch shortcut too?

hollg (approved):
This looks good! `jest-stitches` is cool. I wonder whether in the future we could use it to back our theme with tests and prevent inadvertent changes to it by asserting on what e.g. what `margin-bottom` property `mb: '$3'` comes out as?

andoulla (approved):

package.json
thomasdigby: Yeah that's a good idea, feel free to add it to this branch as I'm focused on campaign work at the moment

thomasdigby:
> This looks good! `jest-stitches` is cool. I wonder whether in the future we could use it to back our theme with tests and prevent inadvertent changes to it by asserting on what e.g. what `margin-bottom` property `mb: '$3'` comes out as?

Yeah that's a good idea, snapshotting the theme to confirm when changes are made would be such a useful validation step

PR #7, opened 2021-01-21
Chore: fix tonal colors order

hollg:
Fixes the name of our `tonal` colors so that they get darker as the number in their names increase.


thomasdigby (approved):
!krow taerg

PR #9, opened 2021-01-22
Code coverage, layout components

thomasdigby:
- Added code coverage for all `/src` files
- Added additional layout components `Flex` and `Grid`
- Simplified `Box` docs examples to use tokens and prevent overlap with flex styles
- Removed legacy or unnecessary storybook config
- Added `bg` shorthand for `backgroundColor`


src/primitives/flex/stories/Flex.stories.mdx
```diff
@@ -0,0 +1,30 @@
+import { Canvas, Meta, Story, ArgsTable } from '@storybook/addon-docs/blocks'
+
+import { Flex } from '../'
+
+<Meta title="Primitives/Flex" component={Flex} />
+
+# Flex
+
+The `Flex` component builds on top of `Box` to apply flex styles to a container component.
```
hollg, line 9: 💬 This seems to suggest that `Flex` renders a `Box`, rather than rendering a `div` directly.

thomasdigby: Agreed, any suggestions to improve it?

src/primitives/flex/stories/Flex.stories.mdx
```diff
@@ -0,0 +1,30 @@
...
+
+<Canvas>
+  <Story name="Basic">
+    <Flex
+      as="aside"
+      css={{
+        backgroundColor: '$primary500',
+        color: 'white',
+        justifyContent: 'space-between',
```
hollg, line 18: 💬 Looking at this example, I'm not 100% sure I understand the reason for including the `Flex` component. If we're going to need to specify properties like `justifyContent`, `alignItems` and `flexDirection` anyway, why not just use a `Box` and add `display: 'flex'`?

Maybe it would be worth defining variants for `Flex`, so that we can effectively use shorthand for useful combinations of those properties?

src/primitives/flex/stories/Flex.stories.mdx
hollg: Well, is there any reason for it _not_ to render a `Box`? Or would that stop it being a primitive? 🤔

src/primitives/flex/stories/Flex.stories.mdx
hollg: Ooh and maybe one of those variants could implement [fukol-grids](https://github.com/Heydon/fukol-grids)!

thomasdigby: I think that's a... _great idea??_

src/primitives/flex/stories/Flex.stories.mdx
thomasdigby: In the Modulz design system, which uses stitches, they export `Box`, `Flex` and `Grid` as lowest-level layout/container components, each setting a display property relevant to their children, `block`, `flex` and `grid`.

In principle I think it's awesome, `<Flex><Box css={{m: 'auto'}}></Box></Flex>` feels like a good use of some default settings e.g. `display: flex` in `Flex` to ease the DX, and it's also (I think) clearer to the user how the elements should render.

However in practice I don't know if this directly translates, I think we'll just have to try it and see how it feels.

hollg: Ohh, okay, that totally makes sense. Thanks!

In that case, I reckon the docs should specify that `Flex` is _analogous_ to `Box`, rather than saying it "builds on" `Box`.

thomasdigby: Yeah bad choice of words fo sho, will update and maybe add `Grid` as part of this PR as well

hollg: Yeah that's a good idea, I think they each make more sense when they can be compared with each other!

src/primitives/flex/stories/Flex.stories.mdx
andoulla: could we catch up on this? I'm fully following the concern here.

Also, shouldn't something like  flex with the type grid be confusing to as of what is actually implementing, flex or grid

hollg: Yeah, I'm sure we could come up with a better/clearer name than that

andoulla: Layout?

thomasdigby:
@hollg Just FYI on the card example changes, I updated it so we didn't need to use `Flex` or `Grid` and utilised the shorthand methods and tokens where possible, hopefully this makes it easier to see a `Box` component in use that doesn't potentially overlap with the other layout components.

thomasdigby:
Also, this branch is a mess of stuff now so we could split it out into better PRs _or just merge it and forget about it 🤞 _

src/primitives/box/stories/Box.stories.mdx
```diff
@@ -41,29 +38,24 @@ addition of the powerful `css` prop and various styling utilities based on our t
   <Story
     name="Basic"
     args={{
-      alignItems: 'flex-start',
       background: 'white',
       border: '#ebebeb',
       borderRadius: 8,
```
andoulla, line 43: should these be in this format `borderRadius: '$3',`

src/stitches.ts
```diff
@@ -142,6 +141,10 @@ export const utils = {
   my: (value: CSSValue) => ({
     marginTop: value,
     marginBottom: value
+  }),
+
+  bg: (value: string) => ({
```
hollg, line 146: 💬 Can I suggest that `bg` should be shorthand for `background` instead? That way we can still just pass a color and get the same functionality, but it also allows us to pass gradient options etc too if we want to

src/primitives/box/stories/Box.stories.mdx
thomasdigby: I think in this instance these are actually arguments for the storybook controls rather than stitches tokens.

hollg (approved):

PR #10, opened 2021-01-22
Panel component

hollg:
Adds a `Panel` component, as seen wrapping the login form. `Panel` is in a new  `components/`directory, but I'm totally open to being convinced it belongs somewhere else.

I originally gave `Panel` three size variants, but we only need one size for the demo and I think it's probably best to wait to get design input on the variants rather than making them up for the sake of it.


sandbox/types.d.ts
```diff
@@ -0,0 +1 @@
+declare module '*.png'
```
andoulla, line 1: Newbie question, why do we need this?

hollg: That's just to stop a TypeScript error about the image we're importing not having any type definitions. It can be removed after the demo!

andoulla: I'm sure we will end up needing it for something else, I just didn't know the terminology to google it

PR #11, opened 2021-01-22
Add size limit validation

thomasdigby:
Running `yarn validate:size` will check the JS output of our design system and allows us also to debug the contents via `--why`.

Screenshot 2021-01-22 at 13 02 18

Additionally uncommenting `.editorconfig` (not sure why it was commented in the first place) and added `.eslintignore` as we were linting the `dist` and `coverage` directories


package.json
```diff
@@ -15,9 +15,10 @@
     "format": "prettier 'src/**/*.{js,ts,tsx}' --write",
     "lint": "eslint 'src/**/*.{js,ts,tsx}' --quiet --fix",
     "validate": "npm-run-all validate:**",
-    "validate:types": "tsc --noEmit",
-    "validate:format": "yarn format --check",
     "validate:dependencies": "yarn audit --groups dependencies",
+    "validate:format": "yarn format --check",
```
andoulla, line 19: Just out of curiosity shouldn't it be running the eslint task too as part of the validation?

andoulla (approved):

package.json
thomasdigby: Yes it should, `validate:format` will validate that the files are formatted correctly, but it's potentially not that valuable... useful on precommit but not publish which is what these scripts will run on. I'll change this now to check the linting.

hollg (approved):

PR #12, opened 2021-01-22
Setup accessibility testing

andoulla:
This branch:
- Adds documentation on accessibility testing approach
- Add jest-axe for unit testing
- Add storybook addon for accessibility validation


.eslintrc
```diff
@@ -30,20 +30,20 @@
...
-    "jsx-a11y/no-noninteractive-element-interactions": "warn",
-    "jsx-a11y/no-onchange": "off",
-    "jsx-a11y/no-static-element-interactions": "warn",
+    // "jsx-a11y/anchor-is-valid": "warn",
+    // "jsx-a11y/click-events-have-key-events": "warn",
+    // "jsx-a11y/mouse-events-have-key-events": "warn",
+    // "jsx-a11y/no-autofocus": "warn",
+    // "jsx-a11y/no-noninteractive-element-interactions": "warn",
+    // "jsx-a11y/no-onchange": "off",
+    // "jsx-a11y/no-static-element-interactions": "warn",
```
thomasdigby, line 46: Can we remove these rules altogether?

andoulla: I will be before merging, I just wanted to show you what I reenabled in the project. Sorry should have left a comment

src/primitives/grid/Grid.test.js
```diff
@@ -11,4 +13,10 @@ describe(`Grid component`, () => {
     await screen.findByText('GRID')
     expect(container).toMatchSnapshot()
   })
+
+  it('has no programmatically detectable a11y issues', async () => {
+    const html = ReactDOMServer.renderToString(<Grid />)
```
thomasdigby, line 18: Why do we need to use `ReactDOMServer` here and not `render`?

README.md
```diff
@@ -34,7 +34,54 @@ _Note:_ Once a branch gets merged into `main`, it is automatically published, th
...
+Different tools are used to validate the components as each serves a different purpose. The a11y storybook addon will validate all the stories and explain each violation, whereas the accessibility unit tests will fail the pipeline and prevent releasing inaccessible components. **However, these tests are only valid with the context they are testing on, so they can't guarantee 100% that a component is fully accessible.** Therefore, please add stories for all the variations a component has. Also, there is a need for manual testing and design validation.
+
+Before developing a component, please take some time to read the following sections and the reading provided.
+
+### Things to consider whilst developing
+
+- Using `React.Fragment` where possible to avoid adding extra <div>
+  > Sometimes we break HTML semantics when we add <div> elements to our JSX to make our React code work, especially when working with lists (<ol>, <ul> and <dl>) and the HTML <table>. In these cases we should rather use React Fragments to group together multiple elements. for examples please look at the [React documentation](https://reactjs.org/docs/accessibility.html)
```
hollg, line 47: 💬 _super_ tiny one: can we wrap names of HTML elements in backticks?

andoulla: of course

src/primitives/box/Box.test.js
```diff
@@ -11,4 +13,10 @@ describe(`Box component`, () => {
     await screen.findByText('BOX')
     expect(container).toMatchSnapshot()
   })
+
+  it('has no programmatically detectable a11y issues', async () => {
```
hollg, line 17: 💬 These tests are awesome. I was expecting them to be much more complicated!

andoulla: I expect they will get a bit trickier when we have something more substantial than a box. Also, I'm not sure if I could have been testing this with some context or not. These type of tests are very dependent on the context and maybe the content

jest.config.js
```diff
@@ -3,5 +3,6 @@ module.exports = {
   snapshotSerializers: ['jest-stitches'],
   collectCoverage: true,
   coveragePathIgnorePatterns: ['/src/stitches.ts'],
-  testPathIgnorePatterns: ['/node_modules/']
+  testPathIgnorePatterns: ['/node_modules/'],
+  setupFilesAfterEnv: ['./setupFiles.js']
```
thomasdigby, line 7: Can we just use `jest-axe/extend-expect` here? I think that should remove the need for an additional set up file

src/primitives/grid/Grid.test.js
andoulla: In the documentation it claims we should be able to just do `const { container } = render(<Box css={{ color: 'red' }}>laskjdl</Box>)` and then use the container in the axe function. 


However, I can't get it to work correctly, as everything passes even when I know it should throw an error

andoulla: Need to add that I have this problem on this type of components that are containers. The above solution works fine for `` `<button>` etc

jest.config.js
andoulla: I would but I can see in the other PR that @hollg has set up the same file, should we keep it since we are adding more than one files to it?

src/primitives/grid/Grid.test.js
thomasdigby: I'm seeing tests pass for both of these examples, I assumed these would be picked up by `jest-axe` as they're getting picked up by the storybook a11y integration

```js
  it('has no programmatically detectable a11y issues', async () => {
    const { container } = render(
      <Flex css={{ bg: 'red', color: 'red' }} />
    )
    const results1 = await axe(container)
    expect(results1).toHaveNoViolations()

    const html = ReactDOMServer.renderToString(
      <Flex css={{ bg: 'red', color: 'red' }} />
    )
    const results2 = await axe(html)
    expect(results2).toHaveNoViolations()
  })
```

jest.config.js
thomasdigby: Yep that's fine then, just update this file to use that naming convention 👍

src/primitives/grid/Grid.test.js
andoulla: Does above test pass because there is no text as content?

andoulla: Rewriting it like this works without the ReactDOMServer
```it('has no programmatically detectable a11y issues', async () => {
    render(<Box />, document.body)

    const results = await axe(document.body)
    expect(results).toHaveNoViolations()
  })```

Comment (thomasdigby) [reply src/primitives/grid/Grid.test.js:18]: Yeah both of these work, I was just testing the wrong thing. Let's run with `render` if we can

```jsx
  it('has no programmatically detectable a11y issues', async () => {
    const { container } = render(
      <Flex css={{ bg: 'red', color: 'red' }}>
        <img src="" />
        Hello
      </Flex>
    )
    const results1 = await axe(container)
    expect(results1).toHaveNoViolations()

    const html = ReactDOMServer.renderToString(
      <Flex css={{ bg: 'red', color: 'red' }}>
        <img src="" />
        Hello
      </Flex>
    )
    const results2 = await axe(html)
    expect(results2).toHaveNoViolations()
  })
```

.eslintrc
```diff
@@ -30,20 +30,13 @@
     "plugin:@typescript-eslint/recommended",
     "plugin:import/errors",
     "plugin:import/warnings",
-    "plugin:jsx-a11y/recommended",
+    "plugin:jsx-a11y/strict",
```
thomasdigby, line 33: Can we disable this for `*.test.js` files?

README.md
```diff
@@ -34,7 +34,54 @@ _Note:_ Once a branch gets merged into `main`, it is automatically published, th
   - `@semantic-release/github` - publishes version to `GitHub Releases`
 - `husky` - precommit hook to prevent incomplete commit messages
 
-### How to setup (To be removed once process is setup)
+## Accessibility
```
thomasdigby, line 37: This whole update is great 👍

.eslintrc
andoulla: oh, probably, I will have a look

src/primitives/grid/Grid.test.js
andoulla: It seems that the axe-core is reported to be bad at catching contrast issues when running on `jsdom` which is jest is defaulting. This could explain why on the browser it picks up the colour contrast issues and in the tests it doesn't.

To be honest, we know that at best we could catch about 50% of the issues. There is still a need for complementary testing in the CI maybe (for contrast at leasT)

.eslintrc
thomasdigby: Let’s resolve later, no need to block this PR

thomasdigby (approved):

.eslintrc
andoulla: I found a workaround I would like to try but it is more of a hack so I would prefer to do it in a separate MR

PR #15, opened 2021-01-25
Feat: Input components

hollg:
Adds `Input` and `Label` primitives, `InputField` and `PasswordField` components, `CSSWrapper` util component.

`InputField` is a generic field for text and numbers (we can pass a `type` prop for things like `type="email"`).

`PasswordField` has a fixed `type` and label, and includes a "Forgot your password?" prompt and visibility toggle.

`CSSWrapper` takes a `css` prop and, if `css` is truthy, wraps its `children` in a `Box` which has that CSS applied. If `css` is falsy, it just renders its `children`.

I'll leave the link styling for the links ticket, during which I can come and update this component.

Also adds some jest utils for asserting on HTML attributes etc.

`PasswordField` uses a couple of icons from FontAwesome. If and when we switch to another icon provider this will need updating.

### `TextField`
Screenshot 2021-01-25 at 12 41 43

### `PasswordField`
image
image


src/primitives/input/stories/Input.stories.mdx
```diff
@@ -0,0 +1,30 @@
...
+
+<Meta title="Primitives/Input" component={Input} />
+
+# Input
+
+The `Input` component is a very thin wrapper around an HTML `input` element. It handles some basic styling and some UX / accessibility concerns for number inputs.
+
+<Canvas>
+  <Story name="Basic">
+    <Flex
```
andoulla, line 14: Why is Flex used here?

hollg: To provide spacing between the `Input`s without passing props to them, so that if someone looks at the code via Storybook, they see the code to create the `Input`s they've seen, without extra noise from `css={{mb: '$3'}}` etc

src/components/input-field/InputField.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react'
+
+import { Box, Input, Label } from '../../primitives'
+import { CSSBlob } from '../../stitches'
+
+type InputFieldProps = {
+  label: string
+  name: string
+  type: string
```
thomasdigby, line 9: Can we validate by actual input types available in HTML? I assume React has those available or we can just grab a list from somewhere.

src/components/input-field/InputField.tsx
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export const InputField = ({
+  label,
+  css,
+  name,
+  ...props
+}: InputFieldProps): React.ReactElement => {
+  return (
+    <Box css={css}>
```
thomasdigby, line 20: I'm not 💯 on this `<Box css={css}>` convention as it may just render an attribute-less `div` if no `css` is passed. Perhaps we could create a conditional box wrapper e.g. `css ? <Box  css={css}>{children}</Box> : children`

src/components/input-field/stories/InputField.stories.mdx
```diff
@@ -0,0 +1,30 @@
...
+
+The `InputField` component combines an `Input` and `label` in a standardised way, linking them and styling both. Its `css` prop applies styling to a containing `Box`,
+making it easy to size the `InputField` and position it relative to other components.
+
+<Canvas>
+  <Story name="Basic">
+    <InputField
+      label="Email address"
+      name="email"
+      css={{ width: '250px', mb: '$3' }}
```
thomasdigby, line 17: I think we need to be careful in examples that we write CSS that we consider valid in this context. I wouldn't necessarily recommend passing a `width` prop to an input, but we can have that discussion when reviewing the docs as a whole

src/components/password-field/PasswordField.tsx
```diff
@@ -0,0 +1,62 @@
...
+      >
+        <Label htmlFor={name}>Password</Label>
+        <a href="https://atomlearning.co.uk/forgotten">Forgot your password?</a>
+      </Flex>
+      <Input
+        type={isPasswordVisible ? 'text' : 'password'}
+        name={name}
+        id={name}
+        css={{ mb: '$1' }}
+      />
```
thomasdigby, line 41: I think we could add some extra input attributes here to provide browsers with a better idea that this is a password field. https://www.chromium.org/developers/design-documents/form-styles-that-chromium-understands

src/components/password-field/PasswordField.tsx
```diff
@@ -0,0 +1,62 @@
...
+          fontFamily: 'sans',
+          color: '$secondary500'
+        })}
+      >
+        <StyledIcon
+          icon={isPasswordVisible ? faEyeSlash : faEye}
+          css={generateCss({ mr: '$2' })}
+        />
+        {`${isPasswordVisible ? 'Hide' : 'Show'} password`}{' '}
+      </button>
```
thomasdigby, line 59: This whole thing should be an interactive icon from a `styled('button')` and we should separate it from this component in future

thomasdigby (approved):

src/components/input-field/stories/InputField.stories.mdx
hollg: Oh really, why's that? I think that field length should reflect the length of the expected value (hence showing two different lengths here). Would you prefer a wrapper around an `input` just to determine the width?

src/components/input-field/InputField.tsx
hollg: yeah this is way better, nice one

src/components/input-field/stories/InputField.stories.mdx
thomasdigby: Ah I don't know really, I think you're right that you'd apply it here, I just think it might also wrongly suggest that you _need_ to add a width.

src/components/input-field/InputField.tsx
thomasdigby: Not sure how that would look and what the rules are around applying it to components but it could be a useful util

hollg: Yeah I was just thinking it could be a higher-order component. I'm tempted to make the change as you suggested here and in `PasswordField` for now, and see whether we find ourselves repeating it before bothering to codify it into a util that we reuse a lot

src/components/input-field/InputField.tsx
```diff
@@ -0,0 +1,39 @@
...
+  css?: CSSBlob
+}
+
+export const InputField = ({
+  label,
+  css,
+  name,
+  ...props
+}: InputFieldProps): React.ReactElement => {
+  const Field = () => (
```
hollg, line 19: @thomasdigby what do you think; does this add so much noise that it's worth building this behaviour into a util now?

thomasdigby: Well, I mean I don't really know how I'm reviewing this still, all of my suggestions are not necessary at all for this sprint. But potentially worth creating this now just so we have an idea of how our utils might be structured in `src`

src/components/password-field/PasswordField.tsx
hollg: Yes, totally agree with this. I thought about creating a very simple one in this PR but then I wasn't sure whether `InteractiveIcon` is just a species of `Icon` or whether they're two separate components, and I figured those conversations are worth having before building it.

src/components/password-field/PasswordField.tsx
hollg: Nice one, added!

src/components/input-field/InputField.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export const InputField = ({
+  label,
+  css,
+  name,
+  ...props
+}: InputFieldProps): React.ReactElement => {
+  return (
+    <CSSWrapper css={css}>
```
hollg, line 21: @thomasdigby I know you approved this ages ago, but I can't help myself. What do you think of this, rather than an HOC?

thomasdigby (approved):

PR #16, opened 2021-01-25
Image component

hollg:
Adds `Image` primitive


src/primitives/image/Image.tsx
```diff
@@ -0,0 +1,4 @@
+import { styled } from '../../stitches'
+
+export const Image = styled('img', {})
```
thomasdigby, line 3: Might be a good place to add some defaults:

```js
styled('img', {
  verticalAlign: 'middle',
  maxWidth: '100%'
})
```

thomasdigby: Also, do we still get access to `StitchesProps<typeof Image>;` without exporting them here?

thomasdigby (approved):
There's a lot of potential with this component to optimise the critical path and improve the DX of assets as a whole, but I'll save those for another conversation!

src/primitives/image/Image.tsx
hollg: Are `verticalAlign` and `maxWidth` specifically things you'd recommend setting? Or were you just giving an example? Definitely think we should set defaults where they apply, I just wasn't really aware of any that would be good here.

And yep, we can use `StitchesProps<typeof Image>` anywhere both `StitchesProps` and `Image` are in scope. Since we're doing named exports, maybe it's worth exporting the props types as a convention too? I'm not exactly sure when/how much we'd use them, but I'm sure the answer is not "never".

hollg:
> optimise the critical path

_phwoar_

src/primitives/image/Image.tsx
thomasdigby: `verticalAlign` helps to remove the default white space underneath an image without making it full width (a la display block) and `maxWidth: 100%` is just a good property to set to ensure that images never render larger than their intended size. A 200px wide image rendering at 400px is going to be _bad_ whatever the context.

Regarding the types, do we not need to export them to ensure that when using `<Image />` we get access to the correct prop autocompletion?

hollg: > Regarding the types, do we not need to export them to ensure that when using <Image /> we get access to the correct prop autocompletion?

Nope, it's ✨magic✨! Those prop types are already built into the definition of `Image`, which is enough for VS Code to catch on. The only situation I've come across so far where we need to access them manually is if we want to wrap an `Image` in another component but be able to pass all of the `Image`'s possible props down to it.

PR #17, opened 2021-01-26
Feat: add Link component

hollg:
Adds the `Link` primitive and updates the `PasswordField` component to use it.


src/primitives/link/Link.tsx
```diff
@@ -0,0 +1,11 @@
...
+
+export const Link = styled('a', {
+  color: '$primary500',
+  fontFamily: 'sans',
+  textDecoration: 'none',
+  cursor: 'pointer',
+  ':hover': { color: '$primary900' },
+  ':active': { color: '$primary900' },
+  ':visited': {}
+})
```
thomasdigby, line 11: I think we should add logic here similar to the `Button` component that renders a `button`, `Link` or `a` depending on passed props. The use-case for the login page is that it should render react routers `Link` component.

We should definitely talk about `Link` and `Button` usage and whether we can bind rules around usage to the components (e.g. `Link` never has an `onClick` or any other JS behaviour added to it) but I think the best thing for now is to allow `Link` and `Button` to render the most relevant element semantically and functionally, with the requested style of a link or a button. We could potentially split out that logic into a utility to share across both.

src/primitives/link/stories/Link.stories.mdx
```diff
@@ -0,0 +1,41 @@
...
+    >
+      <Link href="https://app.atomlearning.co.uk/forgotten">
+        Forgot your password?
+      </Link>
+      <Link
+        href="https://app.atomlearning.co.uk/forgotten"
+        css={{
+          fontSize: '$7',
+          color: '$secondary700',
+          ':hover': { color: '$secondary300' }
```
thomasdigby, line 30: These colour variants (and size variants I think) should be handled by stitches variants to encapsulate those theming decisions within the component

src/primitives/link/Link.tsx
```diff
@@ -0,0 +1,11 @@
+import { styled } from '../../stitches'
+
+export const Link = styled('a', {
+  color: '$primary500',
+  fontFamily: 'sans',
+  textDecoration: 'none',
+  cursor: 'pointer',
+  ':hover': { color: '$primary900' },
```
andoulla, line 8: Just coming off reading lots of accessibility stuff, could we implement a pseudo-class for `:focus` too?

src/primitives/link/stories/Link.stories.mdx
```diff
@@ -0,0 +1,41 @@
...
+        }}
+      >
+        Forgot your password?
```
andoulla, line 33: Not sure if it is needed, but wouldn't be better if our examples are a bit more generic, like "this is a link" type of copy

src/primitives/link/Link.tsx
hollg: I think this is a really good idea, but I'm not sure how that should be styled. I think it should probably be applied consistently to interactive components that are `:focus`ed though. Do you have any suggestions?

src/primitives/link/Link.tsx
hollg: What would be the use case for rendering a `button` in this component? My understanding is that we should style an `a` to look like a button—if that's what we want—rather than render a `button` than navigates the user somewhere else.

src/primitives/link/Link.tsx
andoulla: At this point we need Chema's or Amit's input, to be honest, but if they are not available I would just use what we have, for example using the hover colour-variant (either making both states the same or as a border).

src/primitives/link/Link.tsx
thomasdigby: For example, if the link opened a modal. I think we can make a decision either way (text links are only links, or text links can be anything etc.) but with so many variations in what actions the link could take, it makes sense to me to bake in that flexibility.

andoulla: I would consider some less flexibility (Ie: a link is a link ) to begin with as it helps keep the complexity low. If we find that our consumers need more flexibility then we can add more to it.

hollg: Yeah I would feel very strongly that a `Button` should be used for that.

Coupling this component to `react-router` also feels like quite a big statement to me. It adds a peer dependency to the package and although we use `react-router` in our app, we've done a lot so far to keep the package decoupled from our specific implementation, which I think is really worthwhile.

I'm pretty sure the `as` prop on `react-router`'s `Link` would work with our `Link` (although maybe ours needs renaming to avoid confusion!). `<Link to="/stats" as={OurLinkWithADifferentName} />`

thomasdigby: I think generally the rule that `link` navigates and `button` does an action on the current page is fine.

However, I disagree that we must use the `Button` component to complete an action (like open a modal window) as from the users perspective it still shares the same fundamental principle as a `link` would, navigating them somewhere. The distinction between which actions are classed as navigation are blurry in my opinion and in this case I don't think that we should strictly define the functional properties of the component based on whether it looks like a `link` or a `button`.

In a pure and impeccably planned UI system you may be able to follow those rules to the letter, but we _need_ to bake in a modicum of flexibility otherwise we're unnecessarily constraining ourselves. We have a big team and a lot of apps to convert over the next 6/12 months and I think there are other, more impactful decisions that we can make around semantics and restricting component usage than this.

Totally agree on the react-router point, can you test if `<Link as={OurLink} />` works as expected?

hollg: Ah, I'm sorry, I was being very dense when reading your previous comments! I totally agree that we should support a text `Link` that opens a modal, and that it should render an HTML `button`. What I _don't_ think we should do is use `Link` to render something that _looks like_ a button. But I realise now that's not what you were suggesting!

Re `react-router`, it seems like they've changed the API to be `<Link component={OurLink}/>` now but it works fine!

src/primitives/link/stories/Link.stories.mdx
hollg: I think it's important to keep examples simple, but I also think it's valuable to have them reflect something real so as to give the reader a clearer idea of how the component fits into a view / other components.

I wouldn't have thought it makes much of a difference in this case (but please correct me if you disagree on that!), but that's the principle I was thinking about.

thomasdigby (approved):

src/primitives/link/stories/Link.stories.mdx
andoulla: It is a tiny detail so we could go with either. I was simply basing it on the examples of other DSs we shared last week. They tend to give a very generic copy.

PR #18, opened 2021-01-26
Feat: add Text component

hollg:
Adds `Text` component for body text. Its three `size` variants have default styling based on the typography in the current app.


src/primitives/text/Text.tsx
```diff
@@ -0,0 +1,42 @@
+import * as React from 'react'
+
+import { CSSBlob,styled } from '../../stitches'
+
+const BaseText = styled('p', {
+  fontFamily: 'sans',
+  color: '$tonal900',
+  fontWeight: 400,
+  maxWidth: '100ch',
```
thomasdigby, line 9: Hoo boy am I keen to reduce this value and _enforce it_. Let's chat with Chema & Amit when we implement it proper

thomasdigby (approved):

PR #23, opened 2021-01-27
Local environment and output fixes

thomasdigby:
- Set correct `NODE_ENV` value for builds
- Pointing exports to the correct output files
- Fixed the package name
- Added `babel-plugin-react-remove-properties` to remove any `testid`s from the bundle and `babel-plugin-transform-stitches-display-name` to automatically add `displayName` to `styled()` components
- Updated the github action install command to remove the `--frozen-lockfile` arg

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,13 @@
 {
   "env": {
+    "development": {
+      "plugins": ["transform-stitches-display-name"]
+    },
+    "production": {
+      "plugins": [
+        ["react-remove-properties", { "properties": ["data-testid"] }]
+      ]
+    },
     "test": {
       "presets": ["@babel/preset-react", "@babel/preset-typescript"],
       "plugins": ["@babel/plugin-transform-modules-commonjs"]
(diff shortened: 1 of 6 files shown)


hollg:
Is it worth changing the sandbox output in this PR? It sort of falls under the "output fixes" heading, and my tree-shaking exploration has hit a bit of a dead end, so otherwise I'll just make a PR that just changes the sandbox output.

I changed the `start:sandbox` command to: `"parcel ./sandbox/index.html --out-dir sandbox/build --open"`

.babelrc
```diff
@@ -1,5 +1,13 @@
 {
   "env": {
+    "development": {
+      "plugins": ["transform-stitches-display-name"]
```
andoulla, line 4: does this mean we not longer have to do `Input.displayName= input`

hollg:
> Updated the github action install command to remove the --frozen-lockfile arg

Is this PR missing a change to `.github/workflows.main.yml`?

thomasdigby:
> Is it worth changing the sandbox output in this PR? It sort of falls under the "output fixes" heading, and my tree-shaking exploration has hit a bit of a dead end, so otherwise I'll just make a PR that just changes the sandbox output.
> 
> I changed the `start:sandbox` command to: `"parcel ./sandbox/index.html --out-dir sandbox/build --open"`

Have added to this PR

andoulla (approved):

PR #24, opened 2021-01-27
Buttons

andoulla:
- [x] Added initial 3 variations of the button (PRIMARY, SECONDARY, TERTIARY)
- [x] Added initial tests
- [x] Added basic story
- [x] Update sandbox to use the button

I have completed the outlined variants as they are not needed for the demo, still have to investigate how to get them to work in the stitches recommended format.


src/primitives/button/Button.tsx
```diff
@@ -0,0 +1,62 @@
...
+  color: 'white',
+  fontSize: 'md',
+  height: '40px',
+  // TODO missing disabled states
+  variants: {
+    variant: {
+      primary: {
+        backgroundColor: '$primary500',
+        color: 'white',
+        ':hover': {
```
andoulla, line 20: It seems like a shame, that we can't combine the `hover` and `focus` state as they have exactly the same styling.

maybe we need a nice way of generating these repetitive props as when we add the disabled states it will mean that we will also have to duplicate those states too.  (Attaching the code as taken from the frontend for context)
`&:not([disabled]):hover,
  &:not([disabled]):focus {
    background-color: $color-dark;
    color: white;
    text-decoration: none;
  }
  `

src/primitives/button/Button.tsx
```diff
@@ -0,0 +1,62 @@
...
+      }
+    }
+  }
+}
+
+const BaseButton = styled('button', StyledButton)
+
+type ButtonProps = StitchesProps<typeof BaseButton>
+
+export const Button = ({ ...props }: ButtonProps): React.ReactElement => {
```
hollg, line 54: 💬 the props here can be simplified to `props: ButtonProps` — there's no need to use the rest operator if we're taking the whole object

src/primitives/button/stories/Button.stories.mdx
```diff
@@ -0,0 +1,44 @@
...
+  title="Primitives/Button"
+  component={Button}
+  argTypes={{
+    variant: 'string'
+  }}
+/>
+
+# Button
+
+The `Button` component is a very thin wrapper around an HTML `button` element. It adds default styling and the `css` prop.
```
hollg, line 18: As we already have some defined variants, it would be good to write something in the docs about the intended use case for each one. I'm not sure whether we need to get input from Chema/Amit before writing anything like that though.

src/primitives/button/Button.tsx
```diff
@@ -0,0 +1,62 @@
...
+  return <BaseButton {...props} />
+}
+
+Button.displayName = 'Button'
+
+Button.defaultProps = {
```
andoulla, line 60: Made the executive decision to default to primary. Any thoughts

src/primitives/button/Button.tsx
```diff
@@ -0,0 +1,62 @@
...
+        },
+        ':focus': {
+          backgroundColor: '$tertiary700'
+        }
+      }
+    }
+  }
+}
+
+const BaseButton = styled('button', StyledButton)
```
hollg, line 50: 💬 does it look messy if we pass the default styles object into `styled` directly? If we define the object elsewhere, we don't get the benefit of autocomplete on our styling variables. Unless there's a type we can import from Stitches and apply to the object to enable them? 🤔

src/primitives/button/Button.tsx
hollg, line 20: 💬 Could we also add styling for `:active`:, to give feedback that the button has been pressed? For the `Link` component I changed the `color` back to `$primary500` on `:active`, which I think would also work here. There may well be better patterns for that feedback though.

hollg (changes requested):

src/primitives/button/Button.tsx
hollg: I think that makes total sense. 

I've seen some Stitches examples where the default values aren't in a variant at all, but I think the way you've done it makes everything more explicit and predictable. 👍🏻

src/primitives/button/Button.tsx
andoulla: yes, sorry this is left-over from me spreading to get the variant which I don't need as I have added a default for now

src/primitives/button/stories/Button.stories.mdx
andoulla: it would be good but I'm not 100% sure when to use what. I think we should do this part of the documentation, once all the variants are in (now I only did the basic ones). 

And you are definitely right, we do need Chema's /and Amit's input

src/primitives/button/Button.tsx
hollg: I've been wondering about this too. We could create mixins as JS objects, maybe?

Re the styling from the main app, I tried to find a way to use `not()` in Stitches but didn't find anything, but now it occurs to me that it could be as simple as wrapping the key in a string: `'&:not([disabled]):hover' : { textDecoration: "underline" }`

andoulla: Thanks, I didn't know what colour to set it to. in `Button.module.scss` I couldn't find that info

thomasdigby: Can we not do `":hover,:focus": {}`?

src/primitives/button/Button.tsx
thomasdigby: I think as a rule we should always pass the object directly into the `styled()` method, we don't really gain anything by using this pattern

src/primitives/button/stories/Button.stories.mdx
```diff
@@ -0,0 +1,44 @@
...
+        justifyContent: 'space-between',
+        width: '500px',
+        height: '200px'
+      }}
+    >
+      <Button css={{ width: '150px' }}>Primary</Button>
+      <Button css={{ width: '150px' }} variant={'secondary'}>
+        Secondary
+      </Button>
+      <Button css={{ width: '150px' }} variant={'tertiary'}>
```
thomasdigby, line 35: We shouldn't need to define `width`s here and I think it's not a good example to show, perhaps a layout specific prop like `mb` or `alignSelf` would be more useful here

src/primitives/button/Button.test.js
```diff
@@ -0,0 +1,21 @@
...
+
+import { Button } from '.'
+
+describe(`Button component`, () => {
+  it('renders a button', async () => {
+    const { container } = render(<Button css={{ width: 100 }}>BUTTON</Button>)
+
+    await screen.getByText('BUTTON')
+    expect(container).toMatchSnapshot()
+  })
```
thomasdigby, line 13: Do we think it's useful to add tests for our variants? It's built into the styling library so it may not be necessary here.

hollg: I've been wondering about this too. I think in this case it would amount to testing Stitches, but that if we write a component that has any logic in it about which variant of a primitive to render, that logic should be tested.

src/primitives/button/Button.tsx
andoulla: `":hover,:focus": {}`  doesn't work correctly. we do need something more sophisticated than just typing out all of these.

src/primitives/button/Button.tsx
andoulla: I have changed it

src/primitives/button/Button.test.js
andoulla: I was wondering if I should add something here but not sure if more snapshots are useful or I should just check if the variant prop is being used (as this is testing the internals of the component)

Happy to adopt any suggestions as it feels wrong to only test the default

src/primitives/button/stories/Button.stories.mdx
andoulla: fair enough. do you think I should have a max-width set on the button default styling? or is it ok to let them take the full width of the parent

src/primitives/button/Button.tsx
thomasdigby: https://github.com/modulz/stitches/issues/278#issuecomment-707374961 This comment seems to indicate that it is possible, I guess this is the issue with using a not prod ready package at this early stage. Maybe try `&:hover, &:focus`?

src/primitives/button/stories/Button.stories.mdx
thomasdigby: Ah I think it's worth adding a `width: max-content` to our default styles then

src/primitives/button/Button.test.js
thomasdigby: I think what @hollg said is right, unless we do anything to change the behaviour of the actual component there's no point testing it. Feels wrong, but probably is the right thing to do.

src/primitives/button/Button.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { styled } from '../../stitches'
+
+const StyledButton = {
+  borderRadius: '$1',
+  borderWidth: 0,
+  padding: '$2 $4',
+  width: '100%',
+  color: 'white',
+  fontSize: 'md',
+  height: '40px',
```
thomasdigby, line 13: We could use one of the sizes defined in our theme here https://github.com/Atom-Learning/theme/blob/main/src/index.ts#L56-L61

I've tried to keep those values close to our 8px grid so `48px` or `$2` should probably do the job 👍

src/primitives/button/Button.tsx
```diff
@@ -0,0 +1,62 @@
...
+import * as React from 'react'
+
+import { styled } from '../../stitches'
+
+const StyledButton = {
+  borderRadius: '$1',
+  borderWidth: 0,
+  padding: '$2 $4',
+  width: '100%',
+  color: 'white',
```
thomasdigby, line 11: This should be defined in our variant, not in the default styles

src/primitives/button/Button.test.js
andoulla: I didn't refresh the page so didn't see Gary's comment

src/primitives/button/stories/Button.stories.mdx
andoulla: Thanks, I'm taking your mb suggestion too

src/primitives/button/Button.tsx
andoulla: sorry about this, this is what happens when you copy paste components

src/primitives/button/Button.tsx
andoulla: actually, it is defined in the variants, making this line redundant. I have removed

src/primitives/button/Button.tsx
andoulla: spot on! `'&:hover, &:focus'` works. But we should use the comma notation as soon as it is available

thomasdigby (approved):

hollg (approved):

PR #25, opened 2021-01-28
Buttons advanced

andoulla:
- [x] add outlined versions of each button variant
- [x] add disabled states for each button
- [x] add some documentation (initial until we get Chema's/Amit's feedback on how they should be used)

image


thomasdigby (approved):
👍 👍

PR #27, opened 2021-01-28
Clean up sandbox

andoulla:


README.md
```diff
@@ -4,6 +4,56 @@
...
 This package is one part of the Atom Learning design system
 
+## Getting started
+
+After cloning the repository please run the following commands:
+
+```
+yarn
+yarn build
+yarn dev
```
thomasdigby, line 14: It would be great to describe what these commands achieve, same with all the commands you've listed below. One for next week I reckon 👍

sandbox/Sandbox.tsx
```diff
@@ -1,7 +1,15 @@
...
+import { Box } from '../dist'
+/**
+ * Component library sandbox
+ *
+ * Useful documentation:
+ * - https://stitches.dev
+ * - https://github.com/typescript-cheatsheets/react/blob/main/README.md#basic-cheatsheet-table-of-contents
+ *
+ * Note: Please don't commit the changes in /sandbox/index.tsx
+ */
```
thomasdigby, line 12: Would be great to include the text in sandbox as well:

```
Use this app to quickly render components in an environment where you have more control than in Storybook. But don't commit changes to this app into the main branch!
```

sandbox/Sandbox.tsx
```diff
@@ -14,48 +22,10 @@ const Container = (props) => (
...
-
-const Body = (props) => <Box css={{ flexGrow: 1, p: '$3' }} {...props} />
+const Body = (props) => <Box css={{ flexGrow: 1 }} {...props} />
 
 export const Sandbox = ({ children }) => (
   <Container>
-    <Header />
     <Body>{children}</Body>
   </Container>
 )
```
thomasdigby, line 31: I would potentially just remove this file altogether and keeping a simple file in our index.tsx

```jsx

import { Box } from '../dist'

css.global(reset)()

// Add your new comments here with the links

const App = () => {
  return (
    <Box>
      Edit me
    </Box>
  )
}

```

README.md
andoulla: I know I just wanted to get a skeleton down so we can add things as we go along. and then the `write README` ticket becomes clean up/sort the readme

sandbox/Sandbox.tsx
andoulla: OK will do, it makes sense now that we have removed the header etc

thomasdigby (approved):

PR #28, opened 2021-02-01
contribution_guide

andoulla:


README.md
```diff
@@ -4,6 +4,105 @@
 
 This package is one part of the Atom Learning design system
 
+## Component API
```
andoulla, line 7: Not sure if this should be changed to contributing and moved down to line 37.

README.md
```diff
@@ -4,6 +4,105 @@
...
+- A link to the ticket for the work
+- A brief description of the work you've done
+- Links to any designs provided
+- Screenshots of the built component
+- An explanation of any decisions that were made.
+
+If any of the patterns stated above weren't followed, please explain the reasons for it.
+
+---
+
```
thomasdigby, line 105: Can we move all of this to a `CONTRIBUTING.md`? https://github.blog/2012-09-17-contributing-guidelines/

README.md
hollg: I think it would be good to add `## Contributing` and move all of this content under that heading (and therefore moving these headings down a level). To me, `Versioning` and `Accessibility` feel like they're at the same level as `Contributing`. `Contributing` just has a bunch of subsections, but they could all be lifted out and would make sense on their own, e.g. to go into their own doc.

thomasdigby (approved):
Other than the file change I think this reads well, it might be a tad prescriptive but I don't think thats a bad thing at this stage, it's really clear what we validate as a complete and ready component 👍

README.md
andoulla: I'm wondering now that I see it in the `README.md` and how long it is if we should separate it out.

README.md
```diff
@@ -4,6 +4,38 @@
 
 This package is one part of the Atom Learning design system
 
+## Component API
```
hollg, line 7: Is this section not going in the contribution guide?

andoulla: yes sorry, you are right

PR #29, opened 2021-02-01
Feat: add path aliases

hollg:
Adds `~/*` as an alias for `src/*`, allowing this:
```ts
import { Box } from '~/components'
import { styled } from '~/stitches`
```

But not allowing the following:
```ts
import { Box, styled } from '~'
```

All imports are updated and `tsconfig-paths-webpack-plugin` has been added to allow Storybook to follow the new imports. (This is annoying since we're moving away from Storybook, but we can remove the dependency when we remove Storybook).


.storybook/main.js
```diff
@@ -1,9 +1,15 @@
+const TsconfigPathsPlugin = require('tsconfig-paths-webpack-plugin')
```
andoulla, line 1: Should we have a TODO here to remind us to remove the package?

I'm not the biggest fan of leaving TODOs lying around but might be a good short-term reminder

hollg: I certainly can do!

andoulla (approved):

thomasdigby:
@hollg Do we need to consider adding a `jsconfig` here to allow module resolution in VSCode? https://code.visualstudio.com/docs/languages/jsconfig

hollg:
> @hollg Do we need to consider adding a `jsconfig` here to allow module resolution in VSCode? https://code.visualstudio.com/docs/languages/jsconfig

Nope, it Just Works™

PR #31, opened 2021-02-03
Feat: add input component

hollg:
Adds `Input` component.
Ticket: https://app.asana.com/0/1199598847899587/1199672175559115

## Screenshots
Default state
image

Focussed
image

Disabled
image


src/components/input/Input.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+        placeholder="001"
+      />
+    )
+
+    const input = await findByPlaceholderText('001')
+
+    expect(input).toHaveAttribute('inputmode', 'numeric')
+    expect(input).toHaveAttribute('pattern', '[0-9]*')
+    expect(container).toMatchSnapshot()
+    expect(await axe(container)).toHaveNoViolations()
```
hollg, line 33: I've inlcluded the `axe` assertion in each test rather than in its own test, because we need to make sure that each variant passes. I'd be interested to hear other people's thoughts on whether this is the right approach or whether we should have one accessibility test which renders all variants.

src/utilities/types.ts
```diff
@@ -0,0 +1,24 @@
+/**
```
hollg, line 1: I'm very keen to hear thoughts on this explanation / any ideas for how to improve it!

src/components/input/Input.stories.mdx
```diff
@@ -0,0 +1,23 @@
...
+it might be worth adding a new one.
+
+`Input` accepts a subset of the `types` that an HTML `input` element because we have more specialised
+components for the others (e.g. `Checkbox`, `Radio`).
+
+## Accessibility
+
+While `Input` accepts `"number"` for its `type` prop, this does not render `<input type="number"/>`. Instead,
+it renders `<input type="text" inputmode="numeric" pattern=[0-9]*/>` to tell browsers to show a numeric keyboard
+while avoiding the [UX and accessibility problems associated with `type="number"`](https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/).
```
andoulla, line 23: Really like that you added this section here

src/components/input/Input.test.tsx
andoulla: I was about to ask you and Tom, about our approach on this. If we could combine the snapshot with a few assertions. 
I have the experience where each thing would be it's own unit tests and I understand the principle behind it that. Each unit should address only one concern.
The only pitfalls are that it leads with ending up with lots of tests to go through and maintain, and when you are talking in 1000s of tests (the last project had over 1200) it slowed down the dev experience as the milliseconds eventually add up.

src/components/input/Input.stories.mdx
hollg: Ah, that's useful to hear! I wasn't sure whether to include it or not because on one hand it feels like the sort of implementation detail that we're trying to "protect" users of the design system from, but on the other hand I think it's kind of interesting 🤷🏻‍♂️

src/components/input/Input.tsx
```diff
@@ -0,0 +1,53 @@
...
+  color: '$tonal900',
+  cursor: 'text',
+  display: 'block',
+  fontFamily: 'sans',
+  fontSize: 'md', // necessary to prevent iOS zooming on focus
+  height: '50px',
+  width: '100%',
+  p: '$3',
+  transition: 'all 100ms ease-out',
+  ':focus': {
```
andoulla, line 21: do we need to add states like `hover` and `active`?

src/components/input/Input.tsx
```diff
@@ -0,0 +1,53 @@
+import * as React from 'react'
+
+import { StitchesProps, styled } from '~/stitches'
+import { Override } from '~/utilities/types'
+
+const BaseInput = styled('input', {
```
thomasdigby, line 6: I wonder if it's best to use a particular convention like `StyledInput` instead? I think it probably says more about what the value is than `BaseInput`

src/components/input/Input.tsx
```diff
@@ -0,0 +1,53 @@
...
+  border: '1px solid $tonal500',
+  borderRadius: '$1',
+  boxShadow: 'none', // necessary to prevent default iOS default styling
+  boxSizing: 'border-box',
+  color: '$tonal900',
+  cursor: 'text',
+  display: 'block',
+  fontFamily: 'sans',
+  fontSize: 'md', // necessary to prevent iOS zooming on focus
+  height: '50px',
```
thomasdigby, line 17: We have a `size` scale that we can access here, I think it'll be useful to ensure that buttons, inputs etc. render at the same height if they're ever used side by side. I think you can get `48px` on the size scale.

src/components/input/Input.tsx
```diff
@@ -0,0 +1,53 @@
...
+// override default 'type' property to prevent Input from being used to render
+// checkboxes, radios etc — we will have dedicated components for them
+type InputProps = Override<
+  StitchesProps<typeof BaseInput>,
+  {
+    type: 'text' | 'number' | 'email' | 'password' | 'tel' | 'url'
+  }
+>
+
+export const Input = ({ type, ...props }: InputProps): React.ReactElement => {
```
thomasdigby, line 42: Can we apply the default prop here?

src/utilities/types.ts
thomasdigby: Only think I'd add is to use full names for our params, any reason why `T` can't be `BaseType`?

src/components/input/Input.tsx
```diff
@@ -0,0 +1,53 @@
...
+  appearance: 'none',
+  border: '1px solid $tonal500',
+  borderRadius: '$1',
+  boxShadow: 'none', // necessary to prevent default iOS default styling
+  boxSizing: 'border-box',
+  color: '$tonal900',
+  cursor: 'text',
+  display: 'block',
+  fontFamily: 'sans',
+  fontSize: 'md', // necessary to prevent iOS zooming on focus
```
thomasdigby, line 16: This key has changed since we published the theme, can you install the latest `@atom-learning/theme` and check that the keys all match up.

src/utilities/types.ts
hollg: `T` and `U` are just the conventional names for types in a generic type, in the same way that `i` and `j` are the conventional shorthand for `index` in (nested) loops. They are what you'd see in the language docs (https://www.typescriptlang.org/docs/handbook/generics.html) and what you'd expect to see in the wild.

src/components/input/Input.test.tsx
hollg: Yeah, I think this is worth a proper discussion.

> Each unit should address only one concern.
100% agree. But I think it's reasonable to say that each variant of the component is one concern, and asserting on both accessibility concerns and the rendered output isn't much different from asserting on two different HTML attributes.

src/components/input/Input.tsx
hollg: We don't have any styling/behaviour for those states in the main app currently (which is broadly what I've based this on so far). Is there anything you'd suggest adding for them?

src/components/input/Input.tsx
hollg: I definitely think we should stick to a convention on this but don't have particularly strong feelings about what it is to be honest! Happy to change this to `StyledInput`.

src/components/input/Input.tsx
hollg: Oh yeah nice one, I've definitely come across (and probably written) some v questionable CSS to make buttons and inputs line up properly, so anything we can do to avoid that is 💯.

src/components/input/Input.tsx
hollg: Certainly can do. I've probably been inconsistent on this so far actually. Is the default paramater syntax The Done Thing™ these days?

src/utilities/types.ts
thomasdigby: I haven't written enough Typescript to have much of a valid opinion, but I do think that writing actual names can help document the behaviour? Anecdotally, I have seen more people talk about being verbose with names recently https://fettblog.eu/tidy-typescript-name-your-generics/ https://twitter.com/kentcdodds/status/1348016701632221184

src/components/input/Input.tsx
hollg: What key should it be using now? `size`?

src/components/input/Input.stories.mdx
thomasdigby: It's definitely worth documenting, users don't need that detail when they use the design system, it should just work. But for the long-term health of the project, ensuring contributions are valuable etc. it's important that we document our decisions as well 👍

src/components/input/Input.tsx
thomasdigby: I think the cursor changing is enough visual aid that you're hovering on an input 👍 

I don't think `active` is actually relevant on an input?

src/components/input/Input.tsx
thomasdigby: I mean only from what I've read on the internet, which is not particularly reliable 🤷  Although I haven't seen a TS + React example where they use `Component.defaultProps`, they always lean towards destructuring props and providing defaults there.

src/components/input/Input.tsx
thomasdigby: As in the theme key, so all keys are now prefixed with `$`, so this would be `$md` and `fontFamily` would be `$sans`

src/utilities/types.ts
hollg: Interesting! I'm a big KCD fan but find that tweet kind of funny because this isn't a TypeScript phenomenon at all; it's been the convention since before TS existed!

I definitely don't think we hsould should stick to bad conventions just because they exist, but I think there's a balance to be struck with the benefit of people already knowing conventions.

I find that article much more compelling, but I'm not sure how we'd name the second type parameter here based on its advice. Do you have any thoughts?

Overall I'd like to keep our TypeScript as consistent as possible across our code bases, so I'll ping Brendan on this in Slack to get his opinion / hear what's being done in the existing TypeScript code that we have, and I'll document whatever we all land on in the TS handbook in Notion.

src/components/input/Input.test.tsx
andoulla: It makes sense. I was already doing it the same thing on my branch, but I wasn't sure about it

src/utilities/types.ts
bjshort: Hello GitHub friends.

So I would say that more recently I tend to learn towards naming my generics in situations where it's not blindingly obvious what the generic represents (I've done this recently in persist). But in other situations i think it's fine to use T & U which are pretty well accepted across most generic languages. 

It's a lot harder to name generics when the function is 'super generic' though.

src/components/input/Input.tsx
andoulla: I was just thinking this is similar to the button where we added the missing `:active`  despite the original implementation not having it.

@thomasdigby The browser implements the `:active` state but I don't know if we need it
image

src/utilities/types.ts
hollg: That's useful, thanks! Do you have any feelings about which category this type falls into?

src/components/input/Input.test.tsx
andoulla: Btw have you been able to write failing `axe` test for the input?

src/utilities/types.ts
bjshort: You could rename T -> BaseType i suppose (Not sure about U) - but for me this kinda does feel a bit like a 'super' generic function and therefore I wouldn't have major reservations about leaving it as is.

hollg: Yeah I wasn't sure what `U` would become either. `Overrides` would make some sense to me, but then I'm not sure whether that causes extra confusion by being so similar to the name of the generic type. But we could change that too!

src/components/input/Input.tsx
hollg: The `:active` state is triggered when you click on or tab into the `input`, but the `:focus` state is triggered then too so we are already giving feedback on, effectively, the same event.

src/components/input/Input.tsx
hollg: I think that makes sense with React's push towards "components are just functions". Have updated!

src/utilities/types.ts
thomasdigby: Happy to retain `T` and `U` from this conversation, thanks both!

thomasdigby (approved):

andoulla (approved):

src/components/input/Input.tsx
```diff
@@ -0,0 +1,57 @@
...
+// checkboxes, radios etc — we will have dedicated components for them
+type InputProps = Override<
+  StitchesProps<typeof StyledInput>,
+  {
+    type: 'text' | 'number' | 'email' | 'password' | 'tel' | 'url'
+  }
+>
+
+export const Input = ({
+  type = 'text',
```
hollg, line 43: @thomasdigby can you see anything wrong here? My tests are now failing to build because I'm not explicitly passing the `type` prop in them. Works fine with `.defaultProps`

thomasdigby: Huh, no that should be fine. The issue seems to be with the theme import again 🤔 Can you update all instances of that import to a default import?

hollg: Sorry, I meant my tests running locally (where I have fixed the import / file in `node_modules`)

PR #32, opened 2021-02-03
Feat:Button Primitive

andoulla:
Finished the following options for the button:
- Themes -> Primary, Secondary, Tertiary, Success, Warning, Danger
- Variants -> Outline
- Disabled states

I also added test for 1 of each of those options 

Screenshots

Screenshot 2021-02-03 at 15 02 58

hover:
Screenshot 2021-02-03 at 14 59 06

focus:
Screenshot 2021-02-03 at 15 01 05

---
Screenshot 2021-02-03 at 15 03 07
hover:

Screenshot 2021-02-03 at 15 03 36
---

disabled:
image


src/components/button/Button.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+    )
+
+    const button = await screen.getByText('BUTTON')
+
+    expect(button).toBeDisabled()
+    expect(container).toMatchSnapshot()
+    expect(await axe(container)).toHaveNoViolations()
+  })
+
+  // TODO: figure out accessibility test issue
```
andoulla, line 48: will remove this before merging, I just want to spend a bit of time on this

src/components/button/Button.tsx
```diff
@@ -0,0 +1,179 @@
...
+  width: 'max-content',
+  '&[disabled]': {
+    backgroundColor: '$tonal300',
+    color: '$tonal600',
+    opacity: 0.35,
+    cursor: 'not-allowed'
+  },
+  variants: {
+    theme: {
+      primary: {
```
thomasdigby, line 33: Can we add the `block` or `solid` appearance variant as compound variants and move all of these colour combinations over as well? I think it makes sense to declare them all consistently, whether its variants or compoundVariants.

src/components/button/Button.tsx
```diff
@@ -0,0 +1,179 @@
...
+    }
+  }
+})
+
+StyledButton.compoundVariant(
+  {
+    theme: 'primary',
+    variant: 'outline'
+  },
+  {
```
thomasdigby, line 113: It seems like we might be able to optimise this for both solid/outline variants and all colours by using a little utility to handle the general schema e.g.

```js
const getButtonOutlineVariant = (baseColor, interactColor) => ({
  color: baseColor,
  backgroundColor: "white",
  "&:not([disabled]):hover, &:not([disabled]):focus": {
    color: interactColor,
    backgroundColor: "white",
  },
  ":active": {
    color: baseColor,
  },
  "&[disabled]": {
    color: interactColor,
  },
});

StyledButton.compoundVariant(
  { theme: "primary", variant: "outline" },
  getButtonOutlineVariant("$primary500", "$primary900")
);
StyledButton.compoundVariant(
  { theme: "secondary", variant: "outline" },
  getButtonOutlineVariant("$secondary500", "$secondary700")
);
```

src/components/button/Button.tsx
```diff
@@ -0,0 +1,179 @@
...
+      color: '$tertiary700'
+    }
+  }
+)
+
+type ButtonProps = StitchesProps<typeof StyledButton>
+
+export const Button = ({
+  theme = 'primary',
+  ...rest
```
thomasdigby, line 174: Would be good to add `appearance` or `variant` (shall we agree not to use `variant` as it overlaps with stitches and might be confusing for first time implementers?) to the props and add a default here. Just to make it nice and clear what the default button will return.

src/components/button/Button.tsx
```diff
@@ -0,0 +1,179 @@
...
+}: ButtonProps): React.ReactElement => {
+  return <StyledButton theme={theme} {...rest} />
+}
+
+Button.displayName = 'Button'
```
thomasdigby, line 179: This will be automated so no need to include it

src/components/button/Button.tsx
```diff
@@ -0,0 +1,179 @@
...
+  alignItems: 'center',
+  fontFamily: 'sans',
+  fontSize: 'md',
+  fontWeight: 500,
+  height: '$2',
+  letterSpacing: '0.02em',
+  lineHeight: 1.4,
+  transition: 'all 125ms ease-out',
+  textDecoration: 'none',
+  padding: '$2 $4',
```
thomasdigby, line 22: I think it's worth setting a convention to use shorthand style properties where we can, e.g. `px: '$2', py: '$4'`

src/components/button/Button.tsx
```diff
@@ -0,0 +1,179 @@
...
+const StyledButton = styled('button', {
+  background: 'unset',
+  border: 'none',
+  borderRadius: '$1',
+  cursor: 'pointer',
+  display: 'flex',
+  justifyContent: 'center',
+  alignItems: 'center',
+  fontFamily: 'sans',
+  fontSize: 'md',
```
thomasdigby, line 15: Mentioned to @hollg about the updated naming conventions in the updated theme, worth merging `main` into this branch and following his updates once #31 has been merged

src/components/button/Button.tsx
```diff
@@ -0,0 +1,179 @@
+import { StitchesProps } from '@stitches/react'
+import * as React from 'react'
+
+import { styled } from '../../stitches'
```
thomasdigby, line 4: We should have aliases now for this import right?

thomasdigby (commented):
Looking good, just need to tweak how we apply those variants I reckon 👍

src/components/button/Button.tsx
andoulla: I really like this, I will adopt this and create something for the solid variant too

src/components/button/Button.tsx
andoulla: should we use `theme` for colours and `appearance` for solid/outline?

src/components/button/Button.tsx
hollg: Could we derive `interactColor` from `baseColor` programatically, rather than passing it in? 🤔

src/components/button/Button.tsx
hollg: yep, `import { styled } from '~/stitches'`

andoulla: - []

andoulla: done

src/components/button/Button.tsx
andoulla: ok, it is just force of habit. I'm not really sure about these values though. I took a bit of a guess what they should be as I the `button.module.scss in `atom-core` doesn't specify it, but I can clearly see some padding being applied

thomasdigby: Yeah I think the values themselves are fine, it's just about how we apply them. I think if you look under the sizes section in `button.module.scss` you should see padding.

src/components/button/Button.tsx
andoulla: do you mean passing the theme in and the concatenate the value with the darkness?

src/components/button/Button.tsx
thomasdigby: Yeah I like that 👍

src/components/button/Button.tsx
thomasdigby: This is what I mentioned the other day, unless you mean something like `StyledButton.compoundVariant(getButtonOutlineVariant('secondary'))` and store the base and interact colors in an object elsewhere in the file? e.g. `const appliedThemes = { primary: { base: '$primary500', interact: '$primary700' }}`

andoulla: I'm not sure that benefit this pattern gives us? We are not going to need to use these mappings anywhere else, and it does add a 3rd layer of abstraction (which can be both good and bad at the same time)

hollg: I was initially thinking of something like this:
```ts
type Theme = 'primary' | 'secondary' | 'success' /*  we could */

const getButtonOutlineVariant = (theme: Theme ) => {
const baseColor = `${theme}500`
const interactColor = `${theme}700`

/* a bunch of stuff*/

}
```

But I think @andoulla 's point about a third abstraction is a very good one. As far as I know, the strings "primary", "secondary" etc don't exist anywhere in the theme. If they did, we could derive the `Theme` type above from them, and then _maybe_ this would be OK? 

But as long as they don't, we'd be relying on this type definition being updated alongside the theme, which isn't great. And maybe my whole concept is way too opinionated about the theme's tokens anyway!

The benefit I was thinking of is that we'd be enforcing a specific relationship between `baseColor` and `interactColor`, but maybe that's not the place of this package anyway.

andoulla: It would be lovely that doesn't work for all the variations. 
For instance `primary` goes up to `$primary900` and `success` doesn't have a numeric scale it uses `$successDark`.

We could change the names of the colours in `theme` to make them consistent but it still leaves the fact that we will have code an exception for primary

hollg: Oh, yes, in that case it definitely doesn't work!!

andoulla: Let's get to the middle ground then. I created 2 functions `getButtonOutlineVariant` and `getButtonSolidVariant` which reduced the code duplication by over 80 lines already. Then I will push it and see what everyone things of that version and take it from there

src/components/button/Button.tsx
```diff
@@ -0,0 +1,114 @@
...
+    theme: 'tertiary',
+    variant: 'outline'
+  },
+  getButtonOutlineVariant('$tertiary500', '$tertiary700')
+)
+
+type ButtonProps = StitchesProps<typeof StyledButton>
+
+export const Button = ({
+  theme = 'primary',
```
hollg, line 110: Rather than setting the default value here, could we do the following?

```tsx
Button.defaultProps = {
  theme: 'primary'
} as Partial<ButtonProps>
```

There seems to be a bug in TS where default prop values assigned inline aren't recognised. This probably wouldn't cause a problem in the `Button` component, but it causes build errors for components with required props. Since those components (e.g. `Input`, which has a required `type` prop defaulted to `'text'`) need to use the long form, I think it would be good to use it everywhere so that we're consistent and if a new contributor looks at how we've done it before, they'll find a pattern that definitely works.

The `as Partial<ButtonProps>` casting ensures that we don't default to an invalid prop.

Here's an issue in the TS repo about this problem: https://github.com/microsoft/TypeScript/issues/31247

andoulla: Yes Tom mentioned this yesterday when we talked. will do

src/components/button/Button.tsx
```diff
@@ -0,0 +1,179 @@
...
+  appearance,
+  ...rest
+}: ButtonProps): React.ReactElement => {
+  return <StyledButton theme={theme} appearance={appearance} {...rest} />
+}
+
+Button.defaultProps = {
+  theme: 'primary',
+  appearance: 'solid'
+}
```
hollg, line 179: Can we cast the default props `as Partial<ButtonProps>`? Without it, TypeScript isn't actually checking the types here.

andoulla: ok. I got thrown off text PR so I added it and then removed it. Sorry

hollg:
I can't see in your code where the blue border/outline on `:focus` is coming from. Is that just a browser default? This is probably for a future ticket, but it would be good to use more neutral `:focus` styling that works equally well with all the `Button` `theme`s.

thomasdigby (approved):

andoulla:
> I can't see in your code where the blue border/outline on `:focus` is coming from. Is that just a browser default? This is probably for a future ticket, but it would be good to use more neutral `:focus` styling that works equally well with all the `Button` `theme`s.

It is exactly the same as it is in the current system. It comes from the browser as said. I would rather change it when Chema and Amit are able to give us a better alternative.

PR #33, opened 2021-02-03
Feat: add Text component

hollg:
Adds `Text` component

Screenshot 2021-02-04 at 10 41 06

### DIFF
--- a/src/components/text/Text.stories.mdx
+++ b/src/components/text/Text.stories.mdx
@@ -1,0 +1,25 @@
+import { Canvas, Meta, Story, ArgsTable } from '@storybook/addon-docs/blocks'
+
+import { Text } from '.'
+
+<Meta title="Primitives/Text" component={Text} />
+
+# Text
+
+`Text` is our basic text component. Use it for any non-heading text.
+
+## Sizes
+
+### `sm`
+
+<Text size="sm">
+  Anything smaller than body text. E.g. "forgot your password?" prompts
+</Text>
+
+### `md`
+
+<Text> Body text. Most text in the app will use this. </Text>
+
+### `lg`
+
+<Text size="lg"> Large, non-heading text. E.g. a drop quote in an article</Text>
--- a/src/components/text/Text.stories.mdx
+++ b/src/components/text/Text.stories.mdx
@@ -2,7 +2,7 @@ import { Canvas, Meta, Story, ArgsTable } from '@storybook/addon-docs/blocks'
 
 import { Text } from '.'
 
-<Meta title="Primitives/Text" component={Text} />
+<Meta title="Text" component={Text} />
 
 # Text
 
--- a/src/components/text/Text.test.tsx
+++ b/src/components/text/Text.test.tsx
@@ -8,7 +8,7 @@ describe(`Text component`, () => {
   it('renders some text', async () => {
     const { container } = render(<Text>TEXT</Text>)
 
-    await screen.getByPlaceholderText('TEXT')
+    await screen.getByText('TEXT')
 
     expect(container).toMatchSnapshot()
     expect(await axe(container)).toHaveNoViolations()
--- a/src/components/text/Text.test.tsx
+++ b/src/components/text/Text.test.tsx
@@ -1,0 +1,16 @@
+import { render, screen } from '@testing-library/react'
+import { axe } from 'jest-axe'
+import * as React from 'react'
+
+import { Text } from '.'
+
+describe(`Text component`, () => {
+  it('renders some text', async () => {
+    const { container } = render(<Text>TEXT</Text>)
+
+    await screen.getByPlaceholderText('TEXT')
+
+    expect(container).toMatchSnapshot()
+    expect(await axe(container)).toHaveNoViolations()
+  })
+})
--- a/src/components/text/Text.tsx
+++ b/src/components/text/Text.tsx
@@ -4,7 +4,7 @@ import { StitchesProps, styled } from '~/stitches'
 
 type Override<T, U> = Omit<T, keyof U> & U
 
-const BaseText = styled('p', {
+const StyledText = styled('p', {
   color: '$tonal900',
   fontFamily: 'sans',
   fontWeight: 400,
@@ -31,7 +31,7 @@ const BaseText = styled('p', {
 })
 
 type TextProps = Override<
-  StitchesProps<typeof BaseText>,
+  StitchesProps<typeof StyledText>,
   {
     as: 'p' | 'span'
     size: 'sm' | 'md' | 'lg'
@@ -39,7 +39,7 @@ type TextProps = Override<
 >
 
 export const Text = (props: TextProps): React.ReactElement => {
-  return <BaseText {...props} />
+  return <StyledText {...props} />
 }
 
 Text.defaultProps = {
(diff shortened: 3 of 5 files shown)


src/components/text/Text.tsx
```diff
@@ -0,0 +1,48 @@
...
+        lineHeight: 1.4
+      }
+    }
+  }
+})
+
+type TextProps = Override<
+  StitchesProps<typeof StyledText>,
+  {
+    as: 'p' | 'span'
```
hollg, line 36: @thomasdigby @andoulla do you think we want to limit what elements `Text` can be rendered `as`? (And if so, do you have any suggestions as to which elements should be allowed?)

src/components/text/Text.tsx
```diff
@@ -0,0 +1,48 @@
+import * as React from 'react'
+
+import { StitchesProps, styled } from '~/stitches'
+
+type Override<T, U> = Omit<T, keyof U> & U
+
+const StyledText = styled('p', {
```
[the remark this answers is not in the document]
thomasdigby: Gah, I would probably use `StyledText` but I get your point. Up to you.

src/components/text/Text.tsx
andoulla: I hate answering with a question but do we have specific designs (in the current system) for blockquote and other types of text?

If we do then we should restrict the types in this component as we want people to use the separate components that we will provide for them

If not, then I don't mind leaving this a bit more flexible

src/components/text/Text.tsx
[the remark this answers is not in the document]
hollg: (I'm sorry, I know we keep having long conversation about naming conventions, but) this component isn't the `Text` component + styles. It's `Text` minus everything else. And given that it accepts the `as` prop, it's not _really_ a `p` + styles, either.

I think the point you made this morning about having a consistent convention and the name being an accurate description is a really good one, I'm just not sure how to apply it going forward. 

My original thinking with the `BaseText` pattern was that this component is the foundation of `Text`; `Text` just adds some logic in the form of prop types. I do think that name captures the relationship between this component and `Text`, but can also see that it's less clear what the second part of its name refers to when the component and element share a name.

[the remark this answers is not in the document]
thomasdigby: I'd argue that in this context it is a `p` + styles, and `StyledParagraph` might make more sense. `Text` will consume `StyledParagraph` and the `as` prop but it won't affect the initial definition of `StyledParagraph`

src/components/text/Text.tsx
thomasdigby: We don't really have the concept of `blockquote` or anything else at the moment as we don't really have any long form copy. I think we would create a separate component for it though anyway.

However, I still think we should separate the semantics from the style, this could be a quote that we want to render as a `blockquote` element, but with standard `Text` styles.

src/components/text/Text.tsx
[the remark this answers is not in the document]
hollg: Yeah okay nice, I can definitely get on board with that. Should be easy to apply consistently too. Thanks for the input!

src/components/text/Text.tsx
hollg: Ooh, great question! Yes, we have at least one instance of that in the app, and another in the marketing site. I'll add `blockquote` (and `caption`, `figcaption`, they're the ones that I can think of right now)

hollg: Actually, this list is probably a good starting point: https://developer.mozilla.org/en-US/docs/Web/HTML/Element#text_content

thomasdigby: Haha yeah I just went through React types instead https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/v16/index.d.ts#L2993

src/components/text/Text.tsx
```diff
@@ -0,0 +1,57 @@
+import * as React from 'react'
+
+import { StitchesProps, styled } from '~/stitches'
+
+type Override<T, U> = Omit<T, keyof U> & U
```
thomasdigby, line 5: We can use the utility here I assume?

src/components/text/Text.test.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { axe } from 'jest-axe'
+import * as React from 'react'
+
+import { Text } from '.'
+
+describe(`Text component`, () => {
+  it('renders some text', async () => {
+    const { container } = render(<Text>TEXT</Text>)
+
+    await screen.getByPlaceholderText('TEXT')
```
thomasdigby, line 11: This should be `getByText`

src/components/text/Text.tsx
hollg: Ah yes, I've updated this now. I started work on this branch before the one with `Override` was merged so I copied it in temporarily!

src/components/text/Text.test.tsx
hollg: Huh, yes it should! It seems weird that this wasn't failing ...

src/components/text/Text.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+import { StitchesProps, styled } from '~/stitches'
+import { Override } from '~/utilities/types'
+
+const StyledParagraph = styled('p', {
+  color: '$tonal900',
+  fontFamily: '$sans',
+  fontWeight: 400,
+  margin: 0,
+  maxWidth: '60ch',
```
andoulla, line 11: Why `maxWidth` is 60ch when in the existing system is 100?

hollg: We agreed that we wanted to decrease it. From what I could find onliny, 55  - 65 characters seemed to be the recommended range for readability. This is likely to change as we get design input, but I thought this seemed like a good starting point. Happy to alter it if you have another suggestion though!

src/components/text/Text.tsx
```diff
@@ -0,0 +1,56 @@
...
+type TextProps = Override<
+  StitchesProps<typeof StyledParagraph>,
+  {
+    as:
+      | 'blockquote'
+      | 'caption'
+      | 'dd'
+      | 'dt'
+      | 'figcaption'
+      | 'li'
```
andoulla, line 42: I wondering if it would be better to have the li as not part of this component. As a first time user would expect to find the necessary styling within the `List` or `ListItem` component.

hollg: Oh I absolutely think we should have `List` and `ListItem` components with standard styling but, as with `blockquote`, I don't think that necessarily means we shouldn't be able to create `li`s _without_ that styling.

src/components/text/Text.tsx
andoulla: Thanks, I didn't know

src/components/text/Text.stories.mdx
```diff
@@ -0,0 +1,25 @@
+import { Canvas, Meta, Story, ArgsTable } from '@storybook/addon-docs/blocks'
+
+import { Text } from '.'
+
+<Meta title="Primitives/Text" component={Text} />
```
andoulla, line 5: nitpick, can we skip the Primitives label so everything is all under the same grouping. once we have the categories sorted with can move it under there

hollg: Great catch, thanks!

thomasdigby (approved):

PR #35, opened 2021-02-04
Feature: Export prop types for documentation

thomasdigby:
- Added a script to run `react-docgen-typescript` on prepublish, this will export each components props in a JSON file for us to consume in our docs site. I added an example here for our `Text` component, the `css` and `as` types from `stitches` aren't particularly useful at the moment (potentially a new beta build might improve this) so I'm thinking of visualising these props differently as a link out to the stitches docs instead.
- Changed the `size` preset script to only check the size of the output and not the runtime cost, this should reduce the build/validate time
- Changed our scripts slightly to accomodate this step:
  - `build` is now `build:lib`
  - Added `clean` to remove the `dist` directory
  - Added `prepublishOnly` to run clean, tests, build lib and build docgen

## Additionally
Made some changes to our component structure to greatly improve build time and prop discovery, keen to discuss them in more detail with you both.
- All components are now defined as `React.FC`, it removes TS type inference and really speeds up the build time and VSCode autocomplete time, it also allows our prop defaults to be defined within the destructured prop arguments, _which in turn_ allows react docgen to extract them so we can show them in our `PropsTable`. I'm aware that it introduces a few oddities including allowing `children` when not used, but I think the benefits (we can actually build the thing 😅) outweigh the negatives for now. 
- All components now have `displayName` set, our `stitches-display-name` plugin only runs on `styled()` calls and not on any components that consume those components, so it works on `const Box = styled('div', {})` but not:
  ```
  const StyledDiv = styled('div', {})
  export const Box = (props) => <StyledDiv {...props} />
  ```
  Again this helps with `react-docgen` as it assumes a display name is set so we can now access the component name in docs.
- Updated props that have a default to be optional
- Added `never` to `as` if we don't want it updating, e.g. `Input`

```json5
{
  "tags": {},
  "description": "",
  "displayName": "Text",
  "methods": [],
  "props": {
    "css": {
      "defaultValue": null,
      "description": "",
      "name": "css",
      "declarations": [
        {
          "fileName": "@atom-learning-components/node_modules/@stitches/react/dist/declarations/src/index.d.ts",
          "name": "TypeLiteral"
        }
      ],
      "required": false,
      "type": {
        "name": "enum",
        "raw": "(string & { motion?: (string & {}) | (TFlatCSS<{ tokens: any; utils: { p: (value: CSSValue) => CSSBlob; pt: (value: CSSValue) => CSSBlob; pr: (value: CSSValue) => CSSBlob; ... 11 more ...; bg: (value: string) => CSSBlob; }; breakpoints: { ...; }; }> & TTopUtils<...> & { ...; }) | (TFlatCSS<...> & ... 1 more ... & { ...",
        "value": [
          {
            "value": "undefined"
          },
          {
            "value": "string & { motion?: (string & {}) | (TFlatCSS<{ tokens: any; utils: { p: (value: CSSValue) => CSSBlob; pt: (value: CSSValue) => CSSBlob; pr: (value: CSSValue) => CSSBlob; ... 11 more ...; bg: (value: string) => CSSBlob; }; breakpoints: { ...; }; }> & TTopUtils<...> & { ...; }) | (TFlatCSS<...> & ... 1 more ... & { ...."
          },
          // etc.
        ]
      }
    },
    "as": {
      "defaultValue": {
        "value": "p"
      },
      "description": "",
      "name": "as",
      "declarations": [
        {
          "fileName": "@atom-learning-components/src/components/text/Text.tsx",
          "name": "TypeLiteral"
        }
      ],
      "required": false,
      "type": {
        "name": "enum",
        "raw": "\"blockquote\" | \"caption\" | \"dd\" | \"dt\" | \"figcaption\" | \"li\" | \"p\" | \"span\" | undefined",
        "value": [
          {
            "value": "undefined"
          },
          {
            "value": "\"blockquote\""
          },
          {
            "value": "\"caption\""
          },
          {
            "value": "\"dd\""
          },
          {
            "value": "\"dt\""
          },
          {
            "value": "\"figcaption\""
          },
          {
            "value": "\"li\""
          },
          {
            "value": "\"p\""
          },
          {
            "value": "\"span\""
          }
        ]
      }
    },
    "size": {
      "defaultValue": {
        "value": "md"
      },
      "description": "",
      "name": "size",
      "declarations": [
        {
          "fileName": "@atom-learning-components/src/components/text/Text.tsx",
          "name": "TypeLiteral"
        }
      ],
      "required": false,
      "type": {
        "name": "enum",
        "raw": "\"sm\" | \"md\" | \"lg\" | undefined",
        "value": [
          {
            "value": "undefined"
          },
          {
            "value": "\"sm\""
          },
          {
            "value": "\"md\""
          },
          {
            "value": "\"lg\""
          }
        ]
      }
    }
  }
}

```


scripts/generate-component-props.js
```diff
@@ -0,0 +1,34 @@
...
+  shouldExtractLiteralValuesFromEnum: true,
+  propFilter: filterReactTypes
+})
+
+const run = async () => {
+  try {
+    const componentFilePaths = await glob.sync(`${__dirname}/../src/**/*.tsx`)
+    const ouput = componentFilePaths
+      .filter((path) => !path.includes('test.tsx'))
+      .filter((path) => !path.includes('stories'))
```
andoulla, line 25: We haven't applied this yet but in the contribution doc we discuss how this should `.docs` so it is agnostic of the tool we are using and to ease switching from storybook to the new documentation style. 

For that, I would suggest checking for both, for now until we eliminate `.stories.mdx` from everywhere

src/components/select/Select.tsx
```diff
@@ -16,7 +16,10 @@ export type Option = {
 
 export type SelectProps = StitchesProps<typeof StyledSelect> & {
   name: string
-  options: Array<Option>
```
andoulla, line 19: Should the type `Option` can be removed from line 12?

thomasdigby: Yeah for sure, although we should probably delete the Select component anyway as it was a test, it somehow stuck around into our production component set 😬

scripts/generate-component-props.js
thomasdigby: This is to ensure that we get props from the main component file and not from any test `.tsx` files, this PR doesn't touch anything `.mdx` related.

Agreed on your first point though, we should be clear about the process involved in outputting these props, how you can write components to enable better parsing of the component and document how to create the correct doc format for our documentation site.

andoulla: Sorry, I thought it was filtering and selecting the stories not excluding them.  it still should exclude the `.docs` pattern to make it future proof

src/components/select/Select.tsx
andoulla: it is because we really like the colours used there! :)

scripts/generate-component-props.js
thomasdigby: Sorry I wasn't clear, this _isn't_ related to any documentation formats, `.md` or `.mdx`, this is just to ensure that when we go and find our components in `src` we get the correct ones to create the props for.

For his `Box` PR, Gary created a `CardExample.tsx` within the `stories` directory which our `react-docgen` is picking up as a component to parse and find props for, which we want to avoid.

src/components/button/Button.tsx
```diff
@@ -82,98 +82,64 @@ const StyledButton = styled('button', {
...
   { theme: 'primary', appearance: 'solid' },
   getButtonSolidVariant('$primary500', '$primary900')
 )
-
 StyledButton.compoundVariant(
-  {
-    theme: 'secondary',
-    appearance: 'solid'
-  },
+  { theme: 'secondary', appearance: 'solid' },
```
andoulla, line 91: I'm wondering why we are getting these whitespaces changes. We have the same linting settings..

thomasdigby: Sorry, I manually updated these just to reduce the height of them all, should have added to the description!

andoulla: Oh I thought my linter wasn't working properly as it just let it handle these completely

andoulla:
I think it is ok, but I would definitely want @hollg to have a chance to review it on Monday before merging

andoulla (approved):

thomasdigby:
> I think it is ok, but I would definitely want @hollg to have a chance to review it on Monday before merging

For sure, no rush on this one. 👍

src/components/input/Input.tsx
```diff
@@ -35,27 +35,19 @@ const StyledInput = styled('input', {
 type InputProps = Override<
   StitchesProps<typeof StyledInput>,
   {
+    as: never
```
hollg, line 38: This is cool!

hollg (approved):

PR #36, opened 2021-02-04
Basic Link component

andoulla:
It returns a simple link


src/components/link/Link.test.tsx
```diff
@@ -0,0 +1,18 @@
+import { render, screen } from '@testing-library/react'
+import { axe } from 'jest-axe'
+import React from 'react'
+
+import { Link } from './'
+
+describe.only(`Link component`, () => {
```
andoulla, line 7:
```suggestion
describe(`Link component`, () => {
```

src/components/link/Link.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+export const Link = styled('a', {
+  color: '$primary500',
+  fontFamily: '$sans',
+  fontSize: '$sm',
+  textDecoration: 'none',
+  cursor: 'pointer',
+  ':visited': { color: '$primary500' },
+  ':focus': { color: '$primary900', textDecoration: 'underline' },
+  ':hover': { color: '$primary900', textDecoration: 'underline' },
```
thomasdigby, line 11: We can combine these two rules right? `':hover,:focus': { color: '$primary900', textDecoration: 'underline' }`?

src/components/link/Link.tsx
```diff
@@ -0,0 +1,13 @@
+import { styled } from '../../stitches'
+
+export const Link = styled('a', {
+  color: '$primary500',
+  fontFamily: '$sans',
+  fontSize: '$sm',
```
thomasdigby, line 6: We should potentially create variants for size, like we have in the `Text` component

andoulla: I was wondering about that. Happy to add it. 
Should the default be `$sm` like it is in the Login form or should it be `$md` which is the base and then the login form can override it?

src/components/link/Link.tsx
andoulla: yes, we should thanks.

src/components/link/Link.tsx
thomasdigby: I think we should default to `$md`, and then the login form can override it. I think it keeps it consistent with other components defaults.

src/components/link/Link.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+import { Link } from './'
+
+describe(`Link component`, () => {
+  it('renders an anchor', async () => {
+    const { container } = render(<Link href="https://google.com/">GOOGLE</Link>)
+
+    const renderedLink = (await screen.getByText('GOOGLE')) as HTMLAnchorElement
+
+    expect(renderedLink.href).toBe('https://google.com/')
```
hollg, line 13: We should be able to avoid avoid casting to `HTMLAnchorElement` by asserting `expect(renderedLink).toHaveAttribute('href', 'https://google.com/')`.

We should get better error messages for the more specific assertsions, too: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-the-wrong-assertion

src/components/link/Link.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+    expect(renderedLink.text).toBe('GOOGLE')
+    expect(container).toMatchSnapshot()
+    expect(await axe(container)).toHaveNoViolations()
+  })
+
+  it('renders an large anchor', async () => {
```
hollg, line 19: How much value are we getting from this test? We don't seem to be asserting anything specific to the `size="lg"` (and I don't think we've been testing different sizes for other components).

src/components/link/Link.tsx
```diff
@@ -0,0 +1,48 @@
...
+      lg: {
+        fontSize: '$lg',
+        letterSpacing: '0.02em',
+        lineHeight: 1.4
+      }
+    }
+  }
+})
+
+type LinkProps = Override<
```
hollg, line 35: Do we need to use `Override` here? Won't `StitchesProps<typeof StyledLink>` already include the `size` field that we want?

andoulla: This actually solves the issue was I trying to debug with the CI too. Thanks

src/components/link/Link.test.tsx
andoulla: Thanks I will have a read

src/components/link/Link.test.tsx
andoulla: I wanted to follow the pattern we agreed, that we should have a test for each group of variants. For instance, in the button, we have one test for the solid appearance and one for the outline. 

I was going to add one for sizes once we added different sizes to it.

I did it as a snapshot as i don't think doing a test that checks whether the prop I'm passing is applied is valuable

thomasdigby (approved):

thomasdigby (approved):

PR #37, opened 2021-02-05
docs: removed unnecessary stories folders for single mdx files

andoulla:
Have removed the unnecessary folders and updated the contribution doc to explain the structure as we agreed on it

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,7 +17,7 @@ src/
 	    index.ts
 ```
 
-**Note:** Documentations is nested in a folder only if there are more than 1 files. If there is only 1 file then it should be in the component folder.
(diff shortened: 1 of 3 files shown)


thomasdigby (approved):

CONTRIBUTING.md
```diff
@@ -17,6 +17,8 @@ src/
 	    index.ts
 ```
 
+**Note:** Documentations is nested in a folder only if there are more than 1 files. If there is only 1 file then it should be in the component folder.
```
andoulla, line 20:
```suggestion
**Note:** Documentation is nested in a folder only if there are more than 1 files. If there is only 1 file then it should be in the component folder.
```

PR #38, opened 2021-02-08
Feat: add heading component

hollg:
Adds basic `Heading` component.

## Screenshots
Screenshot 2021-02-08 at 10 44 19


src/components/heading/Heading.tsx
```diff
@@ -0,0 +1,35 @@
...
+>
+
+export const Heading: React.FC<HeadingProps> = (props) => {
+  return <StyledHeading {...props} />
+}
+
+Heading.defaultProps = {
+  as: 'h1',
+  size: 'lg'
+} as Partial<HeadingProps>
```
thomasdigby, line 35: Can we add these to destructured props? Not sure whether you saw the explanation I added on my previous PR but we're able to switch to that now and TS is happy with it 👍

src/components/heading/Heading.tsx
```diff
@@ -0,0 +1,35 @@
...
+    size?: 'sm' | 'md' | 'lg'
+  }
+>
+
+export const Heading: React.FC<HeadingProps> = (props) => {
+  return <StyledHeading {...props} />
+}
+
+Heading.defaultProps = {
+  as: 'h1',
```
thomasdigby, line 33: Could we default this to `h2`? Just in case people don't override, I think it's best to avoid `h1`

src/components/heading/Heading.tsx
```diff
@@ -0,0 +1,35 @@
...
+const StyledHeading = styled('h1', {
+  color: '$tonal-900',
+  fontFamily: '$sans',
+  fontWeight: '700',
+  lineHeight: '1.4',
+  variants: {
+    size: {
+      sm: { fontSize: '16px', letterSpacing: '0.01em' },
+      md: { fontSize: '28px', letterSpacing: '0.02em' },
+      lg: { fontSize: '37px', letterSpacing: '0.02em' }
```
thomasdigby: Can we use font size scale values here?

src/components/heading/Heading.tsx
hollg: Oh no I didn't see that. Nice! What made it work?

thomasdigby: `React.FC` I think

src/components/heading/Heading.tsx
hollg: Good call. I'll change the default `size` value to `md` too then as those two are more likely to used together.

src/components/heading/Heading.tsx
hollg: That was my first guess, but it seems to work when I remove that too. Can't rule out the possibility that I was just doing something totally wrong when I said it didn't work before!!

src/components/heading/Heading.tsx
```diff
@@ -0,0 +1,32 @@
...
+    as?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6'
+    size?: 'sm' | 'md' | 'lg'
+  }
+>
+
+export const Heading: React.FC<HeadingProps> = ({
+  as = 'h2',
+  size = 'md',
+  ...props
+}) => <StyledHeading as={as} size={size} {...props} />
```
andoulla, line 32: @thomasdigby Do we need to add back the `displayName` property?

It was reintroduced in your MR

thomasdigby (commented):
We're missing an export from `components/index.ts`

thomasdigby (approved):

PR #39, opened 2021-02-08
feat: Label component

andoulla:
In this PR:
- Added basic label component with 3 sizes
- Includes test and documentation
- switched deprecated rule `jsx-a11y/label-has-for` to new one `jsx-a11y/label-has-associated-control`


src/components/label/Label.tsx
```diff
@@ -0,0 +1,42 @@
...
+      md: { fontSize: '$md' },
+      lg: { fontSize: '$lg' }
+    }
+  }
+})
+
+type LabelProps = Override<
+  StitchesProps<typeof StyledLabel>,
+  {
+    as?: 'label'
```
thomasdigby, line 26: We should use `as: never` to prevent anyone overwriting the correct element

src/components/label/Label.tsx
```diff
@@ -0,0 +1,42 @@
...
+    size?: 'sm' | 'md' | 'lg'
+  }
+>
+
+export const Label: React.FC<LabelProps> = ({
+  size = 'md',
+  htmlFor,
```
thomasdigby, line 33: Not sure if this is useful or not, I would write this as `for` to hide the React specific method of linking input & label and generally because I think `<Label for="" />` would be a nicer API

src/components/label/Label.tsx
```diff
@@ -0,0 +1,42 @@
...
+  fontFamily: '$sans',
+  fontWeight: 500,
+  maxWidth: '60ch',
+  letterSpacing: '0.02em',
+  lineHeight: '1.4',
+  variants: {
+    size: {
+      sm: { fontSize: '$sm' },
+      md: { fontSize: '$md' },
+      lg: { fontSize: '$lg' }
```
thomasdigby, line 18: We don't need a `lg` label, only `md` and `sm`, unless you've seen a place where it's used like that?

src/components/label/Label.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+import { StitchesProps, styled } from '~/stitches'
+import { Override } from '~/utilities/types'
+
+const StyledLabel = styled('label', {
+  color: '$secondary300',
+  display: 'inline-block',
+  fontFamily: '$sans',
+  fontWeight: 500,
+  maxWidth: '60ch',
```
thomasdigby, line 11: I think we can lose the `maxWidth` and `display` properties on this, they shouldn't be necessary

src/components/label/Label.stories.mdx
```diff
@@ -0,0 +1,29 @@
+import { Canvas, Meta, Story, ArgsTable } from '@storybook/addon-docs/blocks'
+
+import { Label } from '.'
+
+<Meta title="Label" component={Label} />
+
+# Label
+
+`Label` is a thin wrapper around the basic HTML `<label/>` element. It adds default styling and the `css` prop. Use it to compose more complex field components and form related component.
```
thomasdigby, line 9: It might be worth adding details about our `Field` components where we combine `Input` and `Label`s, when we get around to building those.

src/components/label/Label.tsx
andoulla: It is implemented in the current codebase, but I don't know if it is used anywhere.

thomasdigby: Looks like we only use `sm` and `md` so lets stick with them for now!

src/components/label/Label.tsx
andoulla: I can switch it but then I will need to add the for prop in the `LabelProps` type. Also, does this mean that then I need to set `htmlFor: never` ?

src/components/label/Label.tsx
andoulla: I was just looking it up too. Ok I will remove

src/components/label/Label.tsx
thomasdigby: Yeah good point, maybe it's worth leaving as is, we can remove it from our destructured props though as it's getting passed through `{...rest}`

src/components/label/Label.stories.mdx
andoulla: What about this now?

src/components/label/Label.stories.mdx
```diff
@@ -0,0 +1,29 @@
...
+
+For accessibility purposes the component needs either to have the `htmlFor` prop set to link it to the control it is associated with or to wrap around the component that it is labelling.
+
+## Sizes
+
+The default size is `md`.
+
+### `sm`
+
+<Label size="sm" for="controlId">
```
thomasdigby, line 23:
```suggestion
<Label size="sm" htmlFor="controlId">
```

thomasdigby (approved):

PR #40, opened 2021-02-08
Fix: Input snapshot

thomasdigby:
Unsure why this snapshot kept failing so I've updated it here and it should 🤞 fix ours tests on incoming PRs

### DIFF
--- a/src/components/input/__snapshots__/Input.test.tsx.snap
+++ b/src/components/input/__snapshots__/Input.test.tsx.snap
@@ -93,20 +93,20 @@ exports[`Input component renders a number input 1`] = `
   box-shadow: none;
 }
 
-._cmjVcV {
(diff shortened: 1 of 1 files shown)


src/components/input/__snapshots__/Input.test.tsx.snap
```diff
@@ -93,20 +93,20 @@ exports[`Input component renders a number input 1`] = `
   box-shadow: none;
 }
 
-._cmjVcV {
-  border-bottom-right-radius: var(--radii-1);
+._gpXcaM {
+  border-bottom-right-radius: var(--radii-0);
```
andoulla, line 97: it seems that the radius is changing from `$1` to `$0`. According to the styling, it should be 0 so I wonder why it was passing in the first place

andoulla (approved):

PR #41, opened 2021-02-09
feat: add CSSWrapper

hollg:
Adds `CSSWrapper` component.


thomasdigby (approved):
Need to remove the DSstore file, also do we want a root export of utilities?

src/utilities/css-wrapper/CSSWrapper.stories.mdx
```diff
@@ -0,0 +1,21 @@
+import { Meta } from '@storybook/addon-docs/blocks'
+
+import { CSSWrapper } from '../'
+
+<Meta title="Util/CSSWrapper" component={CSSWrapper} />
+
+# CSSWrapper
+
+If `CSSWrapper` is passed a truthy `css` prop, it renders its children wrapped in a `Box` with that `css`. If `css` is falsy, `CSSWrapper`
```
andoulla, line 9: What is the main difference of using the `CSSWrapper` vs using the `Box`?

src/utilities/css-wrapper/CSSWrapper.tsx
```diff
@@ -0,0 +1,21 @@
...
+export const CSSWrapper = ({
+  css,
+  children
+}: CssWrapperProps): React.ReactElement =>
+  css ? (
+    <Box css={css} data-testid="css-wrapper">
+      {children}
+    </Box>
+  ) : (
+    <>{children}</> // eslint-disable-line
```
andoulla, line 20: Could we jot down what rule are we disabling here, for future reference?

src/utilities/css-wrapper/CSSWrapper.stories.mdx
hollg: The `CSSWrapper` doesn't render anything if it isn't passed a `css` prop, so it offers the option to wrap other components with some styling, but doesn't pollute the DOM when we don't want to.

src/utilities/css-wrapper/CSSWrapper.tsx
hollg: Good call! I've disabled the specific relevant rule instead of everything, and added a comment explaining the reason.

PR #42, opened 2021-02-10
V1 primitives select

andoulla:
Implements `Select` with the following options:
- disabled state
- disabled options
- default option

**Notes:**
- I destructure the aria-label value to address some of the accessibility issues that arise, however, I'm not sure this is the right way to solve this.
- The current styling might have some contrast issues but we because we are using a background-image for the chevron the `axe` library can't evaluate

**Screenshots:**
image
image
image


src/components/select/Select.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+  })
+  it('renders a select with 3 options', async () => {
+    const { container } = render(
+      <Select ariaLabel="dropdown" options={mockOptions} />
+    )
+
+    const options = await screen.getAllByRole('option')
+
+    expect(options.length).toEqual(3)
+    options.forEach((option, i) => {
```
andoulla, line 31: Not sure if using a for loop is a good practice, however, it makes the test shorter with less code duplication. Any preferences/thoughts?

src/components/select/Select.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { StitchesProps, styled } from '~/stitches'
+import { Override } from '~/utilities/types'
+
+// TODO: missing on hover? current system doesn't apply one
+const StyledSelect = styled('select', {
+  appearance: 'none',
+  backgroundColor: 'white',
+  backgroundImage:
+    'url("data:image/svg+xml;charset=US-ASCII,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20width%3D%22292.4%22%20height%3D%22292.4%22%3E%0A%20%20%3Cpath%20fill%3D%22%230f65b2%22%20d%3D%22M287%2069a18%2018%200%2000-13-5H18c-5%200-9%202-12%205a18%2018%200%2000-6%2013c0%205%202%2010%205%2013l128%20128c4%204%208%205%2013%205s9-1%2013-5L287%2095c4-3%205-8%205-13s-2-9-5-13z%22%2F%3E%0A%3C%2Fsvg%3E")',
+  backgroundPosition: `right 16px top 50%, 0 0`, // TODO: layout-scale $3 doesn't work
```
andoulla, line 12: for some reason, it can't match the available design tokens for spaces. Need a workaround for complex css props

src/components/select/Select.tsx
```diff
@@ -0,0 +1,106 @@
...
+  '&::-ms-expand': {
+    display: 'none'
+  },
+  '&:disabled, > option:disabled': {
+    opacity: '0.7',
+    backgroundColor: '$tonal300',
+    cursor: 'not-allowed'
+  },
+  variants: {
+    size: {
```
hollg, line 41: So far haven't we been leaving widths up to the user, via the `css` prop? These set widths feel a bit arbitrary, and we don't want to do anything that encourages  people to create a `Select` that isn't wide enough to displays its content properly.

src/components/select/Select.test.tsx
hollg: 🤷🏻‍♂️ I like it!

src/components/select/Select.tsx
andoulla: We have left them out in `Button` but it was always the intention to add them as a follow-up - as it is listed there. Maybe buttons are a different case as you definitely need to have different sizes on those. 

Happy to remove them if we thing they should be taken out

src/components/select/Select.test.tsx
andoulla: Ok, I will keep it. just because it is a very simple use case. I wouldn't do for loops for anything more complicated than this

src/components/select/Select.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+  ]
+  it('renders select with no options', async () => {
+    const { container } = render(<Select ariaLabel="dropdown" />)
+    const select = screen.getByRole('combobox')
+
+    expect(select).toBeInTheDocument()
+    expect(select.childNodes.length).toEqual(0)
+
+    expect(container).toMatchSnapshot()
+    expect(await axe(container)).toHaveNoViolations()
```
hollg, line 21: What was the solution to getting this to pass despite not having a label for the select in the test?

src/components/select/Select.tsx
```diff
@@ -0,0 +1,106 @@
...
+    ariaLabel: string
+  }
+>
+
+export const Select: React.FC<SelectProps> = ({
+  options,
+  disabled,
+  defaultOption,
+  size = 'full',
+  ariaLabel,
```
andoulla, line 83: Not happy about having to do this, however, I don't know how to enforce a type where one of the following three is provided (id, name or aria-label).

src/components/select/Select.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+type Option = {
+  value: string
+  label: string
+  disabled?: boolean
+}
+type SelectProps = Override<
+  StitchesProps<typeof StyledSelect>,
+  {
+    options: Array<Option>
```
thomasdigby, line 72: Can we inline `Option` here? Our docgen script can't handle multiple types and will display the prop as `Option` instead of the object properties.

```ts
type SelectProps = Override<
  StitchesProps<typeof StyledSelect>,
  {
    options: {
      value: string
      label: string
      disabled?: boolean
    }[]
    defaultOption?: string
    ariaLabel: string
  }
>
```

src/components/select/Select.tsx
thomasdigby, line 83: This is unnecessary if you use a linked `label` element, perhaps we can require that either an `id` or an `aria-label` is passed? That way we're requiring at least one of those conventions.

src/components/select/Select.tsx
```diff
@@ -0,0 +1,106 @@
...
+    defaultOption?: string
+    ariaLabel: string
+  }
+>
+
+export const Select: React.FC<SelectProps> = ({
+  options,
+  disabled,
```
thomasdigby, line 80: Do we need to include a prop here if we're just passing it directly to the element and it doesn't have a default value?

src/components/select/Select.test.tsx
thomasdigby: I think this is fine 👍

src/components/select/Select.tsx
hollg: If I've correctly understood what you want to do, I think this SO answer explains it well: https://stackoverflow.com/a/60617060

But what's the reason for wanting to do that here? In the existing form components in the main app, we just take a `name` and use that as the `id`. Off the top of my head, I think the situations where the `select` needs an `id` are the same as where it needs a `name` (i.e. when it's in a form, which will be most use cases). Would that work here?

src/components/select/Select.test.tsx
andoulla: if there isn't an associated label so there is no `id` or `name`  provide it with, the aria-label can compensate for it. this is why I created the `ariaLabel` prop which is passed down. It is not a great solution to be honest

src/components/select/Select.tsx
andoulla: I will remove them then

src/components/select/Select.stories.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+import { disableStitchesProps } from '~/utilities/storybook'
+
+<Meta title="Select" component={Select} />
+
+# Select
+
+The `Select` component is a very thin wrapper around an HTML `select` element. It adds default styling and the `css` prop.
+
+`Select` need to be associated with a label for accessibility purposes, however before using this component consider using one of the provided `Field` components that are available. Use this `Select` to compose more complex `Field` type components.
```
hollg, line 20: Getting very picky now, but this doesn't read quite right to me.

```suggestion
A `Select` needs to be associated with a label for accessibility purposes, so rather than using the `Select` component directly in a UI, consider using a `SelectField`, which provides a `Label` and displays validation errors. Use this `Select` to compose more complex `Field` type components.
```

andoulla: Not picky at all. thank you, it needs to make sense to everyone, not just me

src/components/select/Select.tsx
andoulla: Don't we need it for the documents, it is explicitly mentioned in the props table? If I understood it wrong then it can go

thomasdigby: We could, but this is just one of a large number of standard attributes that can be added to HTML elements, I think it's best to not include these at all and instead mention separately in the docs that any extra props will be spread onto the underlying HTML element.

andoulla: ok, I will remove it

src/components/select/Select.tsx
andoulla: yes, I think like 95% of the cases there will be a label associated with the `select` so the `name` will be passed down to be used as an `id`. 
You would only pass an `aria-label` prop to it if there isn't a label to associate it with.

I'm just trying to figure out how to make it ask either for an `id` or an `aria-label` as without those all the accessibility tests will fail, and they should fail.

src/components/select/Select.tsx
```diff
@@ -0,0 +1,76 @@
...
+type SelectProps = Override<
+  StitchesProps<typeof StyledSelect>,
+  {
+    options?: {
+      value: string
+      label: string
+      disabled?: boolean
+    }[]
+    defaultOption?: string
+  } & ({ id: string } | { 'aria-label': string })
```
thomasdigby, line 53: Would this also disallow one if you have the other? And both at the same time?

src/components/select/Select.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+import { StitchesProps, styled } from '~/stitches'
+import { Override } from '~/utilities/types'
+
+// TODO: missing on hover? current system doesn't apply one
+const StyledSelect = styled('select', {
+  appearance: 'none',
+  backgroundColor: 'white',
+  backgroundImage:
+    'url("data:image/svg+xml;charset=US-ASCII,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20width%3D%22292.4%22%20height%3D%22292.4%22%3E%0A%20%20%3Cpath%20fill%3D%22%230f65b2%22%20d%3D%22M287%2069a18%2018%200%2000-13-5H18c-5%200-9%202-12%205a18%2018%200%2000-6%2013c0%205%202%2010%205%2013l128%20128c4%204%208%205%2013%205s9-1%2013-5L287%2095c4-3%205-8%205-13s-2-9-5-13z%22%2F%3E%0A%3C%2Fsvg%3E")',
```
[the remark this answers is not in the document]
thomasdigby: We can pop this in for a future iteration, would be good to think about globally as a util rather than specific to this component 👍

thomasdigby (approved):

src/components/select/Select.tsx
andoulla: This is an `inclusive OR` according to the [article](https://stackoverflow.com/questions/40510611/typescript-interface-require-one-of-two-properties-to-exist/60617060#60617060) which means that both could be true at the same time. if we wanted to make it disallow the use of the other then we need to switch to an `exclusive or`. I don't see the need to block the user from passing both, even if it is redundant.

thomasdigby: I think it could introduce duplicated announcements on screenreaders, would need to test but I think it's better for us to allow only one. As we can do that with Typescript I don't see any reason not to add it.

andoulla: I didn't know that. I thought that the screen readers would opt for the `aria-*` props if they were present.

src/components/select/Select.tsx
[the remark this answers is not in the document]
andoulla: I will separate out the `svg` in a local function and then add a todo to iterate and separate it out. I'm hoping that after we have icons ticket done we might be able to utilise some of that work into writing the utils or even not need the util maybe.

thomasdigby (approved):

PR #43, opened 2021-02-10
Feat: add checkbox component

hollg:
Adds `Checkbox` component.

Screen cap showing default/hover/focus/checked styling:
https://user-images.githubusercontent.com/21319237/107526754-15daa980-6bb0-11eb-8273-f78f5247fb08.mov


src/components/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { CheckIcon } from '@radix-ui/react-icons'
+import * as React from 'react'
+
+import { StitchesProps, styled } from '~/stitches'
+
+const StyledCheckbox = styled(RadixCheckbox.Root, {
+  appearance: 'none',
+  backgroundColor: 'transparent',
+  border: 'none',
+  borderRadius: 2,
```
thomasdigby, line 11: Can we use a token? `$0`? I think that's like 3px

src/components/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,48 @@
...
+  boxShadow: 'inset 0 0 0 2px $secondary300',
+  color: 'white',
+  cursor: 'pointer',
+  padding: 0,
```
thomasdigby, line 15: Could `all: unset` work here? I wonder if we're able to avoid multiple resetting CSS declarations, are there any defaults that we want to keep?

src/components/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,48 @@
+import * as RadixCheckbox from '@radix-ui/react-checkbox'
+import { CheckIcon } from '@radix-ui/react-icons'
```
thomasdigby, line 2: As we need to talk icons separately I won't flag this, but we'll need to look into it as this currently doesn't treeshake and includes every icon.

thomasdigby (commented):
Very excited to jump on the Radix UI band wagon. 👍

andoulla (approved):

src/components/checkbox/Checkbox.tsx
hollg: I started with `$0` but changes it because at this size that border radius is...significant. I think that is actually what we have in the app currently though so happy to switch back to it.

For reference:
Screenshot 2021-02-11 at 12 18 17

thomasdigby: Yeah fair, could potentially tweak that value in the theme but best to just match the current app style for now.

src/components/checkbox/Checkbox.tsx
hollg: Oh yeah sorry I meant to bring this up - decisions about icons seemed like a matter for Later™ so I stuck with the Radix example for now. Definitely needs revisiting as soon as we do make those decisions.

thomasdigby (approved):

PR #45, opened 2021-02-12
Feat: add image component

hollg:
Adds `Image` component with the CSS from this article applied: https://www.zachleat.com/web/fluid-images/


src/components/image/Image.tsx
```diff
@@ -0,0 +1,17 @@
+import { styled } from '~/stitches'
+
+// default styling lifted from https://www.zachleat.com/web/fluid-images/#copy-and-paste
+export const Image = styled('img', {
```
thomasdigby, line 4: Does it render with an extra space underneath? May need to add a `display: inline-block` or `vertical-align: middle` to counter that.

src/components/image/Image.tsx
```diff
@@ -0,0 +1,17 @@
...
+  },
+  '&[width][height]': {
+    height: 'auto'
+  },
+  '&[src$=".svg"]': {
+    width: '100%',
+    height: 'auto',
+    maxWidth: 'none'
+  }
+})
```
thomasdigby, line 17: Can we add a `displayName` property

hollg: Oh, I've no idea why, but I assumed the `styled` components got it automatically! Adding now, thanks!

thomasdigby: Ah you're right they do. I reckon it's good to be consistent otherwise some components that just export `styled` components won't have a displayname and others that export a component wrapping a styled component will.

thomasdigby:
@hollg Need to update the snapshot 👍

src/components/image/Image.tsx
hollg: Oh yeah for sure, 100% worth making sure that a contributor can use any of our components as a guide for their own and get the same result 👍🏻

thomasdigby (approved):

PR #46, opened 2021-02-13
Feature: Documentation compatibility and improvements

thomasdigby:
- Update all docs to:
  - Remove `stories` from file name
  - Add frontmatter and replace storybook metadata
  - Update story examples to codeblocks
- Removed all storybook packages & references
- Created a script to copy all `.md`/`.mdx` files to dist so we can include in our docs
- Remove custom code from sandbox to match login demo


CONTRIBUTING.md
```diff
@@ -32,9 +32,9 @@ All components should accept a `css` prop. Single-child components are styled di
 - If the component renders two or more sibling elements/components, wrap them in a `CSSWrapper` and pass the `css` prop to that.
 - Otherwise, pass the `css` prop value directly to the direct child.
 
-```tsx
+```jsx
```
hollg, line 35: Why `jsx` and not `tsx`?

CONTRIBUTING.md
```diff
@@ -58,6 +58,49 @@ Each component should have documentation which covers the following:
 - When it should be used
 - When it _shouldn't_ be used (e.g. instead of using an `Input` directly, we'll often want to use an `InputField`)
 
+We use front matter to add metadata to our documentation, the available fields are:
```
hollg, line 61: Is it worth mentioning that we're using YAML syntax for the frontmatter? YAML, TOML and JSON are all pretty popular so I think it might be worth specifying

CONTRIBUTING.md
```diff
@@ -1,4 +1,7 @@
-# Contribution Guide
+---
+title: Contributing
+description: When contributing to this repository, use the following information, to uphold the standards we have set for this project. Please read through this and the `README.md` before starting.
+---
 
 When contributing to this repository, use the following information, to uphold the standards we have set for this project. Please read through this and the `README.md` before starting.
```
andoulla, line 6: Do we need to have this sentence twice

README.md
```diff
@@ -43,16 +43,14 @@ _Note:_ Once a branch gets merged into `main`, it is automatically published, th
...
 
 The design system is aiming to adhere to strict accessibility standards with AA as the minimum. To enable this, `eslint-plugin-jsx-a11y` is set to strict mode.
-Different tools are used to validate the components as each serves a different purpose. The a11y storybook addon will validate all the stories and explain each violation, whereas the accessibility unit tests will fail the pipeline and prevent releasing inaccessible components. **_However, these tests are only valid with the context they are testing on, so they can't guarantee 100% that a component is fully accessible._** Therefore, please add stories for all the variations a component has. Also, there is a need for manual testing and design validation.
+Different tools are used to validate the components as each serves a different purpose. The accessibility unit tests will fail the pipeline and prevent releasing inaccessible components. **_However, these tests are only valid with the context they are testing on, so they can't guarantee 100% that a component is fully accessible._** Therefore, please add stories for all the variations a component has. Also, there is a need for manual testing and design validation.
```
andoulla, line 46:
```suggestion
Different tools are used to validate the components as each serves a different purpose. The accessibility unit tests will fail the pipeline and prevent releasing inaccessible components. **_However, these tests are only valid with the context they are testing on, so they can't guarantee 100% that a component is fully accessible._** Therefore, manual testing is required along with design validation.
```

src/components/box/Box.mdx
```diff
@@ -0,0 +1,57 @@
...
+      Here is a cute cat
+    </Text>
+  </Box>
+  <Box as="footer" css={{ p: '$3' }}>
+    <Text>Remember: cats have whiskers!</Text>
+  </Box>
+</Box>
+```
+
+`Box` falls back to rendering a `div`, but we can use it to render any other container element with e.g. `<Box as="section">`. In the card, `Box` is used to render a `section`, `header` and `footer`.
```
andoulla, line 39: > In the card, `Box` is used to render a `section`, `header` and `footer`.

Do we need the above, especially when we have been discussing that we would probably not add the card in the DS at all?

src/components/flex/Flex.mdx
```diff
@@ -0,0 +1,13 @@
+---
+title: Flex
+component: Flex
+---
```
andoulla, line 4: Should we add a description here, just to be consistent with the other files?

src/components/select/Select.mdx
```diff
@@ -0,0 +1,67 @@
...
+    { value: 1, label: 'Apples' },
+    { value: 2, label: 'Bananas' }
+  ]}
+  css={{ width: 300 }}
+/>
+```
+
+## Accessibility
+
+For accessibility purposes the component needs to have the `id` prop set, to link it to the label it is associated with. If those are not available, please add an `aria-label` to ensure that the component remains accessible
```
andoulla, line 23:
```suggestion
For accessibility purposes the component needs to have the `id` prop set, to link it to the label it is associated with. If a label is not available, please add an `aria-label` to ensure that the component remains accessible
```

CONTRIBUTING.md
thomasdigby: Just to keep to it consistent, we can go the other way and have them all be `tsx` but that might seem like these components are only usable in a typescript environment.

CONTRIBUTING.md
thomasdigby: Yes it is, although the fields themselves are going to be standardised, it's worth knowing the syntax to write in.

src/components/box/Box.mdx
thomasdigby: This example is for the user creating the card themselves, it's important that we show these examples as we're providing that ability for the to use `<Box />` in this way

src/components/flex/Flex.mdx
thomasdigby: Yeah for sure, not too sure what to write though

src/components/box/Box.mdx
andoulla: Oh, I read it like it was referring to an existing card component. Maybe switch it to something like this:

For example, a card component will use `Box` to render a `section`, `header` and `footer`.

CONTRIBUTING.md
andoulla:
```suggestion

```

src/components/flex/Flex.mdx
thomasdigby: Going to leave this until we do a complete sweep of the docs for consistency and content.

scripts/generate-component-props.mjs
```diff
@@ -18,11 +18,18 @@ const tsConfigParser = docgen.withCustomConfig('./tsconfig.json', {
 
 const run = async () => {
   try {
-    const componentFilePaths = await glob.sync('src/**/*.tsx')
-    const ouput = componentFilePaths
+    const componentFilePaths = await glob
+      .sync('src/**/*.tsx')
       .filter((path) => !path.includes('test.tsx'))
       .filter((path) => !path.includes('stories'))
```
andoulla, line 24: Now that stories are removed, we don't the exclusion for stories right?

andoulla (approved):
Just one small question, but it looks ready

hollg (approved):

PR #47, opened 2021-02-15
Feature: Upgrade stitches to beta

thomasdigby:
- Followed the migration guide to upgrade to the beta version of stitches.
- Had to fork and publish a patch to `jest-stitches` to remove the global `:root` selector from our snapshots
- Included `@types/react` as they were missing for some reason, must have been included as a dependency of another package that we removed
- Included `resize-observer-browser` types to fix Radix UI type issue

https://stitches-site-git-beta.modulz-deploys.com/blog/migrating-from-alpha-to-beta

https://github.com/jaredLunde/jest-stitches/issues/1


src/components/box/__snapshots__/Box.test.tsx.snap
```diff
@@ -1,33 +1,90 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Box component renders 1`] = `
-._Fgykb {
-  margin-top: auto;
+:root {
+  --colors--tonal100: hsl(0, 6%, 97%);
```
andoulla, line 5: Is this correct? is it going to import the theme in every snapshot?

This means that if we add a new theme colour it will break the snapshot even if that colour isn't used in the component

thomasdigby: It's an issue with `jest-stitches`, I've opened an issue and linked it in the description

src/components/checkbox/__snapshots__/Checkbox.test.tsx.snap
```diff
@@ -1,142 +1,115 @@
...
+  padding: 0;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  transition: all 100ms ease-out;
 }
 
-._kfgyEr._kfgyEr[data-state="checked"]:hover,
-._kfgyEr[data-state="unchecked"]:focus {
+.sx7plgw[data-state="unchecked"]:focus,
```
andoulla, line 93: So much better that you only have the class once!

src/components/link/Link.tsx
```diff
@@ -31,10 +31,12 @@ const StyledLink = styled('a', {
   }
 })
 
-type LinkProps = StitchesProps<typeof StyledLink>
+type LinkProps = StitchesComponent<typeof StyledLink>
```
andoulla, line 34: Just for consistency with the other components, we could just do the union here 

`type LinkProps = StitchesComponent<typeof StyledLink> & StitchesVariants<typeof StyledLinked>`

thomasdigby: I'm still testing this, not sure the best way to define types with the beta yet

andoulla: sorry didn't notice the draft label on the PR

src/components/box/__snapshots__/Box.test.tsx.snap
thomasdigby: This is now fixed as I forked and patched the issue

andoulla (approved):

src/components/image/Image.tsx
```diff
@@ -1,7 +1,10 @@
+import * as React from 'react'
+
 import { styled } from '~/stitches'
+import { Override } from '~/utilities/types'
 
 // default styling lifted from https://www.zachleat.com/web/fluid-images/#copy-and-paste
-export const Image = styled('img', {
+export const StyledImage = styled('img', {
```
hollg, line 7: 💬 should we lose the `export` on this now?

src/stitches.ts
```diff
@@ -1,77 +1,74 @@
...
-import { createStyled } from '@stitches/react'
+import atomTheme from '@atom-learning/theme'
+import { createCss } from '@stitches/react'
 
 type CSSValue = number | string
 export type CSSBlob = { [key: string]: CSSValue }
 
 export const utils = {
-  p: (value: CSSValue): CSSBlob => ({
+  p: () => (value: CSSValue): CSSBlob => ({
```
hollg, line 8: What's the reason for these changes? I couldn't see anything about it in the migration guide (apologies if it is there and I just didn't recognise it!)

thomasdigby: That's because some of the alpha versions we used had a different syntax, but the migration guide is from the first alpha version (which had this same syntax) to the beta, so it's not included. Necessary though 👍

PR #48, opened 2021-02-16
Feat migrate to rollup

hollg:
Switches from `microbundle` to `rollup` for bundling the library. The config I've done so far is enough to get a very similar bundle size as we got from `microbundle`, but I'm sure there are more optimisations to be made.

(Format of this info varies slightly, but you can see the gist!)

With `rollup`:
`index.js: 7.81 kB → 2.44 kB (gzip)`
`index.js: 7.81 kB → 2.44 kB (gzip)`
`100.3 KB with all dependencies, minified and gzipped`

With `microbundle`:
` 2.58 kB: index.js.gz`
`2.25 kB: index.js.br`
`2.49 kB: index.modern.js.gz`
`2.15 kB: index.modern.js.br`
`100.35 KB with all dependencies, minified and gzipped`


rollup.config.js
```diff
@@ -0,0 +1,32 @@
+import commonjs from '@rollup/plugin-commonjs'
+import resolve from '@rollup/plugin-node-resolve'
+import bundleSize from 'rollup-plugin-bundle-size'
+import { terser } from 'rollup-plugin-terser'
+// import typescript from 'rollup-plugin-typescript2'
```
andoulla, line 5:
```suggestion

```

hollg: D'oh! Thanks for catching this!

rollup.config.js
```diff
@@ -0,0 +1,28 @@
...
+import pkg from './package.json'
+
+const isProduction = process.env.NODE_ENV === 'production'
+
+const deps = Object.keys(pkg.dependencies || {})
+const peerDeps = Object.keys(pkg.peerDependencies || {})
+
+export default {
+  input: 'src/index.ts',
+  external: [...deps, ...peerDeps],
```
thomasdigby, line 16: We should probably include dependencies in our bundle, but not peer dependencies. Would be interested to see the difference between two builds configured like that.

rollup.config.js
```diff
@@ -0,0 +1,28 @@
...
+  output: [
+    { file: pkg.main, format: 'cjs', sourcemap: true },
+    { file: pkg.module, format: 'esm', sourcemap: true }
+  ],
+  plugins: [
+    bundleSize(),
+    commonjs(),
+    resolve(),
+    isProduction && terser(),
```
thomasdigby, line 25: There are probably some terser configuration updates we can make to improve the output, one for the future 👍

thomasdigby (approved):
Pretty happy with this, not much to add at all. Did we not need to add in any support for SVG or image types? Or is that added in your icon PR?

andoulla:
I just noticed that the validation steps fail on the observer type issue. Before merging can we merge main into it and run the pipeline to see it fix itself?

rollup.config.js
hollg: I'm happy to remove dependencies from this if you think it's best, but just FYI, what I've done here is the default behaviour of `microbundle` (i.e. the behaviour we were using already)

rollup.config.js
hollg: I'm gonna add an Asana ticket for this actually because there are a lot of config options so it definitely seems worth looking into

rollup.config.js
thomasdigby: Yeah I think I disagree with their choice but rollup tends to treat most packages as external. Happy to leave as is for now!

hollg:
@thomasdigby 
> Pretty happy with this, not much to add at all. Did we not need to add in any support for SVG or image types? Or is that added in your icon PR?

Yeah I figured I'd use this PR to just recreate our existing behaviour with rollup, and make any further additions in the icon PR.

@andoulla 
> I just noticed that the validation steps fail on the observer type issue. Before merging can we merge main into it and run the pipeline to see it fix itself?

Ah I didn't realise that was in `main` now! Will do, thanks.

rollup.config.js
hollg: Cool - I've made a ticket for optimising the rollup config and noted that we should include looking into this as well as terser etc!

PR #49, opened 2021-02-17
Feat: Radio button

andoulla:
Added a `RadioButton` and a `RadioButtonGroup` that implements the structure of Radix Radio component.

image


sandbox/index.tsx
```diff
@@ -3,11 +3,32 @@ import * as ReactDOM from 'react-dom'
...
 import { Box, globalCss } from '../dist'
+import { RadioButton } from '../src/components/radio/RadioButton';
+import { RadioButtonGroup } from '../src/components/radio/RadioButtonGroup';
 
 globalCss(reset)()
 
 const App = () => {
-  return <Box />
+  return (
+    <Box
```
andoulla, line 13: I will revert these changes it was by mistake

src/components/radio/RadioButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+  },
+  ':checked + &': { backgroundColor: '$secondary300' },
+  ':disabled + &': {
+    backgroundColor: '$tonal600',
+    border: '2px solid $tonal600'
+  }
+})
+
+const StyledIndicator = styled(RadioGroup.Indicator, {
+  width: 6,
```
andoulla, line 33: we don't have a value of six in the theme, so not sure if we should add it

thomasdigby: Could we use 8? We can tweak the theme to add an extra value to the sizing scale

andoulla: Btw this is how it would look if using 8 
image

thomasdigby: Yeah we haven't used a sizing scale before so the current values are temporary. I think 8/16/24/32/48/64 could work, let's just hard code 8 for now

src/components/radio/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,8 @@
+import * as RadioGroup from '@radix-ui/react-radio-group'
+import * as React from 'react'
+
+export const RadioButtonGroup: React.FC = ({ children }) => (
```
hollg, line 4: Is there any reason not to simply export `RadioGroup.Root`, allowing our users to pass whatever props they want to that component itself? This current version doesn't allow for styling of the group etc

src/components/radio/RadioButtonGroup.test.tsx
```diff
@@ -0,0 +1,28 @@
+import { render, screen } from '@testing-library/react'
+import React from 'react'
+
+import { RadioButtonGroup } from '.'
+
+describe(`RadioButton Groupcomponent`, () => {
```
hollg, line 6: Absolutely tiny point, but we're missing a space between `Group` and `component` in the test name

src/components/radio/RadioButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+  height: '$0',
+  display: 'inline-flex',
+  alignItems: 'center',
+  justifyContent: 'center',
+  verticalAlign: 'middle',
+  '&:focus': {
+    outline: 'none'
+  },
+  ':checked + &': { backgroundColor: '$secondary300' },
+  ':disabled + &': {
```
thomasdigby, line 26: This should be `[disabled]` and our nested selectors require `&` prefix now, so: `&[disabled] + &`

Worth applying the `&` to any other selectors that need it

src/components/radio/RadioButtonGroup.test.tsx
andoulla: Thanks

src/components/radio/RadioButton.tsx
andoulla: I switched if for the disabled selector, however, the others don't seem to work with the attribute selector so I will leave them.

src/components/radio/RadioButtonGroup.tsx
andoulla: Now that I think about it again, not there isn't. I'm going to wrap this in styled so we get the styling options and simplify the component as you suggested

thomasdigby (approved):

PR #51, opened 2021-02-18
docs: updated README

andoulla:


README.md
```diff
@@ -2,11 +2,48 @@
...
+- `format`: Formats the code using `Prettier`
+- `lint`: Lints the code using `Eslint`
+- `test`: Runs the testing suit using `Jest`
+- `test:watch`: Runs the testing suit in watch mode
+- `validate`: Runs all the validate commands - This command is used by the CI too, to check pushed changed
+- `validate:dependencies`: Audits the projects dependencies
+- `validate:linting`: Checks for linting issues using `Eslint`
+- `validate:size`: Build the library and checks its size
+- `validate:types`: Checks for `Typescript` errors
+- `start:sandbox`: Starts up a sandbox
```
thomasdigby, line 42: I would remove a few of these scripts as they're not really supposed to be public facing, e.g. `yarn validate` is the public command but it's broken down into a bunch of scripts to keep it organised for us.

I would remove all named `validate` scripts (but not `validate` itself), `clean` and `prepublishOnly`

andoulla: fair enough, I tend to run a few of the validate ones before I push sometimes (like the types one) so that's why I included them

thomasdigby: Yeah same, but I don't think they all need to be documented here 👍

thomasdigby (approved):

PR #52, opened 2021-02-18
Feat: adding textarea component

andoulla:
Added a simple `textarea` component

image


src/components/textarea/Textarea.mdx
```diff
@@ -0,0 +1,16 @@
+---
+title: Textarea
+component: Textarea
+description: Textare is the lowest-level textare component for longer text.
```
andoulla, line 4:
```suggestion
description: Textarea is the lowest-level textarea component for longer text.
```

src/components/textarea/Textarea.tsx
```diff
@@ -0,0 +1,38 @@
...
+  letterSpacing: '0.01em',
+  lineHeight: 1.4,
+  minHeight: 100,
+  resize: 'vertical',
+  width: '100%',
+  color: '$tonal900',
+  p: '$3',
+  transition: 'all 75ms ease-out',
+  '&:focus': {
+    boxShadow: 'inset 0 0 0 1px $primary900',
```
thomasdigby, line 22: Needs updating to `$colors$primary900`

src/components/textarea/Textarea.tsx
```diff
@@ -0,0 +1,38 @@
+import * as React from 'react'
+
+import { styled } from '~/stitches'
+
+const StyledTextarea = styled('textarea', {
+  appearance: 'none',
+  borderRadius: '$2',
```
thomasdigby, line 7: I think this should be `$0` to match our `Input` component

src/components/textarea/Textarea.mdx
```diff
@@ -0,0 +1,15 @@
+---
+title: Textarea
+component: Textarea
+description: Textarea is the lowest-level textarea component for longer text.
+---
+
+Textare should be accompanied by labels, so rather than using `Textarea` directly in a UI,
```
hollg, line 7: Really tiny one:

```suggestion
Textareas should be accompanied by labels, so rather than using `Textarea` directly in a UI,
```

thomasdigby (commented):
Looking good 👍

src/components/textarea/Textarea.tsx
```diff
@@ -0,0 +1,38 @@
...
+  fontFamily: '$sans',
+  fontSize: '$md', // necessary to prevent iOS zooming on focus
+  fontWeight: 400,
+  letterSpacing: '0.01em',
+  lineHeight: 1.4,
+  minHeight: 100,
+  resize: 'vertical',
+  width: '100%',
+  color: '$tonal900',
+  p: '$3',
```
thomasdigby, line 19: May need to also update this to match `Input` as well, which I think is `px: '$3', py: '$2'`

src/components/textarea/Textarea.tsx
```diff
@@ -0,0 +1,38 @@
...
+  },
+  '&[disabled]': {
+    opacity: 0.7,
+    backgroundColor: '$tonal300'
+  }
+})
+
+type TextareaProps = React.ComponentPropsWithoutRef<typeof StyledTextarea>
+
+export const Textarea: React.FC<TextareaProps> = (props) => {
```
hollg, line 34: Is there a difference between this component and `StyledTextArea`?

src/components/textarea/Textarea.tsx
andoulla: yes sorry I got the value wrong. Will update. I think we need something a bit more readable in `themes` about these values or maybe add comments next to them.

src/components/textarea/Textarea.tsx
andoulla: done

src/components/textarea/Textarea.tsx
andoulla: I'm not sure what the question is, sorry

`Textarea` returns the `StyledTextarea` as we did with every other component

hollg: Sure, but in this case `TextArea` doesn't seem to add any props, logic or functionality to `StyledTextArea`, so I was wondering what the reason is for adding this extra layer, instead of the following.

```ts
export const TextArea = styled('textarea', {/* ... */})
```

hollg (approved):

PR #53, opened 2021-02-18
Feat: Add icon component

hollg:
Adds `Icon` component and updates our rollup config to support importing SVGs as React Components


src/components/icon/Icon.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from 'react'
+
+import { styled } from '~/stitches'
+import { Override } from '~/utilities/types'
+
+const StyledIcon = styled('svg', {
+  color: 'currentcolor',
```
hollg, line 7: @thomasdigby @andoulla Is there any other default styling that we should apply here?

thomasdigby: Potentially `vertical-align: middle` and `display: inline-block`? Worth comparing against our current font awesome implementation on the main site, I think they apply those styles and some sizing attributes

src/components/icon/Icon.tsx
```diff
@@ -0,0 +1,39 @@
...
+      md: { height: 24, width: 24 },
+      lg: { height: 32, width: 32 }
+    }
+  }
+})
+
+type IconProps = Override<
+  React.ComponentPropsWithoutRef<typeof StyledIcon>,
+  {
+    is: React.FC<React.SVGProps<SVGSVGElement>>
```
thomasdigby, line 21: This is _cool_

src/components/icon/svg/index.ts
```diff
@@ -0,0 +1,3 @@
+import Check from './check.svg'
+import Home from './home.svg'
```
thomasdigby, line 2: I know these are temporary but we _might_ be able to update these exports to `export Home from './home.svg'`

.babelrc
```diff
@@ -5,7 +5,8 @@
     },
     "production": {
       "plugins": [
-        ["react-remove-properties", { "properties": ["data-testid"] }]
+        ["react-remove-properties", { "properties": ["data-testid"] }],
+        "inline-react-svg"
```
thomasdigby, line 9: What's the benefit of inlining our SVGs in our `@atom-learning/components` bundle?

thomasdigby: Also, should we not apply the SVG optimisation when adding the icon to the library, rather than on every build?

hollg: This is only relevant for getting Jest to accept the SVGs, but may actually no longer be needed after the changes we made on the call today. I'll see if it works without this

src/components/icon/svg/index.ts
hollg: We can't :-(

thomasdigby: Ah well, no problemo!

.babelrc
andoulla: Could I hear about what you decided in stand up please?

src/components/icon/Icon.tsx
```diff
@@ -0,0 +1,39 @@
...
+    as: never
+  }
+>
+
+export const Icon: React.FC<IconProps> = ({
```
andoulla, line 26: how will disabling an icon work? is it meant to be handled by the parent component that consumes the icon? (specifically, I'm thinking of checkbox and  button)

src/components/icon/Icon.tsx
```diff
@@ -0,0 +1,39 @@
...
+  is: SVG,
+  size = 'sm',
+  as, // prevent as from being included in ...props and overwriting as={SVG} on line 29
+  ...props
```
andoulla, line 30: very tiny: Would it be possible to rename this? I think when I see `props` I straight away think that it is the entire props object as it was passed down from the parent.

src/components/icon/Icon.mdx
```diff
@@ -0,0 +1,21 @@
+---
+title: Icon
+component: Icon
+description: Icon wraps our SVG icons
+---
+
+`Icon` effectively takes an SVG from our icon library and adds standardised `size` options.
+
```
andoulla, line 8: Would we add something here about how to optimise the svgs if we are expecting people to optimise them before they commit them

src/components/icon/Icon.tsx
hollg: What do you mean by disabling an icon?

src/components/icon/Icon.tsx
hollg: Do you have any suggestions for the name?

.babelrc
hollg: We didn't really make any decisions as such; Tom just helped me fix some failing tests by adding a plugin to Jest that lets it support SVG imports. Happy to talk more about it at standup though!

src/components/icon/Icon.mdx
hollg: Oh yeah, that's a great idea. Updating now. Thanks!

src/components/icon/Icon.tsx
andoulla: Making it look greyed out. So if a button gets disabled it turns grey. so what happens if it is consuming an icon, will it have to handle the responsibility of styling that icon?

I'm just asking btw

thomasdigby (approved):

src/components/icon/Icon.tsx
andoulla: I think I have figured out my own question, ignore this comment. because we are using `currentcolor` it will just take the parents' styling

Sorry for confusing you

hollg: Yeah, I would say so. IMO an icon doesn't need a concept of being "disabled", it just needs a concept of its color. Buttons have a concept of being "disabled" because it's the button whose behaviour changes; so the button would be in charge of styling the icon appropriately.

This is very much just intuition though! I'd be keen to hear from both you and @thomasdigby if you have any thoughts on how it should work!

src/components/icon/Icon.tsx
andoulla: I have been using `..rest` just because it is the rest operator. But I have seen examples with `...remainingProps`

hollg: Yeah I don't really like `...rest` because it feels a bit like naming a variable `x` in that it doesn't tell you anything about the value. I'll change this to `...remainingProps`. Thanks!

andoulla: There must be a convention about this somewhere. I have been using `rest` just because I see it everywhere used like that. 

`remainingProps` can become our convention

src/components/icon/Icon.tsx
andoulla: Yes, it was a silly question. Icons are assets so they don't need states.

PR #55, opened 2021-02-19
Feat progress bar

andoulla:
Added a `ProgressBar` component with:
- themes 
- appearance 
~- sizes~

I tried to be consistent in naming the themes (with the button) as I think it will make the DS easier to use. Also, I used the same patterns in generating the different combinations of variants.
image


src/components/progress-bar/ProgressBar.mdx
```diff
@@ -0,0 +1,59 @@
+---
+title: Progress Bar
+component: ProgressBar
+description: ProgressBar provides a styled progress bar.
+---
+
+A `ProgressBar` needs to be associated with a label for accessibility purposes, so rather than using this component directly in a UI, consider using a `ProgressBarField`, which provides a `Label` and displays validation errors. Use this `ProgressBar` to compose more complex `Field` type components.
```
thomasdigby, line 7: Is this correct? I see the `ProgressBar` as a UI element to communicate progress, rather than an item with a form field.

src/components/progress-bar/ProgressBar.tsx
```diff
@@ -0,0 +1,166 @@
...
+    },
+    theme: {
+      primary: {},
+      secondary: {},
+      tertiary: {},
+      success: {},
+      warning: {},
+      danger: {}
+    },
+    size: {
```
thomasdigby, line 38: I don't think we should define widths here, we can defer that to the parent component or allow the user to pass custom css to the component

src/components/progress-bar/ProgressBar.tsx
```diff
@@ -0,0 +1,166 @@
...
+      | 'secondary'
+      | 'tertiary'
+      | 'success'
+      | 'warning'
+      | 'danger'
+    color: string
+    size: 'sm' | 'md' | 'lg' | 'xl'
+  } & (
+    | { id: string; 'aria-label'?: string }
+    | { 'aria-label': string; id?: string }
```
thomasdigby, line 143: I don't think these attributes are relevant to this component

src/components/progress-bar/ProgressBar.tsx
```diff
@@ -0,0 +1,166 @@
...
+  backgroundColor: 'currentcolor',
+  height: '100%'
+})
+
+type ProgressBarProps = Override<
+  React.ComponentPropsWithoutRef<typeof StyledProgressBar>,
+  {
+    value: number
+    appearance: 'outline' | 'solid'
+    theme:
```
thomasdigby, line 132: Should `appearance` and `theme` be optional props as we have default values?

src/components/progress-bar/ProgressBar.tsx
```diff
@@ -0,0 +1,166 @@
...
+const getOutlineVariant = (color) => ({
+  border: '1px solid $tonal400',
+  background: 'white',
+  color
+})
+
+const StyledProgressBar = styled(Progress.Root, {
+  borderRadius: 25,
+  minHeight: 12,
+  maxHeight: 30,
```
thomasdigby, line 21: Let's just set a height here of `16`, we should discuss this with @Chema-Atom when we have the new brand but I think it'd be good to simplify some areas of our components where we can.

src/components/progress-bar/ProgressBar.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+    await screen.getByRole('progressbar')
+
+    expect(container).toMatchSnapshot()
+    expect(await axe(container)).toHaveNoViolations()
+  })
+
+  it('renders a xl progress bar', async () => {
+    const { container } = render(
+      <ProgressBar percentage={50} size="xl" aria-label="label" />
```
thomasdigby, line 42: This should be `value` right?

src/components/progress-bar/ProgressBar.mdx
andoulla: Maybe this should say that it needs to be associated with a label. as a progress bar without a label doesn't explain what it visualises hence accessibility picks it up.

```suggestion
A `ProgressBar` needs to be associated with a label for accessibility purposes
```

src/components/progress-bar/ProgressBar.tsx
andoulla: ok I will remove them.  I'm confused about when I should be transferring the sizes from the existing system and when to leave them.

src/components/progress-bar/ProgressBar.tsx
andoulla: I added them as without one of them accessibility will fail

 https://dequeuniversity.com/rules/axe/4.1/aria-progressbar-name?application=axeAPI

src/components/progress-bar/ProgressBar.test.tsx
andoulla: not it is percentage, I had the type wrong. Updating the type

src/components/progress-bar/ProgressBar.mdx
thomasdigby: Yeah good point, worth providing an example with `aria-label` as this is likely the most common use-case

src/components/progress-bar/ProgressBar.tsx
thomasdigby: Yeah I'd say any set sizes should be not transferred and should instead always be handled by the parent. 👍

src/components/progress-bar/ProgressBar.tsx
thomasdigby: Ah great, makes sense! Thanks

src/components/progress-bar/ProgressBar.tsx
andoulla: Yes it will be good to review them as I think there might be some contrast issues too, that `jest-axe` can't pickup

andoulla: image
They are a bit big now

src/components/progress-bar/ProgressBar.mdx
andoulla: I have changed it completely, hopefully, now it makes more sense as a whole

src/components/progress-bar/ProgressBar.test.tsx
andoulla: I reverted to value for everything as it keeps it consistent with `Radix-ui` type and I don't have to override it.

In Atom-core we are using percentage but it is simpler to change it there than having to write a completed type to

src/components/progress-bar/ProgressBar.tsx
thomasdigby: Yeah I agree, let's make em 12 👍

andoulla: we don't have a 12 so I'm just going to hardcode it until we review the theme layer and add it in

src/components/progress-bar/ProgressBar.mdx
```diff
@@ -0,0 +1,91 @@
...
+<ProgressBar value={20} aria-label="Completion rate" />
+```
+
+## Themes
+
+These are the available `themes` for this component: `Primary` (default), `Secondary`, `Tertiary`, `Success`, `Warning`, `Danger`
+
+```jsx preview center
+<>
+  <ProgressBar value={20}>Primary</ProgressBar>
```
thomasdigby, line 22: Let's not pass in children if the component doesn't accept them. I don't think we need the text here, I think the `theme` prop and the visual example should be clear enough.

thomasdigby (approved):
Awesome, just need that docs change then I think we're all good! 👍

src/components/progress-bar/ProgressBar.tsx
```diff
@@ -0,0 +1,166 @@
...
+    | { id: string; 'aria-label'?: string }
+    | { 'aria-label': string; id?: string }
+  )
+>
+
+export const ProgressBar: React.FC<ProgressBarProps> = ({
+  value,
+  appearance = 'outline',
+  theme = 'primary',
+  size = 'sm',
```
hollg: This is something to discuss as a team rather than to fix right now, but I wonder whether it's worth having a convention about default sizes. I think some components default to `sm` and some to `md`.

src/components/progress-bar/ProgressBar.tsx
andoulla: image

src/components/progress-bar/ProgressBar.tsx
andoulla: I agree, however, I have completed removed sizes after Tom's recommendation, and it is going to be handled by the parent.

PR #57, opened 2021-02-19
feat: add InputField 

hollg:
Adds `InputField` component.


src/components/label/Label.tsx
```diff
@@ -20,7 +21,7 @@ const StyledLabel = styled('label', {
 type LabelProps = Override<
   React.ComponentPropsWithoutRef<typeof StyledLabel>,
   {
-    as: never
+    as?: never
```
hollg, line 24: This is to stop TypeScript telling me that `as` is a required prop (but that no value is valid!)

src/utilities/css-wrapper/CSSWrapper.tsx
```diff
@@ -1,10 +1,10 @@
 import * as React from 'react'
 
 import { Box } from '~/components'
-import { CSSBlob } from '~/stitches'
+import { CSS } from '~/stitches'
 
 type CssWrapperProps = {
-  css?: CSSBlob
+  css?: CSS
```
hollg, line 7: This lets us take the `css` prop from a Stitches component and pass it to the `CssWrapper`.

src/components/input-field/InputField.mdx
```diff
@@ -0,0 +1,17 @@
...
+
+As well as text for its `Label`, `InputField` accepts all the same props as `Input` and will pass them on to the `Input` it renders. However, as with all our composed components,
+`InputField`'s `css` prop will be applied to a containing `CSSWrapper`—the styling of the individual components inside `InputField` cannot be altered. Err on the side of using consistent form fields,
+but if you really need something with different styling then consider composing your own field from the `Input`, `Label` and `ValidationError` components.
+
+```jsx live center
+<InputField placeholder="Placeholder text" css={{ width: 300 }} />
+```
+
+#
```
andoulla, line 17:
```suggestion

```

hollg: Yeah this PR is still a draft so there are lots of unfinished areas like this — I would hold off on reviewing for now!

src/components/box/Box.tsx
```diff
@@ -1,5 +1,13 @@
+import * as React from 'react'
+
 import { styled } from '~/stitches'
 
-export const Box = styled('div', {})
+export const UnstyledBox = styled('div', {})
```
andoulla, line 5: Tiny comment: I think this should still be called `StyledBox` as it is still wrapped in the styled function and has the css prop

src/components/input-field/InputField.test.tsx
```diff
@@ -0,0 +1,53 @@
+import { render, screen } from '@testing-library/react'
+import { axe } from 'jest-axe'
+import * as React from 'react'
+
+import { InputField } from '.'
+
+describe(`InputField component`, () => {
```
andoulla, line 7: Should we have a test for showing the validation error?

src/components/box/Box.tsx
hollg: Hahaha, wow, it didn't really occur to me that we were naming them `styled` because that's what the function was called; I figured it was just because they have styles applied to them 😅. Very happy to change this!

src/components/input-field/InputField.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { ValidationError } from '~/components/validation-error'
+import { theme } from '~/stitches'
+
+type InputFieldProps = InputProps & {
+  label: string
+  error?: string
+  required?: boolean
+}
+
+export const InputField: React.FC<InputFieldProps> = ({
```
thomasdigby, line 15: Would like to hear opinions on this, but I think it's worth using a prefix rather than a suffix on each of these field components. `FieldInput`, `FieldTextarea`, `FieldRange` etc. I think makes it clearer when we're using the grouped versions of the fields, rather than `InputField`, `TextareaField`, `RangeField`. Thoughts

src/components/input-field/InputField.tsx
```diff
@@ -0,0 +1,38 @@
...
+  required = false,
+  ...remainingProps
+}) => (
+  <Box css={css}>
+    <Label css={{ mb: '$1' }} htmlFor={name}>
+      {label}
+      {required && <span style={{ color: theme['colors'].danger }}> *</span>}
+    </Label>
+    <Input
+      css={{ mb: '$1', ...(error && { border: '1px solid $danger ' }) }}
```
thomasdigby, line 29: The error state here should be a variant on the `Input` component, potentially `state`? `<Input state="error" />`

src/components/input-field/InputField.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { Input, InputProps } from '~/components/input'
+import { Label } from '~/components/label'
+import { ValidationError } from '~/components/validation-error'
+import { theme } from '~/stitches'
+
+type InputFieldProps = InputProps & {
+  label: string
+  error?: string
+  required?: boolean
+}
```
thomasdigby, line 13: We need to add `name` here no?

src/components/input-field/InputField.tsx
```diff
@@ -0,0 +1,38 @@
...
+  label,
+  name,
+  error = undefined,
+  required = false,
+  ...remainingProps
+}) => (
+  <Box css={css}>
+    <Label css={{ mb: '$1' }} htmlFor={name}>
+      {label}
+      {required && <span style={{ color: theme['colors'].danger }}> *</span>}
```
thomasdigby, line 26: This is likely a question for Chema/Amit, but I wouldn't style this red, I think just a `*` is fine. Avoids having to access the theme as well.

src/components/input-field/InputField.tsx
andoulla: Happy with either, we should just add a line in the contribution guide just to state the pattern

src/components/box/Box.tsx
andoulla: Maybe I'm wrong, I thought that's why we were naming them Styled

src/components/input-field/InputField.tsx
andoulla: isn't it like a convention that mandatory fields get a red asterisk? It is a Chema/Amit question I agree

src/components/input-field/InputField.tsx
hollg: I was actually just thinking about this. `name` is on the `InputProps`, but it's optional there. So we either need to override it to be required there, or add it as required here. I'm leaning towards the latter on the basis that for some specialised inputs (e.g. DOB), there might not be a 1:1 relationship between labels and inputs, but I'm very open to suggestions

src/components/input-field/InputField.tsx
hollg: Oh yeah, nice! Will make that change now

src/components/input-field/InputField.test.tsx
hollg: Yes, great point! Added.

src/components/input-field/InputField.tsx
thomasdigby: I would add it as required here, as you mentioned it's specific to this usecase and for others, but not all.

src/components/input-field/InputField.tsx
thomasdigby: I haven't seen a red asterisk that often, the asterisk itself I think is the convention.

hollg: I was working off this article: https://www.nngroup.com/articles/required-fields/

> Should the asterisk be red? Not necessarily, but red has become the expected required-marker color on the web, which is a reason in its own right to stick with this choice (according to Jakob’s Law). In any case, there is some value to using different colors for the asterisk and for the field label: it allows users to quickly separate the two and focus on the field label while trying to decode what the field means. While red is somewhat recommended, we have a strong recommendation to avoid pale grays or low-contrast colors for the asterisk. Slightly muted colors can have aesthetic benefits, but truly low contrast symbols constitute an accessibility problem for low-vision or elderly users and slow down visual processing of the form for everybody.

(my annecdotal experience would also be that it's very often red)

src/components/input-field/InputField.tsx
hollg: Personally I think that seems slightly _less_ clear, because it seems to make `Field` take on the adjective role, modifying `Input` / `Textarea` / `Range`. So to me it would suggest that `FieldInput` is a type of `Input`, rather than a type of `Field`. 

If the first exposure someone has to these names is seeing `Input` and `FieldInput` then I don't think those names are doing anything to help them understand what those components are or their relationship with each other.

I think `InputField` / `RangeField` / `DOBField` etc avoids this problem and, more generally, mirrors how we'd describe those things in English anyway.

src/components/input-field/InputField.tsx
hollg: Keeping it red for now per Amit's suggestion

src/components/input-field/InputField.tsx
thomasdigby: See I view it the other way around, `Field` to me is the base type, and `Input`, `Range` etc. is the modifier. When naming things I tend to go less-specific -> more specific, e.g. `color-primary-500`

I think we could go back and forth on this as it's fairly subjective but it's not a dealbreaker for me, I think the main reason I wanted to switch to that convention is to a) put the importance on the `Field` type first, and b) better group them together when used. e.g.

```jsx
<Form>
  <FieldInput name="Email" />
  <FieldInput name="Name" />
  <FieldTextarea name="Description />
</Form>

//

<Form>
  <InputField name="Email" />
  <Input name="Name" />
  <TextareaField name="Description />
</Form>
```

Go with your gut for now and we can always tweak in future versions

thomasdigby (approved):

PR #60, opened 2021-02-23
Feat: add tooltip component

hollg:
- Added a `Tooltip` component
- Exporting `Tooltip`, `Tooltip.Trigger` and `Tooltip.Content` to allow user to compose together for better flexibility
- Added mocks for `ResizeObserver` and `DOMRect` in `jest`
- Added `forwardRef` to `Text` to allow use in `Tooltip`, we ought to add this to most of our components I reckon

Screenshot 2021-03-30 at 14 57 20


src/components/tooltip/Tooltip.mdx
```diff
@@ -0,0 +1,19 @@
+---
+title: Tooltip
+component: Tooltip,Tooltip.Trigger,Tooltip.Content
```
andoulla, line 3:
```suggestion
component: Tooltip, Tooltip.Trigger, Tooltip.Content
```

src/components/tooltip/Tooltip.mdx
```diff
@@ -0,0 +1,19 @@
...
+description: A popup that displays information related to an element when the element receives keyboard focus or the mouse hovers over it.
```
andoulla, line 4: Maybe not use the `popup` term as it is associated with dialog popup in a lot of places

src/components/tooltip/Tooltip.mdx
```diff
@@ -0,0 +1,19 @@
...
+category: Surfaces
+---
+
+`Tooltip` exports 3 components that combine to create our tooltip. The `Tooltip.Trigger` requires a React component child (or passed to `as`) to allow the necessary event bindings and accessible properties to be cloned, also note that the component must accept a `ref`.
```
andoulla, line 8:
```suggestion
`Tooltip` exports 3 components that combine to create our tooltip. The `Tooltip.Trigger` requires a React component child (or passed to `as`) to allow the necessary event bindings and accessible properties to be cloned.
 Also, note that the component must accept a `ref`.
```

src/components/tooltip/Tooltip.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+            <span>TOOLTIP</span>
+          </Tooltip.Trigger>
+          <Tooltip.Content>CONTENT</Tooltip.Content>
+        </Tooltip>
+      </IdProvider>
+    )
+
+    fireEvent.mouseOver(screen.getByText('TOOLTIP'))
+
+    expect(await screen.findByRole('tooltip')).toBeInTheDocument()
```
andoulla, line 23: Should this be `toBeVisible()` as it might be in the document but not visible?

src/components/tooltip/Tooltip.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+      <IdProvider>
+        <Tooltip>
+          <Tooltip.Trigger>
+            <span>TOOLTIP</span>
+          </Tooltip.Trigger>
+          <Tooltip.Content>CONTENT</Tooltip.Content>
+        </Tooltip>
+      </IdProvider>
+    )
+
```
andoulla, line 38: I'm not 100% but I think this does actually run against the Tooltip as the tooltip isn't visible without triggering the `mouseOver` Event

src/components/tooltip/TooltipTrigger.tsx
```diff
@@ -0,0 +1,14 @@
+import { Slot } from '@radix-ui/react-slot'
```
andoulla, line 1: How come we are not using Radix's `Tooltip.Trigger` here

andoulla (commented):
I'm just wondering, now that we have a tooltip that can take any content, do we really need a popover component? If yes, I think we should just switch the popover to use this component and add onClick handlers only.

src/components/tooltip/Tooltip.mdx
thomasdigby: This is metadata to enable multiple props tables so doesn't require more readable spacing, the syntax for frontmatter iterables is comma delimited without a space

src/components/tooltip/Tooltip.test.tsx
thomasdigby: Yeah that's a good shout, will update

src/components/tooltip/TooltipTrigger.tsx
thomasdigby: Slot passes the required props onto the child, this is most useful when you have unknown children (often the case with libraries) https://radix-ui.com/primitives/docs/utilities/slot

src/components/tooltip/Tooltip.test.tsx
thomasdigby: I'll update so that the tooltip is visible

src/components/tooltip/TooltipContent.tsx
```diff
@@ -0,0 +1,55 @@
...
+  borderRadius: '$0',
+  color: 'white',
+  fontFamily: '$sans',
+  fontSize: '$sm',
+  lineHeight: 1.2,
+  p: '$2',
+  transformOrigin: 'var(--radix-tooltip-content-transform-origin)',
+  variants: {
+    size: {
+      sm: { maxWidth: '100px' },
```
hollg, line 23: Is there anything in the theme that would work here?

thomasdigby: I don't think so, these sizes are a bit out of the sizing scale unless we open that up to much larger values. We do have `96` and `128` but `250` and `400` are too large.

src/components/tooltip/Tooltip.test.tsx
thomasdigby: It think Radix is setting some properties that render it "not visible" in jsdom, I'm not sure how best to work around this in this context and I don't want to sink too much time into it. Going to leave this test as is and write a ticket to look at this in future.

thomasdigby (approved):

thomasdigby:
> I'm just wondering, now that we have a tooltip that can take any content, do we really need a popover component? If yes, I think we should just switch the popover to use this component and add onClick handlers only.

There's a great thread in the Radix UI Discord as to the differences between the two and why they're relevant https://discord.com/channels/752614004387610674/803656831704629298/824724691574456352

Copied the main reply from one of the lib authors below if you can't access

```
Popover and Tooltip are vastly different concepts in terms of accessibility. Thus why they both exist.
They aren't even built the same way internally.
It's one of these things that may "look" the same on the surface visually, but really isn't the same.
For a start, Tooltip cannot be interactive.
That already makes them completely different from Popovers.
They are only meant for small bits of information. Focus remains on the trigger, only one can be opened at a time, etc.

On the contrary, Popover are closer to a a dialog but with positioning logic similar to tooltip. Focus moves into the Popover and doesn't remain on the trigger, focus may be trapped, etc.
There are quite a few components that we may feel could be bundled into one, but our approach is actually to offer more specialised components which cater for each use case better than if it was all mashed into one.
For example, we're working on a Panel (name may change) primitive, which is also similar to Popover or Dialog but offers slightly different behavior, initial positioning, but not attached to target, draggable, etc
I hope this helps clarify our approach and why both exist. :pray:
```

PR #61, opened 2021-02-23
Feat video

andoulla:
https://recordit.co/a29k44UJuD


src/components/video/Video.tsx
```diff
@@ -0,0 +1,47 @@
...
+  width = '100%',
+  height = 405,
+  lightPlayer = false,
+  ...remainingProps
+}) => {
+  const showControls = !lightPlayer
+
+  return (
+    <StyledVideo
+      role="figure"
```
andoulla, line 34: I'm not sure about this role, but it was the closest I could find that indicates content in a different form.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Figure_Role

As react player returns a `div` without a role I think it is better to have one, but happy to take it out

src/components/video/Video.tsx
```diff
@@ -0,0 +1,47 @@
...
+    lightPlayer?: boolean
+    as: never
+  }
+>
+
+export const Video: React.FC<VideoProps> = ({
+  externalId,
+  width = '100%',
+  height = 405,
+  lightPlayer = false,
```
hollg, line 27: What does "light" mean in this context? This toggle seems to do quite a lot of work but I don't think its name makes it clear what to expect. 

I think it would be nice to have a prop called something like `type` or `display` here with two possible values and default to the non-light player. To me that would feel more consistent with our other components where we've based the API on stitches variants.

src/components/video/Video.tsx
```diff
@@ -0,0 +1,47 @@
...
+}) => {
+  const showControls = !lightPlayer
+
+  return (
+    <StyledVideo
+      role="figure"
+      url={`https://player.vimeo.com/video/${externalId}`}
+      width={width}
+      height={height}
+      playing={lightPlayer || undefined} // true or false to pause or play the media
```
hollg, line 38: What's the effect of the OR expressions here? `true || undefined === true` and `false || undefined === undefined`, so I can't figure out how this is any different in effect than just passing `lightPlayer`.

src/components/video/Video.tsx
thomasdigby: I actually think it's worth removing this `lightPlayer` feature for now, it's not implemented how we'd want it to be on the main app so keen to not copy over that API.

src/components/video/Video.tsx
```diff
@@ -0,0 +1,47 @@
...
+    height?: number | string
+    lightPlayer?: boolean
+    as: never
+  }
+>
+
+export const Video: React.FC<VideoProps> = ({
+  externalId,
+  width = '100%',
+  height = 405,
```
thomasdigby, line 26: I think we should remove these props, and provide a `ratio` prop instead (defaulting to 9/16). We could use [this technique](https://github.com/CookPete/react-player#responsive-player) to define a wrapper component with a ratio, and then render the video component as full width within it.

The max width constraint should come from the parent container of this component.

src/components/video/Video.tsx
thomasdigby: Can we test if just `light` works? https://github.com/CookPete/react-player#light-player

andoulla: Light implies less available features, for example, no controls.

Happy to remove it but just to note that this option is used in atom-core. we would just have to remove it from there when we integrate the DS

src/components/video/Video.tsx
andoulla: The component internally is already at 100% for height and width so I will just add the wrapper to have the ratio. Should we initially have only the 9/16 ratio?

thomasdigby: Yeah we should default to 9/16 and then the video user can pass in a different ratio if necessary `<Video ratio={2/3} id="sdfsd" />`

src/components/video/Video.tsx
andoulla: removed the use of lightPlayer and removed all the custom logic we had. If it is needed, it can be passed as a prop to control the controls, muted and playing options

src/components/video/Video.tsx
andoulla: `false || undefined` returns `undefined`. However, I have to admit I don't know why in `atom-core` we were setting it as undefined.

I would say since I'm removing this completely (from the previous comment) the consumer can pass a true or false value as a prop if they need it

src/components/video/Video.tsx
andoulla: Have done this change. Please, let me know if looks ok

src/components/video/Video.tsx
```diff
@@ -0,0 +1,37 @@
...
+  }
+>
+
+export const Video: React.FC<VideoProps> = ({
+  externalId,
+  ratio = 9 / 16,
+  ...remainingProps
+}) => {
+  return (
+    <CSSWrapper css={{ paddingTop: ratio }}>
```
thomasdigby, line 25: This could just be `<Box>` as we're definitely applying styles, we'll also need to add the `position: relative` here and `postition: absolute; top: 0; left: 0;` to the video component itself. https://css-tricks.com/aspect-ratio-boxes/

thomasdigby: We'll need to apply the ratio as a percentage as well, e.g. `paddingTop: '${ratio}%'`

src/components/video/Video.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { styled } from '~/stitches'
+import { Override } from '~/utilities/types'
+
+const StyledVideo = styled(ReactPlayer, {})
+
+type VideoProps = Override<
+  React.ComponentPropsWithoutRef<typeof StyledVideo>,
+  {
+    externalId: string
+    ratio: number
```
thomasdigby, line 15: Need to make this optional

thomasdigby (approved):
Looking good, just that one type update 🙌

src/components/video/Video.tsx
andoulla: oh yes sorry, forgot that props that have a default should be optional, will update now

PR #64, opened 2021-02-26
Feat loading icon

hollg:
Adds `Loader` component

https://user-images.githubusercontent.com/21319237/109299251-b83f9180-782c-11eb-9f5c-4d1815a997de.mov


src/components/loader/Loader.mdx
```diff
@@ -0,0 +1,14 @@
+---
+title: Loader
+component: Loader
+description: Loader provides feedback that an action is in progress
+---
+
+Render a `Loader` to show that content is loading or that a user's action is in progress.
+
+```jsx preview content
```
thomasdigby, line 9: Do you mean `center` here?

src/components/loader/Loader.tsx
```diff
@@ -0,0 +1,90 @@
...
+        marginLeft: '3px',
+        marginRight: '3px',
+        width: '8px'
+      },
+      lg: {
+        height: '10px',
+        marginLeft: '5px',
+        marginRight: '5px',
+        width: '10px'
+      }
```
thomasdigby, line 58: This is a good place to use our shorthand `mx` and I wonder if we'll find it valuable to create a `size` utility to return height and width? Might be nice to define these variants as: 

```js
lg: {
  mx: 5,
  size: 10
}
```

thomasdigby: Also, I think it's worth removing the `lg` variant, it's too big, and move each variant up the scale by one, `sm`, `md` and `lg` should be useful enough for us.

src/components/loader/Loader.tsx
```diff
@@ -0,0 +1,90 @@
...
+        justifyContent: 'center',
+        whiteSpace: 'nowrap',
+        // Stitches doesn't recognise CSS as the right type here. The Modulz team doesn't know why.
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        ...(css as any)
+      }}
+      role="alert"
+      aria-live="assertive"
+    >
+      <VisuallyHidden.Root>{message}</VisuallyHidden.Root>
```
thomasdigby, line 86: I wonder if it might be worth creating our own utility for this, this implementation adds the relevant CSS as a style attribute but we could write it as a stitches component and get a reusable class and simplify the HTML when debugging. Can add to the backlog if we don't want to tackle now 👍

src/components/loader/Loader.tsx
```diff
@@ -0,0 +1,90 @@
...
+      <Dot size={size} /> <Dot size={size} /> <Dot size={size} />
```
thomasdigby, line 87: Does this layout `<Dot size={size} /> <Dot size={size} /> <Dot size={size} />` introduce any unintended layout issues with whitespace? Or does the `whiteSpace: 'nowrap'` take care of that?

src/components/loader/Loader.mdx
hollg: Honestly I've no idea - I think I just copied and pasted this! _Should_ it be centered?

src/components/loader/Loader.tsx
hollg: There's no whitespace between the rendered `div`s. I've just tested again and actually `whiteSpace: 'nowrap'` can go — there's no whitespace for it to alter anyway

src/components/loader/Loader.tsx
hollg: Oh yeah `size` is cool! Added and refactored here.

src/components/loader/Loader.tsx
hollg: Are you thinking of basically writing this out ourselves? https://github.com/radix-ui/primitives/blob/main/packages/react/visually-hidden/src/VisuallyHidden.tsx

src/components/loader/Loader.mdx
hollg: It's copied from the `Link` documentation, where apparently you put `jsx preview content` in a commit called `feat: updated documentation files to fit new format` :D I can change them both to `center` if that's what it should be

thomasdigby: Oh yeah, should be `center` 👍

src/components/loader/Loader.tsx
thomasdigby: Basically yeah, just to convert that darn `style` attribute to `css`. For a future sprint if we think it's worth it.

src/components/loader/Loader.tsx
```diff
@@ -0,0 +1,79 @@
...
+  return (
+    <Flex
+      css={{
+        justifyContent: 'center',
+        // Stitches doesn't recognise CSS as the right type here. The Modulz team doesn't know why.
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        ...(css as any)
+      }}
+      role="alert"
+      aria-live="assertive"
```
thomasdigby, line 71: Can you talk me through this attribute and what it adds to our loading component?

src/components/loader/Loader.tsx
```diff
@@ -0,0 +1,79 @@
...
+      sm: {
+        size: '4px',
+        mx: '2px'
+      },
+      md: {
+        size: '6px',
+        mx: '2px'
+      },
+      lg: {
+        size: '8px',
```
andoulla, line 44: Sorry to join the party late but does `size` do?

andoulla: Ignore me I found it in the code

src/stitches.ts
```diff
@@ -52,9 +52,12 @@ export const utils = {
     marginTop: value,
     marginBottom: value
   }),
-
   bg: () => (value: string): CSSBlob => ({
     background: value
+  }),
+  size: () => (value: CSSValue): CSSBlob => ({
+    height: value,
```
andoulla, line 59: My only concern is that this doesn't straight away show when reading the code that height and width is the same value. Anyone trying to use it will need to come into this utils file to learn what size is and then, use it.

hollg: Yeah, I kind of share this concern too. But presumably it's also true of (for example) `bg`, `px` etc which we've been okay with. I'm not really sure which way to go with it.

andoulla: For bg, I think it is ok as you can pass any of the possible values for background and it will work. It assigns only one prop. Also, bg is a bit of an accepted shortening for background 

I was trying to think of a better name like squareSize or widthHeightSize to make it a bit more obvious but they don't sit well with me. if we can't get a better name I would either suggest to add a TODO to come back to it before the beta release or add a comment where it is used that it adjusts both the width and height.

hollg: Personally if we think it needs a comment everywhere it's used then I'd rather get rid of it as it wouldn't be saving any lines of code anyway. I guess the naming of these utils is something we should try to get feedback on as the wider team starts using the library. I can add a `TODO` now to the effect of "remove this if it's confusing"?

andoulla: Exactly, so if people don't understand it then we can get rid of it. Maybe we should start a list of things we are going to ask targeted feedback

andoulla: btw if you prefer to get rid of it, go ahead. it is just 2 props

thomasdigby (approved):

PR #65, opened 2021-02-26
Feat: add popover component

andoulla:
image

The right-aligned triangle is missing a border


src/components/popover/Popover.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+type PopoverProps = Override<
+  React.ComponentPropsWithoutRef<typeof StyledPopover>,
+  {
+    show?: boolean
+    className?: string
+  }
+>
+
+export const Popover: React.FC<PopoverProps> = ({
```
hollg, line 105: I don't think we discussed this component (we definitely should have), but I was thinking that it would take a prop for the popover content and render its children "as normal", so that this component is the one in charge of positioning the popover in relation to its trigger. Does that make sense?

src/components/popover/Popover.tsx
```diff
@@ -0,0 +1,122 @@
...
+  align = 'center',
+  show = true,
+  children,
+  ...remainingProps
+}) => {
+  return (
+    <StyledPopover
+      align={align}
+      className={show ? 'visible' : ''}
```
thomasdigby, line 114: Could we use a boolean variant here? `variants: { visible: { true: { ...styles } } }`

src/components/popover/Popover.tsx
hollg: (This is the pattern I followed in the `Tooltip` component, but maybe I'm actually misunderstanding what we want from the `Popover`! For reference: https://github.com/Atom-Learning/components/blob/198d3b8d18b6e8f6d72ad4d48b9a15c0a40812a8/src/components/tooltip/Tooltip.tsx#L31)

src/components/popover/Popover.tsx
andoulla: I had it like that and changed it as visible and hidden didn't really felt like variants. Happy to revert it if we think we should but here is what the code looks like:
https://github.com/Atom-Learning/components/pull/65/commits/f378bb80cc2d2e0f78a7eb8e0eb3e23b5cc6fe0d#diff-008256a6a960945554dfffc4988ad92580829eeca4e4d3ca426779ffe7575a11

thomasdigby: I think you're right that that particular pattern doesn't work as well, what I suggested is slightly different in that it's a boolean variant so more like:

```ts
const StyledPopover = styled('div', {
  // styles
  variants: {
    // more variants
    show: {
      true: {
        opacity: 1,
        visibility: 'visible'
      }
    }
  }
})

<StyledPopover show />
```

This should mean that you only need to declare one variant, and the prop to enable that variant is a boolean 👍

src/components/popover/Popover.tsx
```diff
@@ -0,0 +1,122 @@
...
+  React.ComponentPropsWithoutRef<typeof StyledPopover>,
+  {
+    show?: boolean
+    className?: string
+  }
+>
+
+export const Popover: React.FC<PopoverProps> = ({
+  align = 'center',
+  show = true,
```
thomasdigby, line 107: I would probably default to `false`? or maybe not default at all and require a value passing in

src/components/popover/Popover.tsx
```diff
@@ -0,0 +1,122 @@
...
+    '&::before': {
+      borderStyle: 'solid',
+      borderColor: 'transparent',
+      content: "''",
+      height: 0,
+      pointerEvents: 'none',
+      position: 'absolute',
+      top: '100%',
+      width: 0,
+      borderTopColor: border,
```
thomasdigby, line 39: So, cool thing with the new stitches beta, we can declare locally scoped tokens. e.g. 

```
'$$borderColor: '$tonal400'
'&::before': {
  borderTopColor: '$$borderColor'
}
```

I'm not sure how that will fit with the convention that you've set up here, I think you'll have to declare those tokens within the `styled('div', {})` and then they should be inherited properly within the pseudo elements.

src/components/popover/Popover.tsx
```diff
@@ -0,0 +1,122 @@
...
+// TODO: accessibility
+
+const getTriangle = (position) => {
+  const size = 8
+  const background = 'white'
+  const border = '$tonal400'
+
+  return {
+    '&::after': {
+      // Pseudo-classes need to be separated ($::after, $::before: doesn't work in stitches work)
```
thomasdigby, line 16: Might be a silly suggestion, but can we try `&::after,&::before` without the space between?

src/components/popover/Popover.tsx
```diff
@@ -0,0 +1,122 @@
...
+    }
+  }
+}
+
+const Visibility = () => ''
+
+Visibility.toString = () => '.visible' // Gives the ability to apply a class to show and hide the popover
+
+const StyledPopover = styled('div', {
+  boxShadow: `0px 2px 1px -1px rgba(0, 0, 0, 0.2),
```
thomasdigby, line 52: We should have a box shadow token value to use, try using `$1` or `$0` to access a box shadow value from our elevation scale

src/components/popover/Popover.tsx
andoulla: oh, didn't think of this thanks. Will apply

src/components/popover/Popover.tsx
andoulla: yes sorry, I had it for testing purposes and forgot to add a note next to it. I can make it mandatory too

src/components/popover/Popover.tsx
andoulla: I have tried that too, it can't handle that either. but I'm not sure if the problem is the space or the fact that afterwards, I need to have a separate `::after`( and `::before`) that has the props that are not in common. This part of stitches is very brittle

src/components/popover/Popover.tsx
andoulla: Unfortunately, these don't work as they supposed to. it successfully creates a locally scoped token but it doesn't pass it correctly in the function. for example, the `borderColor` is set to `var(--borderColor)` but that computes to black instead of the value I'm passing.

To save some time, I will just skip this and we can come back to it. I just don't think it is a major issue to spend more time on it.

src/components/index.ts
```diff
@@ -11,9 +11,10 @@ export { InputField } from './input-field'
 export { Label } from './label'
 export { Link } from './link'
 export { Loader } from './loader'
-export { ProgressBar } from './progress-bar'
 export { Select } from './select'
 export { Text } from './text'
+export { ProgressBar } from './progress-bar'
```
hollg, line 16: Can we keep these alphabetical? :D

src/components/popover/Popover.mdx
```diff
@@ -0,0 +1,36 @@
+---
+title: Popover
+component: Popover
+description: Popover provides a styled actionable popover.
+---
+
+The `Popover` can display any type element as a trigger and has the popover hidden by default.
```
hollg, line 7:
```suggestion
The `Popover` can display any type of element as a trigger and has the popover hidden by default.
```

src/components/popover/Popover.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+```jsx live center
+<Popover aria-label="Password hints" content="Here is some content">
+  <Button>Click me</Button>
+</Popover>
+```
+
+## variants
+
+Has 3 different `align` variants `center` (default), `left` and `right`:
```
hollg, line 17:
```suggestion
Has 3 different `align` variants: `center` (default), `left` and `right`:
```

src/components/index.ts
andoulla: of course, we should find the right eslint plugin to do this for us

hollg (approved):

PR #66, opened 2021-02-26
Chore: Publishing setup

thomasdigby:
- Set initial version to 0.1.0 alpha
- Fixed test and dev dependency
- Specified output files for package
- Removed source maps

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -15,10 +15,10 @@ const peerDeps = Object.keys(pkg.peerDependencies || {})
 
 export default {
   input: 'src/index.ts',
-  external: [...deps, ...peerDeps],
+  external: [...deps, ...peerDeps, 'react-player/vimeo'],
   output: [
-    { file: pkg.main, format: 'cjs', sourcemap: true },
-    { file: pkg.module, format: 'esm', sourcemap: true }
+    { file: pkg.main, format: 'cjs' },
+    { file: pkg.module, format: 'esm' }
   ],
   plugins: [
     bundleSize(),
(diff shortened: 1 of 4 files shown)


rollup.config.js
```diff
@@ -15,10 +15,10 @@ const peerDeps = Object.keys(pkg.peerDependencies || {})
 
 export default {
   input: 'src/index.ts',
-  external: [...deps, ...peerDeps],
+  external: [...deps, ...peerDeps, 'react-player/vimeo'],
   output: [
-    { file: pkg.main, format: 'cjs', sourcemap: true },
-    { file: pkg.module, format: 'esm', sourcemap: true }
+    { file: pkg.main, format: 'cjs' },
+    { file: pkg.module, format: 'esm' }
```
andoulla, line 21: how come we don't need the sourcemaps anymore? what am I missing here?

thomasdigby: It's unnecessary for third-party libs, we wouldn't (shouldn't) see the source code for an external package within the exported source maps for first party code.

andoulla: oh yes, why didn't I think of it? thanks

tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "outDir": "./dist/", // path to output directory
-    "sourceMap": true, // allow sourcemap support
+    "sourceMap": false, // allow sourcemap support
```
hollg, line 4: This defaults to `false` so we can remove the whole line. Alternatively, if we want to be explicit about not including source maps then I think we should update the comment on this line

hollg (approved):

PR #67, opened 2021-03-03
Feat: add form component

hollg:
Adds the `Form` component and updates `InputField` to work with it.

`Form` renders a `form` hooked into `react-hook-form` and automatically registers fields to track and validate their values. Any child with a `name` prop is assumed to be a field and any child without a `name` prop is rendered as normal.

There a lot of optimisations and improvements to be made further down the line (e.g. allowing fields to be nested inside non-fields!) but this should work for the public app rebuild.


package.json
```diff
@@ -80,7 +80,7 @@
     "npm-run-all": "^4.1.5",
     "parcel-bundler": "^1.12.4",
     "prettier": "^2.2.1",
-    "react": "^17.0.1",
+    "react": "^16.8 || ^17.0",
```
andoulla, line 83: I got that once too, and I didn't accept the change as I don't understand why we would need to use the older version

src/components/form/Form.mdx
```diff
@@ -0,0 +1,90 @@
...
+description: Form abstracts form validation logic away from the view code
+category: Forms
+---
+
+`Form` is a wrapper around an HTML `form` element and the `react-hook-form` library. It manages the `useForm()` instance and handles the validation logic.
+
+`Form` is compatible with all our field components (i.e. `InputField`, not `Input`) and is also compatible with custom fields so long as they accept the following props:
+
+- `error`: `string` (for validation errors)
+- `register`: `React.ForwardedRef<HTMLInputElement>` (only used internally by the `Form` to keep track of the field's value)
```
andoulla, line 13: I'm assuming that it will take other HTML element types right? If yes maybe we should write it like this
```suggestion
- `register`: `React.ForwardedRef<HTML*Element>` (only used internally by the `Form` to keep track of the field's value)
```

src/components/form/Form.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+```tsx center
+<Form>
+  <InputField type="number" name="numCats" label="Number of cats" validation={{min: {value: 3, message: "You must have at least 3 cats!"}}}/>
+  <InputField name="faveName" label="Favourite cat name" validation={{maxLength: {{value: 20, message: "Cats hate long names!"}}}} />
+  <Button>Submit</Button>
+</Form>
+```
+
+If you pass one of those properties the `value` directly, you'll get a default error message:
```
andoulla, line 51: Wondering if down the line we should block this as we wouldn't be able to translate these defaults. but it is definitely a follow-up ticket we should add to the backlog to remind us

src/components/form/Form.mdx
```diff
@@ -0,0 +1,90 @@
...
+    validation={{
+      validate: (value) =>
+        ['A', 'E', 'I', 'O', 'U'].contains(value) || "That's not a vowel!"
+    }}
+  />
+  <Button>Submit</Button>
+</Form>
+```
+
+### Basic processing
```
andoulla, line 81: I don't fully understand this, lets walk through this during our call

src/components/form/Form.test.tsx
```diff
@@ -0,0 +1,16 @@
+import { render, screen } from '@testing-library/react'
+import { axe } from 'jest-axe'
+import * as React from 'react'
+
+import { Form } from '.'
+
+describe(`Form component`, () => {
+  it('renders a form', async () => {
```
andoulla, line 8: should we add a test for when an element without the `name` prop is pass? Could be good to test these  type of decisions

src/components/form/Form.tsx
```diff
@@ -0,0 +1,82 @@
...
+        const fieldError = errors[name] as ValidationError
+        return React.createElement(child.type, {
+          ...{
+            ...child.props,
+            error: fieldError ? getError(fieldError, validation) : undefined,
+            register: validation ? register(validation) : register,
+            key: name,
+            required: !!validation.required
+          }
+        })
```
andoulla, line 52: Wondering if it would be more readable like this, I had to look at it a couple of times to understand what is being spread where. But it might be a preference thing

```
const element = {
            ...child.props,
            error: fieldError ? getError(fieldError, validation) : undefined,
            register: validation ? register(validation) : register,
            key: name,
            required: !!validation.required
          }
return React.createElement(child.type, {
          ...element
        })

```

src/components/form/Form.tsx
```diff
@@ -0,0 +1,82 @@
...
+  validationOptions: ValidationOptions
+): string => {
+  console.log('error in getError:', error)
+  if (error.message) return error.message
+
+  switch (error.type) {
+    case 'required':
+      return 'This field is required'
+    case 'max':
+      return `The maximum allowed value is ${validationOptions.max}`
```
andoulla, line 71: do we need full stops after the messages as they are sentences? 🤔

src/components/form/Form.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+const getError = (
+  error: ValidationError,
+  validationOptions: ValidationOptions
+): string => {
+  console.log('error in getError:', error)
+  if (error.message) return error.message
+
+  switch (error.type) {
+    case 'required':
```
andoulla, line 68: Do we need to add something for pattern validation?

src/components/form/Form.tsx
```diff
@@ -0,0 +1,82 @@
...
+      return 'This field is required'
+    case 'max':
+      return `The maximum allowed value is ${validationOptions.max}`
+    case 'maxLength':
+      return `The maximum allowed length is ${validationOptions.maxLength}`
+    case 'min':
+      return `The minimum allowed value is ${validationOptions.min}`
+    case 'minLength':
+      return `The minimum allowed length is ${validationOptions.minLength}`
```
andoulla, line 77: I think it is worth covering this function with a test or two? Ie: getting one of the errors and maybe the default, any thoughts?

src/components/form/Form.tsx
```diff
@@ -0,0 +1,82 @@
...
+          }
+        })
+      })}
+    </StyledForm>
+  )
+}
+
+Form.displayName = 'Form'
+
+const getError = (
```
andoulla, line 60: Just for consistency with other files should this function be moved to the top?

src/components/form/ValidationOptions.ts
```diff
@@ -0,0 +1,35 @@
+// Returns `true` if a value passes validation, or a string with the error if not
+type Validator = (value: any) => boolean | string
+
+// Processes a value after validation
+type Processor = (value: any) => any
```
andoulla, line 5: Lets walk through this during our call, I don't fully get it

src/components/input-field/InputField.tsx
```diff
@@ -1,37 +1,43 @@
...
+type InputFieldProps = InputProps & {
+  label: string
+  error?: string
+  required?: boolean
+  name: string
+  register?: React.ForwardedRef<HTMLInputElement>
+  validation?: ValidationOptions
+}
+
+const RedSpan = styled('span', { color: '$danger' })
```
andoulla, line 19: Do `StyledAsteriskWrapper` or `StyledAsterisk` work? I think we don't really need the colour in the description, and if we decide to go with an uncoloured one we don't need to update the name

src/components/input/Input.tsx
```diff
@@ -40,21 +40,29 @@ const StyledInput = styled('input', {
...
-  }
+export const Input: React.FC<InputProps> = React.forwardRef(
+  ({ type = 'text', ...rest }, ref) => {
+    if (type === 'number') {
+      return (
+        <StyledInput
+          type="text"
+          inputMode="numeric"
+          pattern="[0-9]*"
+          {...rest}
```
andoulla, line 58: Very minor, could we move the `{...rest}`  after the ref. I believe it is nicer that away as I can read all the separated values and then it is everything else

src/components/form/Form.test.tsx
hollg: Oh yeah the tests are very far from finished! (Sorry, I meant to make this PR a draft!) I spoke to Marty after publishing this to get an idea of what would be a good testing strategy. I'm going to implement that today.

src/components/form/Form.tsx
hollg: `pattern` and `validate` fall through to the `default` branch because I didn't think there was a good way to write a more specific default error message for them. I'm happy to add branches for them if you think there are, though!

hollg: I also agree that it feels like we shouldn't actually allow this behaviour. I felt conflicted because until the main app is using TypeScript, we can't enforce that specific messages are provided, but maybe that's something we just need to catch at review until then.

src/components/input-field/InputField.tsx
hollg: Upudated 👍🏻

src/components/input/Input.tsx
hollg: updated 👍🏻

package.json
hollg: Updated. I'm actually not sure react should be in both dev dependencies and peer dependencies though. Possibly worth looking into later.

src/components/form/Form.tsx
andoulla: I'm really conflicted about this, as we can't translate these and if people let it go to default we could end up getting a fully translated form with just the errors in English

src/components/form/Form.mdx
hollg: Happy to talk through it in more depth, but basically these options allow the form to process `string` data into another type (`number` or `date`). Or you can provide a custom function to process it.

This is an option provided by `react-hook-form`. I'm not totally convinced that it's any more useful than manually processing the data in the `onSubmit` function, but I thought we might as well support it given that it's so easy to.

src/components/form/Form.mdx
hollg: Updated 👍🏻

src/components/form/Form.test.tsx
```diff
@@ -1,16 +1,39 @@
...
 import * as React from 'react'
 
+import { Button, InputField } from '../'
 import { Form } from '.'
 
 describe(`Form component`, () => {
   it('renders a form', async () => {
-    const { container } = render(<Form />)
+    const { container } = render(
+      <Form onSubmit={console.log}>
```
andoulla, line 12: I'm not sure about using `console.log` in tests. I had various problems in the past. Could we use `jest.fn()` or just `()=>()`

src/components/form/__snapshots__/Form.test.tsx.snap
```diff
@@ -1,10 +1,102 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Form component renders a form 1`] = `
+@keyframes sxwho2l {
+  0% {
+    opacity: 0;
+  }
+
+  30% {
```
andoulla, line 9: where are these keyframes coming from?

src/components/form/Form.test.tsx
hollg: oh yeah sure, I'll change that now

src/components/form/__snapshots__/Form.test.tsx.snap
hollg: Looks like the loading icon animation

andoulla: but loading icon isn't imported anywhere is it?

hollg: No it isn't. I can't explain why it's here haha.

andoulla: the good news is that our snapshots are working. we caught something!!!!

hollg: It's definitely from the `Loader`. Only appears in this snapshot and the snapshot for the `Loader` itself. I am completely baffled...

andoulla: it shows up in my MR too but at least I'm using the loader.

src/components/form/Form.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+export const Form: React.FC<FormProps> = ({
+  children,
+  defaultValues = {},
+  onSubmit,
+  ...remainingProps
+}) => {
+  const { errors, handleSubmit, register } = useForm({
+    defaultValues,
+    mode: 'onBlur'
```
andoulla, line 27: oh one more little thing, should this be a parameter so we could accommodate `onSubmit` too?

src/components/form/Form.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+    await screen.getByRole('form')
+
+    expect(container).toMatchSnapshot()
+    expect(await axe(container)).toHaveNoViolations()
+  })
+
+  it('passes error messages to fields', async () => {
+    render(
+      <Form onSubmit={console.log}>
```
andoulla, line 25:
```suggestion
      <Form onSubmit={jest.fn()}>
```

src/components/form/Form.tsx
hollg: I think I'd rather leave that until we need it. IMO our forms should behave consistently, so I think we'd need to find a good reason to put off validation in a particular case

andoulla: I'm a bit torn about this, as I agree, we need all the forms to behave consistently but also, I think it is important to keep the DS a bit generic, as different apps might have different needs. We can always come back to this, to be honest

package.json
```diff
@@ -81,7 +81,7 @@
     "npm-run-all": "^4.1.5",
     "parcel-bundler": "^1.12.4",
     "prettier": "^2.2.1",
-    "react": "17.0.1",
+    "react": "^16.8.0 || ^17",
```
andoulla, line 84: It's back! what is it that keeps switching it to the older version

src/components/form/Form.tsx
hollg: Ah yeah that is definitely a fair point. I'll change it to accept either `onSubmit` or `onBlur` but to default to `onBlur`.

andoulla (approved):

PR #68, opened 2021-03-04
feat: Button loading state

andoulla:
https://recordit.co/8nFPSGstjb

### DIFF
--- a/src/components/button/Button.test.tsx
+++ b/src/components/button/Button.test.tsx
@@ -1,4 +1,4 @@
-import { render, screen } from '@testing-library/react'
+import { fireEvent, render, screen } from '@testing-library/react'
 import { axe } from 'jest-axe'
 import * as React from 'react'
 
@@ -13,6 +13,7 @@ describe(`Button component`, () => {
     expect(container).toMatchSnapshot()
     expect(await axe(container)).toHaveNoViolations()
   })
+
   it('renders a outline button', async () => {
     const { container } = render(<Button appearance="outline">BUTTON</Button>)
 
@@ -43,5 +44,35 @@ describe(`Button component`, () => {
     expect(button).toBeDisabled()
     expect(container).toMatchSnapshot()
     expect(await axe(container)).toHaveNoViolations()
+  })
+  it('renders a loading button', async () => {
+    const { container } = render(<Button isLoading>BUTTON</Button>)
+
+    expect(await screen.queryByText('BUTTON')).not.toBeInTheDocument()
+    expect(await screen.getByRole('alert')).toBeInTheDocument()
+
+    expect(container).toMatchSnapshot()
+    expect(await axe(container)).toHaveNoViolations()
+  })
+  it('onClick is called when button is clicked', async () => {
+    const handleClick = jest.fn()
+    render(<Button onClick={handleClick}>BUTTON</Button>)
+
+    fireEvent.click(screen.getByRole('button'))
+
+    expect(handleClick).toHaveBeenCalled()
+  })
+
+  it('onClick is prevented if button is in loading state', async () => {
+    const handleClick = jest.fn()
+    render(
+      <Button onClick={handleClick} isLoading>
+        BUTTON
+      </Button>
+    )
+
+    fireEvent.click(screen.getByRole('button'))
+
+    expect(handleClick).toHaveBeenCalledTimes(0)
   })
 })
--- a/src/components/button/Button.test.tsx
+++ b/src/components/button/Button.test.tsx
@@ -5,8 +5,10 @@ import * as React from 'react'
 import { Button } from '.'
 
 describe(`Button component`, () => {
+  const props = { onClick: jest.fn() }
+
   it('renders a button', async () => {
-    const { container } = render(<Button>BUTTON</Button>)
+    const { container } = render(<Button {...props}>BUTTON</Button>)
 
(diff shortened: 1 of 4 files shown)


src/components/button/Button.tsx
```diff
@@ -127,13 +140,43 @@ const StyledButton = styled('button', {
...
+  {
+    isLoading?: boolean
+    onClick: () => void
+  }
+>
 
 export const Button: React.FC<ButtonProps> = ({
   theme = 'primary',
   appearance = 'solid',
+  isLoading = false,
```
hollg, line 155: I'm interested to hear your (and @thomasdigby's) thoughts on this, but this is the sort of prop where I think even though it's a boolean, it's nice to just name it after the state (i.e. `loading` rather than `isLoading`). IMO it feels more natural, perhaps because it matches `disabled`.

Adopting this pattern would mean slightly complicating the existing convention on boolean names. Personally I think it's worth it and easy enough to grasp/document, but it would be good to hear what other people think.

src/components/button/Button.tsx
```diff
@@ -127,13 +140,43 @@ const StyledButton = styled('button', {
   ]
 })
 
-type ButtonProps = React.ComponentPropsWithoutRef<typeof StyledButton> &
-  StitchesVariants<typeof StyledButton>
+type ButtonProps = Override<
```
hollg, line 143: Do we need `Override` here? Can we not just use a union of the two types (`&`)?

src/components/button/Button.tsx
```diff
@@ -127,13 +140,43 @@ const StyledButton = styled('button', {
...
+
+  return (
+    <StyledButton
+      className={isLoading ? 'isLoading' : ''}
+      theme={theme}
+      appearance={appearance}
+      onClick={() => handleClick(onClick)}
+      {...rest}
+    >
+      {isLoading ? <Loader /> : children}
```
hollg, line 177: This is perhaps just something to think about in the future (and to talk to Chema about) but it would be nice if we could keep some text in the button in the loading state so that the user can still see _what_ is in progress.

src/components/button/Button.tsx
```diff
@@ -60,6 +62,17 @@ const StyledButton = styled('button', {
     opacity: 0.35,
     cursor: 'not-allowed'
   },
+  '&.isLoading': {
+    cursor: 'not-allowed',
```
hollg, line 66: This is a cool idea!

src/components/button/Button.test.tsx
```diff
@@ -44,4 +45,34 @@ describe(`Button component`, () => {
     expect(container).toMatchSnapshot()
     expect(await axe(container)).toHaveNoViolations()
   })
+  it('renders a loading button', async () => {
```
hollg, line 48: Very minor point: could we re-order these tests so that anything related to the loading state is kept together?

src/components/button/Button.tsx
```diff
@@ -127,13 +140,43 @@ const StyledButton = styled('button', {
...
+
+    callback()
+  }
+
+  return (
+    <StyledButton
+      className={isLoading ? 'isLoading' : ''}
+      theme={theme}
+      appearance={appearance}
+      onClick={() => handleClick(onClick)}
```
hollg, line 174: It might be quicker to parse if we simplified this to something like `onClick={isLoading ? null : onClick}`. I _think_ it should have the same overall effect ...

src/components/button/Button.tsx
andoulla: I agree but I didn't want to deviate from the original too much. I read that having some descriptive text works well (like `<button isloading>Loading <loading icon></button>`)

I will keep it like this for now if that's ok

src/components/button/Button.tsx
andoulla: you are probably right, I will try it as soon as I have a moment

src/components/button/Button.tsx
andoulla: thanks

src/components/button/Button.test.tsx
andoulla: of course

andoulla: actually would it be better if we grouped them in a describe? that way we have a clearer grouping?

hollg: Oh, yes, that would make sense!

src/components/button/Button.tsx
hollg: Yeah absolutely!

src/components/button/Button.tsx
andoulla: I like having a distinction between props we are defining and HTML attributes. In this example, `disabled` is a valid HTML option and isLoading is something we are adding to the button.

src/components/button/Button.tsx
andoulla: for future reference: TS throws an error because of the `IntrinsicAttributes` definition of the onClick method

src/components/button/Button.tsx
hollg: Sorry, I don't think I follow. I don't understand how the names of any of our props would distinguish them from native HTML attributes. Is there a specific reason people might think `loading` is an HTML attribute?

hollg (approved):

PR #69, opened 2021-03-05
Feat: add password field

hollg:
Adds `PasswordField` component


src/components/password-field/PasswordField.tsx
```diff
@@ -0,0 +1,80 @@
...
+          mb: '$1',
+          fontFamily: 'sans',
+          fontSize: 'sm'
+        }}
+      >
+        {/* TODO: standardise asterisk -- should this be an option in the Label component? */}
+        <Label htmlFor={name}>{`${label} ${required ? '*' : ''}`}</Label>
+        {forgotPasswordURL && (
+          <Link href={forgotPasswordURL} css={{ fontSize: '$sm' }}>
+            Forgot your password?
```
andoulla, line 47: Should we make this copy for the link a prop that gets passed in? Once again we will run into translation issues if we dont

src/components/password-field/PasswordField.tsx
```diff
@@ -0,0 +1,80 @@
...
+        css={{
+          justifyContent: 'space-between',
+          alignItems: 'center',
+          mb: '$1',
+          fontFamily: 'sans',
+          fontSize: 'sm'
+        }}
+      >
+        {/* TODO: standardise asterisk -- should this be an option in the Label component? */}
+        <Label htmlFor={name}>{`${label} ${required ? '*' : ''}`}</Label>
```
andoulla, line 44: I think we could make this a nice util that can be imported into any component we want. 

Also there are some interesting thoughts for required fields on this https://www.deque.com/blog/anatomy-of-accessible-forms-required-form-fields/

src/components/password-field/PasswordField.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+  it('renders a password field', async () => {
+    const { container } = render(
+      <PasswordField css={{ m: 'auto', height: 100, width: 100 }} />
+    )
+
+    expect(container).toMatchSnapshot()
+    expect(await axe(container)).toHaveNoViolations()
+  })
+
+  it('toggles password visibility', async () => {
```
andoulla, line 18: Could we add a test for validating that the error is shown?

src/components/icon/Icon.tsx
```diff
@@ -21,7 +21,7 @@ type IconProps = Override<
   React.ComponentPropsWithoutRef<typeof StyledIcon>,
   {
     is: React.FC<React.SVGProps<SVGSVGElement>>
-    as: never
+    as?: never
```
andoulla, line 24: Just for my knowledge, why do we need this to be optional?

src/components/password-field/PasswordField.tsx
hollg: That's a really good point! Maybe this component should live in the main app, where it can be colocated with any language variants. That way, we could also define the URL as part of the component and we wouldn't have to pass it in every time we used it. What do you think?

src/components/password-field/PasswordField.tsx
hollg: That's a really good point! Maybe this component should live in the main app, where it can be colocated with any language variants. That way, we could also define the URL as part of the component and we wouldn't have to pass it in every time we used it. What do you think?

src/components/password-field/PasswordField.tsx
hollg: Oh yeah that is interesting! I didn't even know about the `required` HTML attribute. If we add that then I think we should be OK — do you agree?

src/components/icon/Icon.tsx
hollg: If it's required then I get a TS error for not providing it when I render an `<Icon />`. But there's no value you can pass that would satisfy the type `never`. (Maybe this isn't the best pattern, long-term!)

andoulla: hm, this really messes up with what I thought `never` does in TS. Up to this point, I thought it just set something to not be assignable

src/components/password-field/PasswordField.tsx
andoulla: Yes, of course, I didn't know it either to be honest until now

src/components/password-field/PasswordField.tsx
andoulla: We could move it but every time we keep agreeing that it is a good component to include in the DS. A different option would be to define the forgot password as 
```forgotPassword = { link: string, label: string}``` and in that case, it will accommodate having it in the DS system and decoupled enough from copy to not cause problems.

What do you think?

Comment (hollg) [reply src/components/password-field/PasswordField.tsx:47]: I'd never thought about this problem when we discussed it before haha. If we do keep it in the DS, then I think that pattern makes a lot of sense as it will make sure nobody passes a link without a label etc. 

The thing that makes me hesitant is that every time we use it in the main app, we'd be passing the same values in. Obviously one of the things we want to achieve with our components is to make the developer experience as pleasant as possible, and I think having to pass that data in manually even though it never changes would just kind of suck. So I think it would make sense to make a wrapper around it in the main app that automatically passes those values in. In which case, we end up with a component in the DS and one in the main app anyway. And at that point we're inviting a situation where nobody can remember whether to use the password field from the library, or whether there's an Atom-specific one etc.

The thing is, this logic can probably be applied to most or maybe all the form fields we might make that are any more complex than an input (or checkbox, or slider, etc) + a label. E.g. for the DOB field, we ideally wouldn't want to pass the same three labels in each time for the three inputs. 

I'm now leaning towards including those labels in the category of "Atom logic" that disqualifies a component from living in the DS library, but this is definitely something we should have a proper conversation about. I'll tag @thomasdigby here now so that he hopefully sees this thread, and then let's discuss it next week when we're all here.

Comment (andoulla) [reply src/components/password-field/PasswordField.tsx:47]: ok let's extend the standup on Monday to have this conversation.

File: src/components/password-field/PasswordField.tsx [hunk 5]
```diff
@@ -0,0 +1,86 @@
...
+
+  return (
+    <Box css={{ position: 'relative', ...(css as any) }}>
+      <Flex
+        css={{
+          justifyContent: 'space-between',
+          alignItems: 'center',
+          mb: '$1',
+          fontFamily: 'sans',
+          fontSize: 'sm'
```
thomasdigby, line 45: Are these font styles required here? Are they not already covered by `Label` and `Link`?

src/components/password-field/PasswordField.tsx
```diff
@@ -0,0 +1,86 @@
...
+        }}
+      >
+        {/* TODO: standardise asterisk -- should this be an option in the Label component? */}
+        <Label htmlFor={name}>{`${label} ${required ? '*' : ''}`}</Label>
+        {prompt && (
+          <Link href={prompt.link} css={{ fontSize: '$sm' }}>
```
thomasdigby, line 51: This could have `size="sm"` instead of `css={{ fontSize: '$sm' }}`

src/components/password-field/PasswordField.tsx
```diff
@@ -0,0 +1,86 @@
...
+            {prompt.label}
+          </Link>
+        )}
+      </Flex>
+
+      <Input
+        type={isPasswordVisible ? 'text' : 'password'}
+        css={{ mb: '$1', paddingRight: 24 }}
```
thomasdigby, line 66: Keen to use our shorthand keys where possible: `pr: 24`. Could we also include a comment to communicate why we're not using the scale? e.g. `// input padding to offset the exact size of the icon`

src/components/password-field/PasswordField.tsx
```diff
@@ -0,0 +1,86 @@
...
+      <Flex
+        css={{
+          justifyContent: 'space-between',
+          alignItems: 'center',
+          mb: '$1',
+          fontFamily: 'sans',
+          fontSize: 'sm'
+        }}
+      >
+        {/* TODO: standardise asterisk -- should this be an option in the Label component? */}
```
thomasdigby, line 55: Add a ticket to the component optimisation section 👍

thomasdigby (approved):

PR #73, opened 2021-03-09
chore: update version for npm release

andoulla:


package.json
```diff
@@ -84,7 +84,7 @@
     "npm-run-all": "^4.1.5",
     "parcel-bundler": "^1.12.4",
     "prettier": "^2.2.1",
-    "react": "17.0.1",
+    "react": "^16.8.0 || ^17",
```
thomasdigby, line 87: What's the reason for this change? This is a dev dependency so shouldn't affect how it's used in other projects

andoulla, line 87:
```suggestion
    "react": "17.0.1",
```

andoulla: I have reverted it, it randomly comes up and we don't know what's triggering it, to be honest. we had this problem last week

thomasdigby (approved):

PR #75, opened 2021-03-09
Feat: add list component

hollg:
Adds `List` component. Currently only supports unordered list with coloured bullets as that's what we need in the public Atom app. As soon as we have need/designs for an ordered list, it will be very easy to add with a `styled(ol, {/* ... */})` and a `type` prop for `List`.

Screenshot 2021-03-10 at 08 53 33


src/components/list/List.tsx
```diff
@@ -0,0 +1,48 @@
...
+          '&::marker': { color: '$primary500' }
+        }
+      },
+      secondary: {
+        [`& ${StyledListItem}`]: {
+          '&::marker': { color: '$secondary500' }
+        }
+      }
+    },
+    size: {
```
hollg, line 30: I added `size` as a variant because it `fontSize` affects the size of the bullets, so I thought it made sense to declare it once for the whole list. I'd be interested to hear any other thoughts on this though.

src/components/list/List.test.tsx
```diff
@@ -0,0 +1,19 @@
...
+      <List>
+        <List.Item>Item 1</List.Item>
+        <List.Item>Item 2</List.Item>
+      </List>
+    )
+
+    expect(container).toMatchSnapshot()
+    expect(await axe(container)).toHaveNoViolations()
+  })
+})
```
andoulla, line 19: Not sure but should we have an empty list test too? (I have been caught out by empty states in the past

Also, we could check that it renders 2 `List.Items` (not sure again)

src/components/list/List.tsx
```diff
@@ -0,0 +1,48 @@
...
+      sm: { fontSize: '$sm' },
+      md: { fontSize: '$md' },
+      lg: { fontSize: '$lg' }
+    }
+  }
+})
+
+type ListProps = React.ComponentProps<typeof StyledUL> & {
+  theme?: 'primary' | 'secondary'
```
andoulla, line 39: Quick q, but do we ever allow bullets that are the same colour as the text? Do we need to accommodate for that?

src/components/list/List.test.tsx
hollg: If we test that two `List.Item`s are rendered here, wouldn't we just be testing React? We're adding those manually ourselves; there's no logic in the `List` to render them.

Re a test for an empty `List`: what would we test?

src/components/list/List.tsx
hollg: No, not currently

src/components/list/List.mdx
```diff
@@ -0,0 +1,14 @@
+---
+title: List
+component: List
+description: List renders an unordered list
+---
```
andoulla, line 5: Should the category be added here? content is a good fit for it

src/components/list/List.tsx
andoulla: Great, I was just checking we didn't miss anything

src/components/list/List.test.tsx
andoulla: Maybe you are right, this way where it is just rending the children prob, there is no code in the component that handles the items. so yes maybe both my suggestions are redundant. I wasn't sure when I was asking it.

src/components/list/List.tsx
```diff
@@ -0,0 +1,48 @@
...
+      content: '"•"',
+      fontWeight: 'bold'
+    }
+  },
+  variants: {
+    theme: {
+      primary: {
+        [`& ${StyledListItem}`]: {
+          '&::marker': { color: '$primary500' }
+        }
```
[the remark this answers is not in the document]
hollg: I think it seems like it should work, but I can't get it to ...

There's some unintuitive rule about using `&` when "chaining" (that's what one of the Modulz people called it when I asked for help on their Discord yesterday to get this to work at all). 

Using this seems to behave the same way that my original `&`-less attempt did, but I can't find a way to add the missing `&` in to your suggestion without it all styling disappearing.

src/components/list/List.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import { styled } from '~/stitches'
+
+const StyledListItem = styled('li', {})
+
+const StyledUL = styled('ul', {
+  fontFamily: '$sans',
+  pl: '$2',
+  [`& ${StyledListItem}`]: {
+    pl: '$1',
```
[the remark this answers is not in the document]
hollg: Yeah, good question. I don't feel particularly strongly either way. I think there's a decent argument both for moving it (and thereby keeping the declaration for `StyledUL` smaller) and for keeping it here (and thereby keeping all the `::marker` shennanigans together, perhaps making it easier to grok what's going on).

Happy to move if you think makes more sense!

hollg:
@thomasdigby I've changed the `size` variants to match the styling of the `Text` components and updated the screenshot - could you take a look and let me know what you think?

src/components/list/List.mdx
hollg: Oh, yes, probably! I must have copied this from somewhere that was missing `category`. I'll add it now.

thomasdigby (approved):

PR #76, opened 2021-03-09
Fix: Minor documentation and component updates

thomasdigby:
Apologies, this is a bit of a mixed bag PR, hopefully easy enough to parse the changes I've made:
- Added categories to all components, "General" is placeholder for now
- Fixed `Button` `isLoading` prop in docs and made `onClick` an optional prop
- Allowed `aria-label` to be overwritten in `Form` and simplifed the passing of child props whilst removing the validation prop that was being added to the DOM
- Added `required` prop to the `Label` component as per our todo
- Moved margin to the `ValidationError` in our field components to ensure consistent spacing between form fields without errors
- Added a test for the `PasswordField` prompt
- Fixed a bug on the `Video` component where it wasn't rendering at all, required `width: 100%`


src/components/button/Button.mdx
```diff
@@ -66,5 +66,5 @@ Each variation has an `disabled` stated, by setting the `disabled` property.
 When using a button to request data or fire an action that has a potential delay, including a loading state to the button can be a useful indicator that stuff is happening. The `isLoading` prop must be a boolean value to activate the loading state.
 
 ```jsx live center
-<Button isloading> Hello world</Button>
+<Button isLoading> Hello world</Button>
```
andoulla, line 69: Thanks for catching this, it was my bad!

src/components/label/Label.test.tsx
```diff
@@ -13,4 +13,11 @@ describe(`Label component`, () => {
     expect(container).toMatchSnapshot()
     expect(await axe(container)).toHaveNoViolations()
   })
+  it('renders an asterisk if required', async () => {
+    const { container } = render(<Label required>TEXT</Label>)
+
+    await screen.getByText('TEXT')
+
```
andoulla, line 20: I'm not sure that this assesses effectively whether the asterisk is shown?

Would it be better if we added a ` expect(screen.getByText("*").toBeInDocument()`? I believe this way if it doesn't get rendered we get a more specific error.

src/components/label/Label.test.tsx
```diff
@@ -13,4 +13,11 @@ describe(`Label component`, () => {
     expect(container).toMatchSnapshot()
     expect(await axe(container)).toHaveNoViolations()
   })
+  it('renders an asterisk if required', async () => {
```
andoulla, line 16:
```suggestion

  it('renders an asterisk if required', async () => {
```

It is a small nitpick, sorry

src/components/password-field/PasswordField.test.tsx
```diff
@@ -15,6 +15,15 @@ describe(`Password component`, () => {
     expect(await axe(container)).toHaveNoViolations()
   })
 
+  it('renders with a prompt', async () => {
+    const { container } = render(
+      <PasswordField prompt={{ label: 'Hello', link: '/somewhere' }} />
+    )
+
```
andoulla, line 22: Should we add this too? 

`expect(screen.getByRole('link')).toBeInoDocument();`
or
`expect(screen.getByText('Hello').toBeInDocument()`

it will give us a more targeted assertion than the snapshot

src/components/popover/Popover.mdx
```diff
@@ -2,6 +2,7 @@
 title: Popover
 component: Popover
 description: Popover provides a styled actionable popover.
+category: General
```
andoulla, line 5:
```suggestion
category: Surfaces
```

As we discussed in the final Taxonomy doc

src/components/progress-bar/ProgressBar.mdx
```diff
@@ -2,6 +2,7 @@
 title: Progress Bar
 component: ProgressBar
 description: ProgressBar provides a styled progress bar.
+category: General
```
andoulla, line 5:
```suggestion
category: Feedback
```
as we discussed in the final Taxonomy
https://www.notion.so/atomlearning/Final-Proposal-49adc18286f04ae89cc67c9c32a78aae

src/components/radio/RadioButton.mdx
```diff
@@ -1,7 +1,7 @@
 ---
 title: Radio button
 component: RadioButton
-description: The `RadioButton` component implements the Radio component from [Radix](https://radix-ui.com/primitives/docs/components/radio-group) with default styling and the `css` prop.
+description: The RadioButton component implements the Radio component from Radix with default styling and the css prop.
 category: Forms
```
andoulla, line 5:
```suggestion
category: Inputs
```

src/components/button/Button.mdx
thomasdigby: No problem, just noticed it while updating the docs

src/components/radio/RadioButton.mdx
thomasdigby: Going to leave this in Forms for now, can review when the docs site is up

src/components/button/Button.tsx
```diff
@@ -142,7 +142,7 @@ const StyledButton = styled('button', {
 type ButtonProps = React.ComponentPropsWithoutRef<typeof StyledButton> &
   StitchesVariants<typeof StyledButton> & {
     isLoading?: boolean
-    onClick: () => void
+    onClick?: () => void
```
hollg, line 145: What's the use case for a `Button` with no `onClick`? Is this in anticipation of rendering as an `<a>` etc?

thomasdigby: Yeah exactly, we should be able to just render a button with a `to` but we haven't got there yet, can remove this.

src/components/input-field/InputField.tsx
```diff
@@ -29,19 +25,17 @@ export const InputField: React.FC<InputFieldProps> = ({
...
     <Input
-      css={{ mb: '$1' }}
       id={name}
       name={name}
       ref={register}
       {...(error && { state: 'error' })}
       {...remainingProps}
     />
-    {error && <ValidationError>{error}</ValidationError>}
+    {error && <ValidationError css={{ mt: '$1' }}>{error}</ValidationError>}
```
hollg, line 38: Is there a specific reason for adding margin to the top of the `ValidationError` rather than the buttom of the `Input`? If not, I think layouts tend to be easier to work with if we stick with one direction for margin on each axis (and we have generally used `mb` so far, I think)

src/components/button/Button.tsx
thomasdigby: Ah actually, just thought of one, `<Button>` within a `<Form>`

src/components/input-field/InputField.tsx
thomasdigby: Yep, that's to remove the unecessary `mb` space on the input when there isn't a error. We're trying to avoid leaking external layout styles that affect other sibling components and this is one of those.

We could look at adding the margin bottom on the input when there's an error but its simpler to add it this way.

src/components/button/Button.tsx
thomasdigby: Have removed for now, will add and change this behavior alongside the `to` `<a>` behavior in another PR.

src/components/input-field/InputField.tsx
hollg: Ohh, gotcha. Hadn't considered that sitch at all. Great point!

PR #78, opened 2021-03-11
Fix tweaks for field components

andoulla:
Adjusts the space between input and label to be the same as in the atom core


src/components/label/Label.tsx
```diff
@@ -10,6 +10,7 @@ const StyledLabel = styled('label', {
   fontWeight: 500,
   letterSpacing: '0.02em',
   lineHeight: '1.4',
+  margin: 0,
```
andoulla, line 13: I had to add this, as in `atom-core` we get some margin added from the general stylesheet

hollg (approved):

PR #79, opened 2021-03-11
Feat: polymorphic button

hollg:
Makes `Button` polymorphic by overriding its `as` prop to accept `React.ElementType` | `React.ComponentType`.

This allows us to wrap `Button` with logic to render as an `a` or `react-router` `Link` in `seahorse`, but is probably over-permissive. We should revisit this as we get more experience with realistic requirements and hopefully we can narrow the scope of the `as` prop.

I also updated the `onClick` behaviour so that no function is called at all if no `onClick` is passed. This is to ensure that the click handler in `Button` doesn't get in the way of the default behaviour of an `a`.

### DIFF
--- a/src/components/button/Button.test.tsx
+++ b/src/components/button/Button.test.tsx
@@ -65,6 +65,19 @@ describe(`Button component`, () => {
     expect(handleClick).toHaveBeenCalled()
   })
 
+  it.only('is polymorphic', async () => {
+    render(
+      <Button as="a" href="https://app.atomlearning.co.uk">
+        BUTTON
+      </Button>
+    )
+
+    const link = screen.getByRole('link')
+
+    expect(link).toBeInTheDocument()
+    expect(link).toHaveAttribute('href', 'https://app.atomlearning.co.uk')
+  })
+
   describe('Loading state', () => {
     it('renders a loading button', async () => {
       const { container } = render(
--- a/src/components/button/Button.tsx
+++ b/src/components/button/Button.tsx
@@ -3,6 +3,7 @@ import * as React from 'react'
 
 import { Loader } from '~/components/loader'
 import { styled } from '~/stitches'
+import { Override } from '~/utilities'
 
 const getButtonOutlineVariant = (baseColor: string, interactColor: string) => ({
   boxShadow: 'inset 0 0 0 2px',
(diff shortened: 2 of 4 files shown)


src/components/button/Button.test.tsx
```diff
@@ -65,6 +65,19 @@ describe(`Button component`, () => {
     expect(handleClick).toHaveBeenCalled()
   })
 
+  it.only('is polymorphic', async () => {
```
andoulla, line 68: let's not forget 
```suggestion
  it('is polymorphic', async () => {
```

src/components/button/Button.tsx
```diff
@@ -139,19 +140,22 @@ const StyledButton = styled('button', {
...
 })
 
-type ButtonProps = React.ComponentPropsWithoutRef<typeof StyledButton> &
+type ButtonProps = Override<
+  React.ComponentPropsWithoutRef<typeof StyledButton>,
   StitchesVariants<typeof StyledButton> & {
     isLoading?: boolean
-    onClick: () => void
+    onClick?: () => void
+    as: React.ComponentType | React.ElementType
```
andoulla, line 148: Should the component have only `button` and `a` as options? (if it is even possible, to be honest)

I just wonder what will happen if a user passes a `p`  which is technically an ElementType

src/components/button/Button.tsx
```diff
@@ -139,19 +140,22 @@ const StyledButton = styled('button', {
...
+>
 
 export const Button: React.FC<ButtonProps> = ({
   theme = 'primary',
   appearance = 'solid',
   isLoading = false,
   type = 'button',
   children,
-  onClick,
+  onClick = undefined,
```
andoulla, line 158: Why is this undefined? So I understand it better

src/components/button/Button.tsx
hollg: Yeah, this is what I was talking about  when I described this approach as "over-permissive" in the PR description. Unfortunately if we restrict it to `"button" | "a"` then we can render it as a `Link` from `react-router` (and we can't explicitly refer to `Link` without making `react-router` a dependency etc).

I definitely see this as the first step towards the final implementation rather than the final implementation itself.

src/components/button/Button.tsx
hollg: Oh, it doesn't need to be! I'm just in the habit of defaulting optional props, but the way it's being handled at line `176` means this isn't necessary. I've removed it.

src/components/button/Button.tsx
andoulla: ok I get it. We can catch any cases in the code reviews, to be honest. Or just add a line in the docs to ask people not to do it

andoulla (approved):
As this is just a step in the process, it looks good for now

thomasdigby:
@hollg Can we also make this change to the `Link` component?

thomasdigby (approved):

PR #81, opened 2021-03-11
Minor docs update

thomasdigby:
- Removing `center` option as all previews now center
- Changed one or two categories (all tbc just conforming to our current structure)

### DIFF
--- a/src/components/popover/Popover.mdx
+++ b/src/components/popover/Popover.mdx
@@ -7,7 +7,7 @@ category: Surfaces
 
 The `Popover` can display any type of element as a trigger and has the popover hidden by default.
 
-```jsx live center
+```jsx live
 <Popover aria-label="Password hints" content="Here is some content">
   <Button>Click me</Button>
 </Popover>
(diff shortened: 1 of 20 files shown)


src/components/popover/Popover.mdx
```diff
@@ -17,19 +17,21 @@ The `Popover` can display any type of element as a trigger and has the popover h
 
 Has 3 different `align` variants: `center` (default), `left` and `right`:
 
-```jsx live center
-<Popover aria-label="Password hints" content="Hello World"/>
-<Popover aria-label="Password hints" align="left" content="Hello World"/>
-<Popover aria-label="Password hints" align="right" content="Hello World"/>
+````jsx live
+<>
```
andoulla, line 21: This is good to know, that we need to use a fragment here

andoulla (approved):

src/components/popover/Popover.mdx
thomasdigby: Yeah I thought I had managed to get around this in the `CodeBlock` component in our docs site but it didn't work as I expected, so unfortunately it's necessary

PR #82, opened 2021-03-11
Fix: Optional `as` for `Button` and `Link`

thomasdigby:
These should be optional to allow standard component rendering e.g. `<Link href="http://hello.com">Hello</Link>`

### DIFF
--- a/src/components/button/Button.tsx
+++ b/src/components/button/Button.tsx
@@ -145,7 +145,7 @@ type ButtonProps = Override<
   StitchesVariants<typeof StyledButton> & {
     isLoading?: boolean
     onClick?: () => void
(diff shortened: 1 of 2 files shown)


andoulla (approved):
I'm approving, but I just have a thought. Do we want to set a default for the `as` prop? It will only affect readability, to be honest.

thomasdigby:
> I'm approving, but I just have a thought. Do we want to set a default for the `as` prop? It will only affect readability, to be honest.

I'm not sure, it might be adding extra functional complexity for the sake of highlighting one props value. Going to merge for now as I need this for the docs deploy but happy to discuss!

PR #85, opened 2021-03-16
Fix: Minor component updates

thomasdigby:
- Fixed `Textarea` padding
- Fixed `PasswordField` eye icon positioning
- Tweaked `ValidationError` size
- Fixed/updated some component previews in docs

Screenshot 2021-03-16 at 14 08 05
Screenshot 2021-03-16 at 14 08 29


hollg:
In the screenshots in the PR description, the `Textarea` isn't horizontally aligned with the `PasswordField`. Is that just a side-effect of them being in two separate images?

src/components/validation-error/ValidationError.tsx
```diff
@@ -1,4 +1,7 @@
+import * as React from 'react'
+
 import { Text } from '~/components/text'
-import { styled } from '~/stitches'
 
-export const ValidationError = styled(Text, { color: '$danger' })
+export const ValidationError = ({ css, ...remainingProps }) => (
+  <Text {...remainingProps} css={{ color: '$danger', ...css }} size="sm" />
```
hollg, line 6: I had assumed the previous version was functionally the same as this. What's the difference?

Also don't we want to type `ValidationError` as `React.FC`?

src/components/password-field/PasswordField.tsx
```diff
@@ -60,28 +60,30 @@ export const PasswordField: React.FC<PasswordFieldProps> = ({
...
+          onMouseDown={(e) => e.preventDefault()} // prevent focus being lost from password input
+          type="button"
+        >
+          <Icon
+            css={{ color: '$tonal700' }}
+            is={isPasswordVisible ? Eye : EyeOff}
+          />
+        </InvisibleButton>
+      </Box>
+      {error && <ValidationError css={{ mt: '$2' }}>{error}</ValidationError>}
```
andoulla, line 86: I think that maybe we need to add a test to assert that turning off the visibility works as expected, ie: the aria-label is correct, the icon changes and the input type is set to text. Considering how many little bits and bobs change it would be good to test it.

src/components/validation-error/ValidationError.tsx
thomasdigby: The minor difference is settings the `size` to small, otherwise yeah the previous example would have sufficed 👍

thomasdigby:
> In the screenshots in the PR description, the `Textarea` isn't horizontally aligned with the `PasswordField`. Is that just a side-effect of them being in two separate images?

This is just a side-effect of taking two separate images so alignment in the description will often be incorrect

src/components/password-field/PasswordField.tsx
thomasdigby: This is already in the test suite; and to clarify, I only added a wrapping `Box` around the input and button to allow the button to position itself correctly relative to the input

andoulla (approved):

PR #86, opened 2021-03-16
Feature: Automate new component creation

thomasdigby:
- Create a new utility to provide a step by step creation of a new component or utility
- Choose type of component (Component or Utility), the component name and the category (for documentation)
- It will create an `index.ts`, `Component.tsx`, `Component.test.tsx` and `Component.mdx`

https://user-images.githubusercontent.com/3226804/111365080-f4c31800-8689-11eb-8e31-86462b4ba02b.mov


scripts/add-component.mjs
```diff
@@ -0,0 +1,182 @@
...
+  if (confirmation && !fs.existsSync(directory)) {
+    // safe to do as we know the directory does not exist
+    fs.mkdirSync(directory)
+    // go and write files from templates with content
+    await Promise.all([
+      writeFile('index.ts', templates.index, response),
+      writeFile(`${name}.tsx`, templates.component, response),
+      writeFile(`${name}.test.tsx`, templates.test, response),
+      writeFile(`${name}.mdx`, templates.documentation, response)
+    ])
```
andoulla, line 174: Very tiny, but maybe a `catch` here would be a nice addition as we can handle how the error is displayed

scripts/add-component.mjs
```diff
@@ -0,0 +1,182 @@
+// eslint-disable-next-line
+import { default as changeCase } from 'change-case'
```
hollg, line 2: Is this different to `import changeCase from 'change-case'`?

hollg:
This is _so cool_!

hollg (approved):

scripts/add-component.mjs
thomasdigby: I was having a few issues with this import as `change-case` doesn't technically have a default import, node is shimming it. Reckon we can go back to that import statement though 👍

PR #87, opened 2021-03-16
Feature: Apply capsize to text

thomasdigby:
- Applied [`capsize`](https://seek-oss.github.io/capsize/) calculations to offset the unnecessary extra line-height with text
- Updated `line-height` values and added `capsize` to all sizes (apart from `sm`) to `Text`, `Heading`, `List`, `Label` and `Link` and ensure they're all consistent
- Added an additional smaller size to `Heading` as was necessary in the documentation and likely necessary within the main app

(Apologies for the many, many snapshot updates 😬)


src/components/link/Link.tsx
```diff
@@ -13,21 +15,8 @@ const StyledLink = styled('a', {
...
-        letterSpacing: '0.02em',
-        lineHeight: 1.4
-      },
-      lg: {
-        fontSize: '$lg',
-        letterSpacing: '0.02em',
-        lineHeight: 1.4
-      }
+      sm: textVariantSize.sm,
+      md: textVariantSize.md
```
andoulla, line 19: Just out of curiosity, why remove the `lg` size here?

src/utilities/style/capsize.ts
```diff
@@ -0,0 +1,15 @@
+import { CSS } from '~/stitches'
+
+// https://seek-oss.github.io/capsize/
+export const capsize = (margin: string): Record<string, CSS> => ({
```
andoulla, line 4: Wondering if we could standardise the margin values, so we can make them more consistent.
`
export const marginSizes= {
sm: -0.4489em,
md: -0.4489em,
......
`

This could also mean that when we are using capsize we don't have these very specific values everywhere.

src/components/password-field/PasswordField.tsx
```diff
@@ -47,10 +47,10 @@ export const PasswordField: React.FC<PasswordFieldProps> = ({
         css={{
           justifyContent: 'space-between',
           alignItems: 'center',
-          mb: '$1'
+          mb: '$2'
         }}
       >
-        <Label htmlFor={name} css={{ mb: '$2' }} required={required}>
```
andoulla, line 53: We need the margin here as it makes it consistent with the `InputField`. Also I made these to match what it is in `atom-core`

andoulla (changes requested):

src/components/password-field/PasswordField.tsx
thomasdigby: The margin needs to be solely applied to the containing `Box` to line up correctly with the prompt, otherwise you have margin on both label and container

src/utilities/style/capsize.ts
thomasdigby: It would be great to handle these values differently but they're very specifically tied to font size and line height, which is different to each size.

We also can't create the calculations ourselves as the font size value is stored in our theme and used in stitches, so we only have the key to use (rather than the value to calculate capsize with)

src/components/link/Link.tsx
thomasdigby: Ah I shouldn't have committed this, will change

src/components/password-field/PasswordField.tsx
andoulla: Does this means that this change needs to happen on the Inputfield then?

src/components/heading/Heading.tsx
```diff
@@ -1,19 +1,35 @@
...
-  lineHeight: '1.4',
   variants: {
     size: {
-      sm: { fontSize: '$lg', letterSpacing: '0.01em' },
-      md: { fontSize: '$xl', letterSpacing: '0.02em' },
-      lg: { fontSize: '$xxl', letterSpacing: '0.02em' }
+      xs: {
+        fontSize: '$md',
+        lineHeight: 1.625,
+        ...capsize('-0.4489em')
```
hollg, line 16: Were these values arrived at by trial and error, or is there some maths here that I don't understand?

src/components/password-field/PasswordField.tsx
thomasdigby: Nope, because it doesn't render a `prompt`, we can make that change in a separate PR if need be

src/components/heading/Heading.tsx
thomasdigby: They come from the capsize tool, I added a comment above the utility with the URL, I'll add some more detail to the comment

hollg (approved):

andoulla (approved):

PR #88, opened 2021-03-17
Feat: component updates for registration views

hollg:
* adds `mb` to `List.Item`s inside a `List`
* adds `fullWidth` variant to `Button`

### DIFF
--- a/src/components/list/List.tsx
+++ b/src/components/list/List.tsx
@@ -10,6 +10,7 @@ const StyledUL = styled('ul', {
   fontFamily: '$sans',
   pl: '$3',
   [`& ${StyledListItem}`]: {
+    mb: '$2',
     pl: '$2',
     '&::marker': {
       content: '"•"',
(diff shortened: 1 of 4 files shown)


src/components/list/List.tsx
```diff
@@ -10,6 +10,7 @@ const StyledUL = styled('ul', {
   fontFamily: '$sans',
   pl: '$3',
   [`& ${StyledListItem}`]: {
+    mb: '$2',
```
thomasdigby, line 13: Do we need to remove it from `:last-child`?

hollg: Yes, good catch! I don't suppose you know what combination of `&`s and `:`s I need to make that work in stitches? I assumed it would  be this, but no joy:

```
'&:not(:last-child)': {
      mb: '$2'
    }
```

thomasdigby: Huh I would have assumed that would work, let me have a quick check

hollg: Never mind — I just wasn't running `yarn dev`! Fixed now.

thomasdigby (approved):
Need to remove the sandbox changes, then good to go

PR #90, opened 2021-03-17
Feature: More documentation!

thomasdigby:
Added fresh words to our documentation.

- Moved most of `Readme` into `Introduction` and created a simpler readme for github/npm etc. without frontmatter
- Moved `Accessibility` and `Versioning` into their own pages
- Created `Core concepts` and `Styling` pages to give a more in-depth look at stitches and the concepts behind the design system components
- Excluding the readme and changelog from the documentation export

If there's anything that you vehemently disagree with, please feedback and comment on this PR. I'm keen to get this merged and deployed before Friday so let me know if there are any changes that could be deferred to a separate PR 👍 

Screenshot_2021-03-18 Screenshot


src/documentation/coreconcepts.mdx
```diff
@@ -0,0 +1,77 @@
+---
+title: Core concepts
+description: There are a few key concepts that underpin our component library, these should help with component composition, encapsulated styling and consistent UI
+category: Overview
+priority: 2
+---
+
+These components are intended to provide an abstraction over HTML elements and native components, providing an interface to ease creation of user interfaces in React. This library isn't comprehensive, but it should provide enough primitive, low-level components to be able to easily create pages, layouts, content, forms, as well as using them to compose more complex components with additional logic.
+
+Each component in the design system is built from a `styled()` function and this gives them access to a series of common props.
```
andoulla, line 10: Could we link the stitches.mdx here?

andoulla (commented):
I understand why the changelog is excluded, but I wonder if we should have somewhere showing the current version and any migration instructions. It is up for discussion though

thomasdigby:
> I understand why the changelog is excluded, but I wonder if we should have somewhere showing the current version and any migration instructions. It is up for discussion though

Yep agreed, will add a ticket 👍

src/documentation/coreconcepts.mdx
thomasdigby: That sentence is more in relation to the following `as` and `css` props, I can tweak this sentence to make that clear.

andoulla (approved):

src/documentation/coreconcepts.mdx
```diff
@@ -0,0 +1,77 @@
+---
+title: Core concepts
+description: There are a few key concepts that underpin our component library, these should help with component composition, encapsulated styling and consistent UI
```
hollg, line 3: This isn't quite grammatical. I'm not sure whether we specifically want these descriptions to be one sentence, so I'd suggest either changing to something like this if not:
```
There are a few key concepts that underpin our component library. These should help with component composition, encapsulated styling and consistent UI.
```

Or something like this if we do:

```
Our component library is underpinned by concepts that help with component composition, encapsulated styling and consistent UI
```

src/documentation/coreconcepts.mdx
```diff
@@ -0,0 +1,77 @@
...
+priority: 2
+---
+
+These components are intended to provide an abstraction over HTML elements and native components, providing an interface to ease creation of user interfaces in React. This library isn't comprehensive, but it should provide enough primitive, low-level components to be able to easily create pages, layouts, content, forms, as well as using them to compose more complex components with additional logic.
+
+Each component in the design system is built from a `styled()` function and this gives them access to a series of common props.
+
+## Semantic elements with `as`
+
+To apply the correct semantic element to one of our components, you can use `as` and pass in a string HTML element. For example, when using the `Heading` component you can pass in `h3` to use that element.
```
hollg, line 14: This doesn't seem to take into account components like `InputField` where multiple elements are rendered. Is it worth making this more specific?

src/documentation/coreconcepts.mdx
```diff
@@ -0,0 +1,77 @@
...
+<h3>Hello World</h3>
+```
+
+Some components have a restricted set of valid elements to pass into `as`, for example, `Heading` only allows `h1` through to `h6`, and some components remove the ability to change the HTML element altogether, like `<Select />` or `<Icon />`, which require a certain element to work correctly. Each of these restrictions are detailed on the component documentation under `API Reference`.
+
+Abstracting the underlying HTML from the component is beneficial as it allows us to control the document outline whilst preserving the ability to use a visually distinct style. This is often the case with container elements where you need to render a `section`, `aside`, `nav` or `main`, but still utilise a particular styled component, e.g. `<Panel as="aside" />`
+
+## Contextual styling with `css`
+
+The `css` prop is available on all of our components to allow you to style them. You can pass in an object of CSS key value pairs and the component will apply those styles as a class on the rendered component.
```
hollg, line 30: Could we add something about multi-element components and the fact that any styling that affects their internal spacing etc can't be overwritten because they get wrapped in a `Box` / `CSSWrapper` which receives the `css` prop?

src/documentation/stitches.mdx
```diff
@@ -0,0 +1,149 @@
+---
+title: Stitches
+description: Stitches is our styling library of choice, it leans on modern CSS best practices, has near-zero runtime, and provides a great developer experience
```
hollg, line 3:
```suggestion
description: Stitches is our styling library of choice(. I | because i)t leans on modern CSS best practices, has near-zero runtime, and provides a great developer experience
```

src/documentation/stitches.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+Stitches enables you to create shorthand utilities to improve the CSS authoring experience within `styled()` or `css`. We have provided an initial set of utilities that should be a useful starting point when composing components with a need to provide contextual layout or styles.
+
+A quick example using `bg` and `size` to create a little blue square.
+
+```jsx preview
+<Box css={{ bg: '$primary200', size: 100 }} />
+```
+
+The full list of utilties below
```
hollg, line 24: I feel slightly wary of reproducing the whole list here because it feels like we're setting ourselves up to forget to update it if we update the list in code. Could we give an example and then link to the code for the full list?

src/documentation/versioning.mdx
```diff
@@ -0,0 +1,36 @@
+---
+title: Versioning
+description: The project follows the SemVer standard [Semantic Versioning](https://semver.org/spec/v2.0.0.html) to make using the library easy.
```
hollg, line 3: I'm being _super_ picky here but we've got inconsistencies here about where descriptions end with a `.` or not. Can we make them consistent?

hollg (changes requested):

src/documentation/coreconcepts.mdx
andoulla: I had the same thought yesterday, but I didn't know if it would be duplicating the CSSWrapper doc. It does feel there is a gap here

src/documentation/coreconcepts.mdx
thomasdigby: Added to the second paragraph with `InputField` as an example

src/documentation/stitches.mdx
thomasdigby: I'm able to embed the _actual_ list of utilities here instead, but it's formatted production code, e.g. `m: () => (e) => ({ margin: e });`.

How do we feel about that compared to linking out to the file?

thomasdigby: It works well with the responsive conditions but less so with the utilities:

Screenshot 2021-03-19 at 10 12 37

andoulla: I don't mind it as our utils are quite simple at the moment. They don't do anything long or complicated for the formating to be a dealbreaker

thomasdigby: Scrapped both, just copying in the conditions for now.

hollg (approved):

PR #91, opened 2021-03-18
Fix: Link capsize

thomasdigby:
Make capsize conditional on text size variants

### DIFF
--- a/src/components/text/Text.tsx
+++ b/src/components/text/Text.tsx
@@ -1,9 +1,12 @@
 import * as React from 'react'
 
-import { styled } from '~/stitches'
+import { CSS, styled } from '~/stitches'
 import { capsize, Override } from '~/utilities'
 
-export const textVariantSize = {
+export const textVariantSize = ({ applyCapsize = true } = {}): Record<
+  'sm' | 'md' | 'lg',
+  CSS
+> => ({
   sm: {
     fontSize: '$sm',
     lineHeight: 1.53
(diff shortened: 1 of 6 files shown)


hollg (approved):

src/components/text/Text.tsx
```diff
@@ -1,24 +1,27 @@
 import * as React from 'react'
 
-import { styled } from '~/stitches'
+import { CSS, styled } from '~/stitches'
 import { capsize, Override } from '~/utilities'
 
-export const textVariantSize = {
+export const textVariantSize = ({ applyCapsize = true } = {}): Record<
+  'sm' | 'md' | 'lg',
```
andoulla, line 7: Just out of curiosity, how come this isn't in the util, since it is the thing we reuse in the other components?

andoulla (approved):

src/components/text/Text.tsx
thomasdigby: Sorry @andoulla I didn't see this before, the `sm`, `md` or `lg` keys that are returned are specific to the text variants, not the capsize calculation utility. So when other components want to replicate the text sizes used by `Text`, they can import `textVariantSize` (which comes with `capsize`). The benefit of this is collating the line height, font size, letter spacing, font weight into a single place instead of duplicating it across components.

I think the group text styles are separate enough to the capsize utility (you could use either without them depending on each other) to store them separately and not couple them together.

PR #92, opened 2021-03-19
Fix: form validation

hollg:
Fixes `Form` so that validation works properly in custom field components. Also documents changes with examples of custom form field composition.


src/components/input-field/InputField.tsx
```diff
@@ -22,21 +25,33 @@ export const InputField: React.FC<InputFieldProps> = ({
...
-  </Box>
-)
+}) => {
+  let ref
+  if (register && validation) {
+    ref = register(validation)
+  } else if (register) {
+    ref = register
+  } else {
+    ref = undefined
```
andoulla, line 37: we could do `let ref = undefined` and then remove lines 36-38. It think it reads better but it could be just me

andoulla (approved):

src/components/form/Form.tsx
```diff
@@ -47,9 +47,12 @@ export const Form: React.FC<FormProps> = ({
         return React.createElement(child.type, {
           ...childProps,
           error: fieldError ? fieldError.message : undefined,
-          register: validation ? register(validation) : register,
+          register: register,
           key: childProps.name,
-          required: validation ? !!validation.required : false
+          // ensure that any field marked as required in the validiation object
```
thomasdigby, line 52:
```suggestion
          // ensure that any field marked as required in the validation object
```

src/components/input-field/InputField.tsx
hollg: Oh yeah I think you're right. I hate having to use `let` at all to be honest, but the nested ternary alternative was gross. I think your suggestion is the lesser evil! I'll update it now.

thomasdigby: I agree with this 👍

src/components/form/Form.tsx
hollg: Lol, good catch. Thanks!

src/components/input-field/InputField.tsx
andoulla: I struggle with nested ternaries, to be honest. I think if it can't be a simple ternary then we should just use an if condition. That's why they exist anyway.

hollg: Annoyingly TS throws a hissy fit about assigning `register` or `register(validation)` to `undefined`. I've updated it to 

```
 let ref
  if (register && validation) {
    ref = register(validation)
  } else if (register) {
    ref = register
  }
```

... which leaves the `undefined` implicit but compiles.

PR #96, opened 2021-03-23
Feature: Upgrade stitches

thomasdigby:
- Upgrade stitches to latest beta (pretty sure this is the release candidate, aside from hotfixes)
- ~Upgraded Radix packages to include new hotfixes (aside from radiogroup which has an issue when testing)~
- Fixed a TS issue in `PopOver` (we didn't need to spread content and root types into the prop types)
- Removed a storybook utility
- Added `"skipLibCheck": true` to our TS config to bypass some internal TS errors within stitches

The snapshot changes seem to be:
- changed how classes are named
- some class duplication merged into a single selector
- prefixing `appearance`

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -10,6 +10,7 @@
     "lib": ["es6", "dom"],
     "types": ["node", "jest"],
     "baseUrl": "./src",
+    "skipLibCheck": true,
     "paths": {
       "~/*": ["./*"] // allow imports from '~/components', but not '~'
     },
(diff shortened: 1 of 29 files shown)


tsconfig.json
```diff
@@ -10,6 +10,7 @@
     "lib": ["es6", "dom"],
     "types": ["node", "jest"],
     "baseUrl": "./src",
+    "skipLibCheck": true,
```
hollg, line 13: This sounds v promising for unblocking a bunch of Radix stuff!

hollg (approved):

PR #99, opened 2021-03-24
feat: add carousel component

hollg:
Adds the `Carousel` component. The library we're using requires a CSS file to be imported, so this PR also includes a couple of changes to allow that import in jest and rollup.

https://user-images.githubusercontent.com/21319237/112289516-48cf8d00-8c86-11eb-9aa5-4ed2ca46a44b.mov


src/components/carousel/Carousel.tsx
```diff
@@ -0,0 +1,54 @@
+import 'pure-react-carousel/dist/react-carousel.es.css'
+
+import { CarouselProvider, Slider } from 'pure-react-carousel'
+import * as React from 'react'
+
+import { Box } from '~/components/box'
+import { Flex } from '~/components/flex'
+
+import { Dots } from './Dots'
```
andoulla, line 9: Just for my understanding, How come the Dots are separated out? I understand separating the Slide as the client will need it to compose the Carousel. but the dots are always going to be there and I can't think of a use case where we would need to add them separately (sorry if I'm missing something

hollg: No you're not missing anything! I just thought this file was getting kind of long and that it might make the feature easier to understand if this file/component just "brings everything together". Do you think it would be clearer if `Dots` were just written inline? I'm very happy to change this to whatever is simplest to read!

andoulla: No, it is ok, if it just for readability I get that. I was just wondering if I'm missing something

andoulla:
Can we please remove this file `src/components/icon/.DS_Store`

hollg:
@andoulla great catch! Done.

src/components/carousel/Slide.tsx
```diff
@@ -0,0 +1,29 @@
...
+const StyledSlide = styled(BaseSlide, {
+  opacity: '0.7',
+  transition: 'all 0.25s ease-in',
+  '&[aria-selected="true"]': {
+    opacity: 1,
+    // apply box shadow to the Slide content directly,
+    // not the wrappers added internally
+    '& > div > div > *': {
+      boxShadow: '$1'
+    }
```
thomasdigby, line 16: Do we want to bake these styling decisions into the lib itself? Or defer that to the consumer?

hollg: It's pretty complex to apply styles to the selected slide's content as a consumer because of the need to reference the class you can see here and the nesting that goes on to put padding in the right places etc.

I figured that a soon as radix's carousel lands, we'll want to use that one anyway since this library's implementation is not very friendly to the composition model that we're trying to push (you can see the result of this in the version in the main app, which has to accept about five different classNames at the top level and distribute them internally).

Given the above, I've basically been aiming to build something that meets our requirements for the public app without getting into variants or allowing any control that we don't need yet, just to avoid spending lots of time on something we'll bin in the near future. That said, if you think I've missed things we _do_ need then obviously I'll add them!

If we do want options on the styling of the selected / deselected slides, I think the approach that avoids the most fiddly styling would be to have variants of either the `Carousel` or the `Slide`. There could be a variant to control whether opacity changes and one to control whether the box-shadow appears, maybe? Interested to hear your thoughts!

src/components/carousel/Carousel.tsx
```diff
@@ -0,0 +1,54 @@
...
+    >
+      <Flex css={{ flexDirection: 'column' }}>
+        <Box
+          as={Slider}
+          css={{
+            ml: '50%',
+            cursor: 'grab',
+            // override default styling from the carousel library
+            overflow: 'visible !important',
+            transform: 'translateX(-50%)'
```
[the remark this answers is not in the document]
hollg: My reply to your comment about the box-shadow probably applies here too. Short answer: they're required for every use-case that the existing component in the main app is used in.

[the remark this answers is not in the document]
thomasdigby: Could we spread `css` here to give the consumer to ability to add these specific styles?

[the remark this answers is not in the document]
thomasdigby: Default to `cursor` and `overflow`

src/components/carousel/Slide.tsx
thomasdigby: I'm a bit wary of this approach as it'll mean reverting these changes in future but I understand why this is preferable to hard-coding it all in the main app if there's not an easy way to add styling to the carousel or slides.

Although looking through the component again I wonder if we could spread `css` into `StyledSlide` and copy these styles over to the main app? We're not actually using any complex classes, just a few selectors that we can document for now, we would avoid the need to drastically change styles in a future version as this initial one is so simple.

hollg: I've made that change locally now and this is what we end up with when using the `Carousel`. It's not _awful_, but it will be used in both existing places that we have a carousel. Do you think it's OK?

```
<Carousel
  slideWidth={200}
  slideHeight={300}
  name="Example carousel"
  type="overflow"
  >
    {[0, 1, 2, 3].map((num) => (
      <Carousel.Slide
      key={num}
      index={num}
      css={{
        opacity: '0.7',
        transition: 'all 0.25s ease-in',
        '&[aria-selected="true"]': {
          opacity: 1,
          // apply box shadow to the Slide content directly,
          // not the wrappers added internally
          '& > div > div > *': {
            boxShadow: '$1'
           }
         }
       }}
      >
        <Image src="https://placekitten.com/200/300" alt="" />
     </Carousel.Slide>
  ))}
 </Carousel>

```

src/components/carousel/Carousel.tsx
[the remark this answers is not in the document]
hollg: I've since realised that `transform` and `ml` are integral to making it look anything like a carousel, rather than being to do with specific styling for a particular use case

src/components/carousel/Slide.tsx
thomasdigby: Yeah I think so? It looks a bit nasty but it's better for us to defer those styles to the consumer. They could wrap those up in `styled(Carousel.Slide)` to make it a bit nicer. The nested selector isn't _great_ but I'd probably (🤷‍♂️?) apply those styles to the `Image` component itself. e.g.

```jsx
const Slide = styled(Carousel.Slide, {
  opacity: 0.7,
  transition: "all 0.25s ease-in",
  '&[aria-selected="true"]': { opacity: 1 },
});

const Thing = () => (
  <Carousel
    slideWidth={200}
    slideHeight={300}
    name="Example carousel"
    type="overflow"
  >
    {[0, 1, 2, 3].map((num) => (
      <Slide key={num} index={num}>
        <Image
          src="https://placekitten.com/200/300"
          alt=""
          css={{
            '[aria-selected="true"] &': {
              boxShadow: "$1",
            },
          }}
        />
      </Slide>
    ))}
  </Carousel>
);
```

src/components/carousel/Slide.tsx
```diff
@@ -0,0 +1,29 @@
+import { Slide as BaseSlide } from 'pure-react-carousel'
+import * as React from 'react'
+
+import { Box } from '~/components'
+import { styled } from '~/stitches'
+
+const StyledSlide = styled(BaseSlide, {
+  // opacity: '0.7',
```
andoulla, line 8: let's remove the commented our code

src/components/icon/svg/index.ts
```diff
@@ -1,4 +1,6 @@
 export { ReactComponent as Check } from './check.svg'
+export { ReactComponent as ChevronLeft } from './chevron-left.svg'
```
andoulla, line 2: Should these be migrated to use the icons from Ikonate?

src/components/carousel/__snapshots__/Carousel.test.tsx.snap
```diff
@@ -0,0 +1,363 @@
...
+  }
+
+  80% {
+    opacity: 0;
+  }
+
+  100% {
+    opacity: 0;
+  }
+}
```
andoulla, line 24: I thought I added a comment before but I must have forgotten to submit it. We are getting the same issue with the form component here. Where the Loader keyframes are showing up

src/components/carousel/Carousel.tsx
```diff
@@ -0,0 +1,78 @@
...
+                },
+                '& div[class*="slide_"]': {
+                  float: 'left',
+                  pb: '0 !important'
+                },
+                '& div[class*="slideInner"]': {
+                  display: 'flex',
+                  justifyContent: 'center'
+                }
+              }}
```
thomasdigby, line 61: Can we break this out into a `Slider` styled component? I think this will be slightly more readable if it's not used within a `css` prop

src/components/carousel/Slide.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+const StyledSlide = styled(BaseSlide, {})
+
+type SlideProps = React.ComponentProps<typeof StyledSlide> & { index: number }
+
+export const Slide: React.FC<SlideProps> = ({ children, ...rest }) => {
+  return (
+    <StyledSlide {...rest} tag="div">
+      {/* padding to create the gap between slides */}
+      <Box css={{ px: '$3' }}>{children}</Box>
```
thomasdigby, line 15: Can we apply this padding to the `StyledSlide` component? What's the benefit of nesting it within a `Box`?

package.json
```diff
@@ -93,6 +93,7 @@
     "react-dom": "^17.0.1",
     "rollup": "^2.39.0",
     "rollup-plugin-bundle-size": "^1.0.3",
+    "rollup-plugin-embed-css": "^1.0.21",
```
thomasdigby, line 96: This plugin and any changes to our testing set-up to support CSS are no longer relevant right?

thomasdigby (commented):
I think this component might be more useful as a series of exports for the individual parts, that can then be composed together by the consumer. e.g.

```jsx
<Carousel>
  <Carousel.Slider>
    <Carousel.Slide />
    <Carousel.Slide />
    <Carousel.Slide />
  </Carousel.Slider>
  <Carousel.Pagination />
</Carousel>
```

hollg:
@thomasdigby 

> I think this component might be more useful as a series of exports for the individual parts, that can then be composed together by the consumer. e.g.
> 
> ```js
> <Carousel>
>   <Carousel.Slider>
>     <Carousel.Slide />
>     <Carousel.Slide />
>     <Carousel.Slide />
>   </Carousel.Slider>
>   <Carousel.Pagination />
> </Carousel>
> ```

What would be the usecase for doing anything other than putting the components in exactly that order?

thomasdigby:
> @thomasdigby
> 
> > I think this component might be more useful as a series of exports for the individual parts, that can then be composed together by the consumer. e.g.
> > ```js
> > <Carousel>
> >   <Carousel.Slider>
> >     <Carousel.Slide />
> >     <Carousel.Slide />
> >     <Carousel.Slide />
> >   </Carousel.Slider>
> >   <Carousel.Pagination />
> > </Carousel>
> > ```
> 
> What would be the usecase for doing anything other than putting the components in exactly that order?

You can remove the pagination completely, align the arrows with the pagination, tweak the spacing around the arrows, change the responsive styles, etc. etc.

We're baking in a lot of assumptions (including responsive styles) with our component that we could remove just by creating the individual parts and exporting them for the consumer to decide how best to use. I think it's better to allow the consumer to literally just put the components together in that order and gain much more flexibility in any future usecase where it doesn't directly fit this design.

src/components/carousel/Slide.tsx
hollg: Applying it directly to `StyledSlide` breaks the carousel by sometimes forcing some items onto a second line. I couldn't tell you why. `Slide` seems to do a lot both in CSS and JS so it's kind of fragile.

src/components/carousel/Carousel.tsx
```diff
@@ -0,0 +1,67 @@
...
+  Arrows: typeof Arrows
+  Pagination: typeof Dots
+  Slide: typeof Slide
+  Slider: typeof Slider
+} = ({
+  children,
+  name,
+  slideHeight,
+  slideWidth,
+  type = 'arrows',
```
thomasdigby, line 35: Is `type` relevant any more?

hollg: That's TBD. I haven't updated the props yet because I'm not 100% certain whether everything works as before without them being used. Testing it atm!

hollg: Everything seems to work! I've updated this (and hopefully everywhere that references it)

src/components/carousel/Carousel.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+import * as React from 'react'
+
+import { Box, Flex, Image } from '../'
+import { Carousel } from '.'
+
+describe('Carousel component', () => {
+  it('renders', async () => {
+    const { container } = render(
+      <Carousel slideWidth={200} slideHeight={300} numSlides={4}>
+        <Flex css={{ flexDirection: 'column' }}>
```
thomasdigby, line 12: Could we add the `css` to the `Carousel` component here? To simplify it a tad?

src/components/carousel/Arrows.tsx
```diff
@@ -0,0 +1,62 @@
...
+  ...buttonStyles,
+  right: '$2',
+  // moving position:absolute to buttonStyles causes TS errors
+  position: 'absolute',
+  '@md': {
+    right: '$3'
+  },
+  '@lg': {
+    right: '$4'
+  }
```
thomasdigby, line 50: I think we should move these responsive styles out into the consumer layer, potentially even the positioning of them. What we gain by exporting these components separately is the ability to flexibly style each component part so that it suits its context, in some cases (not relevant for the public app but for potentially for others) we may use a carousel in a constrained width container across all breakpoints and won't need these specific styles for changing the arrow position.

What do you reckon? We add more complexity to the integration layer, but we gain the ability to be much more flexible with it and use it in a number of different ways in the consumer layer.

thomasdigby (commented):
This is looking really great, just need to confirm our approach to flexible styling and how that'll effect some of the components 👍

src/components/carousel/Carousel.test.tsx
hollg: Haha yes, that's much better. I knew there was a reason I added the wrapper to the `Carousel` ...

src/components/carousel/Arrows.tsx
hollg: I definitely agree that, given we're exporting the `Carousel`'s constitutent parts separately now, it would make sense to style these directly. But I don't know what the nicest API would be for them / where we as consumers should apply the styling. I'll play around with it!

hollg: This is done now. The example in `Carousel.mdx` is updated to show how it works, let me know what you think!

src/components/carousel/Carousel.mdx
```diff
@@ -0,0 +1,75 @@
...
+      left: '$2',
+      '@md': {
+        left: '$3'
+      },
+      '@lg': {
+        left: '$4'
+      }
+    }}
+  />
+  <Carousel.ArrowNext
```
andoulla, line 38: Just wondering if it makes more sense to have the `ArrowNext` after the `Slider`. as it is where it gets displayed and that's where I would expect to find it. 

It is minor thought, for better readability

andoulla (approved):

PR #100, opened 2021-03-25
Feature: Add icons and upgrade theme

thomasdigby:
- Add `@atom-learning/icons` and replace all icons in use
- Upgrade `@atom-learning/theme` and update tokens where they've changed
- Remove `jest` & `rollup` packages handling `.svg` as we're deferring that to the icons lib
- Update `Icon` component styles/markup to match new icon lib
- Lock Radix package versions
- Minor updates like renaming `{...props}` to `{...remainingProps}` where necessary, update to implicit returns where possible


sandbox/index.tsx
```diff
@@ -1,14 +1,57 @@
...
+} from '../dist'
 globalCss(reset)()
 
 const App = () => {
-  return <Box />
+  return (
+    <Flex
+      css={{
+        flexDirection: 'column',
+        gap: '$4',
```
hollg, line 30: 😍

hollg (approved):

sandbox/index.tsx
thomasdigby: Ah yeah I forgot to remove this, but yeah `gap` hey   * swoon *

PR #101, opened 2021-03-26
Release: Beta

thomasdigby:


package.json
```diff
@@ -3,7 +3,7 @@
   "source": "src/index.ts",
   "main": "dist/index.js",
   "module": "dist/index.modern.js",
-  "version": "0.1.0-alpha.5.1",
+  "version": "0.1.0",
```
andoulla, line 6: shouldn't this be `version: "0.1.0-beta"

thomasdigby: I'm following stitches convention here, we're almost stable so a minor prerelease bump `0.1.0` makes sense to me.

Further updates can be either `minor` or `patch` version bumps now

andoulla: I'm not 100% on conventions here, but I thought that you go from alpha to beta to then having version one ready.
https://stackoverflow.com/questions/51373151/releasing-alpha-to-beta-to-production-with-semver/51432806

thomasdigby: As this is a stable prerelease version with most of the necessary fundamental functionality built it makes sense to release as a prerelease feature `0.1.0`. I don't think the `beta` tag is useful or necessary now, the lib is in a usable state but is not finalised, hence the `0` major version.

I'll be looking to update `@atom-learning/theme` to this soon as well

thomasdigby: If I wasn't clear before (sorry) `beta` isn't necessary, as you probably read from your linked post, loads of variants of semver prerelease versions are valid. From my perspective, `alpha` is unstable & prone to change, `beta` is more stable but could have large changes still, and prerelease versions are stable, likely fewer changes, potentially requiring a locked version.

andoulla: I got that from your first reply hence the 👍

PR #104, opened 2021-03-29
Feature: Stack component

thomasdigby:
- Create new `Stack` component for single axis layouts


src/components/stack/Stack.tsx
```diff
@@ -0,0 +1,57 @@
...
+      }
+    }
+  }),
+  {}
+) as Record<keyof typeof theme.space, CSS>
+
+const StyledStack = styled('div', {
+  display: 'flex',
+  flexWrap: 'wrap',
+  width: 'max-content',
```
andoulla, line 25: sorry If I'm missing something, do we need to set the width? don't we want it to be controlled by the parent or passed it. 

I guess it can be overridden so it is just a wondering

thomasdigby: You're right, but `max-content` in this case just ensures that the container doesn't grow larger than the children. `Stack` is purely for layout purposes so any width that needs to be assigned to the container should be added to a another container along with any other necessary styles.

hollg (approved):
This is dope!

andoulla (approved):

PR #105, opened 2021-03-29
Feat: update grid component

hollg:
Updates `Grid` to implement [fukol grid](https://github.com/Heydon/fukol-grids)

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "parcel-bundler": "^1.12.4",
     "prettier": "^2.2.1",
     "prompts": "^2.4.0",
-    "react": "17.0.1",
+    "react": "^16.8.0",
     "react-docgen-typescript": "^1.21.0",
     "react-dom": "^17.0.1",
     "rollup": "^2.39.0",
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "parcel-bundler": "^1.12.4",
     "prettier": "^2.2.1",
     "prompts": "^2.4.0",
-    "react": "^16.8.0",
+    "react": "17.0.1",
     "react-docgen-typescript": "^1.21.0",
     "react-dom": "^17.0.1",
     "rollup": "^2.39.0",
--- a/sandbox/index.tsx
+++ b/sandbox/index.tsx
@@ -4,19 +4,11 @@ import * as React from 'react'
 import * as ReactDOM from 'react-dom'
 import { reset } from 'stitches-reset'
 
-import { Box, globalCss, Grid, Image } from '../dist'
+import { Box, globalCss } from '../dist'
 globalCss(reset)()
 
 const App = () => {
-  return (
-    <Box>
-      <Grid basis="0.5em" gap={1}>
-        {[0, 1, 2, 3].map((num) => (
-          <Image key={num} src="https://placekitten.com/200/300" alt="" />
-        ))}
-      </Grid>
-    </Box>
-  )
+  return <Box />
 }
 
 ReactDOM.render(<App />, document.getElementById('root'))
--- a/sandbox/index.tsx
+++ b/sandbox/index.tsx
@@ -10,7 +10,7 @@ globalCss(reset)()
 const App = () => {
   return (
     <Box>
-      <Grid basis="0.5em" gap="0.05em">
+      <Grid basis="0.5em" gap={1}>
         {[0, 1, 2, 3].map((num) => (
           <Image key={num} src="https://placekitten.com/200/300" alt="" />
         ))}
--- a/sandbox/index.tsx
+++ b/sandbox/index.tsx
@@ -4,11 +4,19 @@ import * as React from 'react'
 import * as ReactDOM from 'react-dom'
 import { reset } from 'stitches-reset'
 
-import { Box, globalCss } from '../dist'
+import { Box, globalCss, Grid, Image } from '../dist'
 globalCss(reset)()
 
 const App = () => {
-  return <Box />
+  return (
+    <Box>
+      <Grid basis="0.5em" gap="0.05em">
+        {[0, 1, 2, 3].map((num) => (
+          <Image key={num} src="https://placekitten.com/200/300" alt="" />
+        ))}
+      </Grid>
+    </Box>
+  )
 }
 
 ReactDOM.render(<App />, document.getElementById('root'))
--- a/src/components/grid/Grid.mdx
+++ b/src/components/grid/Grid.mdx
@@ -7,6 +7,8 @@ category: Layout
 The `Grid` component implements a simple flex-box based grid as described [here](https://github.com/Heydon/fukol-grids).
 Its `gap` prop specifies the distance between items in the grid using the theme's `space` scale. Its `minItemSize` prop specifies, in any valid CSS units, the minimum size of each grid item (by default this will set their `width`, but if the `css` prop is used to set `flex-direction` to `column` then it will set `height` instead).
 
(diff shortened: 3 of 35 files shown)


package.json
```diff
@@ -88,7 +88,7 @@
     "parcel-bundler": "^1.12.4",
     "prettier": "^2.2.1",
     "prompts": "^2.4.0",
-    "react": "17.0.1",
+    "react": "^16.8.0",
```
andoulla, line 91: how come we are going back a version?

sandbox/index.tsx
```diff
@@ -4,11 +4,19 @@ import * as React from 'react'
 import * as ReactDOM from 'react-dom'
 import { reset } from 'stitches-reset'
 
-import { Box, globalCss } from '../dist'
+import { Box, globalCss, Grid, Image } from '../dist'
 globalCss(reset)()
 
 const App = () => {
-  return <Box />
+  return (
```
andoulla, line 11: Just a reminder not to commit these changes

src/components/grid/Grid.mdx
```diff
@@ -4,24 +4,13 @@ component: Grid
 category: Layout
 ---
 
-The `Grid` component is analagous to `Box` but additionally sets the rendering context of its children to `display: grid`.
+The `Grid` component implements a simple flex-box based grid as described [here](https://github.com/Heydon/fukol-grids).
```
andoulla, line 7: I wonder if the name grid would be confusing as people might expect a grid implementation. But it is very minor. Also, all I can come up with as a suggestion  is `GridLayout` which isn't great

src/components/grid/__snapshots__/Grid.test.tsx.snap
```diff
@@ -1,21 +1,67 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Grid component renders 1`] = `
-.sxsxxpi {
-  display: grid;
+@keyframes sxwho2l {
```
andoulla, line 4: oh no, why is the loader appearing here too now?

src/components/grid/Grid.tsx
```diff
@@ -1,5 +1,39 @@
...
+  gap,
+  css,
+  ...remainingProps
+}) => {
+  return (
+    <Flex
+      css={
+        {
+          flexWrap: 'wrap',
+          margin: `-${gap}`,
```
thomasdigby, line 24: Gap should use one of our `space` scale values, otherwise we'll have hardcoded CSS values in the codebase. Check out the `Stack` component for variants from a scale https://github.com/Atom-Learning/components/pull/104/files#diff-3711ed1a89cfe31e3c95c9c8ae36868128115f4ee00a6d35c3388da49c0e0b6aR6

src/components/grid/Grid.mdx
hollg: Yeah I think this is a really good point, definitely worth thinking about. The `basis` prop name is kinda nasty too imo. It would be good to find a name that conveys what it does without specifically being linked to the `flex-basis` implementation. I haven't thought of anything that meets that criteria yet though :D

src/components/grid/Grid.mdx
```diff
@@ -4,24 +4,13 @@ component: Grid
...
+Its `gap` prop specifies, in any valid CSS units, the distance between items in the grid. Its `basis` prop specifies, in any valid CSS units, the minimum size of each grid item (by default this will set their `width`, but if the `css` prop is used to set `flex-direction` to `column` then it will set `height` instead).
```
thomasdigby, line 8: We probably ought to document how percentage widths will act in this context. As `margins` are applied to the children they'll be additional to the declared width, e.g. `<Grid basis="50%">` will render two children stacked on top of one another

src/components/grid/Grid.mdx
andoulla: what about being very literal and call it `initialSize` or `minSize`

hollg: Part of the issue is that the prop is passed to the `Grid`, but it sets the size on all the grid items. So maybe `minItemSize`?

src/components/grid/Grid.tsx
hollg: 🤯

hollg: What are the names of those variants?

src/components/grid/Grid.mdx
andoulla: yes, that could work

src/components/grid/Grid.tsx
thomasdigby: The key, so `0`, `1`, `2` etc. I don't think it requires it to match a string or number. `<Grid gap={1} />` & `<Grid gap="1" />` should be synonymous

src/components/grid/__snapshots__/Grid.test.tsx.snap
hollg: Eek yeah I can't answer that

andoulla: it is coming up more and more, should we add a ticket to look into it separately?

src/components/grid/Grid.tsx
```diff
@@ -1,38 +1,56 @@
...
 import { Flex } from '~/components/flex'
-import { CSS } from '~/stitches'
+import { CSS, styled } from '~/stitches'
+
+const GridContainer = styled(Flex, {
+  flexWrap: 'wrap',
+  variants: {
+    gap: Object.keys(theme.space).reduce(
+      // output:
+      // { 0: { m: '-$space$0', '& > *': { m: '$space$0 } }, ... }
```
andoulla, line 12:
```suggestion

```

hollg: This explains what the `reduce` below outputs, so that you don't have to open the `theme` repo to figure it out

src/components/grid/__snapshots__/Grid.test.tsx.snap
hollg: Yeah, agreed. I've been presuming that it's a Stitches-specific issue but perhaps not.

src/components/grid/Grid.mdx
hollg: Good point! I've documented this now, but it would be good to find a way to allow percentage based widths in the future if possible.

src/components/grid/Grid.test.tsx
```diff
@@ -2,19 +2,38 @@ import { render, screen } from '@testing-library/react'
...
     expect(container).toMatchSnapshot()
   })
 
   it('has no programmatically detectable a11y issues', async () => {
-    render(<Grid />)
+    render(
+      <Grid
+        as="main"
+        gap="0.05em"
+        basis="0.5em"
```
thomasdigby, line 29: Need to update the tests with the updated API 👍

src/components/grid/Grid.tsx
```diff
@@ -1,5 +1,57 @@
...
+          }
+        }
+      }),
+      {}
+    )
+  }
+})
+
+type GridProps = {
+  gap: string
```
thomasdigby, line 28: We should get `gap` from `React.ComponentProps<typeof GridContainer>`, I've updated the `Stack` component to use this

thomasdigby (commented):
Only those last minor tweaks!

src/components/grid/Grid.tsx
hollg: Oh yeah nice! Updated

src/components/grid/Grid.test.tsx
hollg: Done 👍🏻

thomasdigby (approved):
Noice!

andoulla (approved):

PR #106, opened 2021-03-29
Chore: separate tests

andoulla:
This separates tests into individual concerns (snapshot and axe). 

Note: in a couple of places I have removed the accessibility testing as it was unnecessary
Note2: in a couple of places I have removed unnecessary snapshot tests (ie: in popover)

### DIFF
--- a/src/components/input-field/InputField.test.tsx
+++ b/src/components/input-field/InputField.test.tsx
@@ -25,7 +25,6 @@ describe(`InputField component`, () => {
       <InputField
         label="INPUT FIELD"
         name="INPUT FIELD"
-        css={{ m: 'auto', height: 100, width: 100 }}
         placeholder="INPUT FIELD"
       />
     )
@@ -56,7 +55,6 @@ describe(`InputField component`, () => {
       <InputField
         label="INPUT FIELD"
         name="INPUT FIELD"
-        css={{ m: 'auto', height: 100, width: 100 }}
         type="number"
         placeholder="001"
       />
@@ -86,7 +84,6 @@ describe(`InputField component`, () => {
       <InputField
         label="INPUT FIELD"
         name="INPUT FIELD"
-        css={{ m: 'auto', height: 100, width: 100 }}
         placeholder="INPUT FIELD"
         disabled
       />
@@ -116,12 +113,7 @@ describe(`InputField component`, () => {
     const errorText = 'This field is required'
 
     const { container } = render(
-      <InputField
-        label="INPUT FIELD"
-        name="INPUT FIELD"
-        error={errorText}
-        css={{ m: 'auto', height: 100, width: 100 }}
-      />
+      <InputField label="INPUT FIELD" name="INPUT FIELD" error={errorText} />
     )
 
     expect(await axe(container)).toHaveNoViolations()
--- a/src/components/input-field/InputField.test.tsx
+++ b/src/components/input-field/InputField.test.tsx
@@ -18,10 +18,22 @@ describe(`InputField component`, () => {
     await screen.getByPlaceholderText('INPUT FIELD')
 
     expect(container).toMatchSnapshot()
+  })
+
+  it('renders a field with a text input - has no programmatically detectable a11y issues', async () => {
+    const { container } = render(
+      <InputField
+        label="INPUT FIELD"
+        name="INPUT FIELD"
+        css={{ m: 'auto', height: 100, width: 100 }}
+        placeholder="INPUT FIELD"
+      />
+    )
+
     expect(await axe(container)).toHaveNoViolations()
   })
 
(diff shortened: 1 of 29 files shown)


src/components/input-field/InputField.test.tsx
```diff
@@ -18,10 +18,22 @@ describe(`InputField component`, () => {
...
 
     expect(container).toMatchSnapshot()
+  })
+
+  it('renders a field with a text input - has no programmatically detectable a11y issues', async () => {
+    const { container } = render(
+      <InputField
+        label="INPUT FIELD"
+        name="INPUT FIELD"
+        css={{ m: 'auto', height: 100, width: 100 }}
```
thomasdigby, line 28: Can we remove any custom `css` from `a11y` tests, they pollute the test view and are unnecessary for the test.

src/components/label/Label.test.tsx
```diff
@@ -11,6 +11,11 @@ describe(`Label component`, () => {
     await screen.getByText('TEXT')
 
     expect(container).toMatchSnapshot()
+  })
+
+  it('has no programmatically detectable a11y issues', async () => {
+    const { container } = render(<Label>TEXT</Label>)
+
     expect(await axe(container)).toHaveNoViolations()
```
thomasdigby, line 19: I'm curious if tests like these have any real value? We're just testing a `<label>` element

src/components/radio/RadioButton.test.tsx
```diff
@@ -16,13 +16,26 @@ describe(`Radio component`, () => {
...
+  it('renders a radio button - has no programmatically detectable a11y issues', async () => {
+    render(
+      <IdProvider>
+        <RadioButtonGroup>
+          <RadioButton value="value" aria-label="indicator" />
+        </RadioButtonGroup>
+      </IdProvider>
+    )
+
+    // filtering on hidden is needed to find the visibile item as Radix adds a hidden input before the visible button
```
thomasdigby, line 33: Should this method of rendering a hidden input not be tested by `axe`? Are we not testing the entire container?

src/documentation/accessibility.mdx
```diff
@@ -16,6 +16,16 @@ Before developing a component, please take some time to read the following secti
 - Add unit tests to test accessibility
   Add an accessibility unit test (using the `jest-axe`) to ensure that the components' different variations don't have any accessibility issues. Example:
 
+```jsx
+it('has no programmatically detectable a11y issues', async () => {
+  const { container } = render(<Button />)
+
+  expect(await axe(container)).toHaveNoViolations()
+})
+```
```
thomasdigby, line 25: Should we add this to our `add-compontent` test template?

andoulla: Yes, maybe we do. I will look into it

src/components/label/Label.test.tsx
andoulla: There is a long debate to be had in general, as we might be better off getting a tool and run tests in the CI instead, as `axe` in unit tests only catch a portion of the issues. 

In regards to this particular test, we added it for consistency when I first made this component. I can remove it if we think it is useless.

src/documentation/accessibility.mdx
andoulla: done

src/components/input-field/InputField.test.tsx
andoulla: done

src/components/radio/RadioButton.test.tsx
andoulla: If we run `axe` on the container, because of the hidden fields that Radix adds, the test will always fail on it. This is why I was filtering out hidden inputs.  Also, it will fail if we have a radio indicator without a label which makes sense as we would never have a radio button without some text.

Would we prefer to have no test at all? is that what the ask is here? Sorry I missed your point

thomasdigby: I assume that the hidden fields shouldn't fail an accessibility test, as they're hidden and shouldn't intefere with the accessibility of the component. I think it's worth testing it as it's supposed to be used, so add a label if you need to.

src/components/label/Label.test.tsx
thomasdigby: There are a few components that we ought to double check what we're getting from an automated accessibility test, `Text` and `Heading` come to mind. Let's investigate that separately 👍

andoulla: ok, for now, we can know that this will check for the font size and colour ratio combination.

src/components/radio/RadioButton.test.tsx
andoulla: done, I think this works ok for now

thomasdigby (approved):

PR #108, opened 2021-03-30
Button additional states

andoulla:
image


src/components/button/Button.tsx
```diff
@@ -86,6 +86,28 @@ const StyledButton = styled('button', {
...
+      md: {
+        p: '$2 $3',
+        fontSize: '$md',
+        lineHeight: 2
+      },
+      lg: {
+        p: '$3 $4',
+        fontSize: '$lg',
+        lineHeight: 1.5
+      }
```
thomasdigby, line 105: A few things on our sizes here:
- Can we add a height to each of these so it lines up correctly with sibling elements, e.g. `height: '$3'` for `sm`, `$5` for `lg`? We can remove the `lineHeight` declarations if we're setting heights
- Could we separate out our padding values into `px` and `py`
- Can we remove the `md` size values from the default button?  I think its `p`, `fontSize` & `height`

src/components/button/Button.tsx
```diff
@@ -168,12 +192,36 @@ export const Button: React.FC<ButtonProps> = ({
...
+
+      if (child?.type?.name === 'Icon') {
+        const style = {
+          [i === 0 ? 'marginRight' : 'marginLeft']: ThemeTokens['space'][3]
+            .value
+        }
+
+        return React.cloneElement(child, {
+          style
+        })
```
thomasdigby, line 209: We could use `css` here to pass the margin value e.g.

```js
if (child?.type?.name === 'Icon') {
  return React.cloneElement(child, {
    css: { [i === 0 ? 'mr' : 'ml']: '$3' }
  })
}
```

_stitches is awesome!_

src/components/button/Button.test.tsx
```diff
@@ -110,6 +136,18 @@ describe(`Button component`, () => {
...
+    const { container } = render(
+      <Button isRounded {...props}>
+        <Icon is={Ok} />
+      </Button>
+    )
+
+    await screen.getByRole('button')
+
+    expect(container).toMatchSnapshot()
+  })
```
thomasdigby, line 149: I think we'll create a separate primitive for icon buttons to ensure that the padding is consistent on both x/y, we'll probably bake in a required aria-label/title as well. Worth removing this test for now 👍

src/components/button/Button.mdx
```diff
@@ -68,3 +68,11 @@ When using a button to request data or fire an action that has a potential delay
 ```jsx live
 <Button isLoading> Hello world</Button>
 ```
+
+## Rounded Variant
+
+Setting `isRounded` prop it creates a rounded button
+
+```jsx live
+<Button rounded> OK </Button>
```
thomasdigby, line 77:
```suggestion
<Button isRounded>Rounded Button</Button>
```

src/components/button/Button.mdx
```diff
@@ -68,3 +68,11 @@ When using a button to request data or fire an action that has a potential delay
 ```jsx live
 <Button isLoading> Hello world</Button>
 ```
+
+## Rounded Variant
+
+Setting `isRounded` prop it creates a rounded button
```
thomasdigby, line 74:
```suggestion
Setting `isRounded` creates a rounded button
```

src/components/grid/__snapshots__/Grid.test.tsx.snap
```diff
@@ -1,6 +1,28 @@
...
+
+  80% {
+    opacity: 0;
+  }
+
+  100% {
+    opacity: 0;
+  }
+}
+
```
thomasdigby, line 25: Have you got the latest from `main`? It should have the fix for this issue

andoulla: I did and I got all the updated snapshots for everything else (form etc) but as soon as run the snapshots it gets readded. I'm going to look into it again today.

andoulla: solved it by deleting node_modules and doing a fresh yarn install.  which doesn't make any sense as what was added was a mock

src/components/button/Button.test.tsx
andoulla: instead of removing it, I'm going to add a test for this variant where it is rounded and has text. 
image

I have seen this style used atom-core I think

src/components/button/Button.tsx
andoulla: Thanks, I tried accessing the shorthands but because I was using style instead of css I couldn't get them.

src/components/button/Button.tsx
andoulla: for the large button `$5` seems really big
image

thomasdigby: Yeah I agree, it's the only value that fits in our scale though and it does look like a similar size to our current large button https://storybook.atomlearning.technology/?path=/docs/components-button--default-story#size

We _could_ use something like `height: calc($4 + $2)` to bump up the height by 8px instead of 16px, but then we're going off the scale a touch... what do you think?

yarn.lock
```diff
@@ -8747,7 +8747,7 @@ punycode@^2.1.0, punycode@^2.1.1:
 
 pure-react-carousel@^1.27.6:
   version "1.27.6"
-  resolved "https://registry.yarnpkg.com/pure-react-carousel/-/pure-react-carousel-1.27.6.tgz#d1d726ba653c0a700a6ac5741618b4196c4c6fbb"
+  resolved "https://registry.atomlearning.technology/pure-react-carousel/-/pure-react-carousel-1.27.6.tgz#d1d726ba653c0a700a6ac5741618b4196c4c6fbb"
```
thomasdigby, line 8750: Can we revert this change? And the package.json update as well. Neither change is relevant to this PR but there's obviously something wrong with the order of some dependencies as they keep changing when anyone runs a `yarn install`, I'll do a quick PR with just this update to fix it 👍

src/components/button/Button.tsx
thomasdigby: Also, if you missed it (or haven't done it yet), we can remove the `px: '$4'`, `py: '$2'`, `fontSize: '$md'` and `height: '$4'` from the initial styles for the Button as we'll always overwrite them with our sizes 👍

yarn.lock
thomasdigby: Ah this is also the issue with the `yarn install` on CI, reverting this change will fix it

src/components/button/Button.tsx
andoulla: let me try it out and see it. The visual always helps me

andoulla: $4 + $2
image

$4 + $1
image


I think the second is better

thomasdigby: Yeah you're right, I forgot the values were from the sizing scale not the spacing. `$4 + $1` 👍

thomasdigby (approved):
Happy with this, but just wanted to flag this comment again https://github.com/Atom-Learning/components/pull/108/files#r604706408

PR #109, opened 2021-03-30
Feat: add form field wrapper

hollg:
- Adds `FieldWrapper` to help compose form fields with consistent layouts.
- Updates `InputField` and `PasswordField` to use `FieldWrapper`
- Removes default `name` from `PasswordField` as it has to be set explicitly in order for `Form` to register the field
- Makes `name` a required prop for `Input` (and by extension `InputField`, `PasswordField`) for the same reason


src/components/field-wrapper/FieldWrapper.tsx
```diff
@@ -0,0 +1,35 @@
...
+  error,
+  htmlFor,
+  label,
+  required
+}) => {
+  return (
+    <Box css={css}>
+      <Label required={required} htmlFor={htmlFor} css={{ mb: '$2' }}>
+        {label}
+      </Label>
```
thomasdigby, line 28: Should we include the `prompt` functionality here as well?

src/components/field-wrapper/FieldWrapper.tsx
```diff
@@ -0,0 +1,35 @@
...
+  htmlFor: string
+  label: string
+  required?: boolean
+}
+
+export const FieldWrapper: React.FC<FieldWrapperProps> = ({
+  css,
+  children,
+  error,
+  htmlFor,
```
thomasdigby, line 20: I wonder if this could be `id`? Feels like it could be a nice abstraction from the implementation

andoulla: does fieldId work?

src/components/field-wrapper/FieldWrapper.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+  })
+
+  it('links the label to an input ', async () => {
+    render(<ExampleField />)
+
+    userEvent.click(await screen.findByText('Example Field'))
+
+    expect(await screen.findByRole('textbox')).toHaveFocus()
+  })
+})
```
andoulla, line 35: Should we add a test for the validation error?

src/components/field-wrapper/FieldWrapper.tsx
hollg: Yeah that seems good! I'll update it now. Thanks!

thomasdigby: I think within the context of `FieldWrapper`, `id` should be clear enough, it's the identifier to link the surrounding elements with the passed form element. Happy to run with `fieldId` though 👍

src/components/field-wrapper/FieldWrapper.tsx
hollg: I've added this and updated `PasswordField` to use `FieldWrapper` accordingly. No idea why I decided it wouldn't work yesterday!

src/components/field-wrapper/FieldWrapper.test.tsx
hollg: Done!

thomasdigby (approved):
Would be nice to add a single test for the `prompt`, otherwise good to go 👍

hollg:
@thomasdigby done!

PR #111, opened 2021-03-30
Feature: Popover

thomasdigby:
Screenshot 2021-03-30 at 21 57 17


src/components/popover/Popover.test.tsx
```diff
@@ -7,116 +8,50 @@ import { Popover } from '.'
...
-    fireEvent.keyPress(screen.getByText(/click me/i), {
-      key: 'A',
-      code: 'A'
-    })
+    const trigger = screen.getByText('TRIGGER')
 
-    expect(screen.getByText('Content')).toBeInTheDocument()
-  })
+    expect(await trigger).toBeInTheDocument()
+    expect(await screen.queryByText('CONTENT')).not.toBeInTheDocument()
```
andoulla, line 37: Very minor, not sure if we need the `await` here

src/components/popover/Popover.test.tsx
```diff
@@ -7,116 +8,50 @@ import { Popover } from '.'
...
+      <IdProvider>
+        <Popover defaultOpen>
+          <Popover.Trigger as="button">TRIGGER</Popover.Trigger>
+          <Popover.Content>CONTENT</Popover.Content>
+        </Popover>
+      </IdProvider>
     )
 
-    expect(await axe(container)).toHaveNoViolations()
+    expect(await waitFor(() => axe(container))).toHaveNoViolations()
```
andoulla, line 55: I know there is a ticket for this but I thought it would be good to mention. shouldn't we trigger the popover to open before running axe on the container?

andoulla (approved):

src/components/popover/Popover.test.tsx
thomasdigby: `defaultOpen` is set on the `Popover` so that it should be open in the `axe` test, there are a few oddities with RTL, Radix and async/await that I haven't figured out yet for the snapshots, but running `debug` on render (to see what is tested by `axe` shows the correct markup 👍 

I'm going to update the ticket to take a sweep over all of our interactive components to ensure that they're being tested properly.

src/components/popover/Popover.test.tsx
thomasdigby: I might leave this in and come back to it when we take a look at all of our interactive component tests

hollg (approved):

PR #113, opened 2021-04-01
Feat: loads of form stuff

hollg:
- Adds `CheckboxField` (`Checkbox` + `Label`)
- Adds `RadioButtonField` (`RadioButton` + `Label`)
- Refactors `Form` and our field components to use `react-hook-form`'s context provider rather than prop drilling (WIP)
- Adds`RadioGroupField`

### `RadioGroupField`
Screenshot 2021-04-06 at 08 56 10


src/components/radio-field/RadioField.mdx
```diff
@@ -0,0 +1,10 @@
+---
+title: Radio Field
+component: RadioField
+description: This is the description
```
andoulla, line 4:
```suggestion
description: Combines a RadioButton with a Label and ValidationError
```

src/components/radio-field/RadioField.mdx
```diff
@@ -0,0 +1,10 @@
...
+category: Forms
+---
+
```
andoulla, line 7: Maybe we can add something similar with the Checkbox content here


And also mention that the value needs to be a string

src/components/radio-field/RadioGroupField.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { Controller, useFormContext } from 'react-hook-form'
+
+import { Box } from '~/components/box'
+import { ValidationOptions } from '~/components/form'
+import { Label } from '~/components/label'
+import { RadioButtonGroup } from '~/components/radio'
+import { ValidationError } from '~/components/validation-error'
+import { CSS } from '~/stitches'
+
+import { RadioField } from './RadioField'
```
andoulla, line 11: Just adding a bit of space here
```suggestion
import { RadioField } from './RadioField'

```

src/components/checkbox-field/CheckboxField.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+type CheckboxFieldProps = {
+  css?: CSS
+  error?: string
+  label: string
+  name: string
+  validation?: ValidationOptions
+} & React.ComponentProps<typeof Checkbox>
+
+export const CheckboxField: React.FC<CheckboxFieldProps> = ({
```
andoulla, line 18: Just for my understanding, how come we don't need to pass the ref for these components as we do in the other field components?

src/components/field-wrapper/FieldWrapper.tsx
```diff
@@ -50,3 +50,34 @@ export const FieldWrapper: React.FC<FieldWrapperProps> = ({
...
+
+type InlineFieldWrapperProps = {
+  css?: CSS
+  error?: string
+  label: string
+  fieldId: string
+  required?: boolean
+}
+
+export const InlineFieldWrapper: React.FC<InlineFieldWrapperProps> = ({
```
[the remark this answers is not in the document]
hollg: What do you mean by variant? It isn't a stitches component so it can't have a variant in that sense. I'm reluctant to add a boolean `inline` prop because they're so different. Also the markup for the `FieldWrapper` component is already pretty noisy; I think adding either an early `return` or a bunch of `&&` conditionals would make it even less readable.

What duplication are you thinking of? Maybe there will be another way to avoid it.

src/components/checkbox-field/CheckboxField.tsx
hollg: In most cases, three conditions are met:
- there's a 1:1 relationship between our field component and the HTML `input` (or `select`, or whatever)`
- the `input` (or whatever) is uncontrolled
- we have direct access to the `input` (or whatever) from our code

In those cases, we pass a `ref` to the `input` (or whatever) and `react-hook-form` uses it to keep track of the value. 

If any of the above conditions aren't met then we have to use the `Controller` to connect a field to the form. This lets us track our own state and manually pass it into `react-hook-form`'s `onChange` (without `Controller`, `react-hook-form` can _only_ track the values in the DOM).

Annoyingly, the Radix checkbox and radio components seem to mask the underlying `input`s and prevent us from passing `ref`s to them (i.e. we do not have direct acces to them). So we're manually tracking the `isChecked` state and using `Controller` to keep `react-hook-form` updated.

In the `RadioGroupField` the `Controller` is also allowing us to work around the fact that there are multiple `input`s. In that case, we have a `name` for the whole group and the `value` will depend on which radio has been selected. 

I hope this helps!

andoulla: Thanks that's an amazing explanation. I was just wondering why because they are all inputs. Didn't take into consideration that Radiix might interfere. Maybe it was a trivial question sorry

src/components/checkbox-field/CheckboxField.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+    const { container } = render(
+      //  workaround for failing axe test -
+      //   Radix's checkbox renders an
+      //   <input type="checkbox" hidden="" /> but doesn't give it role="none"
+      //   so axe wants it to have a label
+      <label>
+        <Form>
+          <CheckboxField label="Example" name="example" title="example" />
+        </Form>
+      </label>
```
thomasdigby, line 29: Surely this `<label><form/></label>` markup is flagged by axe? 😅

sandbox/index.tsx
```diff
@@ -4,16 +4,21 @@ import * as React from 'react'
...
+    <Form onSubmit={console.log} css={{ mx: 'auto', width: '500px', p: '$2' }}>
+      <SelectField
+        label="Favourite color"
+        name="color"
+        options={[
+          { value: 1, label: 'papayawhip' },
+          { value: 2, label: 'aliceblue' }
+        ]}
+      />
+    </Form>
```
thomasdigby, line 21: 🙅‍♂️ I'm working on a script to prevent commits when this file is added

src/components/radio-field/RadioGroupField.tsx
```diff
@@ -0,0 +1,57 @@
...
+          <RadioButtonGroup
+            defaultValue={defaultValue}
+            onValueChange={onChange}
+            value={value}
+          >
+            {children}
+          </RadioButtonGroup>
+        )}
+      />
+      {errors[name] && (
```
thomasdigby, line 46: Will errors always be an object in this context? I assume react hook form defaults to an empty object?

src/components/field-wrapper/FieldWrapper.tsx
```diff
@@ -50,3 +50,34 @@ export const FieldWrapper: React.FC<FieldWrapperProps> = ({
...
+  required
+}) => {
+  return (
+    <Box css={css}>
+      <Flex css={{ alignItems: 'center' }}>
+        <Box css={{ mr: '$2' }}>{children}</Box>
+        <Label htmlFor={fieldId} required={required}>
+          {label}
+        </Label>
+      </Flex>
```
thomasdigby, line 77: One thing that we could tweak here to slightly improve the UX is to replace the Flex wrapper with the Label so it wraps the RadioButton. That way you have an implicit link between the label and the form element(without requiring htmlfor) _and_ the entire block becomes clickable, not just the label/input but also the space in-between.

thomasdigby (commented):
Mega work here, a really neat set of changes to reduce & simplify the form usage and handle the tricky integration between 3rd party libs

src/components/field-wrapper/FieldWrapper.tsx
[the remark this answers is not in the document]
andoulla: Ok, I went through it again and can see the differences in structure now. The file is so busy some of the differences weren't very visible. Any reason why we can't spilt these 2 into their own files to improve readability? (They can still be in the same folder)

src/components/checkbox-field/CheckboxField.test.tsx
hollg: lmao you'd think so, right?

sandbox/index.tsx
hollg: 🥳

src/components/radio-field/RadioGroupField.tsx
hollg: Yep, exactly

src/components/field-wrapper/FieldWrapper.tsx
hollg: Oh yeah nice one. Updated!

src/components/checkbox-field/CheckboxField.tsx
hollg: Not trivial at all! I spent a lot of Thursday last week figuring this stuff out!

src/components/field-wrapper/FieldWrapper.tsx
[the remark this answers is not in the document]
hollg: No reason at all — I'll do that now

src/components/checkbox-field/CheckboxField.tsx
andoulla: Maybe a line explaining when to use the controller will be a good addition to the mdx file

src/components/checkbox-field/CheckboxField.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+        <CheckboxField label="Example" name="example" />
+      </Form>
+    )
+
+    expect(container).toMatchSnapshot()
+  })
+
+  it('has no programmatically detectable a11y issues', async () => {
+    const { container } = render(
+      //  workaround for failing axe test -
```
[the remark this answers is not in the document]
hollg: I'm not sure I see the benefit of writing a util just to add a `<label>` since we'll still have to come back and remove all the instances of the util from the tests anyway

src/components/checkbox-field/CheckboxField.test.tsx
```diff
@@ -0,0 +1,34 @@
+import { render } from '@testing-library/react'
+import { axe } from 'jest-axe'
+import * as React from 'react'
+
+import { Form } from '../'
+import { CheckboxField } from '.'
+
+describe('CheckboxField component', () => {
+  it('renders', async () => {
```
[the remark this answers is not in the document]
hollg: Which styling are you thinking of? Should this not be covered in the `Checkbox` tests rather than the `CheckboxField`?

src/components/checkbox-field/CheckboxField.test.tsx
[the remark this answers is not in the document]
andoulla: fair enough, I was just thinking it might make the refactoring easier. but it was just a thought

thomasdigby (approved):

PR #115, opened 2021-04-03
Feature: Dialog component

thomasdigby:
- Created a `Dialog` component that mimics the `Modal` that we currently use on the main application
- We avoid having to use `react-transition-group` as Radix handles unmounting once a css animation has finished 🙏 
- Tweaked the structure of the test slightly to enable validating a11y when open and sharing the UI code
- Added a new `inset` utility to match the new CSS property https://developer.mozilla.org/en-US/docs/Web/CSS/inset

https://user-images.githubusercontent.com/3226804/113474114-c2753100-9465-11eb-8528-228347ed4920.mov


src/components/dialog/Dialog.mdx
```diff
@@ -0,0 +1,34 @@
+---
+title: Dialog
+component: Dialog,Dialog.Trigger,Dialog.Content,Dialog.Close
+description: The Dialog component renders its children inside a modal and puts a dimmer over the rest of the screen
+category: Surfaces
+---
+
+`Dialog` exports many components that combine to create a modal. The `Dialog.Trigger` requires a React component child (or passed to `as`) to allow the necessary event bindings and accessible properties to be cloned.
```
hollg, line 8: How are we defining "dialog" and "modal"?

thomasdigby: I decided to stick with `Dialog` as the component name to make it consistent with Radix and to match with the html `dialog` element, `modal` as a term is still useful for explaining the concept but I can understand if this is confusing based on our previous naming conventions.

src/components/dialog/DialogContent.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+const StyledDialogClose = styled(Close, {
+  all: 'unset',
+  alignItems: 'center',
+  color: '$tonal800',
+  cursor: 'pointer',
+  display: 'flex',
+  justifyContent: 'center',
+  position: 'absolute',
+  right: '0px',
```
andoulla, line 38: we don't need the `px` here or in top

src/stitches.ts
```diff
@@ -53,6 +53,12 @@ export const utils = {
     marginTop: value,
     marginBottom: value
   }),
+  inset: () => (value: CSSValue): CSSBlob => ({
```
hollg, line 56: What's the generalised use case for setting all four of these properties at once? I don't think I've ever used all four at once so I'm unsure in what situation you'd want this vs e.g. padding

thomasdigby: It's also essentially synonymous to `top: 0; left: 0; height: 100%; width: 100%;` which we've use a few times on the main app for overlays or filling content. I think it's worthwhile creating a few utilities that mirror modern CSS properties, as well as those which are useful shorthands (in this case I think it's both), we could even expand to use other parts of the logical properties spec and introduce `block` and `inline` spacing and sizing in future iterations.

src/components/dialog/DialogContent.tsx
thomasdigby: I think Gary and I discussed this a little while ago and we decided to include the unit for clarity.

src/stitches.ts
hollg: Would we ever give it a value other than `0`? I'm definitely in favour of these utilities in general, I'm just not grasping this one

thomasdigby: Personally I think even if that is the case, I still see it having value. I can remove it if we want to be stricter about our utilities, we can discuss its inclusion in a separate PR

hollg (approved):

andoulla (approved):

PR #116, opened 2021-04-03
Feature: Alert dialog component/hook

thomasdigby:
- Created the `AlertDialog` component and a `useAlert` hook to programmatically show the alert. The `AlertDialog` itself can be used as a component to show/hide as needed, but the main value from this component comes with pairing it with the `AlertProvider` and using `useAlert()` to show an alert
- `useAlert` can handle multiple calls to `showAlert()`, they'll be queued and each will show once the previous one has been actioned
- Had to make some TS tweaks to `Heading`/`Text` components to allow `AlertDialog.Title` and `AlertDialog.Description` to be passed into the `as` prop
- Fixed the potential to override the negative margin that offsets the gap property on `Stack` component
- Fixed `Button` `children` type to allow any React content (and made a few stylist and structural tweaks)
- Updated `test:watch` to not show coverage

Screenshot 2021-04-03 at 16 21 20


src/components/alert-dialog/alert-context/AlertDialog.tsx
```diff
@@ -0,0 +1,60 @@
...
+  cancelActionText,
+  confirmActionText,
+  onClose,
+  ...remainingProps
+}) => (
+  <AlertDialog defaultOpen>
+    <AlertDialog.Content
+      onEscapeKeyDown={(e) => e.preventDefault()}
+      onPointerDownOutside={(e) => e.preventDefault()}
+      css={{ textAlign: 'center' }}
```
andoulla, line 28: Apologies if it is a silly question, but doesn't this need to be merged with the css that might be passed as a prop?

something like this: ` css={{textAlign: "center", ...css}}`

src/components/alert-dialog/alert-context/AlertDialog.tsx
```diff
@@ -0,0 +1,60 @@
...
+        </Text>
+      )}
+      <Stack gap="3" css={{ mx: 'auto' }}>
+        <Button
+          size="sm"
+          theme="danger"
+          as={AlertDialog.Cancel}
+          onClick={() => onAction(false)}
+        >
+          {cancelActionText || 'Cancel'}
```
andoulla, line 47: Maybe we don't have a default and we force the user to pass the value as this could lead to accidentally leaving the word untranslated and the rest of the content would be translated. That always looks really awkward to a user.

src/components/button/Button.tsx
```diff
@@ -40,23 +40,18 @@ const getButtonSolidVariant = (baseColor: string, interactColor: string) => ({
...
   fontFamily: '$sans',
-  fontSize: '$md',
   fontWeight: 500,
-  height: '$4',
   letterSpacing: '0.02em',
   lineHeight: 1.4,
   transition: 'all 125ms ease-out',
   textDecoration: 'none',
-  px: '$4',
-  py: '$2',
```
andoulla, line 59: how come we are not adding any padding on the y axis (out of curiosity)

src/components/heading/Heading.tsx
```diff
@@ -3,7 +3,7 @@ import * as React from 'react'
 import { styled } from '~/stitches'
 import { capsize, Override } from '~/utilities'
 
-const StyledHeading = styled('h1', {
+const StyledHeading = styled('h2', {
```
andoulla, line 6: Again out of curiosity, why are we defaulting to h2?

src/components/input-field/__snapshots__/InputField.test.tsx.snap
```diff
@@ -17,50 +17,50 @@ exports[`InputField component renders a field in error state 1`] = `
...
+}
+
+.sxegjw0xyrog--c2 {
   background-color: var(--colors-primary500);
   color: white;
 }
 
-.sxjzq4uxyrog--c2:not([disabled]):hover,
-.sxjzq4uxyrog--c2:not([disabled]):focus {
+.sxegjw0xyrog--c2:not([disabled]):hover,
```
andoulla, line 54: I don't understand why this snapshot is changing? There are no changes in the `Input` or `Label` or `ValidationError`.

Also, why is the css from the button included here? This component doesn't import the button component at all

package.json
```diff
@@ -19,8 +19,8 @@
     "clean": "del ./dist/",
     "format": "prettier 'src/**/*.{js,ts,tsx}' --write",
     "lint": "eslint 'src/**/*.{js,ts,tsx}' --fix",
-    "test": "jest",
-    "test:watch": "yarn test -- --watch",
+    "test": "jest --coverage",
```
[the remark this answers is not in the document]
thomasdigby: Yeah for sure, want to put in a PR today for it?

src/components/alert-dialog/alert-context/AlertContext.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+    fireEvent.click(trigger)
+
+    fireEvent.click(await screen.findByText('Confirm'))
+
+    expect(onActionMock).toHaveBeenCalledWith(true)
+  })
+
+  it('dialog is not closed with ESC', async () => {
+    const { container } = render(<AlertContextTest />)
```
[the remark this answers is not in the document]
thomasdigby: Yeah potentially worth testing in addition, I think anything more is just testing the library itself

src/components/alert-dialog/alert-context/AlertDialog.tsx
thomasdigby: Yeah I had it like this before but it isn't necessary as this component is used only within the `AlertProvider` so it isn't user modifiable.

src/components/alert-dialog/alert-context/AlertDialog.tsx
thomasdigby: Yeah I wasn't sure what to do here, I think you're probably right, will update! 👍

package.json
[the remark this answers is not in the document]
andoulla: Ok I will add it in my todo list as soon as I'm done with my ticket

src/components/button/Button.tsx
thomasdigby: Because we're always setting a height and the text will vertically align by default anyway so no need to include y axis padding.

src/components/alert-dialog/alert-context/AlertContext.test.tsx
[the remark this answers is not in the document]
andoulla: No sorry, I should have deleted to comment. The test file came first in the code review so when I looked at the component the first time around I missed the line. then I found it

src/components/heading/Heading.tsx
thomasdigby: We were already defaulting to `h2` in the `as` prop in this component but it wasn't reflected in the original `styled` function.

On the question of h1 or h2; there should only ever be a single h1 on a page so defaulting to that makes it more likely that multiple h1s are rendered, however much we audit it, it's still likely that `Heading`s will be used without `as` so we need to safeguard against it.

src/components/input-field/__snapshots__/InputField.test.tsx.snap
thomasdigby: Not too sure, probably something upstream but will need to check!

src/components/alert-dialog/alert-context/AlertContext.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+      </IdProvider>
+    )
+    const trigger = screen.getByText('TRIGGER')
+
+    fireEvent.click(trigger)
+
+    expect(await screen.queryByText('TITLE')).toBeInTheDocument()
+    expect(await screen.queryByText('TITLE 2')).not.toBeInTheDocument()
+
+    const cancelButton = await screen.getByText('Cancel')
```
hollg, line 107: `getByX` queries aren't `async`, so I don't think this `await` is doing anything

src/components/alert-dialog/alert-context/reducer.ts
```diff
@@ -0,0 +1,21 @@
...
+  | { type: 'REMOVE'; payload: string }
+
+export const initialState = []
+
+export const reducer = (state: State, action: Action): alert[] => {
+  switch (action.type) {
+    case 'ADD':
+      return [...state, { ...action.payload, id: uid() }]
+    case 'REMOVE':
+      return state.filter((t) => t.id !== action.payload)
```
hollg, line 17: What does `t` stand for here? I'm sure it does make sense and that I'm just missing something :D

thomasdigby: Ah yes, forgot to update 🤦

src/components/input-field/__snapshots__/InputField.test.tsx.snap
thomasdigby: `Button` was being used in the test, replaced with a standard `<button>`

andoulla: oh sorry I just checked the component, didn't think to check the test.

hollg (approved):

PR #120, opened 2021-04-08
Semantic release

hollg:
Adds this action to our GitHub workflow to integrate `semantic-release`: https://github.com/cycjimmy/semantic-release-action


.github/workflows/main.yml
```diff
@@ -23,3 +26,28 @@ jobs:
...
+        uses: actions/setup-node@v2.1.4
+        with:
+          node-version: '14'
+      - run: yarn install
+
+      - name: Test
+        run: yarn test
+
+      - name: Validate
+        run: yarn validate
```
thomasdigby, line 47: Both of these steps should be automatically run when `npm publish` is called. Our `prepublishOnly` script handles that step.

Is there a way to run this action with `--dry-run` on the publish command? Then we could just see the logs instead of publishing it proper

package.json
```diff
@@ -134,5 +137,11 @@
     "extends": [
       "@commitlint/config-conventional"
     ]
+  },
+  "release": {
+    "plugins": [
+      "@semantic-release/changelog",
+      "@semantic-release/git"
```
thomasdigby, line 144: Could we add a slack webhook here to notify the DS pod or a release channel of the update? I think there's a plug-in for it

thomasdigby: Also just FYI, you can test this locally with `npx semantic-release`. Not sure how you pass dry run but worth trying!

andoulla: I think the dev-team-updates might be more appropriate as everyone will be on that one or the front-end one

.github/workflows/main.yml
```diff
@@ -23,3 +26,33 @@ jobs:
...
+        with:
+          node-version: '14'
+      - run: yarn install
+
+      - name: Semantic Release
+        uses: cycjimmy/semantic-release-action@v2.5.0
+        with:
+          branch: 'main'
+          extra_plugins: |
+            @semantic-release/changelog
```
andoulla, line 48: how are changes in the changelog handle (if you found it in the docs as I couldn't) does it do an extra commit?

package.json
hollg: Hm I've added the Slack app and have the webhook URL, but it comes with a warning that it's a private value and shouldn't be shared. Do we have anything set up that would allow us to use the value here without committing it into the public repo?

andoulla:
Should this be documented somewhere? The logic for version bumps etc

package.json
thomasdigby: You can use a github secret like your others e.g. `${{ secrets.SLACK_WEBHOOK }}`

hollg: Even though it's not in the workflow file?

thomasdigby: Oh I thought that's where you needed to access it, let me take a look at the docs

thomasdigby: https://github.com/juliuscc/semantic-release-slack-bot#environment-variables You should be able to set them in github and they'll be accessable for the slack-bot

thomasdigby:
@hollg Can we ensure that our first release has an "🎉 Initial release" style message in the changelog instead of all of our previous commits? Is that possible with this automated system? Perhaps V1 is manually released with that change and then we use sematic-commit after then?

hollg:
@thomasdigby I'd probably recommend manually releasing v1 if you want that message. You can hardcode a message in the config, but then we'd just have 2 extra commits adding and them removing it :D

.github/workflows/main.yml
```diff
@@ -23,3 +26,29 @@ jobs:
 
       - name: Validate
         run: yarn validate
+  publish:
+    name: publish
+    needs: test
```
thomasdigby, line 31: @hollg Just wanted to check, is this `needs: test` necessary as we have a specific script `prepublishOnly` that `npm` will always run before publishing. Or should we remove the prepublish step if it duplicates the behaviour from `test`

hollg: Good question! My instinct is to remove `prepublishOnly` and keep this here because I think it makes it clearer why something didn't release and it feels slightly less risky to me to not even try to run the publish step if test didn't pass.

Both of these are pretty negligible differences though so I'm happy either way if you have a preference!

thomasdigby: I think we should definitely keep the `test` requirement as part of the PR, we should be able to see why it wasn't run as part of the Github CI. We could remove the `test` and `validate` parts of the `prepublishOnly` script so it's solely a preparatory step. `"prepublishOnly": "run-s clean build:lib build:docs"` should work for us.

thomasdigby (approved):

PR #121, opened 2021-04-09
Chore: Prevent commit when sandbox has changes

thomasdigby:
Screenshot 2021-04-09 at 13 25 13


andoulla:
How will we make changes to the sandbox if we need to update it? just use `--no-verify`?

thomasdigby:
> How will we make changes to the sandbox if we need to update it? just use `--no-verify`?

Yeah exactly, this handles most use-cases but if we do need to update it we can just skip this check

hollg:
Do we not just want to prevent merges into `main` with sandbox changes, rather than prevent commits on WIP branches? I almost always keep changes in the sandbox over many commits, right up until the PR is ready. Could we run a script during the CI process to check this?

thomasdigby:
> Do we not just want to prevent merges into `main` with sandbox changes, rather than prevent commits on WIP branches? I almost always keep changes in the sandbox over many commits, right up until the PR is ready. Could we run a script during the CI process to check this?

Ideally we wouldn't commit those changes to the remote at all, even in a WIP branch. The only change to your git workflow locally is to not stage the sandbox when committing changes.

hollg:
What's the reason for not wanting changes to the sandbox in the remote at all? One reason to support it imo is that it's handy to be able to say to ask someone else to pull a branch and look at the sandbox either to help debug a problem or to try the API of a new component and give feedback

thomasdigby:
> What's the reason for not wanting changes to the sandbox in the remote at all? One reason to support it imo is that it's handy to be able to say to ask someone else to pull a branch and look at the sandbox either to help debug a problem or to try the API of a new component and give feedback

As far as I'm concerned, Sandbox should be a local only testing solution as we should avoid committing WIP or redundant changes. Now we're using commits to build our changelog it's more important that our commit history is aligned closely with features and fixes that we release.

Also just to clarify this change only prevents a commit when any `sandbox` file is _staged for commit_, so it's a safety net more than preventing using the sandbox at all.

thomasdigby:
@hollg @andoulla As we have more contributors now I think we're more at risk of devs unintentionally updating the sandbox. Do either of you have an issue with merging it? A reminder that it'll only block a commit where the sandbox is staged, and hopefully the error message is clear enough

hollg:
I have actually done a total 180 on this: I accidentally commit changes to it every time I make a PR to this repo and think it would be _great_ to have something that stops me.

PR #124, opened 2021-04-12
docs: update contributing quidelines

andoulla:


CONTRIBUTING.md
```diff
@@ -1,12 +1,16 @@
 ---
 title: Contributing
-description: When contributing to this repository, use the following information, to uphold the standards we have set for this project. Please read through this and the `README.md` before starting.
+description: When contributing to this repository, use the following information to uphold the standards we have set for this project.
 category: Overview
 ---
 
+Everyone is encouraged to contribute to the development of this project. If you spot a missing component or an additional feature, please raise a ticket to track the changes by using this [form](https://www.notion.so/atomlearning/Need-a-new-component-4431d2db39ee47ecb9af2be4c77c6130#f006b07e3922444d85f8a2742eac21a6)
```
hollg, line 7: Should we be linking to Notion here? Is that not private?

CONTRIBUTING.md
```diff
@@ -48,81 +52,119 @@ const ComposedComponent = ({ css }) => (
...
+  it('has no programmatically detectable a11y issues', async () => {
+    render(<Box />, document.body)
+
+    const results = await axe(document.body)
+    expect(results).toHaveNoViolations()
+  })
+})
+```
+
+## Commits
```
hollg, line 87: Can I suggest putting this at (or near) the top? If someone only reads the first couple of sections, I think this should be one of them.

CONTRIBUTING.md
andoulla: I think the form is open but the board is private, need to check. I can take it off I just wanted to mention that it needs a ticket

CONTRIBUTING.md
andoulla: Yes of course, I'm still thinking of the most sensible way of laying out the sections

CONTRIBUTING.md
hollg: Hmm, given that the library is open source, I feel like people outside of Atom should not only be able to make requests but also see them. Maybe we should be using GitHub issues?

andoulla: The problem with using GitHub issues is that we are going to then have to maintain 2 sources  of truths. Something is always going to be out-of-date and we will have lots of duplication.

Switching, only to GitHub issues for everyone, might have the problem that we will be posting company links in to the public, which might be a problem not 100% sure

andoulla: btw, I had the wrong link, that was meant to be the form, not the notion page

CONTRIBUTING.md
andoulla: done

CONTRIBUTING.md
```diff
@@ -1,12 +1,52 @@
...
+description: When contributing to this repository, use the following information to uphold the standards we have set for this project.
 category: Overview
 ---
 
+Everyone is encouraged to contribute to the development of this project. If you spot a missing component or an additional feature, please raise a ticket to track the changes by using the "Need a component" form
+In this document, you will find all the necessary information to develop and test the features. Therefore, please review this document and the `README.md` before you get started.
+
+## Commits
+
+This section is **very important**! Our releases and version numbers follow [Semantic Versioning](https://semver.org/) and are generated from the commit messages in PRs when they get merged into `main`, so make sure you follow our conventions.
```
thomasdigby, line 12: **very important** 👍  👍

CONTRIBUTING.md
```diff
@@ -1,12 +1,52 @@
 ---
 title: Contributing
-description: When contributing to this repository, use the following information, to uphold the standards we have set for this project. Please read through this and the `README.md` before starting.
+description: When contributing to this repository, use the following information to uphold the standards we have set for this project.
 category: Overview
 ---
 
+Everyone is encouraged to contribute to the development of this project. If you spot a missing component or an additional feature, please raise a ticket to track the changes by using the "Need a component" form
```
thomasdigby, line 7: Having tried to read this from the perspective of a new user, not linking the "Need a component" form here doesn't really make sense, even though I understand why we haven't. I think we should recreate this form within github and create a "Request a component" issue template so we can easily link to and keep track of requests in the open.

CONTRIBUTING.md
```diff
@@ -1,12 +1,52 @@
...
+<footer>
+```
+
+All commit messages must have a type (a word followed by a colon at the start of the message) from this [list](https://github.com/conventional-changelog/commitlint/tree/master/@commitlint/config-conventional#type-enum) (the scope is optional but strongly encouraged). Either of the following types will cause a new release to be published when your PR is merged into `main`:
+
+- `fix` for bug fixes (patch version)
+- `perf` for performance improvements (patch version)
+- `feat` for new features (minor version)
+
+Other subjects (such as `chore`) will not cause a new release **unless** the commit footer starts with `BREAKING CHANGE:` (followed by an explanation of the breaking change). This **must** be included with **any** commit that introduces a breaking change. Breaking changes will cause a major version increase.
```
thomasdigby, line 32: I'm keen to also adopt the `!` breaking change notation e.g. `feat!: this a breaking change`. Can we also document it here

thomasdigby (commented):
Looking really good, happy to get this in once we potentially look at the github issue template

CONTRIBUTING.md
andoulla: this is all @hollg

CONTRIBUTING.md
andoulla: does this work?

CONTRIBUTING.md
```diff
@@ -1,12 +1,54 @@
 ---
 title: Contributing
-description: When contributing to this repository, use the following information, to uphold the standards we have set for this project. Please read through this and the `README.md` before starting.
+description: When contributing to this repository, use the following information to uphold the standards we have set for this project.
 category: Overview
 ---
 
+Everyone is encouraged to contribute to the development of this project. If you spot a missing component or an additional feature, please raise an issue in GitHub using the feature request template.
```
thomasdigby, line 7: Can we link to the new issue page? https://github.com/atom-learning/components/issues/new?template=request-a-feature.md

thomasdigby (approved):

PR #128, opened 2021-04-14
RFC: Select API

thomasdigby:
Interested in your thoughts on this change to the `Select` component, I've been mulling this over for a few weeks and I don't think there's any reason to abstract the options into a prop, we can use the standard `<option>` HTML syntax and pass them as `children`.

Currently each options attributes are passed one by one onto the internal `<option>` element, which prevents using other standard attributes like `selected` (an easy fix by spreading the attributes but highlights the larger issue of flexibility/customisation). We also prevent using other standard `select` children like `optgroup` by abstracting `options`.

### DIFF
--- a/src/components/select/Select.mdx
+++ b/src/components/select/Select.mdx
@@ -20,12 +20,12 @@ A `Select` needs to be associated with a label for accessibility purposes, so ra
 
 For accessibility purposes the component needs to have the `id` prop set, to link it to the label it is associated with. If a label is not available, please add an `aria-label` to ensure that the component remains accessible
 
-## Default option
+## Placeholder
 
-The component can display a default option without a value by passing the `defaultOption` property.
+The component can display a placeholder without a value by passing the `placeholder` property.
 
 ```jsx preview
-<Select default="Please select a fruit" css={{ width: 300 }}>
+<Select placeholder="Please select a fruit" css={{ width: 300 }}>
   <option value="apples">Apples</option>
   <option value="bananas">Bananas</option>
 </Select>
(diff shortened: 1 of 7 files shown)


src/components/select/Select.mdx
```diff
@@ -10,62 +10,47 @@ It adds default styling and the `css` prop.
...
+  <option value="bananas">Bananas</option>
+</Select>
 ```
 
 ## Accessibility
 
 For accessibility purposes the component needs to have the `id` prop set, to link it to the label it is associated with. If a label is not available, please add an `aria-label` to ensure that the component remains accessible
 
-## Default Option
+## Default option
```
hollg, line 23: Maybe we could even get rid of this prop?

```
<Select css={{ width: 300 }}>
    <option value="" disabled selected hidden>Please select a fruit</option>
    <option value="apples">Apples</option>
    <option value="bananas">Bananas</option>
</Select>
```

hollg: Either way, I think we should rethink these examples because we're using `default` as a placeholder rather than a real default value, which likely to cause confusion.

thomasdigby: Yeah I considered that but I think that abstraction is a useful one if we're piecing together a few required attributes for that specific purpose, `placeholder` is a better name though I agree 👍

hollg: Yeah cool, if it is just for a placeholder rather than a default selected option, I think that makes sense 👍🏻

hollg:
I like it

andoulla:
> I like it

Same here. 
Relooking at this code I think, I shouldn't have just lifted it as-is from the old stuff.

thomasdigby:
> > I like it
> 
> Same here.
> Relooking at this code I think, I shouldn't have just lifted it as-is from the old stuff.

I thought that was the best API as well at the time, taking a step back and comparing against other components we've made makes it a lot more obvious

thomasdigby:
@hollg @andoulla This PR is good to merge if you're both happy with the API change

andoulla (approved):

hollg (approved):

PR #130, opened 2021-04-14
Feature: Minor component updates

thomasdigby:
- Upgraded stitches to 0.1.7 (all snapshots have been updated as the custom properties are now prefixed)
- Remove `pointer-events` from disabled button
- Ensure `Divider` renders correctly when centered
- Add support for `<Link>` rendering as `button` and removed default browser and visited styles
- Remove `fieldset` and `legend` default styles
- Updated `Grid` to use `display: grid` and `auto-fit` columns to apply same layout but with modern CSS (also fixes an issue with negative margins and the gap between items being double the requested size)
- Refactored `Grid` to improve the exported types
- Updated rendered `Grid` in tests 
- Removed wrapping widths from `Stack`
- Upgraded `react-docgen` to access default props correctly
- Removed default styles from `List`
- Fixed `generate-component-props` script to allow multiple component exports from a single file
- Added missing component exports from documentation
- Updated all instances of `{...rest}` to `{...remainingProps}`
- Removed unnecessary imports
- Updated the `Video` `externalId` to `id`


src/components/link/Link.test.tsx
```diff
@@ -42,7 +42,7 @@ describe(`Link component`, () => {
   })
 
   it('is polymorphic', async () => {
-    render(<Link as="button">See more</Link>)
+    render(<Link>See more</Link>)
```
andoulla, line 45: I find it a bit confusing, the way it is now. If I don't pass anything in it I would still expect to see a link. 

Before having the `as=button` made it very clear that the output would be

src/components/link/Link.tsx
```diff
@@ -6,15 +6,18 @@ import { Override } from '~/utilities'
 import { textVariantSize } from '../text'
 
 const StyledLink = styled('a', {
+  bg: 'unset',
```
andoulla, line 9: would `all:unset` work here? (just taking your recommendation from a different MR)

src/components/link/Link.tsx
```diff
@@ -6,15 +6,18 @@ import { Override } from '~/utilities'
...
+  cursor: 'pointer',
   fontFamily: '$sans',
   textDecoration: 'none',
-  cursor: 'pointer',
   '&:focus, &:hover': {
     color: '$primary900',
     textDecoration: 'underline'
   },
-  '&:active, &:visited': {
+  '&:active': {
```
andoulla, line 20: what colour should `visited` links should be? Now it will just take the browser default which might not look good

src/components/link/__snapshots__/Link.test.tsx.snap
```diff
@@ -35,37 +37,39 @@ exports[`Link component renders an anchor 1`] = `
...
   color: var(--colors-primary500);
+  cursor: pointer;
   font-family: var(--fonts-sans);
   text-decoration: none;
-  cursor: pointer;
 }
 
-.sx8xsnx:focus,
-.sx8xsnx:hover {
+.sxx4uif:focus,
```
andoulla, line 50: researching the serialisers all week, I think there is a way that we can adjust them to not flag up class changes and to only show styling changes. Doesn't belong in this MR of course but I thought it would improve the reviewing experience

src/components/list/List.tsx
```diff
@@ -4,51 +4,55 @@ import { styled } from '~/stitches'
...
           '&::marker': { color: '$tonal900' }
         }
       },
       primary: {
-        [`& ${StyledListItem}`]: {
+        [`& ${StyledLi}`]: {
           '&::marker': { color: '$primary500' }
         }
       },
       secondary: {
```
andoulla, line 38: I think we should add a variant that doesn't have a marker, it is going to be useful soon

src/components/link/Link.test.tsx
thomasdigby: Yeah actually looking at this test I think you're right, and there's no extra logic to handle, just the `as`

src/components/link/Link.tsx
thomasdigby: As we're setting a colour on the default link it will always use that colour, never the browser default. It actually overrides any colour set via `css` as it's more specific which is why we should remove it

src/components/link/__snapshots__/Link.test.tsx.snap
thomasdigby: Oh interesting, what did you find? these changes happen because any CSS change will update the underlying ID for the styles, therefore updating _all_ the classnames for the component

src/components/list/List.tsx
thomasdigby: Ah yes the List with just reset styles right? Worth adding a feature request to the repo and we can prioritise and delegate it to one of us

src/components/link/Link.tsx
thomasdigby: Yeah it's actually pretty dangerous, I found when used with other components it overwrites any less specific styles 😬 Need to change on the other components where it's used

andoulla: oh no! We added it in a few  places

src/components/list/List.tsx
andoulla: https://github.com/Atom-Learning/components/issues/131

src/components/link/__snapshots__/Link.test.tsx.snap
andoulla: I know, I can't find the blog now but I think it was by adding a classname prefix (50% sure it is this). Will try and find it and share

src/components/divider/Divider.tsx
```diff
@@ -1,11 +1,10 @@
...
 import { styled } from '~/stitches'
 
 export const Divider = styled('hr', {
   border: 0,
   borderTop: '1px solid $tonal400',
   my: '$4',
   mx: 'auto',
-  maxWidth: '50%'
+  maxWidth: '50%',
+  width: '100%'
```
hollg, line 9: What's the effect of these two rules being applied together?

hollg (approved):
LGTM!

src/components/divider/Divider.tsx
thomasdigby: Yeah this some nonsense CSS, I'll update width to be 50%

src/components/link/Link.tsx
thomasdigby: Yeah I'm going to revert those now and replace with some specific keys

src/components/link/__snapshots__/Link.test.tsx.snap
thomasdigby: Ping me if you find the article!

PR #134, opened 2021-04-21
Feat: update Form to take render prop

hollg:
Adds `render` prop to `Form` component, giving access to everything returned by the internal `useForm()` call. 

I added it as a prop rather than using `children` because I found [this article](https://americanexpress.io/faccs-are-an-antipattern/) on the subject pretty compelling and because it's consistent with how `react-hook-form`'s own components (`Controller`) achieve the same thing.

### DIFF
--- a/src/components/form/Form.mdx
+++ b/src/components/form/Form.mdx
@@ -228,3 +228,48 @@ const EmailField = ({ css, name, validation, required, ...rest }) => {
   )
 }
 ```
+
+## Accessing form data outside of fields
+
+The same object that `useFormContext` returns is available in an optional render prop function, making it easy to use the current form state to determine whether a button should be disabled or read the current value of a specific field. Note that if `render` is provided, `Form` should not be given any children. If both are provided, only `render` will be used.
+
+```tsx
+const SomeForm = () => {
+  return (
+    <Form
+      onSubmit={console.log}
+      render={({ formState }) => (
+        <>
+          <InputField
+            label="Name"
+            name="name"
+            validation={{ required: 'This field is required!' }}
+          />
+          <Button type="submit" disabled={!formState.isValid}>
+            Submit
+          </Button>
+        </>
+      )}
+    />
+  )
+}
+```
+
+You can also name your `render` function:
+
+```tsx
+const SomeFormContent = ({ formState }) => (
+  <>
+    <InputField
+      label="Name"
+      name="name"
+      validation={{ required: 'This field is required!' }}
+    />
+    <Button type="submit" disabled={!formState.isValid}>
+      Submit
+    </Button>
+  </>
+)
+
+const SomeForm = () => <Form onSubmit={console.log} render={SomeFormContent} />
+```
--- a/src/components/form/Form.mdx
+++ b/src/components/form/Form.mdx
@@ -234,25 +234,21 @@ const EmailField = ({ css, name, validation, required, ...rest }) => {
 The same object that `useFormContext` returns is available in an optional render prop function, making it easy to use the current form state to determine whether a button should be disabled or read the current value of a specific field. Note that if `render` is provided, `Form` should not be given any children. If both are provided, only `render` will be used.
 
 ```tsx
-const SomeForm = () => {
-  return (
-    <Form
-      onSubmit={console.log}
-      render={({ formState }) => (
-        <>
-          <InputField
-            label="Name"
-            name="name"
-            validation={{ required: 'This field is required!' }}
-          />
-          <Button type="submit" disabled={!formState.isValid}>
-            Submit
-          </Button>
-        </>
-      )}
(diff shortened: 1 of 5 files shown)


src/components/form/Form.tsx
```diff
@@ -12,30 +13,41 @@ type FormProps = Override<
     defaultValues?: { [key: string]: string | number }
     onSubmit: (data: any) => void
     validationMode: 'onBlur' | 'onSubmit'
-  }
+  } & (
+    | { children: React.ReactNode; render?: never }
+    | { children?: never; render: (methods: UseFormMethods) => React.ReactNode }
```
thomasdigby, line 18: 👍 👍

src/components/form/Form.tsx
```diff
@@ -12,30 +13,41 @@ type FormProps = Override<
...
   defaultValues = {},
   onSubmit,
   validationMode = 'onBlur',
+  render,
   ...remainingProps
 }) => {
+  if (process.env.NODE_ENV === 'development' && children && render)
+    console.warn(
+      '`Form` should only be given one of `children` or `render`. When both are provided, `render` will be used and `children` will be ignored.'
+    )
```
thomasdigby, line 33: I wonder if this is a good time to introduce `invariant` to consistently handle conditionals like these and provide a useful error message?

```jsx
invariant(
  children && render,
  '`Form` should only be given one of `children` or `render`. When both are provided, `render` will be used and `children` will be ignored.'
)
```

https://github.com/zertosh/invariant

hollg: Oh that looks cool! That does the environment check for us?

src/components/form/Form.mdx
```diff
@@ -228,3 +228,48 @@ const EmailField = ({ css, name, validation, required, ...rest }) => {
...
+          />
+          <Button type="submit" disabled={!formState.isValid}>
+            Submit
+          </Button>
+        </>
+      )}
+    />
+  )
+}
+```
```
thomasdigby, line 256: I'd remove the wrapping function here to keep it consistent with our other `<Form>` examples

thomasdigby (commented):
Looking good 👍

src/components/form/Form.mdx
hollg: Fixed!

src/components/form/Form.tsx
```diff
@@ -12,30 +14,42 @@ type FormProps = Override<
...
 
 export const Form: React.FC<FormProps> = ({
   children,
   defaultValues = {},
   onSubmit,
   validationMode = 'onBlur',
+  render,
   ...remainingProps
 }) => {
+  console.log('children:', children)
```
thomasdigby, line 31: ☠️

thomasdigby (approved):

src/components/form/Form.tsx
```diff
@@ -12,30 +14,42 @@ type FormProps = Override<
...
+  invariant(
```
andoulla, line 32: does it really help to throw this error in production? If someone passes both TS will throw an error, and in JS projects children will be ignored so they will no be able to see what children they are passing

hollg: Oh I misread the description of the library when Tom linked me to it — I thought it wouldn't throw at all in production. I think I agree with you that we probably don't want to throw in prod. @thomasdigby what do you think?

andoulla: maybe I'm looking at the wrong package

andoulla: > The invariant is thrown when the invariant condition is false.

“Invariant” is just a way of saying “this condition always holds true”. You can think about it as making an assertion.

> It is important to keep development and production behavior similar, so invariant throws both in development and in production. The error messages are automatically replaced with error codes in production to avoid negatively affecting the byte size.

https://reactjs.org/docs/codebase-overview.html

andoulla: wondering if this will be caught in sentry or will it be in the console. if it is in the console the chance is we wont even see it

hollg: I feel a bit conflicted about this, but 

> we should be very clear with the user that the behaviour is not supported

is pretty hard to argue with imo. 

I've updated the docs to explicitly state that an error will be thrown too. I guess someone can always raise an issue if they're getting errors in prod and don't think they should 😅

thomasdigby: @andoulla It will throw an error, not just log it. This will be caught by Sentry and display a generic error on the front-end using our error boundaries. I agree that a `console.log` is not good enough in most instances to flag that an error has occurred.

@hollg Yeah good idea to add to the documentation as well, if they're getting errors in prod then it's likely they're not using the component correctly and the errors have successfully communicated that 👌

PR #135, opened 2021-04-22
Feat: Switch component

mel-liow:


thomasdigby:
Hey @mel-liow, thanks for the contribution! It would be great if you could write a description about the component you're adding as well as some screenshots or videos of the behaviour.

I need to fix the documentation process so it's easy to preview the docs you've written so let me pull that down and confirm all is good.

package.json
```diff
@@ -115,6 +115,7 @@
     "@radix-ui/react-progress": "0.0.11",
     "@radix-ui/react-radio-group": "0.0.14",
     "@radix-ui/react-slot": "^0.0.9",
+    "@radix-ui/react-toggle": "^0.0.6",
```
thomasdigby, line 118: I assume you intended to use the "Switch" component from Radix? https://radix-ui.com/primitives/docs/components/switch

I think this is the correct component to use as the underlying properties and rendered elements meet the requirements for accessibility

src/components/toggle/Toggle.tsx
```diff
@@ -0,0 +1,46 @@
...
+    backgroundColor: '$success',
+    justifyContent: 'flex-end'
+  },
+  '&[data-state=off]': {
+    justifyContent: 'flex-start'
+  }
+})
+
+export const Switch = styled('span', {
+  size: '20px',
```
thomasdigby, line 30: Where we can use values from our tokens we should try to do so, `$2` would return `24px` which is close enough to `20px`

src/components/toggle/Toggle.tsx
```diff
@@ -0,0 +1,46 @@
...
+  appearance: 'none',
+  backgroundColor: '$tonal300',
+  border: 'none',
+  padding: '2px',
+  boxShadow: 'inset 0 0 0 1px $tonal300',
+  overflow: 'hidden',
+  borderRadius: '$round',
+  cursor: 'pointer',
+  display: 'inline-flex',
+  width: '50px',
```
thomasdigby, line 16: `$3` here would be `48px` and would match nicely with the `24px` we used for the `span`

src/components/toggle/Toggle.tsx
```diff
@@ -0,0 +1,46 @@
+import * as RadixToggle from '@radix-ui/react-toggle'
+import * as React from 'react'
+
+import { styled } from '~/stitches'
+
+const StyledToggle = styled(RadixToggle.Root, {
+  appearance: 'none',
+  backgroundColor: '$tonal300',
+  border: 'none',
+  padding: '2px',
```
thomasdigby, line 10: We could use `$0` instead of `2px`

src/components/toggle/Toggle.tsx
```diff
@@ -0,0 +1,46 @@
+import * as RadixToggle from '@radix-ui/react-toggle'
+import * as React from 'react'
+
+import { styled } from '~/stitches'
+
+const StyledToggle = styled(RadixToggle.Root, {
+  appearance: 'none',
+  backgroundColor: '$tonal300',
```
thomasdigby, line 8: I think we may need to bump this up to `$tonal400` to increase the contrast a touch

src/components/toggle/Toggle.tsx
```diff
@@ -0,0 +1,46 @@
...
+  overflow: 'hidden',
+  borderRadius: '$round',
+  cursor: 'pointer',
+  display: 'inline-flex',
+  width: '50px',
+  '&:focus': {
+    outline: 'none'
+  },
+  '&[data-state=on]': {
+    backgroundColor: '$success',
```
thomasdigby, line 21: Could we add a `transition` for this property?

thomasdigby (commented):
A few changes to use our design tokens, a switch to `Switch` and some documentation for usage and we're all looking good 👍

src/components/toggle/Toggle.mdx
```diff
@@ -0,0 +1,6 @@
+---
+title: Toggle
+component: Toggle
+description: A toggle is a digital on/off switch. Toggles are used to change the state of system functionality and preferences. They prompt users to choose between two mutually exclusive options and always have a default value.
```
andoulla, line 4: I think everything after the first sentence would be better if they got moved out of the description and into the main documentation. We generally kept descriptions quite short.

src/components/toggle/Toggle.test.tsx
```diff
@@ -0,0 +1,29 @@
+import { render, screen } from '@testing-library/react'
+import { axe } from 'jest-axe'
+import React from 'react'
+
+import { Toggle } from '.'
+
+describe(`Toggle component`, () => {
```
andoulla, line 7: It might be good to have a test with the toggle in the on state. Just so we are covering both states

src/components/toggle/Toggle.mdx
```diff
@@ -0,0 +1,6 @@
...
+category: Forms
+---
```
andoulla, line 6: Could we add a couple of lines here for more information? Examples of what could be added:

- It uses the Radix-Switch and a link to take the reader there
- How to use this component and where (please see my comment in line 4)
- What is the default state for this component? Can we set the default state?
- Accessibility - what is aria-label expecting?

package.json
hollg: Oh boy this is at least 99% my fault. Mel and I were on a call together looking through the radix components and found the toggle but didn't even look at the switch. Sorry Mel!! 😬

mel-liow: No worries! I'll make the `switch` 😉

mel-liow:
Implementation of Radix Switch button: 

https://user-images.githubusercontent.com/40150248/116394215-8cd02780-a81a-11eb-9bb7-9f65a3b0b52e.mov

Screenshot 2021-04-28 at 12 09 59

src/components/switch/Switch.mdx
```diff
@@ -0,0 +1,29 @@
+---
+title: Switch
+component: Switch
+description: An implementation of the Switch component from Radix with default styling and the css prop
+category: Forms
+---
+
+The Switch component implements [Radix's Switch component](https://radix-ui.com/primitives/docs/components/switch) where the default value can be set via `default` props.
+
+Switches are used to change the state of functionality and preferences. They prompt users to choose between two mutually exclusive options and will always have a default value.
```
hollg, line 10: Including these details about proper usage patterns is 🔥 but I think we can tighten it up a little to make it clearer:

```suggestion
Switches are used to toggle application state between two mutually exlusive values. They always have a default value.
```

src/components/switch/Switch.mdx
```diff
@@ -0,0 +1,29 @@
...
+
+The Switch component should not be used to handle data (e.g. where we submit data in a form it's encouraged to use checkboxes or radio buttons). Instead it should be used interactively where the effect is an immediate display or state change.
+
+
+## Appearance
+```tsx live
+<Switch />
```
hollg, line 17: The docs say that we need a default value but one isn't provided here. Can we clarify whether we should be providing one manually / what happens if we don't?

andoulla:
> Implementation of Radix Switch button:
> 
>  Screen.Recording.2021-04-28.at.12.08.42.mov 
> Screenshot 2021-04-28 at 12 09 59


It looks great! 

A couple of final comment from me is that the focus outline looks really thick and dark.  Should we consider making it thinner or at least use a primary hue instead?

Do we need more themes or is it always going to be green? Just wondering, if we should have a list of in which background colours this component comes in (like with button and badges)

src/components/switch/Switch.mdx
```diff
@@ -0,0 +1,29 @@
+---
+title: Switch
+component: Switch
+description: An implementation of the Switch component from Radix with default styling and the css prop
+category: Forms
+---
+
+The Switch component implements [Radix's Switch component](https://radix-ui.com/primitives/docs/components/switch) where the default value can be set via `default` props.
```
hollg, line 8: I think the comment about the `default` prop feels like a bit of a non sequitur here. It's definitely worth covering in the docs (see other comments on this file) but I don't think it fits in this para.

```suggestion
The Switch component implements [Radix's Switch component](https://radix-ui.com/primitives/docs/components/switch).
```

src/components/switch/Switch.mdx
```diff
@@ -0,0 +1,29 @@
...
+
+Switches are used to change the state of functionality and preferences. They prompt users to choose between two mutually exclusive options and will always have a default value.
+
+The Switch component should not be used to handle data (e.g. where we submit data in a form it's encouraged to use checkboxes or radio buttons). Instead it should be used interactively where the effect is an immediate display or state change.
```
hollg, line 12: This is all v good info but I think we can be a little clearer about what our users should take away from it:

```suggestion
The Switch component should not be used to collect form data — use checkboxes or radio buttons for that. Only use Switch inside a form if it is serving its primary purpose of toggling some state with immediate effect (e.g. to conditionally render a section of the form).
```

hollg (changes requested):
This looks great!

I have been super pedantic about docs changes (sorry!) but to be honest it's already one of the better documented components in the library. Unbelievable first contribution.

hollg:
Can we wrap instances of `Switch` in the docs in backticks?

src/components/switch/Switch.tsx
```diff
@@ -0,0 +1,49 @@
...
+  display: 'block',
+  width: '$1',
+  height: '$1',
+  backgroundColor: 'white',
+  borderRadius: '$round',
+  transition: 'transform 100ms',
+  transform: 'translateX(1px)',
+  willChange: 'transform',
+  '&[data-state="checked"]': {
+    transform: 'translateX(27px)'
```
thomasdigby, line 37: We could make a few tweaks to the styles to tighten up the visual output of this component. I'd recommend:

- Setting a `size: $2` to the `Thumb` to increase the height hit area
- Using tokens within the transform to couple the parent/child together `transform: 'translateX(calc($sizes$2 - $space$1))'` (we want to translate it horizontally by its own width and negate the internal padding of the parent, this could also be applied to the width of the parent to add that extra 4px to the width)
- Remove the `boxShadow` on the parent when focused, a standard outline should be fine. That should also make the default `boxShadow` redundant
- Adding a transition to the `backgroundColor` change on the parent

https://user-images.githubusercontent.com/3226804/116401332-09670400-a823-11eb-9881-ea6f8381e46d.mov

src/components/switch/Switch.mdx
```diff
@@ -0,0 +1,29 @@
...
+
+
+## Appearance
+```tsx live
```
thomasdigby, line 16: `live` isn't appropriate here, `preview` should be fine 👍 

I'd also maybe remove the `## Appearance` title to keep it consistent with the other docs

src/components/switch/Switch.mdx
mel-liow: Yeah that's my bad, that's an incorrect explanation and a hangover from the Toggle. I've changed the documentation to clarify

mel-liow:
Uploading Screen Recording 2021-04-28 at 14.31.33.mov…

Thanks @thomasdigby @hollg @andoulla for the comments!
I've made the changes so should be good for another review.

@andoulla re switch themes - I don't think that's necessary. I think we should strive to create familiar design and having different colours would add confusion to what should be a simple logic decision (e.g. having a red "on" state would be confusing for users).

andoulla:
> Uploading Screen Recording 2021-04-28 at 14.31.33.mov…
> 
> Thanks @thomasdigby @hollg @andoulla for the comments!
> I've made the changes so should be good for another review.
> 
> @andoulla re switch themes - I don't think that's necessary. I think we should strive to create familiar design and having different colours would add confusion to what should be a simple logic decision (e.g. having a red "on" state would be confusing for users).

oh, that's not what I meant. so sorry. I meant is there a chance where we will need a primary coloured switch or a secondary coloured switch. These can be added later

andoulla (approved):

src/components/switch/Switch.tsx
```diff
@@ -0,0 +1,47 @@
...
+  }
+})
+
+const StyledThumb = styled(RadixSwitch.Thumb, {
+  display: 'block',
+  size: '$2',
+  backgroundColor: 'white',
+  borderRadius: '$round',
+  transition: 'transform 100ms',
+  transform: 'translateX(1px)',
```
thomasdigby, line 32: What's the reason for the `1px` offset at the start? I think we can remove this

thomasdigby (approved):

hollg (approved):

PR #137, opened 2021-04-28
Feat: Add Combobox component

hollg:
Adds `Combobox` based on Reach UI's component: https://reach.tech/combobox/Ï

https://user-images.githubusercontent.com/21319237/116424431-6b326880-a839-11eb-91bf-5870533365f5.mov


src/components/combobox/ComboboxOption.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { styled } from '~/stitches'
+
+export const ComboboxOption = styled(BaseComboboxOption, {
+  '&[data-reach-combobox-option]': {
+    cursor: 'pointer',
+    margin: 0,
+    padding: '$2',
+
+    '&[aria-selected="true"]': {
+      background: 'hsl(211, 10%, 95%)',
```
andoulla, line 12: Should these be moved into the theme? or are they too specific?

hollg: Oh, these should be replaced with things from the theme! They came from the Reach UI example. I'll update them now :)

src/components/combobox/ComboboxPopover.tsx
```diff
@@ -0,0 +1,24 @@
+import { ComboboxPopover as BaseComboboxPopover } from '@reach/combobox'
+
+import { keyframes, styled } from '~/stitches'
+
+const slideDown = keyframes({
+  '0%': { transform: 'scaleY(0)' },
+  '100%': { transform: 'scaleY(1)' }
+})
```
thomasdigby, line 8: I'm not a big fan of this animation style, I understand that it's difficult to animate the height correctly but I'd rather avoid squashing and expanding the content.

For now I think we should remove the animation and have it snap in normally, I think that'll still look smart and feel nice and reactive for the users.

src/components/combobox/ComboboxOption.tsx
andoulla: ha I assumed that these were additional values... 🤦

src/components/combobox/ComboboxPopover.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export const ComboboxPopover = styled(BaseComboboxPopover, {
+  '&[data-reach-combobox-popover]': {
+    boxShadow: 'inset 0 0 0 1px $colors$primary900',
```
thomasdigby, line 12: Do we have any guidance from the design team on how this should look? If not I'd reduce these styles to something a bit simpler, mocked something up below:

Screenshot 2021-04-28 at 12 15 22

(Just an example but can provide the code if you need)

src/components/combobox/ComboboxPopover.tsx
```diff
@@ -0,0 +1,24 @@
...
+    outline: 'none',
+    boxSizing: 'border-box',
+    fontFamily: '$sans',
+    fontSize: '$sm', // prevent iOS zooming on focus
```
thomasdigby, line 16: `'$sm'` won't prevent zooming so we should remove the comment

src/components/combobox/ComboboxPopover.tsx
```diff
@@ -0,0 +1,24 @@
...
+    p: '$2',
+    border: 'solid 1px $primary900',
+    borderTop: 'none',
+    background: 'hsla(0, 100%, 100%, 0.99)',
```
thomasdigby, line 20: What's the reasoning for this white and _almost_ opaque background?

thomasdigby (commented):
Looking good and works well 👍

src/components/combobox/ComboboxPopover.tsx
hollg: Updated! 

https://user-images.githubusercontent.com/21319237/116424540-81d8bf80-a839-11eb-97d7-623d7d1c5973.mov

src/components/combobox/ComboboxList.tsx
```diff
@@ -0,0 +1,12 @@
+import { ComboboxList as BaseComboboxList } from '@reach/combobox'
+
+import { styled } from '~/stitches'
+
+export const ComboboxList = styled(BaseComboboxList, {
+  '&[data-reach-combobox-list]': {
```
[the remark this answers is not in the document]
hollg: I was, but it works without them and now I can't remember exactly what problem I solved by adding them 🤔. Updated!

src/components/combobox/ComboboxPopover.tsx
hollg: Ah that's just a holdover from the default `@reach/combobox` styling which I used as a starting point. Updated!

src/components/combobox/Combobox.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { axe } from 'jest-axe'
+import * as React from 'react'
+
+import { Label } from '../label'
+import { Combobox } from '.'
+
+const SimpleExample = () => (
+  <>
+    <Label htmlFor="fruit">What's your favourite fruit?</Label>
+    <Combobox openOnFocus>
```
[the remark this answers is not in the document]
hollg: Oh yeah that's much nicer, thanks!

src/components/combobox/ComboboxPopover.tsx
```diff
@@ -0,0 +1,23 @@
...
+  outline: 'none',
+  p: '$2',
+
+  // top and position are set by @reach/ui and
+  // override our styling without !important
+  // ... but !important causes a compilation error on position!
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  position: 'relative !important',
+  top: '$1 !important'
```
thomasdigby, line 22: Could we use `marginTop` instead to avoid this override?

src/components/combobox/ComboboxOption.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+import { styled } from '~/stitches'
+
+export const ComboboxOption = styled(BaseComboboxOption, {
+  cursor: 'pointer',
+  margin: 0,
+  padding: '$2',
+
+  '&:hover': {
+    bg: '$alpha100'
```
thomasdigby, line 11: I think `bg: '$tonal200'` and `borderRadius: '$0'` should polish it off nicely

src/components/combobox/ComboboxList.tsx
```diff
@@ -0,0 +1,10 @@
+import { ComboboxList as BaseComboboxList } from '@reach/combobox'
+
+import { styled } from '~/stitches'
+
+export const ComboboxList = styled(BaseComboboxList, {
+  listStyle: 'none',
+  margin: 0,
+  padding: 0,
```
thomasdigby, line 8: Just for consistency can we ensure that all available utilities are used when possible, e.g. `m: 0`, `p: 0`

src/components/combobox/ComboboxPopover.tsx
hollg: `@reach/ui` adds `position: absolute` and some magically-arrived-at `top` value. If we use `marginTop` here then those styles are still applied and still need removing.

src/components/combobox/ComboboxOption.tsx
hollg: Are both of those suggestions specifcally for `:hover`?

src/components/combobox/ComboboxPopover.tsx
thomasdigby: Do we want to remove the absolute position of the popover? I would have assumed that it was crucial to position the box relative to the input?

Potentially could use `transform: translateY($1)` to jus' bump it down a tad

src/components/combobox/ComboboxPopover.tsx
```diff
@@ -0,0 +1,16 @@
...
+  bg: 'white',
+  border: 'solid 1px $tonal500',
+  borderRadius: '$0',
+  boxShadow: '$0',
+  boxSizing: 'border-box',
+  fontFamily: '$sans',
+  fontSize: '$md',
+  outline: 'none',
+  p: '$2',
+  transform: 'translateY(4px)'
```
thomasdigby, line 15: This should be a layout value `transform: 'translateY($1)'`

thomasdigby (approved):
One minor change, all looking good 👍 👁️ 👁️ 👍

andoulla (approved):

src/components/combobox/ComboboxPopover.tsx
hollg: Yeah I tried that yesterday but it didn't work. The style in the browser is `transform: translateY(var(--sx-1))` but `--sx-1` isn't defined.

thomasdigby: Ok, may have to be `transform: 'translateY($space$1)'` or `transform: 'translateY($spaces$1)'` (I can never recall)

hollg: Both of those have the same problem (`--sx-space1 is not set`, `--sx-spaces1 is not set` respectively).

A while ago there was some other CSS property that I couldn't get these values to work with, though now I can't remember what it was. Do we expect them to work with all properties?

thomasdigby: Oh interesting, they should work with all properties. Let me test quickly

thomasdigby: Screenshot 2021-04-29 at 09 15 06

It's working ok for me locally? What environment are you testing in?

hollg: Locally on both firefox and chrome... let me just delete all the built stuff and try again

hollg: Screenshot 2021-04-29 at 09 26 07

thomasdigby (approved):

PR #141, opened 2021-05-05
Feature: Update sandbox to use `vite`

thomasdigby:
Updates our sandbox to use `vite` instead of `parcel`, this improves on our current sandbox with quicker builds and allows us to import from `./src`.

Had to make a few minor adjustments to our exports to ensure that `vite` could run correctly.

Screenshot 2021-05-05 at 18 47 05


hollg (approved):

hollg:
Looks good! What is it about `vite` that means we can nor import from `src`?

sandbox/vite.config.js
```diff
@@ -0,0 +1,21 @@
+import reactRefresh from '@vitejs/plugin-react-refresh'
+import path from 'path'
+
+/**
+ * @type {import('vite').UserConfig}
+ */
+const config = {
+  plugins: [reactRefresh()],
```
andoulla, line 8: Glad to see this one!

andoulla (approved):

thomasdigby:
> Looks good! What is it about `vite` that means we can nor import from `src`?

Honestly I think it's just the path aliases which I'm sure are configurable in parcel but we never got around to it. I thought I'd take the opportunity to test the switch to vite as it's _so quick_ and with a few minor updates to get `./src` imports working I thought it was worth PRing.

PR #144, opened 2021-05-06
Feature: Stack Content component

### LINKED ISSUE #139: Feature request: Stack Content
## Component Name(\*) or Feature Name

`StackContent` or something similar 🤷 

## Description(\*)

This component should wrap standard content components like `Text`, `List`, `Heading`, and add bottom margins to space them properly. e.g.

```jsx
<StackContent>
  <Heading size="lg">Something</Heading> {/* has margin-bottom related to size */}
  <Text>Hello</Text> {/* has no margin bottom */}
</StackContent>
```

## Designs and UX wireframes(\*)

No designs

thomasdigby:
- Applies spacing and layout styles to `Heading`, `Text`, `List`, `Image` and `Divider` child components
- Max width `65ch` and `75ch` are there to preserve readability, I can see it being a property that others might like to remove but I'm feeling pretty dogmatic about the benefits http://webtypography.net/2.1.2
- Display block for `Image` ensures that margins are collapsed above/below it

Screenshot 2021-05-06 at 17 27 11

Closes #139


src/components/stack-content/StackContent.tsx
```diff
@@ -0,0 +1,67 @@
...
+    {React.Children.map(children, (child) => {
+      if (!React.isValidElement(child)) {
+        return child
+      }
+
+      const type = child?.type
+
+      if (type === Heading) {
+        return cloneWithStyle(child, {
+          maxWidth: '75ch',
```
rebeccahuseyina: I'm wondering how using the `ch` unit would work if the user set their browser text size in `px`? Like if they set their text size to `1px` would they still only be able to see 75 characters max even on a very wide screen?

andoulla (approved):

src/components/stack-content/StackContent.tsx
thomasdigby: Good question, the answer is yes! And that's a good thing, there are a few edge-cases where text size will impact potentially how many characters we might want to limit the line-length to, but in general, 55-75 characters is a good, readable line-length for any text size.

andoulla (approved):

PR #145, opened 2021-05-07
Feature: Toast package update

thomasdigby:


src/components/toast/Toast.mdx
```diff
@@ -5,26 +5,16 @@ description: A toast notification that gives non-intrusive feedback to the user
 category: Feedback
 ---
 
-`Toast` has 2 exports, the `ToastProvider` component which is the container for all the notifications and the `toast` function that allows triggering the notification. It sets all the default behaviour, like pausing on hover and positioning and styles the different types of notification (`info` (default), `error`, `warning` and `success`).
+`Toast` has 2 exports, the `ToastProvider` component which is the container for all the notifications and the `toast` function that allows triggering the notification. It sets all the default behaviour, like pausing on hover and positioning and styles the different types of notification (`blank` (default), `error` and `success`).
```
andoulla, line 8: I'm not the biggest fan of `blank`. Why not use `info`? most notification/alert patterns have an info variant

src/components/toast/Toast.tsx
```diff
@@ -1,127 +1,121 @@
...
+  p: 'unset',
+  size: '$3'
 })
 
-type ToastProps = React.ComponentPropsWithoutRef<typeof StyledToastContainer>
+type ToastProps = React.ComponentProps<typeof StyledToast> &
+  ToastInterface & {
+    calculateOffset: (
+      id: string,
+      opts?: { reverseOrder?: boolean; margin?: number }
```
andoulla, line 73: could we use the full name here? options seems a bit more readable to me

src/components/toast/Toast.tsx
```diff
@@ -1,127 +1,121 @@
...
+    message,
+    role,
+    type = 'blank',
+    visible,
+    calculateOffset,
+    updateHeight
+  }) => {
+    const offset = calculateOffset(id, {
+      reverseOrder: true,
+      margin: 8
```
andoulla, line 92: is there a way to use the space token here instead of a hardcoded 8?

src/components/toast/Toast.mdx
thomasdigby: It's the toast package that supplies these statuses, I _could_ map them onto something different for us but as `blank` is the only one we're not 💯 on, I don't think we need to worry as that'll just be the default. e.g. `toast()` rather than `toast.blank()`

src/components/toast/Toast.tsx
thomasdigby: Yeah sure, I just copied this from the package (annoyed that they didn't export them!)

src/components/toast/Toast.tsx
thomasdigby: Yes there should be, although I think `calculateOffset` requires a number, which the theme doesn't export. It exports the values as rems.

src/components/toast/ToastProvider.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { useToaster } from 'react-hot-toast'
+
+import { styled } from '~/stitches'
+
+import { Toast } from './Toast'
+
+export { default as toast } from 'react-hot-toast'
+
+const Container = styled('div', {
+  left: 'calc(50% - 360px / 2)',
```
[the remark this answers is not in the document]
thomasdigby: Just to make it clear what the number is in relation to, I might pull out that `360` value and name it, that might make more sense within this context.

.eslintrc
```diff
@@ -46,7 +46,11 @@
   "rules": {
     "no-prototype-builtins": "off",
     "no-unneeded-ternary": ["warn", { "defaultAssignment": false }],
-    "no-unused-vars": ["warn", { "ignoreRestSiblings": true }],
+    "no-unused-vars": "off",
```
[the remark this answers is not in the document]
thomasdigby: The line below it switches on the `no-unused-vars` TS variant which prevents those annoying warning with unsed vars in type/interface definitions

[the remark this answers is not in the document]
hollg: OH! That's awesome.

hollg (approved):

andoulla (approved):

PR #147, opened 2021-05-09
Feature: ActionIcon

thomasdigby:
- New component `ActionIcon` for wrapping `Icon`s that need to be interactive
- Created `outline` and `solid` appearance variants (which tend to be the common conventions in the main app), also added an `appearance="none"` for flexibility
- Decided against coupling `Icon` size with `ActionIcon` as we want a bit of flexibility with icon size, especially if we want to use it in the main app as they are currently styled
- Replaced the button in `PasswordField` with `ActionIcon`

Screenshot 2021-05-09 at 17 36 23


src/components/action-icon/ActionIcon.tsx
```diff
@@ -0,0 +1,94 @@
...
+    ref
+  ) => {
+    const INVALID_CHILDREN_MESSAGE = `A single ${Icon.displayName} component is permitted as a child of ${ActionIcon.displayName}`
+
+    invariant(React.Children.count(children) === 1, INVALID_CHILDREN_MESSAGE)
+
+    return (
+      <StyledButton
+        type="button"
+        {...remainingProps}
```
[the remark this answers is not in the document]
thomasdigby: Yeah it doesn't make any difference as those props are extracted from `remainingProps` anyway.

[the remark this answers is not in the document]
hollg: Ah cool, that makes sense! Thanks.

src/components/action-icon/ActionIcon.mdx
```diff
@@ -0,0 +1,35 @@
+---
+title: Action Icon
```
andoulla, line 2: Interactive Icon fits better in my opinion as better describes what it does

src/components/action-icon/ActionIcon.mdx
```diff
@@ -0,0 +1,35 @@
...
+  <ActionIcon label="Attach a file" appearance="outline">
+    <Icon is={Clip} size="sm" />
+  </ActionIcon>
+  <ActionIcon label="Attach a file" appearance="solid">
+    <Icon is={Clip} size="sm" />
+  </ActionIcon>
+</>
+```
+
+It is recommended to pair `ActionIcon` with `Tooltip` to provide a clear visual description of the action the user is about to perform.
```
andoulla, line 24: I always hesitate to suggest using the tooltip as they don't work on mobile devices well. It would be better to update the designs to remove the need for a tooltip

src/components/action-icon/ActionIcon.tsx
```diff
@@ -0,0 +1,94 @@
...
+      outline: {
+        border: '1px solid $tonal500',
+        bg: 'transparent',
+        color: '$tonal600',
+        '&:hover,&:focus': {
+          borderColor: '$primary900',
+          color: '$primary900'
+        }
+      },
+      solid: {
```
hollg, line 33: To get a `solid` `ActionButton` with a different bg color (e.g. a trash can icon on a red bg), would we do `<ActionIcon css={{bg: '$danger500'}}>` ?

src/components/action-icon/ActionIcon.tsx
```diff
@@ -0,0 +1,94 @@
...
+        color: '$tonal800',
+        '&:hover,&:focus': {
+          color: '$primary900'
+        }
+      },
+      outline: {
+        border: '1px solid $tonal500',
+        bg: 'transparent',
+        color: '$tonal600',
+        '&:hover,&:focus': {
```
andoulla, line 28:
```suggestion
        '&:hover, &:focus': {
```

src/components/action-icon/ActionIcon.tsx
```diff
@@ -0,0 +1,94 @@
...
+}
+
+export const ActionIcon: React.FC<ActionIconProps> = React.forwardRef(
+  (
+    { children, appearance = 'none', size = 'md', label, ...remainingProps },
+    ref
+  ) => {
+    const INVALID_CHILDREN_MESSAGE = `A single ${Icon.displayName} component is permitted as a child of ${ActionIcon.displayName}`
+
+    invariant(React.Children.count(children) === 1, INVALID_CHILDREN_MESSAGE)
```
andoulla, line 64: shouldn't this be `!==1`

src/components/action-icon/ActionIcon.mdx
thomasdigby: @hollg Any thoughts?

@andoulla I've seen `ActionIcon` a few times in other design systems and I prefer `ActionIcon` as I think "action" is better term to describe the intent than "interactive", the user will be performing an _action_ with this icon. It's also less verbose which helps when you're likely to have a few icons/action-icons as siblings in a view

src/components/action-icon/ActionIcon.tsx
hollg: I think it's worth building those three in (and wouldn't add any more than that), but would the above example _work_ to override it?

src/components/action-icon/ActionIcon.mdx
thomasdigby: I understand where you're coming from and in an ideal world I'd agree and remove it, but this recommendation is borne from the potential for some developers to use this without adding a text description at all. A tooltip is a nice compromise that gets us most of the way there.

Something for @Amit-AtomLearning and @Chema-Atom to consider when writing their guidelines for sure 👍

src/components/action-icon/ActionIcon.tsx
thomasdigby: This initially threw me a few times as well as `invariant` doesn't test for a negative but instead defines a required truthy value for the function to work. It's a slightly more involved way of validating your props before use.

src/components/action-icon/ActionIcon.tsx
```diff
@@ -0,0 +1,94 @@
...
+        color: '$tonal600',
+        '&:hover, &:focus': {
+          borderColor: '$primary900',
+          color: '$primary900'
+        }
+      },
+      solid: {
+        bg: '$tonal200',
+        color: '$tonal800',
+        '&:hover,&:focus': {
```
andoulla, line 36:
```suggestion
        '&:hover, &:focus': {
```

src/components/action-icon/ActionIcon.tsx
thomasdigby: Added 4 themes, `primary`, `success`, `warning` and `error`, and created a new variant called `simple` which just changes the icon color.

You can pass in `bg` to change the background but it will be overridden when the user hovers/focuses the icon.

thomasdigby:
@hollg @andoulla Any last thoughts before I merge?

src/components/action-icon/ActionIcon.tsx
hollg: Sounds good! (And yes, you can do all the nested styling shenanigans in a `css` prop that you can inside a `styled()` call!)

andoulla (approved):

Issue #148, opened 2021-05-10
Feature request: Polymorphic Link and Button


hollg:
## Component Name(\*) or Feature Name

Polymorphic `Link` and `Button`.

## Description(\*)
From a Slack message:
> I saw another #opinion of Jamies that I’m keen on (we had this change in the DS but removed it before merging) https://twitter.com/buildsghost/status/1389294014222934018 - Does anyone have thoughts on this pattern?

>>Jamie Kyle 🏳️‍🌈Jamie Kyle 🏳️‍🌈 @buildsghost
>>Design System opinion:
Don't make your <Link> and <Button> components support both link and button visuals
Do make your <Link> and <Button> components renders either <a> or <button> based on the props passed to it ("to" should render an anchor, "onClick" should render a button) https://pbs.twimg.com/media/E0fC-UlVoAIzf9l.jpg
TwitterTwitter | May 3rd (49 kB)

## Designs and UX wireframes(\*)

N/A

PR #154, opened 2021-05-18
feat: polymorphic link and button

michalrozenek:
This PR is an answer on following issue:
https://github.com/Atom-Learning/components/issues/148

Changes:
Support Polymorphic feature for Link and Button.

Link can render anchor or button depends on passed properties (onClick / href)
Button can render button or anchor depends on passed properties (onClick / href)

### DIFF
--- a/src/components/button/Button.mdx
+++ b/src/components/button/Button.mdx
@@ -76,3 +76,19 @@ Setting `isRounded` creates a rounded button
 ```tsx live
 <Button isRounded>Rounded Button</Button>
 ```
+
+## Polymorphism
+
+The `Button` component supports polymorphism, therefore depending on whether it receives an `onClick`/`href` as a prop, it will produce a `button` or `link` respectively 
+
+```tsx preview
+<Button href="http://example.com/">
+  I'm a link
+</Button>
+```
+
+```tsx preview
+<Button onClick={() => console.log('clicked')}>
+  I'm a button
+</Button>
+```
--- a/src/components/button/Button.tsx
+++ b/src/components/button/Button.tsx
@@ -4,6 +4,7 @@ import * as React from 'react'
 import { Box } from '~/components/box'
 import { Loader } from '~/components/loader'
 import { styled } from '~/stitches'
+import { NavigatorActions } from '~/types'
 import { Override } from '~/utilities'
 
 const getButtonOutlineVariant = (baseColor: string, interactColor: string) => ({
@@ -195,18 +196,16 @@ type ButtonProps = Override<
     as?: React.ComponentType | React.ElementType
     children: React.ReactNode
     isLoading?: boolean
-    onClick?: () => void
-    to?: string
-  }
+  } & NavigatorActions
 >
 
-export const Button: React.FC<ButtonProps> = React.forwardRef(
+export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
   (
     {
       children,
       isLoading,
       onClick,
-      to,
+      href,
       appearance = 'solid',
       size = 'md',
       theme = 'primary',
@@ -215,7 +214,13 @@ export const Button: React.FC<ButtonProps> = React.forwardRef(
     },
     ref
   ) => {
-    const optionalLinkProps = to ? { as: 'a', href: to } : {}
+    const optionalLinkProps = href
+      ? {
+          as: 'a',
+          href,
+          onClick: undefined
+        }
+      : {}
 
     // Note: button is not disabled when loading for accessibility purposes.
     // Instead the click action is not fired and the button looks faded
@@ -239,6 +244,6 @@ export const Button: React.FC<ButtonProps> = React.forwardRef(
       </StyledButton>
     )
   }
-)
+) as React.FC<ButtonProps>
 
 Button.displayName = 'Button'
--- a/src/components/link/Link.mdx
+++ b/src/components/link/Link.mdx
@@ -35,3 +35,19 @@ It adds default styling and the `css` prop.
   7-12+. Prepare your child for secondary school entry and beyond.
 </Text>
 ```
+
+## Polymorphism
+
+The `Link` component supports polymorphism, therefore depending on whether it receives an `onClick`/`href` as a prop, it will produce a `button` or `link` respectively 
+
+```tsx preview
+<Link href="http://example.com/">
+  I'm a link
+</Link>
+```
+
+```tsx preview
+<Link onClick={() => console.log('clicked')}>
+  I'm a button
+</Link>
+```
--- a/src/components/link/Link.tsx
+++ b/src/components/link/Link.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react'
 
 import { styled } from '~/stitches'
+import { NavigatorActions } from '~/types'
 import { Override } from '~/utilities'
 
 import { StyledHeading } from '../heading/Heading'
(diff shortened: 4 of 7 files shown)


thomasdigby:
@michalrozenek Do we want to limit users use of `onClick` based on whether it's a link or not? We may have cases where we want to fire an `onClick` on a link, for example https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon for event tracking

src/components/button/Button.mdx
```diff
@@ -76,3 +76,20 @@ Setting `isRounded` creates a rounded button
 ```tsx live
 <Button isRounded>Rounded Button</Button>
 ```
+
+## Polymorphic
+
+We support polymorphic feature, so depends on passed property
```
andoulla, line 82: I think this reads a bit better and is more specific to the behaviour implemented
```suggestion
The `Button` component supports polymorphism, therefore depending on whether it receives an `onClick`/`href` as a prop, it will produce a `button` or `link` respectively 
```

src/components/button/Button.mdx
```diff
@@ -1,7 +1,7 @@
 ---
 title: Button
 component: Button
-description: The Button component is a light wrapper around an HTML button element
+description: The Button component is a light wrapper around an HTML button(default) or HTML anchor element
```
andoulla, line 4: I don't think we need to include the anchor in the description, it is a secondary use case of the component

src/components/link/Link.mdx
```diff
@@ -1,7 +1,7 @@
 ---
 title: Link
 component: Link
-description: Link is a light wrapper around the HTML anchor element
+description: Link is a light wrapper around the HTML anchor(default) or HTML button element
```
andoulla, line 4: Same as with the button description. I think we could omit secondary use cases in the description, to keep the documentation simple and easy to scan

src/components/link/Link.mdx
```diff
@@ -24,3 +24,19 @@ It adds default styling and the `css` prop.
   I'm a link
 </Link>
 ```
+## Polymorphic
+
+We support polymorphic feature, so depends on passed property
```
andoulla, line 29: same as above

src/components/button/Button.mdx
michalrozenek: I think we should because it's true. We render HTML `a` or `button` depends on properties.

michalrozenek:
> @michalrozenek Do we want to limit users use of `onClick` based on whether it's a link or not? We may have cases where we want to fire an `onClick` on a link, for example https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon for event tracking

@thomasdigby I think we do not limit users since they can use `onClick` rather than `href` for `<Link />` component.
So they can pass a function which for example trigger event tracking but also navigate the user to some URL (e.g. `window.location.href = URL`)

Did I miss something?

src/components/button/Button.mdx
andoulla: yes but I think it is confusing if you don't have any context. In the section below were it explains why and how it makes more sense.

src/components/button/Button.mdx
michalrozenek: Thank you

thomasdigby:
> > @michalrozenek Do we want to limit users use of `onClick` based on whether it's a link or not? We may have cases where we want to fire an `onClick` on a link, for example https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon for event tracking
> 
> 
> 
> @thomasdigby I think we do not limit users since they can use `onClick` rather than `href` for `<Link />` component.
> 
> So they can pass a function which for example trigger event tracking but also navigate the user to some URL (e.g. `window.location.href = URL`)
> 
> 
> 
> Did I miss something?

I think I'd prefer to retain a consistent API for navigating to pages e.g. keeping href. Might just need to update the types to allow both `onClick` and `href` for both components?

src/components/button/Button.mdx
hollg: I agree with @andoulla. I think the `description` should be enough to help someone recognise that this is the component they're looking for; nothing more. We can probably improve the `description` fields for pretty much all our components in this sense, to be honest. Ideally our users won't even need to worry about what element is rendered — they should just find the component that meets their functionality requirement and be able to trust that the correct element will be rendered in the same way they should be able to trust that it's well-tested, meets the relevant accessibility requirements etc.


What I think we should communicate here is something to the effect of "`Button` renders something that looks like a button and can handle both click events and navigation".

src/components/button/Button.mdx
```diff
@@ -76,3 +76,20 @@ Setting `isRounded` creates a rounded button
 ```tsx live
 <Button isRounded>Rounded Button</Button>
 ```
+
+## Polymorphic
```
hollg, line 80: I think the headings should probably be the noun form of the word
```suggestion
## Polymorphism
```

src/components/button/Button.tsx
```diff
@@ -189,24 +189,26 @@ const getChildren = (children) =>
...
     children: React.ReactNode
     isLoading?: boolean
-    onClick?: () => void
-    to?: string
-  }
+  } & ButtonActions
 >
 
-export const Button: React.FC<ButtonProps> = React.forwardRef(
+export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
```
hollg, line 205: What's the reason for this chance and casting to `as React.FC<ButtonProps>` at line 250?

src/components/link/Link.tsx
```diff
@@ -25,17 +25,30 @@ const StyledLink = styled('a', {
   }
 })
 
+type LinkActions =
```
hollg, line 28: If this type is identical to `ButtonActions`, is there a sensible way to re-use one type between both files?

src/components/button/Button.tsx
michalrozenek: The reason is that we use forwardRef, so whole component is wrapped by React.forwardRef which is kind of `React.ForwardRefExoticComponent`.

`React.forwardRef` is a generic function that has type parameters for the type of the ref and the props, so I typed it 
```
React.forwardRef<HTMLButtonElement, ButtonProps>
```

but then I had problems with properties so I wanted to tell TS that all that is inside forwardRef is type `React.FC< ButtonProps>`

src/components/link/Link.tsx
michalrozenek: Actually, it is. We can re-use it.
Do you have a place for a shared type? I did not check yet.

src/components/button/Button.mdx
michalrozenek: Cool, that's fine for me :)

src/components/button/Button.tsx
hollg: Huh. What I'm not understanding is that in the existing version of the `Button` component, we do `export const Button: React.FC<ButtonProps> = React.forwardRef( /* stuff/* )`, which seems to suggest that the output of `React.forwardRef` _is_ of type `React.FC` (or at least can be implicitly cast to that type), so it's unclear to me why we need the more verbose/explicit syntax after these changes.

Or are you saying that, unrelated to the polymorphism changes, the more explicit syntax gives us other benefits? E.g. I don't think the compiler can properly determine the type of `ref` in the previous version, but I suppose it probably can with these changes?

src/components/link/Link.tsx
hollg: This is a very good question! We have a `types` directory, but it's outside of `src`... I'm not sure why that is, but it seems odd to me. My intuition is that a shared type for this should be inside `src/components/types`, but I would be keen to hear other people's thoughts on that.

thomasdigby: Maybe `src/types`?

src/components/button/Button.tsx
michalrozenek: Zrzut ekranu 2021-05-18 o 17 04 49

When I tried to leave types as it was before ☝️. And btw if I removed forwardRef error disappeared. So I checked forwardRef types and I tried to handle that issue

michalrozenek: So @hollg what do you think?
I would say - I haven't better idea how we can type it to make it typed properly.
After my types implementation component is aware of all properties as we expect and not throwing any error.

Maybe you have another idea how can we achieve it. Let me know.

hollg: 😨 yeah fair enough, your version certainly makes sense and that error looks horrible so getting rid of it like this seems like a win!

hollg (approved):

andoulla (approved):

PR #156, opened 2021-05-18
Fix: Stitches theme autocomplete

thomasdigby:
Fixes autocomplete for theme values and utilities in `styled()`.

Requires an update to `@atom-learning/theme` and for someone to figure how how to get the generic `keyof typeof atomTheme[T]` to work 🤷


thomasdigby:
Have requested reviews from y'all in case you have any idea how I can get the `UtilTheme` type to work

src/stitches.ts
```diff
@@ -1,80 +1,78 @@
...
+  inset: () => (value: UtilTheme<'space'>): CSSBlob => ({
     top: value,
     right: value,
     bottom: value,
     left: value
   }),
-  bg: () => (value: string): CSSBlob => ({
-    background: value
+  size: () => (value: UtilTheme<'sizes'>): CSSBlob => ({
+    height: value,
```
andoulla, line 69: how is the condition for different size and width remove (for my FYI)?

Wondering if this should be square_size so to show that it will be the same width And Height

thomasdigby: As we're not using it currently I thought it might be better to simplify this method and be explicit about height/width differences. I don't think we necessarily need to change the key, as I think `size` has worked well so far?

andoulla: I haven't gotten the chance to use it yet but let leave it so people can let us know if it is confusing

thomasdigby:
Unnecessary now we're upgrading to Stitches v1

PR #161, opened 2021-05-25
feat: add PasswordInput

### LINKED ISSUE #159: Feature request: <Input type=password/> hide/show value
## Description(\*)

Our `PasswordField` component has some logic to show or hide the entered password when the user clicks an eye icon. I think we should move this logic down to the underlying `Input` component so that we get this behaviour whenever `type = "password"`. 

This will allow us to compose other, more specific types of password field and get consistent show/hide behaviour for free.

For example, the `CreatePasswordField` would get it for free.

hollg:
Closes #159

Refactors `PasswordField` to use a new `PasswordInput` which provides the visibility toggling functionality. `PasswordInput` will be used in the main app to compose things like the `CreatePasswordField` while getting the visibility toggling for free.


src/components/password-input/PasswordInput.tsx
```diff
@@ -0,0 +1,52 @@
...
+      showPasswordText = 'Show password',
+      ...restProps
+    },
+    ref
+  ) => {
+    const [isPasswordVisible, setIsPasswordVisible] = useState<boolean>(false)
+    const togglePasswordVisibility = () =>
+      setIsPasswordVisible((currentState) => !currentState)
+
+    console.log(restProps)
```
andoulla, line 28:
```suggestion

```

andoulla (approved):

src/components/password-field/PasswordField.tsx
```diff
@@ -26,52 +23,34 @@ export const PasswordField: React.FC<PasswordFieldProps> = ({
...
-    </Box>
+    <FieldWrapper
+      label={label}
+      fieldId={name}
+      prompt={prompt}
+      css={{ ...css, position: 'relative' } as CSS}
+      error={error}
+    >
+      <PasswordInput
+        css={{ pr: '$sizes$2' }}
```
thomasdigby, line 45: This `css` should live with the `PasswordInput` as we need it to offset the icon (which is coupled with `PasswordInput`

src/components/password-field/PasswordField.tsx
```diff
@@ -1,14 +1,11 @@
...
 
-import { ActionIcon } from '~/components/action-icon'
-import { Box } from '~/components/box'
 import { FieldWrapper } from '~/components/field-wrapper'
 import type { ValidationOptions } from '~/components/form'
-import { Icon } from '~/components/icon'
-import { Input, InputProps } from '~/components/input'
+import { InputProps } from '~/components/input'
+import { PasswordInput } from '~/components/password-input'
+import { CSS } from '~/stitches'
```
thomasdigby, line 8: Do we need to import this as a type? Not sure what we're doing elsewhere in the codebase

src/components/password-field/PasswordField.tsx
hollg: 🤔 it works fine with this regular `import` but I'm not really sure what the difference is

thomasdigby: I assume you meant to reply to the other comment 😄 but yeah we'll keep as is, I think a good convention (and what vite requires for _some_ imports 🤔) is to define imports as `type`s to make the relationship explicit

hollg: Lol yes, sorry! The comment about the `css` prop is definitely correct! 😅

thomasdigby: @hollg I think you've removed the style, rather than move it. We still need to apply to bigger `pr` to offset the icon in the `PasswordInput` component

hollg: lmao it never made it out of my clipboard. I'm finding this comment very challenging!

thomasdigby (approved):

PR #163, opened 2021-06-01
fix: updated condition to apply margin on Icon inside Button

Mhoog:
When using the library in another project, the margin on `<Icon />` inside a `<Button />` was not applied. This was because the margin was applied on components with the name `'icon'`. This name was minified, so the condition always resulted in a `false` when using the library.

Updated the condition to check if the child is an instance of the `Icon` component instead of checking the name.

_Note: the snapshot test for `PasswordField` failed. Decided to update that and include it here as it only changed the order of CSS._


hollg (approved):

andoulla (approved):
Thanks for fixing my code!

PR #164, opened 2021-06-08
Fix: add z-index to styled toast

mel-liow:
Fix to add a z-index to the toast component


thomasdigby:
Can we add a snapshot test for the container? Just realised that we're missing one

thomasdigby (approved):
Perfect, thanks @mel-liow !

Issue #165, opened 2021-06-09
Feature request: Form Component - localStorage persist functionality


LeahB8:
## Add persist to sessionStorage functionality to the form component

Form.tsx

## Description

- Integrate persist functionality for the form using sessionStorage
- As a USER I want to be able to keep my form data between refreshes
- As a DEV I want an integrated and easy to use option to retain form data between refreshes so I don't have to write a custom approach each time

andoulla:
Hi @LeahB8, could we get more details on this? 

What are the use cases where we need to add this in the form component? Ie, what problem are we trying to address

LeahB8:
> Hi @LeahB8, could we get more details on this?
> 
> What are the use cases where we need to add this in the form component? Ie, what problem are we trying to address

Tom suggested adding this as I was in the process of writing a custom approach to doing this in the new sign up journey forms. I know multiple places on the app that could use this functionality where I or someone else have written a custom approach (localStorage or redux) which can get quite messy.

thomasdigby:
@hollg @LeahB8 Should we close this issue or keep it open until we fix the issues with it?

LeahB8:
> @hollg @LeahB8 Should we close this issue or keep it open until we fix the issues with it?

We could keep it open until the issues are fixed

hollg:
I think it's worth keeping this open.

The issues that I'm aware of with the current implementation are:
- `react-hook-form` itself doesn't provide a way to read default values via `useFormContext()`
- `react-hook-form-persist` doesn't interact at all with default values for the form


Consequently it's difficult to get persisted values into our complex field components (i.e. the ones that involve a `Controller`) because in those ones we have to manage default values ourselves.

To solve this issue I think we should:
- add functionality to our `Form` to make default values accessible to the nested field components (maybe a custom hook?)
- add our own persistence logic that correctly updates default values for complex field components

andoulla:
@hollg do you think this is still relevant to keep open?

PR #168, opened 2021-06-09
feat(persist-form): adding react-hook-form-persist package

LeahB8:
Linked issue - https://github.com/Atom-Learning/components/issues/165

## In this PR
- [x] Integrated persist functionality for the form in sessionStorage using `react-hook-form-persist`
- [x] Allow for `shouldPersist` prop to be sent to the form to indicate whether it should be used
- [x] If `shouldPersist` is true, persisted data in sessionStorage will populate `defaultValues` in the form

### DIFF
--- a/src/components/form/Form.mdx
+++ b/src/components/form/Form.mdx
@@ -153,17 +153,16 @@ The `process` property accepts a function of type `(value: any) => any`. It can 
 Here's an example using the `persist` prop:
 
 ```tsx
-<Form
-  persist={{ id: 'nameAndPassword', exclude: ['password'], storage: 'local' }}
->
+<Form persist={{ id: 'nameAndSecret', exclude: ['secret'], storage: 'local' }}>
   <InputField
     name="name"
     label="Name"
     validation={{ required: 'Name is required' }}
   />
-  <PasswordField
-    name="password"
-    validation={{ required: 'Password is required' }}
+  <InputField
+    name="secret"
+    label="Secret"
+    validation={{ required: 'Secret is required' }}
   />
   <Button type="submit">Submit</Button>
 </Form>
--- a/src/components/form/Form.mdx
+++ b/src/components/form/Form.mdx
@@ -154,7 +154,6 @@ Here's an example using the `persist` prop:
 
 ```tsx
 <Form
-  onSubmit={console.log('')}
   persist={{ id: 'nameAndPassword', exclude: ['password'], storage: 'local' }}
 >
   <InputField
@@ -166,9 +165,7 @@ Here's an example using the `persist` prop:
     name="password"
     validation={{ required: 'Password is required' }}
   />
-  <Button type="submit" onClick={console.log('')}>
-    Submit
-  </Button>
+  <Button type="submit">Submit</Button>
 </Form>
 ```
 
--- a/src/components/form/Form.mdx
+++ b/src/components/form/Form.mdx
@@ -154,7 +154,7 @@ Here's an example using the `persist` prop:
 
 ```tsx
 <Form
-  onSubmit={jest.fn()}
+  onSubmit={console.log('')}
   persist={{ id: 'nameAndPassword', exclude: ['password'], storage: 'local' }}
 >
   <InputField
@@ -166,7 +166,7 @@ Here's an example using the `persist` prop:
     name="password"
     validation={{ required: 'Password is required' }}
   />
-  <Button type="submit" onClick={jest.fn()}>
+  <Button type="submit" onClick={console.log('')}>
     Submit
   </Button>
 </Form>
--- a/src/components/form/Form.mdx
+++ b/src/components/form/Form.mdx
@@ -17,6 +17,13 @@ priority: 1
 
 Custom field components should also accept a `validation` prop of type `ValidationOptions` and apply it to the field when present.
 
+`Form` is integrated with `react-hook-form-persist`'s package to save form data into session storage:
+
+- send `shouldPersist` boolean prop to persist form data to sessionStorage per `react-hook-form-persist`’s docs
+- send `persistOptions` prop with a required `id` and optional `include` or `exclude` arrays with reference to the input's names
+- `include` will only save the listed inputs to session storage and `exclude` will exclude the listed inputs
+- persisted form data in session storage will populate form with `defaultValues` on refresh
+
 Here’s an example using a simplified version of our `InputField`:
 
 ```tsx
--- a/src/components/form/Form.mdx
+++ b/src/components/form/Form.mdx
@@ -17,13 +17,6 @@ priority: 1
 
 Custom field components should also accept a `validation` prop of type `ValidationOptions` and apply it to the field when present.
 
-`Form` is integrated with `react-hook-form-persist`'s package to save form data into session storage:
-
-- send `shouldPersist` boolean prop to persist form data to sessionStorage per `react-hook-form-persist`’s docs
-- send `persistOptions` prop with a required `id` and optional `include` or `exclude` arrays with reference to the input's names
-- `include` will only save the listed inputs to session storage and `exclude` will exclude the listed inputs
-- persisted form data in session storage will populate form with `defaultValues` on refresh
-
 Here’s an example using a simplified version of our `InputField`:
 
 ```tsx
@@ -145,6 +138,39 @@ These properties only apply to text inputs. If `true`, the value will be convert
 ### Custom processing
 
 The `process` property accepts a function of type `(value: any) => any`. It can be used to process the value after validation.
+
+## Persisting form data
+
+`Form` is integrated with `react-hook-form-persist`'s package to save form data into session storage:
+
+- send `persist` prop to Form component to persist form data in sessionStorage
+- `persist` object has a required `id`, an optional `storage` prop and optional `include` or `exclude` arrays with reference to the input's names
+- `include` will only save the listed inputs to session storage and `exclude` will exclude the listed inputs
+- if both `include` and `exclude` are sent it will ignore `include` and use only `exclude`
+- if no `storage` prop is sent, it will default to use sessionStorage, however you can override this by sending `local` to use localStorage
+- persisted form data in sessionStorage or localStorage will populate form with `defaultValues` on refresh
+
+Here's an example using the `persist` prop:
+
+```tsx
+<Form
+  onSubmit={jest.fn()}
+  persist={{ id: 'nameAndPassword', exclude: ['password'], storage: 'local' }}
+>
+  <InputField
+    name="name"
+    label="Name"
+    validation={{ required: 'Name is required' }}
+  />
+  <PasswordField
+    name="password"
+    validation={{ required: 'Password is required' }}
+  />
+  <Button type="submit" onClick={jest.fn()}>
+    Submit
+  </Button>
+</Form>
+```
 
 ## Composing form field components
 
--- a/src/components/form/Form.test.tsx
+++ b/src/components/form/Form.test.tsx
@@ -109,6 +109,29 @@ describe(`Form component`, () => {
     )
   })
 
+  it('checks user input values are being persisted in sessionStorage', async () => {
+    render(
+      <Form onSubmit={jest.fn()} persist={{ id: 'nameForm' }}>
+        <InputField
+          name="name"
+          label="Name"
+          type="text"
+          validation={{ required: 'Name is required' }}
+        />
+        <Button type="submit" onClick={jest.fn()}>
+          Submit
+        </Button>
+      </Form>
+    )
+    const input = screen.getByRole('textbox', { name: 'Name' })
+
+    userEvent.type(input, 'Kyle Lowry')
+    expect(input).toHaveValue('Kyle Lowry')
+    expect(JSON.parse(sessionStorage.getItem('nameForm')).name).toEqual(
+      'Kyle Lowry'
+    )
+  })
+
   it('saves form data to sessionStorage excluding secret field', async () => {
     render(
       <Form
@@ -135,41 +158,14 @@ describe(`Form component`, () => {
 
     const nameInput = screen.getByRole('textbox', { name: 'Name' })
     const secretInput = screen.getByRole('textbox', { name: 'Secret' })
+    const parsedStorage = JSON.parse(sessionStorage.getItem('nameAndSecret'))
 
     userEvent.type(nameInput, 'Kawhi Leonard')
     userEvent.type(secretInput, 'Very secret secret')
 
-    expect(JSON.parse(sessionStorage.getItem('nameAndSecret'))).toEqual(
-      expect.anything()
-    )
-    expect(JSON.parse(sessionStorage.getItem('nameAndSecret')).name).toEqual(
-      'Kawhi Leonard'
-    )
-    expect(
-      JSON.parse(sessionStorage.getItem('nameAndSecret')).secret
-    ).toBeFalsy()
-  })
-
-  it('checks user input values are being persisted in sessionStorage', async () => {
-    render(
-      <Form onSubmit={jest.fn()} persist={{ id: 'nameForm' }}>
-        <InputField
-          name="name"
-          label="Name"
-          type="text"
-          validation={{ required: 'Name is required' }}
-        />
-        <Button type="submit" onClick={jest.fn()}>
-          Submit
-        </Button>
-      </Form>
-    )
-
-    userEvent.type(screen.getByRole('textbox', { name: 'Name' }), 'Kyle Lowry')
-    expect(screen.getByRole('textbox')).toHaveValue('Kyle Lowry')
-    expect(JSON.parse(sessionStorage.getItem('nameForm')).name).toEqual(
-      'Kyle Lowry'
-    )
+    expect(parsedStorage).toEqual(expect.anything())
+    expect(parsedStorage.name).toEqual('Kawhi Leonard')
+    expect(parsedStorage.secret).toBeFalsy()
   })
 
   it('checks data attributes are availabe in the DOM', async () => {
--- a/src/components/form/Form.test.tsx
+++ b/src/components/form/Form.test.tsx
@@ -104,7 +104,9 @@ describe(`Form component`, () => {
       </Form>
     )
 
-    expect(sessionStorage.getItem('nameAndYearGroup')).toBeTruthy()
+    expect(sessionStorage.getItem('nameAndYearGroup')).toEqual(
+      expect.anything()
+    )
   })
 
   it('saves form data to sessionStorage excluding secret field', async () => {
@@ -136,31 +138,48 @@ describe(`Form component`, () => {
     ).toBeFalsy()
   })
 
-  it('checks default values are being persisted in sessionStorage', async () => {
+  it('checks user input values are being persisted in sessionStorage', async () => {
     render(
-      <Form onSubmit={jest.fn()} persist={{ id: 'nameAndTeam' }}>
+      <Form onSubmit={jest.fn()} persist={{ id: 'nameForm' }}>
         <InputField
           name="name"
           label="Name"
-          defaultValue="Kyle Lowry"
           validation={{ required: 'Name is required' }}
-        />
-        <InputField
-          name="team"
-          label="Team"
-          defaultValue="Toronto Raptors"
-          validation={{ required: 'Team is required' }}
         />
         <Button type="submit" onClick={jest.fn()}>
           Submit
         </Button>
       </Form>
     )
-    expect(JSON.parse(sessionStorage.getItem('nameAndTeam')).name).toEqual(
+
+    userEvent.type(screen.getByRole('input'), 'Kyle Lowry')
+    expect(screen.getByRole('form')).toHaveValue('Kyle Lowry')
+    expect(JSON.parse(sessionStorage.getItem('nameForm')).name).toEqual(
       'Kyle Lowry'
     )
-    expect(JSON.parse(sessionStorage.getItem('nameAndTeam')).team).toEqual(
-      'Toronto Raptors'
+  })
+
+  it('checks data attributes are availabe in the DOM', async () => {
+    render(
+      <div>
+        <Form
+          onSubmit={jest.fn()}
+          persist={{ id: 'nameAndTeam' }}
+          data-index-number="12345"
+        >
+          <InputField
+            name="name"
+            label="Name"
+            validation={{ required: 'Name is required' }}
+          />
+          <Button type="submit" onClick={jest.fn()}>
+            Submit
+          </Button>
+        </Form>
+      </div>
+    )
+    expect(screen.getByRole('form').getAttribute('data-index-number')).toEqual(
+      '12345'
     )
   })
 })
--- a/src/components/form/Form.test.tsx
+++ b/src/components/form/Form.test.tsx
@@ -133,8 +133,8 @@ describe(`Form component`, () => {
       </Form>
     )
 
-    const nameInput = screen.getByRole('textbox', { name: 'name' })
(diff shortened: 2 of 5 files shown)


src/components/form/Form.mdx
```diff
@@ -17,6 +17,13 @@ priority: 1
 
 Custom field components should also accept a `validation` prop of type `ValidationOptions` and apply it to the field when present.
 
+`Form` is integrated with `react-hook-form-persist`'s package to save form data into session storage:
```
andoulla, line 20: I would suggest moving this section below the example in its own section. Currently, it may confuse readers that the example below is the way to use these props.

src/components/form/Form.test.tsx
```diff
@@ -84,4 +84,36 @@ describe(`Form component`, () => {
 
     expect(await screen.findByText('Submit')).toHaveAttribute('disabled', '')
   })
+
+  it('saves form data to sessionStorage when shouldPersist prop is true', async () => {
+    render(
+      <Form
+        onSubmit={jest.fn()}
+        shouldPersist
+        persistOptions={{ id: 'nameAndPassword', exclude: ['password'] }}
```
thomasdigby, line 93: I'd be keen to update some of these props to help with the DX, what are your thoughts on changing the props to:

```jsx
<Form persist="user-created-id">
```

to allow simple usage of persisted data with a single string ID, as well as

```jsx
<Form persist={{ id: "user-created-id", exclude: ['password'] }}>
```

to allow the user to modify the persist settings?

src/components/form/Form.test.tsx
```diff
@@ -84,4 +84,36 @@ describe(`Form component`, () => {
...
+          Submit
+        </Button>
+      </Form>
+    )
+
+    expect(
+      sessionStorage.getItem('form-nameAndPassword') !== null ||
+        sessionStorage.getItem('form-nameAndPassword') !== undefined
+    )
+    expect(
```
andoulla, line 114: I find this test hard to follow. Up for discussion but wouldn't it better to go by what the user does. We could trigger a user event to update the name field and then assess that the local storage has the inputted value. I think this way we can truly verify that if I type a name the value is correctly saved.

I would also recommend not using the `PasswordField` as an example as I'm not sure we would ever want to persist the password.

src/components/form/Form.tsx
```diff
@@ -2,30 +2,103 @@ import invariant from 'invariant'
...
+    storage: window.sessionStorage
+  }
+  if (persistOptions?.exclude) {
+    params = { ...params, exclude: persistOptions.exclude }
+  }
+  if (!persistOptions?.exclude && persistOptions?.include) {
+    params = { ...params, include: persistOptions.include }
+  }
+
+  useFormPersist(`form-${persistOptions.id}`, { watch, setValue }, params)
```
thomasdigby, line 69: Can we pass in `persistOptions` (or `persist` if we change the api) here directly? I can't see the benefit of setting params separately, ideally we should also allow the user to override the type of storage and default to sessionstorage.

Also, I don't think there's a need to prefix the `id` with `form-`, the user may want to access the persisted data separately to its use with `Form`, so modifying the `id` will make that little trickier.

src/components/form/Form.tsx
```diff
@@ -37,21 +110,31 @@ export const Form: React.FC<FormProps> = ({
...
 
-  return (
-    <FormProvider {...formMethods}>
-      <StyledForm
-        aria-label="form"
-        {...remainingProps}
-        onSubmit={handleSubmit(onSubmit)}
+  const props = {
+    formMethods,
+    remainingProps,
```
thomasdigby, line 119: `remainingProps` should be spread onto this object, and we should have a test to confirm that a standard `data-attribute`s can be passed onto the DOM. Can you add that test for the `Form` in this PR?

src/components/form/Form.tsx
```diff
@@ -2,30 +2,103 @@ import invariant from 'invariant'
...
+  remainingProps,
+  handleSubmit,
+  onSubmit,
+  formContent
+}) => {
+  return (
+    <FormProvider {...formMethods}>
+      <StyledForm
+        aria-label="form"
+        {...remainingProps}
```
thomasdigby, line 85: Ah I can see how you're handling the other props now, however I think we still ought to ensure that `remainingProps` is always using rest spread to keep it consistent with its usage elsewhere and generally how it's used.

src/components/form/Form.tsx
```diff
@@ -2,30 +2,103 @@ import invariant from 'invariant'
...
 
 import { styled } from '~/stitches'
 import { Override } from '~/utilities'
 
 const StyledForm = styled('form', {})
 
+type ExcludeAndInclude = { exclude?: string[]; include?: string[] }
+type PersistOptions = { id: string } & ExcludeAndInclude
+type FormPersistParams = {
+  storage: Storage
```
andoulla, line 15: Where is the type `storage` coming from? It is not imported or defined anywhere

src/components/form/Form.tsx
thomasdigby: Having read through the persist lib docs I assume you're using `params` to ensure we don't pass the `id` into the options and that we only accept either `exclude` _or_ `include`. I think we can rely on Typescript to notify the users about what options are available and how they should be used, we should lean on the libraries handling of the options. Also you could use `...rest` for the separation of `id` here. e.g.

```js
const { id, ...options } = persistOptions

useFormPersist(id, { watch, setValue }, { storage: window.sessionStorage, ...options })
```

src/components/form/Form.mdx
```diff
@@ -17,6 +17,13 @@ priority: 1
...
+
+- send `shouldPersist` boolean prop to persist form data to sessionStorage per `react-hook-form-persist`’s docs
```
andoulla, line 22: could we just check if persistOptions are available and then apply the hook? it would be much simpler to not have to send 2 props in

src/components/form/Form.tsx
```diff
@@ -2,30 +2,103 @@ import invariant from 'invariant'
...
+  formContent
+}) => {
+  return (
+    <FormProvider {...formMethods}>
+      <StyledForm
+        aria-label="form"
+        {...remainingProps}
+        onSubmit={handleSubmit(onSubmit)}
+      >
+        {formContent}
```
thomasdigby, line 88: Why are we not using `children` here?

src/components/form/Form.mdx
```diff
@@ -17,6 +17,13 @@ priority: 1
...
+- send `persistOptions` prop with a required `id` and optional `include` or `exclude` arrays with reference to the input's names
```
andoulla, line 23: Would `LocalStorageOptions` be a better name for it?

src/components/form/Form.test.tsx
```diff
@@ -84,4 +84,36 @@ describe(`Form component`, () => {
...
+      JSON.parse(sessionStorage.getItem('form-nameAndPassword')).password ===
+        undefined
+    )
```
thomasdigby, line 117: We need to change these assertions so that they test something, e.g. `expect(sessionStorage.getItem('form-nameAndPassword')).toBeFalsy()`

It would be good to test having default values and user input to see that the values are persisted.

src/components/form/Form.mdx
```diff
@@ -17,6 +17,13 @@ priority: 1
...
+- `include` will only save the listed inputs to session storage and `exclude` will exclude the listed inputs
+- persisted form data in session storage will populate form with `defaultValues` on refresh
+
```
thomasdigby, line 26: This documentation should probably sit just above `Composing form field components` and be its own section titled `Persisting form data` or something similar, we should also add a code example to show the API in use.

src/components/form/Form.tsx
```diff
@@ -2,30 +2,103 @@ import invariant from 'invariant'
...
     validationMode: 'onBlur' | 'onSubmit'
+    shouldPersist?: boolean
+    persistOptions?: PersistOptions
   } & (
     | { children: React.ReactNode; render?: never }
     | { children?: never; render: (methods: UseFormMethods) => React.ReactNode }
   )
 >
 
+type FormContentProps = Override<
```
andoulla, line 32: I wonder if all of these should have been separated out in a different file and then imported and used when needed. This file is getting quite large but also the form is already a complicated component to comprehend. 

I'm very keen to hear other people's opinion on this.

src/components/form/Form.tsx
```diff
@@ -37,21 +110,31 @@ export const Form: React.FC<FormProps> = ({
...
-        {formContent}
-      </StyledForm>
-    </FormProvider>
-  )
+        <FormContent {...props} />
+      </PersistFormWrapper>
+    )
+  }
+
+  return <FormContent {...props} />
```
thomasdigby, line 137: Related to some of my other comments, if we can use `formContent` as children directly here I think it'll help make it clearer how its used e.g.

```jsx
return <FormContent {...props}>{formContent}</FormContent>
```

And maybe just while we're here, `formContent` could be renamed to `resolvedChildren` or similar to retain a link with "children"

thomasdigby (commented):
I think generally this is the correct approach to implement, so good job on that. Mostly minor comments about composition, naming conventions and tests so hopefully not too much work to update.

It would be good to get @andoulla's thoughts as part of this review process.

src/components/form/Form.tsx
```diff
@@ -2,30 +2,103 @@ import invariant from 'invariant'
...
 import type { UseFormMethods } from 'react-hook-form'
 import { FormProvider, useForm } from 'react-hook-form'
+import useFormPersist from 'react-hook-form-persist'
 
 import { styled } from '~/stitches'
 import { Override } from '~/utilities'
 
 const StyledForm = styled('form', {})
 
+type ExcludeAndInclude = { exclude?: string[]; include?: string[] }
```
andoulla, line 12: In combination with a comment above, maybe this could be LocalStorageConfig or PersistConfig if we don't like the other suggestion

src/components/form/Form.test.tsx
LeahB8: Yeah I'm on board with sending fewer props, but I think it might be good to keep the persist prop type consistent, especially as we're using ts. So it could be just an object with an id or an object with an id and exclude/include array.

thomasdigby: Yeah I can get on board with that, leaves us room to introduce that option down the line as well if we think it's worth it (without a breaking change)

LeahB8: For sure

src/components/form/Form.tsx
LeahB8: It's a global interface made available to reference web storage
Screenshot 2021-06-10 at 11 11 13

src/components/form/Form.tsx
LeahB8: Yeah for sure. Do you have a convention yet in the DS for storing types? I have just made a file at the same level called `Forms.types.ts`. Does that sound alright?

src/components/form/Form.tsx
LeahB8: Have switched it out

src/components/form/Form.tsx
LeahB8: I'm not sure what you mean about adding a test for data-attributes being passed onto the DOM. Could you elaborate?

src/components/form/Form.test.tsx
```diff
@@ -84,4 +84,36 @@ describe(`Form component`, () => {
...
+          validation={{ required: 'Password is required' }}
+        />
+        <Button type="submit" onClick={jest.fn()}>
+          Submit
+        </Button>
+      </Form>
+    )
+
+    expect(
+      sessionStorage.getItem('form-nameAndPassword') !== null ||
```
Mhoog, line 111: You should use `expect()` with a _matcher_ function:

```js
expect(sessionStorage.getItem('form-nameAndPassword')).toEqual(expect.anything()) // checks if the value is anything else than undefined or null
```

src/components/form/Form.test.tsx
LeahB8: This sounds like a good idea. Do you know the best way to trigger a user event in a test like this? I'm not too familiar.

andoulla: you can find an example in this component https://github.com/Atom-Learning/components/blob/main/src/components/input-field/InputField.test.tsx

```
  const { container, findByText, getByRole } = render(
      <Form>
        <InputField
          label="INPUT FIELD"
          name="INPUT FIELD"
          validation={{ required: errorText }}
        />
        <button type="submit">Submit</button>
      </Form>
    )

    userEvent.click(getByRole('button'))
    await findByText(errorText)

```

src/components/form/Form.tsx
thomasdigby: So the user should be able to write `<Form data-hello="world">` and see that atttribute appear in the DOM, e.g. `<form data-hello="world">`. My concern was that we had inadvertently removed that ability by not spreaing `remainingProps`.

`expect(getByRole('form').getAttribute('data-hello')).toEqual('world')` (as an example)

src/components/form/Form.test.tsx
andoulla: and a more relevant example from the documentation:
```
test('type', () => {
  render(<textarea />)

  userEvent.type(screen.getByRole('textbox'), 'Hello,{enter}World!')
  expect(screen.getByRole('textbox')).toHaveValue('Hello,\nWorld!')
})
```

src/components/form/Form.tsx
thomasdigby: This will unfortunately affect the props-table in the documentation site as our types lose their detail and are displayed as just their name. Let's go with your suggestion @LeahB8 and separate the types into `Form.types.ts` for now, I'll take a look on the docs site when I get a chance next week and we can make a call how we want to handle this as a convention across the design system then.

thomasdigby: To clarify, this won't prevent us from merging and publishing this update.

src/components/form/Form.mdx
```diff
@@ -139,6 +139,39 @@ These properties only apply to text inputs. If `true`, the value will be convert
...
+- `include` will only save the listed inputs to session storage and `exclude` will exclude the listed inputs
+- if both `include` and `exclude` are sent it will ignore `include` and use only `exclude`
+- if no `storage` prop is sent, it will default to use sessionStorage, however you can override this by sending `local` to use localStorage
+- persisted form data in sessionStorage or localStorage will populate form with `defaultValues` on refresh
+
+Here's an example using the `persist` prop:
+
+```tsx
+<Form
+  onSubmit={jest.fn()}
```
thomasdigby, line 157: No need to include `jest.fn()` in our docs

src/components/form/Form.tsx
```diff
@@ -1,57 +1,116 @@
...
     !(children && render),
     '`Form` should only be given one of `children` or `render`. When both are provided, `render` will be used and `children` will be ignored.'
   )
 
+  // if (persist) {
+  //   invariant(
+  //     !(persist.include && persist.exclude),
+  //     'If `Form` is sent `persist`, it should only be given one of `include` or `exclude`. When both are provided, `exclude` will be used and `include` will be ignored.'
+  //   )
+  // }
```
thomasdigby, line 88: ☠️

src/components/form/Form.tsx
LeahB8: Ah okay cool. I have now spread the `remainingProps`, but will add a test for this too

src/components/form/Form.types.ts
```diff
@@ -0,0 +1,79 @@
...
+
+type PersistOptionsInclude = {
+  id: string
+  storage?: StorageEnum
+} & IncludeConfig
+
+type PersistOptionsExclude = {
+  id: string
+  storage?: StorageEnum
+} & ExcludeConfig
```
thomasdigby, line 24: Are these two types used? `PersistOptionsInclude` & `PersistOptionsExclude`

src/components/form/Form.types.ts
```diff
@@ -0,0 +1,79 @@
+import type { UseFormMethods } from 'react-hook-form'
+
+import { styled } from '~/stitches'
+import { Override } from '~/utilities'
+
+const StyledForm = styled('form', {})
```
thomasdigby, line 6: ☠️

src/components/form/Form.types.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+>
+
+export {
+  FormPersistParams,
+  PersistFormWrapperProps,
+  FormContentProps,
+  FormProps,
+  StorageEnum
+}
```
thomasdigby, line 79: Alongside our `import type` statement, we can `export type` as well

thomasdigby:
I think this is looking really good @LeahB8, we need two more tests, one for testing the data-attribute and one testing user input, and a few minor bits to tidy up then I think we're good to go 👍

src/components/form/Form.tsx
```diff
@@ -1,57 +1,116 @@
...
-type FormProps = Override<
-  React.ComponentPropsWithoutRef<typeof StyledForm>,
-  {
-    defaultValues?: { [key: string]: string | number }
-    onSubmit: (data: any) => void
-    validationMode: 'onBlur' | 'onSubmit'
-  } & (
-    | { children: React.ReactNode; render?: never }
-    | { children?: never; render: (methods: UseFormMethods) => React.ReactNode }
+import {
```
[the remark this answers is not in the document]
LeahB8: Yeah for sure. It does show eslint errors in the code when import as a type now tho. I tried cleaning the dist folder and rebuilding and still the same 
Screenshot 2021-06-11 at 10 57 53

src/components/form/Form.types.ts
LeahB8: I committed this mid fight with typescript as a WIP.  I couldn't figure out the best way to allow for the persist object type to either optionally have include or exclude, but not both, but also okay to have neither.

src/components/form/Form.types.ts
LeahB8: This is used in some of the below types tho. Should it not be?

src/components/form/Form.types.ts
LeahB8: I also tried this and getting eslint errors again. Couldn't see how to get around this
Screenshot 2021-06-11 at 11 01 44

src/components/form/Form.tsx
LeahB8: Okay sounds good

src/components/form/Form.test.tsx
LeahB8: Nice, thank you

src/components/form/Form.tsx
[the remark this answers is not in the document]
thomasdigby: Huh, I need to take a closer look at these types, are they used at run-time at all?

src/components/form/Form.tsx
LeahB8: Have changed the types, so this is no longer relevant

src/components/form/Form.tsx
LeahB8: I just needed to add my workaround here for the package bug. When it's fixed, I can refactor this part.

src/components/form/Form.types.ts
thomasdigby: Ah yeah you're right, this is likely why we didn't extract the types across other more complex components as they're reliant on this `StyledForm` value. We may need to keep the "component" types co-located in this file, and then combine them with the imported types in `Form.types.ts` e.g.

```ts
// Form.types.ts

type PersistFormWrapper = {
  persist: PersistOptions
  watch
  setValue
}
```

```ts
// Form.ts

const StyledForm = styled('form', {})

type PersistFormWrapperProps = React.ComponentPropsWithoutRef<typeof StyledForm> & PersistFormWrapper

const PersistFormWrapper: React.FC<PersistFormWrapperProps> = ({
```

src/components/form/Form.types.ts
LeahB8: Actually now this isn't relevant until the package bug is fixed, so all good

src/components/form/Form.tsx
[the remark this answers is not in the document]
thomasdigby: `StorageEnum` is used at runtime so we'll need to import and export that separately. e.g.

```ts
import type {
  FormContentProps,
  FormPersistParams,
  FormProps,
  PersistFormWrapperProps
} from './Form.types'

import { StorageEnum } from './Form.types'
```

src/components/form/Form.mdx
```diff
@@ -139,6 +139,39 @@ These properties only apply to text inputs. If `true`, the value will be convert
...
+  <InputField
+    name="name"
+    label="Name"
+    validation={{ required: 'Name is required' }}
+  />
+  <PasswordField
+    name="password"
+    validation={{ required: 'Password is required' }}
+  />
+  <Button type="submit" onClick={console.log('')}>
```
thomasdigby, line 169: Being pedantic here as we want to ensure the docs are 100% correct for our design system, we shouldn't need an `onClick` here at all as the form submission is handled by `Form`.

Also I don't think it's necessary to include the `onSubmit` on the `Form` component here, we don't do it in other parts of the docs and `console.log('')` doesn't add anything to what you've already written 👍

LeahB8: Okay no probs, have removed

src/components/form/Form.tsx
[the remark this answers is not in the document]
LeahB8: Nice, have changed it

src/components/form/Form.mdx
```diff
@@ -139,6 +139,39 @@ These properties only apply to text inputs. If `true`, the value will be convert
...
+<Form
+  onSubmit={console.log('')}
+  persist={{ id: 'nameAndPassword', exclude: ['password'], storage: 'local' }}
+>
+  <InputField
+    name="name"
+    label="Name"
+    validation={{ required: 'Name is required' }}
+  />
+  <PasswordField
```
andoulla, line 165: Could we not use PasswordField as an example here too? I just really want to ensure that we don't give the idea to anyone to store passwords in localstorage

src/components/form/Form.tsx
```diff
@@ -1,31 +1,78 @@
...
+  watch,
+  setValue,
+  children
+}) => {
+  const { id, ...options } = persist
+
+  let params: FormPersistParams = {
+    ...options,
+    storage:
+      options.storage === StorageEnum.local
```
andoulla, line 29: Isn't the convention for Enums to have them uppercased? very minor obviously

src/components/form/Form.tsx
```diff
@@ -1,31 +1,78 @@
...
+        ? window.localStorage
+        : window.sessionStorage
+  }
+
+  if (options.exclude) {
+    // Workaround for package bug
```
andoulla, line 35: Since we are adding a comment here could we make it specific as I don't really know what the package bug is?

src/components/form/Form.tsx
andoulla: Not uppercased, in all capitals, sorry

src/components/form/Form.mdx
LeahB8: For sure, have swapped it out

src/components/form/Form.tsx
LeahB8: True, have changed it

src/components/form/Form.tsx
LeahB8: For sure, have added

src/components/form/Form.types.ts
LeahB8: Have updated

andoulla (approved):

thomasdigby (approved):
Awesome work, thanks @LeahB8 this is good to go 👍

PR #174, opened 2021-06-21
Feature: Tabs

alexandrunemes:
Link to the [ticket](https://app.asana.com/0/0/1200478307675076/f)

### In this PR

- base `Tabs` component, which acts as a wrapper for the smaller building blocks
- `Tab.TriggerList` component, which gets `Tab.Trigger` components as children and renders them
- `Tab.Trigger` component, which holds the content that will be rendered inside the trigger button for a particular tab section
- `Tab.Content` component, which holds the actual section's content
- unit tests 
- documentation
- updates all radix components to their latest versions, in order to avoid breaking the `IdProvider` component, which is needed inside unit tests

*Note: the snapshot test on the Tabs component has been commented out because of a regression introduced by newer versions of radix packages. A GH issue has been submitted on their repo. Once it gets fixed, we will uncomment our test. 

Normal state, with the second tab selected:
Screenshot 2021-06-21 at 14 24 06

First tab selected, second tab disabled:
Screenshot 2021-06-22 at 11 49 39


src/components/tabs/Tabs.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { styled } from '~/stitches'
+
+const StyledRoot = styled(Root, {
+  display: 'flex',
+  flexDirection: 'column'
+})
+
+type TabsProps = React.ComponentProps<typeof StyledRoot>
+
+export const Tabs: React.FC<TabsProps> = ({
```
hollg, line 12: Can we follow the same pattern as `List` and `Carousel` here, and set the sub-components as properties on the container component object?

src/components/tabs/TabContent.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react'
+import { Content } from '@radix-ui/react-tabs'
+import { styled } from '~/stitches'
+
+const StyledTabContent = styled(Content, {
+  flexGrow: 1,
+  padding: 20
```
hollg, line 7: Where they exist, we want to stick to CSS values from our theme (e.g. `$1`, `$2` etc) unless there's a specific reason not to in a given case.

src/components/tabs/TabContent.tsx
```diff
@@ -0,0 +1,24 @@
...
+type TabContentProps = React.ComponentProps<typeof StyledTabContent> & {
+  value: string
+}
+
+export const TabContent: React.FC<TabContentProps> = ({
+  children,
+  css,
+  value
+}: TabContentProps) => (
+  <StyledTabContent value={value} css={css}>
```
hollg, line 19: When you're passing all the props to one component, you can simplify it like this:
```
export const TabContent: React.FC<TabContentProps> = (props) => <StyledContent {...props}
/>
```

This also ensures that _all_ valid props can be used (in the current version, there are some keys from `React.ComponentProps<typeof StyledTabContent>` that aren't accounted for in `{children, css, value}`).


In this particular case though, I think we can perhaps just use `StyledTabContent` directly without this extra layer of abstraction, because as far as I can tell we're not actually adding anything new in this component (`value` is defined as a prop on the Radix component, so `StyledTabContent` already has it).

src/components/tabs/TabTrigger.tsx
```diff
@@ -0,0 +1,57 @@
+import * as React from 'react'
+import { Trigger } from '@radix-ui/react-tabs'
+import { CSS, styled } from '~/stitches'
+
+const StyledTabTrigger = styled(Trigger, {
+  flexShrink: 0,
+  padding: '10px 20px',
```
hollg, line 7: We should use theme values for this padding, just like with all the `color`s.

src/components/tabs/Tabs.mdx
```diff
@@ -0,0 +1,79 @@
...
+    <TabTrigger>Tab 3</TabTrigger>
+  </TabTriggerList>
+</Tabs>
+```
+
+## TabTrigger
+
+The `TabTrigger` component holds the content that will be displayed inside the button that the user would click in order to switch tabs. In can hold either a string, or some other component. It needs to be passed a `value` prop, that would be identical to the `value` prop passed to its corresponding `TabContent` component.
+
+```tsx preview
```
hollg, line 44: `preview` renders the output of the code example in the docs page, but I don't think these examples will work because they have subcomponents (.e.g. `TabTrigger`) without the required wrapper components

src/components/tabs/Tabs.mdx
```diff
@@ -0,0 +1,79 @@
...
+<TabTrigger disabled value="tab1">
+  Tab 1
+</TabTrigger>
+```
+
+## Styling the `TabTrigger`
+
+In order to style the different states of a trigger, the following CSS selectors need to be used when passed to the `css` prop:
```
hollg, line 72: This is _super useful_ documentation 🔥

src/components/tabs/Tabs.tsx
alexandrunemes: Sure thing!

src/components/tabs/TabContent.tsx
alexandrunemes: Got it!

src/components/tabs/TabContent.tsx
alexandrunemes: Then the same logic would apply to the TabTriggerList component. The only one that actually does some extra logic is the Trigger component.

alexandrunemes: And I guess I might as well drop the now unnecessary modules for `TabTriggerList` and `TabContent` and simply move both of the definitions inside the `Tabs` module, as in `List.tsx`.

src/components/tabs/Tabs.mdx
```diff
@@ -0,0 +1,106 @@
+---
+title: Tabs
+component: Tabs
+description: Tabs is a component that provides different sections of content that are displayed one at a time.
+category: Layout
+---
+
+The component provides a set of default styles, which can be overwritten by using the `css` prop. It is composed by combining smallar components, such as `Tabs.TriggerList`, `Tabs.Trigger`, and `Tabs.Content`.
```
andoulla, line 8:
```suggestion
The component provides a set of default styles, which can be overwritten by using the `css` prop. It is composed by combining smaller components, such as `Tabs.TriggerList`, `Tabs.Trigger`, and `Tabs.Content`.
```

src/components/tabs/Tabs.mdx
```diff
@@ -0,0 +1,106 @@
...
+  </Tabs.Content>
+</Tabs>
+```
+
+## Tabs.TriggerList
+
+The `Tabs.TriggerList` component simply holds the individual `Tabs.Trigger` components. It can also get custom styling via the `css` prop.
+
+```tsx preview
+<Tabs>
```
andoulla, line 32: This example is similar to the one above. Could we either add the css prop to show how it works or remove it?

src/components/tabs/Tabs.mdx
```diff
@@ -0,0 +1,106 @@
...
+  <Tabs.TriggerList>
+    <Tabs.Trigger value='tab1'>Tab 1</Tabs.Trigger>
+    <Tabs.Trigger value='tab2'>
+      <Text>Tab 2</Text>
+    </Tabs.Trigger>
+  <Tabs.TriggerList>
+  <Tabs.Content value="tab1">Tab 1 content</Tabs.Content>
```
andoulla, line 56: Same as above, I don't think we need an extra example if it is exactly the same as the previous examples

src/components/tabs/Tabs.mdx
```diff
@@ -0,0 +1,106 @@
...
+  <Tabs.Content value="tab2">Tab 2 content</Tabs.Content>
+</Tabs>
+```
+
+## Tabs.Content
+
+The `Tabs.Content` component, as the name suggests, holds the content for that particular section of the tabs component. It takes a `value` prop that needs to match the `value` prop passed to its corresponding trigger. Its content can be any text or valid component.
+
+```tsx preview
+<Tabs>
```
andoulla, line 66: Could we pass the default value/value pairing to show how it works?

src/components/tabs/Tabs.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+  it('has no programmatically detectable a11y issues', async () => {
+    const { container } = render(<TabsTest />)
+
+    expect(await axe(container)).toHaveNoViolations()
+  })
+
+  it('it makes the correct tab content visible, based on the `defaultValue` prop', async () => {
+    render(<TabsTest defaultValue="tab2" />)
+    const tab1content = await screen.findByText('Important content for tab 1')
+    const tab2content = await screen.findByText('Important content for tab 2')
```
andoulla, line 38:
```suggestion
    const tab2content = await screen.findByText('Important content for tab 2')
    
```

Just adding a bit of space for readability. Very good test by the way

src/components/tabs/Tabs.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+    expect(tab1content).not.toBeVisible()
+    expect(tab2content).toBeVisible()
+  })
+
+  it('allows the user to switch tabs', async () => {
+    render(<TabsTest />)
+
+    const tab1content = await screen.findByText('Important content for tab 1')
+    const tab2content = await screen.findByText('Important content for tab 2')
```
andoulla, line 47:
```suggestion
    const tab2content = await screen.findByText('Important content for tab 2')
    
```

src/components/tabs/Tabs.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+    expect(tab1content).toBeVisible()
+    expect(tab2content).not.toBeVisible()
+    userEvent.click(screen.getByText('Trigger 2'))
```
andoulla, line 50:
```suggestion
    
    userEvent.click(screen.getByText('Trigger 2'))
    
```

src/components/tabs/Tabs.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+        </Tabs>
+      </IdProvider>
+    )
+
+    const tab1content = await screen.findByText('Important content for tab 1')
+    const tab2content = await screen.findByText('Important content for tab 2')
+    expect(tab1content).toBeVisible()
+    userEvent.click(screen.getByText('Trigger 2'))
+    expect(tab1content).toBeVisible()
+    expect(tab2content).not.toBeVisible()
```
andoulla, line 76: Is this correct? When something is disabled it should be visible but not clickable

andoulla (commented):
Good job. Could we please add some screenshots on how the components look in the different use cases,( loading, one selected and disabled)

src/components/tabs/TabTrigger.tsx
```diff
@@ -0,0 +1,57 @@
...
+  '&[data-state="active"]': {
+    color: '$primary900',
+    boxShadow: 'inset 0 -3px 0 0 currentColor'
+  },
+  '&[data-state="active"] *': {
+    color: '$primary900'
+  }
+})
+
+const renderChildren = (children: string | React.ReactNode, css: CSS) =>
```
hollg, line 30: Could you explain why we need this `renderChildren` function?

src/components/tabs/Tabs.test.tsx
alexandrunemes: I can't tell what the change here is :))

andoulla: just adding a bit of space between the lines to make them a bit more legible

src/components/tabs/Tabs.test.tsx
alexandrunemes: Yes, this test verifies that when `tab2` is disabled, even if the trigger gets clicked, the content is not displayed.

src/components/tabs/TabTrigger.tsx
alexandrunemes: I just figured we might want to pass components instead of strings to the `Trigger`, and we might want to pass down whatever styling we set on the `Trigger` to the nested component. But now that you pointed it out, it would reduce the complexity if we just passed `css` to the nested component directly.

andoulla (approved):

src/components/tabs/Tabs.test.tsx
alexandrunemes: I adjusted the spacing on the entire file.

src/components/tabs/Tabs.mdx
alexandrunemes: Removed the example.

src/components/tabs/TabTrigger.tsx
```diff
@@ -0,0 +1,35 @@
...
+  color: '$primary500',
+  userSelect: 'none',
+  '&:hover': { color: '$primary900' },
+  '&[data-disabled]': { color: '$tonal600' },
+  '&[data-disabled]:hover': { color: '$tonal600' },
+  '&[data-state="active"]': {
+    color: '$primary900',
+    boxShadow: 'inset 0 -3px 0 0 currentColor'
+  }
+})
```
thomasdigby, line 17: Can we add a font family to this to prevent rendering as default serif?

src/components/tabs/Tabs.mdx
```diff
@@ -0,0 +1,100 @@
...
+category: Layout
+---
+
+The component provides a set of default styles, which can be overwritten by using the `css` prop. It is composed by combining smaller components, such as `Tabs.TriggerList`, `Tabs.Trigger`, and `Tabs.Content`.
+
+```tsx preview
+<Tabs defaultValue="tab1" css={{ height: '400px', border: '1px solid black' }}>
+  <Tabs.TriggerList>
+    <Tabs.Trigger value="tab1">
+      <Text>Nested component under the Tabs.Trigger component</Text>
```
thomasdigby, line 14: Would we expect the user to supply `Text` in the `Trigger`? Or should we cover that style in the trigger itself?

thomasdigby (approved):
Nice PR 👍 

I have a few questions about how/where these tabs are going to be used and how flexible we may need some of these default styles to be, but I think we can consider that in a future version when we have a better idea of how its used.

src/components/tabs/TabTrigger.tsx
thomasdigby: Might be a redundant comment based on your answer to my other question!

alexandrunemes: Should it be $sans?

src/components/tabs/Tabs.mdx
alexandrunemes: No, Trigger can take either strings, or components as children. If they pass components, they should also style the components accordingly, via the `css` prop.

thomasdigby: Cool ok, so if the user passes a string, we should expect the Trigger to render in the correct font-style; if they use a component then they can override that style if need be.

src/components/tabs/TabTrigger.tsx
thomasdigby: Yep 👍

PR #175, opened 2021-06-22
Feat/table

shaneakins:
### In this PR

Table component with multiple subcomponents:

- Table.Body component, which is comparable to the HTML 'tbody' tag
- Table.Cell, which is comparable to the HTML 'td' tag
- Table.Footer, which is comparable to the HTML 'tfoot' tag
- Table.FooterCell which is comparable to the HTML 'td', but located in the footer
- Table.Header, which is comparable to the HTML 'thead' tag 
- Table.HeaderCell, which is comparable to the HTML 'th' tag 
- Table Row, which is comparable to the HTML 'tr' tag 

Unit tests
Documentation

Normal table with base styles (width added to constrain size for screenshot:
Screen Shot 2021-06-28 at 16 33 32

Table with styling applied to various subcomponents:
Screen Shot 2021-06-28 at 16 33 43

### Context
This is a standard static table made of subcomponents in the spirit of Semantic UI. Using components instead of tags, while being more verbose, improves readability and allows for direct styling consistent with the rest of the design system.


src/components/table/Table.test.tsx
```diff
@@ -0,0 +1,54 @@
+import { render } from '@testing-library/react'
+import { axe } from 'jest-axe'
+// import { render } from 'jest/test-utils'
```
alexandrunemes, line 3: you forgot this commented import

shaneakins: Oops - good eye. Thanks!

src/components/table/Cell.tsx
```diff
@@ -0,0 +1,9 @@
+import { styled } from '~/stitches'
+
+export const Cell = styled('td', {
+  verticalAlign: 'top',
+  textAlign: 'left',
+  p: '$2 $3'
```
alexandrunemes, line 6: Is this something `stitches` recognises as `padding`?

alexandrunemes (approved):

src/components/table/Cell.tsx
alexandrunemes: nevermind :D i gave `stitches.ts` a read

src/components/table/Table.mdx
```diff
@@ -0,0 +1,42 @@
...
+component: Table, Table.Body, Table.Cell, Table.Footer, Table.Header, Table.HeaderCell, Table Row
+description: The Table component displays a collection of data grouped into rows
+category: Layout
+---
+
+The `Table` component structure mirrors that of a regular HTML table, with the smaller `Table.Body`, `Table.Cell`, `Table.Footer`, `Table.Header`, `Table.HeaderCell` and `Table Row` components corresponding to the `<tbody>`, `<td>`, `<tfoot>`, `<thead>`, `<th>` and `<tr>` tags, respectively. 
+
+The default table width is 100% of the parent container, but this and other styles can be overridden with the `css` prop on all components.
+
+Best practice: Due to rendering issues with some browsers, adding borders to `Table.Row` is possible by styling the Table component with `border-collapse: collapse`; however, this practice is generally not recommended. The preferred method for adding borders is applying styles to the `Table.Cell` and `Table.HeaderCell` components;
```
andoulla, line 12:
```suggestion
Best practice: Due to rendering issues with some browsers, adding borders to `Table.Row` is possible by styling the Table component with `border-collapse: collapse`; however, this practice is generally not recommended. The preferred method for adding borders is applying styles to the `Table.Cell` and `Table.HeaderCell` components.
```

src/components/table/Table.tsx
```diff
@@ -0,0 +1,42 @@
+import * as React from 'react'
+
+import { styled } from '~/stitches'
+
+// import { CSSWrapper } from '~/utilities'
```
andoulla, line 5:
```suggestion

```

thomasdigby (commented):
I think we ought to consider how this table will be used, we're keeping flexibility with styling and composing components within a table but the usage is a bit verbose. Can you imagine us using a further abstraction to make it easier to display a set schema of data with this table?

src/components/table/Header.tsx
```diff
@@ -0,0 +1,5 @@
+import { styled } from '~/stitches'
```
andoulla, line 1: I'm wondering if we should change the filename pattern here and use the same as in Carousel for consistency.


I think the file names should be TableFooter, TableHeader etc.

src/components/table/Header.tsx
```diff
@@ -0,0 +1,5 @@
...
+
+export const Header = styled('thead', {})
+
+Header.displayName = 'Header'
```
hollg, line 5: I think these subcomponents' display names should be `TableHeader`, `TableFooter` etc so that their names are clear out of context.

src/components/table/Table.test.tsx
```diff
@@ -0,0 +1,53 @@
+import { render } from '@testing-library/react'
+import { axe } from 'jest-axe'
+import * as React from 'react'
+
+import { Table } from './Table'
+
+describe(`Table component`, () => {
+  it('renders', async () => {
+    const { container } = await render(
+      <Table css={{ height: 100, width: 400, color: '$primary500' }}>
```
hollg, line 10: I'm being _real_ picky here, but can we use values like `"100px"` rather than `100`? `100` is totally valid, but I think it's worth being as inescapably clear as possible since we do have some "magic" numbers in the design system.

src/components/table/Table.tsx
```diff
@@ -0,0 +1,42 @@
...
+  fontFamily: '$sans',
+  fontSize: '$sm'
+})
+
+type TableProps = React.ComponentProps<typeof StyledTable>
+
+export const Table: React.FC<TableProps> & TableSubComponents = ({
+  children,
+  ...restProps
+}: TableProps) => <StyledTable {...restProps}>{children}</StyledTable>
```
hollg, line 33: Since we're passing all the props on to one component here, we can simplify this to:

```tsx
export const Table: React.FC<TableProps> & TableSubComponents = (props: TableProps) => <StyledTable {...props}/>
```

I _think_ you can also get rid of the type declaration for `props` because it's impled by `React.FC<TableProps>`

hollg (changes requested):
This is looking great! Other than the very small changes I've commented, I agree with @thomasdigby that we should add some useful default styles. Ideally we should make it as easy as possible to render these components looking a) good and b) consistent with Atom's branding — otherwise we might end up with the first half dozen `Tables` in the app to all be manually styled totally differently.

I'd be very happy to weigh in on decisions about what those default styles should be if that would be helpful.

src/components/table/Table.tsx
shaneakins: Amazing! Thanks for the tip :)

src/components/table/Header.tsx
shaneakins: Carousel uses dot notation. After I noticed that, I figured it was best to stick with the pattern:
Screen Shot 2021-06-23 at 16 17 14

hollg: The individual subcomponents are called `CarouselPagination`, `CarouselArrows` etc. But then we do `Carousel.Arrows = CarouselArrows`. This ensures that the component has a complete name wherever you look at it (and we could make a separate `Pagination` component later with no clashes).

src/components/table/Table.test.tsx
shaneakins: Sure, will do

src/components/table/Table.mdx
```diff
@@ -0,0 +1,42 @@
...
+      <Table.HeaderCell>First Name</Table.HeaderCell>
+      <Table.HeaderCell>Last Name</Table.HeaderCell>
+      <Table.HeaderCell>Age</Table.HeaderCell>
+    </Table.Row>
+  </Table.Header>
+  <Table.Body>
+    <Table.Row>
+      <Table.Cell css={{border: '1px solid $tonal300'}}>Rakim</Table.Cell>
+      <Table.Cell css={{border: '1px solid $tonal300'}}>Jackson</Table.Cell>
+      <Table.Cell css={{border: '1px solid $tonal300'}}>35</Table.Cell>
```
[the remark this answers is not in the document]
shaneakins: That's a question I meant to ask: should I keep it unstyled except for basics like the text and spacing ,or style it with our standard (blue heading with white text and horizontal borders) and have that as default?

src/components/table/Table.mdx
```diff
@@ -0,0 +1,42 @@
...
+
+The `Table` component structure mirrors that of a regular HTML table, with the smaller `Table.Body`, `Table.Cell`, `Table.Footer`, `Table.Header`, `Table.HeaderCell` and `Table Row` components corresponding to the `<tbody>`, `<td>`, `<tfoot>`, `<thead>`, `<th>` and `<tr>` tags, respectively. 
+
+The default table width is 100% of the parent container, but this and other styles can be overridden with the `css` prop on all components.
+
+Best practice: Due to rendering issues with some browsers, adding borders to `Table.Row` is possible by styling the Table component with `border-collapse: collapse`; however, this practice is generally not recommended. The preferred method for adding borders is applying styles to the `Table.Cell` and `Table.HeaderCell` components;
+
+
+```jsx preview
+<Table css={{ height: 100, width: 400, color: '$primary500' }}>
```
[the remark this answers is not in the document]
shaneakins: Good catch - that was old code just to make sure things worked. Will update

src/components/table/Header.tsx
shaneakins: Ahh gotcha - the actual filenames! Will update

hollg: Have you pushed your latest changes? The filenames don't seem to have been updated

src/components/table/Table.test.tsx
hollg: This also hasn't been updated

src/components/table/Table.tsx
hollg: This also hasn't been updated

src/components/table/Table.mdx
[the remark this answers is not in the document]
hollg: I think this probably needs to be properly answered to get the PR merge-ready. I agree with @thomasdigby that there are some useful styles we can add without getting too opinionated which would be a v good start.

I also think we should have a couple of variants, so that we can do something like `variant="primary"` and get the styling in the first example from the screenshot in this PR's description. But I don't know whether we have any designs that use a table yet, so perhaps it's worth waiting to base variants on those? Keen to hear opinions on that.

hollg:
> I think we ought to consider how this table will be used, we're keeping flexibility with styling and composing components within a table but the usage is a bit verbose. Can you imagine us using a further abstraction to make it easier to display a set schema of data with this table?

I think this is a v important question too. Most times you render a table, you have an array of objects for the rows so you do something like this, which when combined with whatever logic you need in each row and the markup for the header and footer, gets _real_ noisy.

```tsx
<Table>
{
  data.map(item => {
    return (<Table.Row>
      <Table.Cell>{item.id}</Table.Cell>
      { /* etc */}
    </Table.Row>
  }
}
</Table>
```

I've seen lots of table components that take some sort of `data` prop and make some assumptions about how it should be processed, but they've always also had escape hatches so that you can override the default logic, because invariably there will be exceptions to what you want to render.

@thomasdigby did you have any specific idea in mind for the abstraction you mentioned? If you do then I think it could be worth getting it included in this PR but if not then designing a good one might take a while and I'm conscious that this PR has been open a while and we might be blocking some of @shaneakins's work — could be something to revisit?

src/components/table/Table.mdx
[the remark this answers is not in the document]
shaneakins: Atanas created a new table design that I've been using as a point of reference: [Atom Assessment Entrance Exams](https://www.figma.com/file/lXBJN5xPUgN6oUTcfSP01j/Atom-Assessments?node-id=412%3A0). To be clear, are you suggesting to make `variant="primary"` and set that as default, similar to how the Button component functions?

src/components/table/Header.tsx
shaneakins: Filenames have been updated - feel free to recheck

src/components/table/Header.tsx
shaneakins: Display names have been updated - feel free to recheck

src/components/table/Table.test.tsx
shaneakins: Test has been updated - feel free to recheck

src/components/table/Table.tsx
shaneakins: Refactor has been implemented - feel free to recheck

hollg (approved):

src/components/table/Table.mdx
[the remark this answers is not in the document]
thomasdigby: I don't think we should lean on variants until we have multiple to choose from, for now lets style this as per Atanas' designs, applying those styles when using it doesn't feel that great a DX to me.

thomasdigby:
> > I think we ought to consider how this table will be used, we're keeping flexibility with styling and composing components within a table but the usage is a bit verbose. Can you imagine us using a further abstraction to make it easier to display a set schema of data with this table?
> 
> I think this is a v important question too. Most times you render a table, you have an array of objects for the rows so you do something like this, which when combined with whatever logic you need in each row and the markup for the header and footer, gets _real_ noisy.
> 
> ```tsx
> <Table>
> {
>   data.map(item => {
>     return (<Table.Row>
>       <Table.Cell>{item.id}</Table.Cell>
>       { /* etc */}
>     </Table.Row>
>   }
> }
> </Table>
> ```
> 
> I've seen lots of table components that take some sort of `data` prop and make some assumptions about how it should be processed, but they've always also had escape hatches so that you can override the default logic, because invariably there will be exceptions to what you want to render.
> 
> @thomasdigby did you have any specific idea in mind for the abstraction you mentioned? If you do then I think it could be worth getting it included in this PR but if not then designing a good one might take a while and I'm conscious that this PR has been open a while and we might be blocking some of @shaneakins's work — could be something to revisit?

Yeah exactly, I think we should definitely export the table as is, the individual parts of the table fit the primitives model that we're using elsewhere. However for this to be _really_ useful I think we probably ought to provide the abstraction layer that maps data onto the table like @hollg suggests. This could (should) be a separate component and we can discuss the use case for that when delving into filterable/sortable tables, but it will broadly follow the same concept that `react-table` uses e.g. `columns` & `data`.

Let's not hold this PR up any longer, @hollg can you run through @shaneakins replies to your comments and resolve if necessary.

src/components/table/Table.tsx
```diff
@@ -0,0 +1,43 @@
...
+const StyledTable = styled('table', {
+  width: '100%',
+  fontFamily: '$sans',
+  fontSize: '$sm'
+})
+
+type TableProps = React.ComponentProps<typeof StyledTable>
+
+export const Table: React.FC<TableProps> & TableSubComponents = (
+  props: TableProps
```
thomasdigby, line 32: We don't need this second type validation here, `React.FC<TableProps>` should be enough

src/components/table/TableFooterCell.tsx
```diff
@@ -0,0 +1,12 @@
+import { styled } from '~/stitches'
+
+export const TableFooterCell = styled('td', {
+  fontFamily: '$sans',
+  color: '$primary500',
```
thomasdigby, line 5: Is this colour choice reflected in the designs? If not it might be worth defaulting to bold dark grey text instead

thomasdigby (approved):

PR #177, opened 2021-06-25
Feature: 1.0

thomasdigby:
### Remaining tasks
https://github.com/Atom-Learning/components/milestone/1

--- 
https://atomlearning-docs-beta-9kxbs.ondigitalocean.app/components


src/components/field-wrapper/FieldWrapper.tsx
```diff
@@ -23,30 +25,34 @@ export const FieldWrapper: React.FC<FieldWrapperProps> = ({
...
+      </Label>
+      {prompt && (
+        <Link href={prompt.link} size="sm">
+          {prompt.label}
+        </Link>
+      )}
+    </Flex>
+    {children}
+    {error && <ValidationError css={{ mt: '$2' }}>{error}</ValidationError>}
+    {description && (
```
andoulla, line 50: Shouldn't the description be before the error?  I might look a bit off to have  it move to show and hide the error

src/components/field-wrapper/InlineFieldWrapper.tsx
```diff
@@ -10,28 +10,40 @@ type InlineFieldWrapperProps = {
...
   css,
   label,
-  required
-}) => {
-  return (
-    <Box css={css}>
-      <Label
-        css={{ display: 'flex', alignItems: 'center' }}
-        required={required}
+  required,
```
andoulla, line 21: Since we are adding a description to the  FieldWrapper I think we should add it here. these wo components are practically the same other than the alignment

src/components/icon/Icon.tsx
```diff
@@ -13,6 +13,7 @@ const StyledIcon = styled('svg', {
   verticalAlign: 'middle',
   variants: {
     size: {
+      xs: { size: '$space$3' },
```
andoulla, line 16: How come us a different scale is used here?

src/components/list/List.tsx
```diff
@@ -4,71 +4,46 @@ import { styled } from '~/stitches'
...
+  [`& > ${StyledLi}`]: {
+    '&:not(:last-child)': { mb: '$2' },
+    '&:last-child': { mb: 0 }
+  },
   variants: {
-    theme: {
-      tonal: {
-        color: '$tonal900'
+    size: textVariantSize({ applyCapsize: false }),
+    as: {
```
andoulla, line 19: Would it be confusing to have a variant called `as` and the `as` prop every component has?

src/components/select/Select.tsx
```diff
@@ -61,9 +64,9 @@ export type SelectProps = Override<
 
 export const Select: React.FC<SelectProps> = React.forwardRef(
   ({ placeholder, children, ...remainingProps }, ref) => (
-    <StyledSelect {...remainingProps} ref={ref}>
+    <StyledSelect defaultValue="" {...remainingProps} ref={ref}>
```
andoulla, line 67: Is this defaultValue needed? I was looking into this error last week and setting both the defaultValue and the value seems to trigger it

`Warning: Select elements must be either controlled or uncontrolled (specify either the value prop, or the defaultValue prop, but not both). Decide between using a controlled or uncontrolled select element and remove one of these props. More info: https://fb.me/react-controlled-components`

src/documentation/Styling.mdx
```diff
@@ -51,10 +51,16 @@ You can also use them to switch [variants](https://stitches.dev/docs/variants)
 
 ```tsx
 <Button
-  // change the 'theme' to 'secondary' when you trigger the 'md' breakpoint
-  theme={{ '@initial': 'primary', '@md': 'secondary' }}
+  // change the 'theme' to 'success' when you trigger the 'md' breakpoint
```
andoulla, line 54: this is very cool

src/utilities/style/capsize.ts
```diff
@@ -15,15 +15,18 @@ import { CSS } from '~/stitches'
...
  */
-export const capsize = (margin: string): Record<string, CSS> => ({
+export const capsize = (
+  before: number,
+  after?: number
+): Record<string, CSS> => ({
   '&::before': {
     content: "''",
-    marginBottom: margin,
+    marginBottom: `-${before}em`,
```
andoulla, line 24: should `mb` and `mt` be used here?

src/components/field-wrapper/FieldWrapper.tsx
thomasdigby: Yeah that's a fair point, although it may not link to the input field as well? If there's a paragraph of text between the input and the error. Let me get some images back for you!

src/components/field-wrapper/InlineFieldWrapper.tsx
thomasdigby: Yeah I was hesitant about this as I think the block layout for the `FieldWrapper` more easily allows the description, but I think having the description underneath the `InlineFieldWrapper` could still work.

src/components/icon/Icon.tsx
thomasdigby: The sizing scale doesn't have many defined values in the lower sizes, generally the sizing scale works well but for these icons there's a need to be a bit pragmatic about what looks best in relation to other elements. As the icons don't have a defined boundary we can use a value like `12px` here without looking like it's out of sync with a 16px height Text component, or a small 32px button.

src/components/list/List.tsx
thomasdigby: This is a cool side-effect of the stitches variants and the polymorphic `as` prop, you can use that prop just as any other for variants and as we're setting `as` as `ul` or `ol`, it's easier to define the styles based on the element than another generic prop.

The `as` prop isn't used by the user in this case, they'll write `<List ordered />` and it'll render an `ol` with the relevant styles

src/components/select/Select.tsx
thomasdigby: Yeah I remember looking into this issue before but can't recall what came of it, let me see what I can do!

src/documentation/Styling.mdx
thomasdigby: _Right??_ This is the real game changer for me, ❤️ stitches

src/components/field-wrapper/FieldWrapper.tsx
thomasdigby: Screenshot 2021-07-20 at 11 37 38
Screenshot 2021-07-20 at 11 37 05

Out of the two versions, I think having the error closer to the field makes more sense, and I don't think the description jumping after the field is made valid should be toooo much of a concern.

andoulla: Now that I see it out on the screen, I realise that I'm more used to having the description below the heading but that would break most designs we have currently in place.

It feels like it needs something but this is already an improvement so let's go with it. We just need to make sure we don't add massive blocks of text in the description

src/components/icon/Icon.tsx
andoulla: thanks for explaining it

PR #179, opened 2021-07-07
MarkdownContent component

Mhoog:
https://user-images.githubusercontent.com/2501587/124798918-d9875b80-df4b-11eb-9818-c9b8bdc9ffac.mp4


jest.config.js
```diff
@@ -2,13 +2,14 @@ module.exports = {
...
+  testEnvironment: 'jsdom',
   testPathIgnorePatterns: ['/node_modules/'],
   setupFilesAfterEnv: ['./jest-setup.ts'],
   moduleNameMapper: {
     '~/(.*)': '<rootDir>/src/$1', // matches the path alias in tsconfig.json
     '^react$': '<rootDir>/node_modules/react'
   },
   transform: {
-    '^.+\\.(ts|tsx)$': 'babel-jest'
+    '^.+\\.[t|j]sx?$': 'babel-jest'
```
thomasdigby, line 13: Are these jest changes still required?

src/components/list/List.mdx
```diff
@@ -17,6 +17,16 @@ category: Content
 </List>
 ```
 
+```tsx preview
+<List type="ordered">
```
thomasdigby, line 21: Should this be `ordered` as a boolean prop to match the component?

src/components/markdown-content/MarkdownContent.mdx
```diff
@@ -0,0 +1,50 @@
+---
+title: Markdown Content
+component: MarkdownContent
+description: This is the description
```
thomasdigby, line 4: Could potentially use the opening paragraph here

src/components/markdown-content/MarkdownContent.mdx
```diff
@@ -0,0 +1,50 @@
...
+```
+
+The `MarkdownContent` component supports [Markdown directives](https://talk.commonmark.org/t/generic-directives-plugins-syntax/444) via the `handleDirectives` prop. This prop takes a function that returns a `ReactElement`, the function will have access to the Markdown `node` and the `handleNode` function that can be used to transform child nodes.
+
+See the following example that transform the following directive into a `Button` component:
+
+```tsx live
+<MarkdownContent
+  content={`:button[This is a button]{href=https://atomlearning.co.uk theme="secondary" isRounded=true appearance=outline}`}
+  handleDirectives={(node, handleNode) =>
```
hollg, line 44: I don't think I follow how `handleDirectives` works. Do not always want the same behaviour, e.g. mapping `'button'` to `<Button>` etc? Or is the idea that we could use this prop to control which features are available in a particular instance?

src/components/markdown-content/MarkdownContent.tsx
```diff
@@ -0,0 +1,173 @@
...
+export const MarkdownContent: React.FC<MarkdownContentProps> = ({
+  content,
+  handleDirectives = () => null
+}) => {
+  const AST = fromMarkdown(content, {
+    extensions: [syntax()],
+    mdastExtensions: [directive.fromMarkdown]
+  })
+
+  const handleNode = (node) => {
```
hollg, line 29: This function is very cool but also very long. Could we make it a bit easier to read by splitting it up? I think `getSize` and `getAs` could easily become `getHeadingSize` and `getHeadingAs` and be moved outside the function, for example. 

We could possibly also move the `Box` styling for `code` and `inlineCode` into `styled()` calls — normally I suggest using inline `css` props for styling that is only used once, but in this case the two big blocks of styling are noise if you're trying to understnd the logic that maps nodes to components.

src/components/markdown-content/MarkdownContent.tsx
```diff
@@ -0,0 +1,173 @@
...
+    if (node.type === 'textDirective') {
```
hollg, line 30: Are the different node types in any particular order? If not it might be nice to put them into one, even if it's just alphabetical.

jest.config.js
Mhoog: The `transform` changes weren't needed. Changed that one back.

The `testEnvironment` one is needed, as I updated `jest` to v27+ and the [default `testEnvironment` is now `node` instead of `jsdom`](https://jestjs.io/blog/2021/05/25/jest-27#flipping-defaults)

src/components/markdown-content/MarkdownContent.mdx
Mhoog: [Directives are a way of 'extending' Markdown using syntax that you can shape to your needs.](https://talk.commonmark.org/t/generic-directives-plugins-syntax/444)

Because it's very much depending on the use case, I'd prefer not to include (and _force_ any directives) in this library. Instead, this way it's an option to extend the default functionality.

src/components/markdown-content/MarkdownContent.tsx
Mhoog: Nice suggestion, thanks!

Moved the functions for Heading outside and used `styled()` for components with more than 1 line of CSS.

src/components/markdown-content/MarkdownContent.tsx
Mhoog: No particular order before, changed them to alphabetical 🙂

src/components/list/List.mdx
Mhoog: Oops, yeah I've changed that from a `type` prop to `boolean` at some point and forgot to change it here 😅 .

Fixed it.

thomasdigby (approved):

hollg (approved):

PR #182, opened 2021-07-15
Feature: Search input

thomasdigby:
Keen to get some thoughts on creating a search input for the design system, it'll only be an icon and a search input type but I think for consistency it'll be worth including.

Screenshot 2021-07-15 at 17 46 28

### DIFF
--- a/src/components/search-input/SearchInput.mdx
+++ b/src/components/search-input/SearchInput.mdx
@@ -1,10 +1,12 @@
 ---
 title: Search Input
 component: SearchInput
-description: An input used for searching
+description: A search input with matching search icon 
 category: Forms
 ---
 
+`SearchInput` wraps `Input` and adds an icon to visualise the specific input type, it's recommended to use this component when the user is performing a search query. It is functionally identical to the standard `Input` aside from a more specific "type" but makes it clear to the user what the action is and what potential feedback they might receive.
+
 ```tsx preview
-<SearchInput />
+<SearchInput name="search" css={{ width: 320 }} placeholder="Search" />
 ```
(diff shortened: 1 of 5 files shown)


alexandrunemes:
I'm wondering if we should have some docs or at least a test that might show how to use the input in a search context. Like getting its value, doing something when the value changes etc. Or do we just leave it as unopinionated as it currently is.

thomasdigby:
> I'm wondering if we should have some docs or at least a test that might show how to use the input in a search context. Like getting its value, doing something when the value changes etc. Or do we just leave it as unopinionated as it currently is.

Yeah definitely, this will go back to the design team to document fully the use-cases outside of the technical perspective.

I was unsure about whether this could just be an addition to the input to keep it simple but that felt like it would go against our component composition model, then this became more about consistency to ensure that all of our search inputs are the same. So with that in mind, I definitely think it's worth keeping it unopinoinated but there's no harm adding additional docs to show it in combination with a filter method or something similar.

andoulla:
In the designs, the alignment of the icon tends to be on the right. Will it be ok to move it to the left? I like the idea of only having one alignment option to enforce consistency

src/components/search-input/SearchInput.mdx
```diff
@@ -0,0 +1,10 @@
+---
+title: Search Input
+component: SearchInput
+description: An input used for searching
+category: Forms
+---
+
```
andoulla, line 7: It feels a bit bare: 
```suggestion

Renders the Search icon along with the input component
```

thomasdigby:
> In the designs, the alignment of the icon tends to be on the right. Will it be ok to move it to the left? I like the idea of only having one alignment option to enforce consistency

Yeah that's an interesting point, I've generally seen it positioned to the left in most cases on the web. The reason for that is that I think it makes a distinction between the icon representing a potential clickable action or just an indicator of the type of input.

It's easy enough to make a non-breaking change with this in future so let's merge it in and discuss in more detail when we have the time.

PR #183, opened 2021-07-15
Feature: Small form fields

thomasdigby:
Have seen this smaller form field convention used across a few incoming features, not sure it should be applicable to password, checkbox/radio or textarea, I guess combobox should also allow a small size input as well. 

Screenshot 2021-07-15 at 17 59 13

### DIFF
--- a/src/components/field-wrapper/__snapshots__/FieldWrapper.test.tsx.snap
+++ b/src/components/field-wrapper/__snapshots__/FieldWrapper.test.tsx.snap
@@ -1,34 +1,37 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`FieldWrapper component renders 1`] = `
-.sxojaks {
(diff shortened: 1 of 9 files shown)


alexandrunemes:
What's the concern regarding password/checkbox/textarea? Usability/readability?

thomasdigby:
> What's the concern regarding password/checkbox/textarea? Usability/readability?

Yeah exactly, for basic text and selecting from a list of options, they're simple enough to be usable in contexts that require them to be reduced in size. These are mainly for cases where the user may need to type a word to search, or select something to filter a list of items.

Asking the user to enter a password is an important focus for them so shouldn't be used in a reduced format, textarea has a similar issue for me where a reduced size doesn't fit with the type of content you're asking from the user. Checkboxes are already small enough so definitely can't get any smaller!

andoulla (approved):

PR #185, opened 2021-07-15
Fix: Inline field wrapper alignment

thomasdigby:
`Switch` and `Input`/`Select` children within `InlineFieldWrapper` were throwing off the alignment, I updated `Switch` to the new design and tweaked a few layout props to allow more flexibility with alignment.

Screenshot 2021-07-15 at 19 13 50

### DIFF
--- a/src/components/checkbox-field/CheckboxField.tsx
+++ b/src/components/checkbox-field/CheckboxField.tsx
@@ -28,14 +28,14 @@ export const CheckboxField: React.FC<CheckboxFieldProps> = ({
   const error = errors[name]?.message
 
   return (
-    <InlineFieldWrapper label={label} css={css} error={error}>
-      <Controller
-        onChange={() => setIsChecked((current) => !current)}
-        control={control}
-        name={name}
-        defaultValue={isChecked}
-        rules={{ required: true }}
-        render={({ onChange, value, name: innerName }) => (
+    <Controller
+      onChange={() => setIsChecked((current) => !current)}
+      control={control}
+      name={name}
+      defaultValue={isChecked}
+      rules={{ required: true }}
+      render={({ onChange, value, name: innerName }) => (
+        <InlineFieldWrapper label={label} css={css} error={error}>
           <Checkbox
             name={innerName}
             onCheckedChange={() => {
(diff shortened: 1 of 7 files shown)


src/components/checkbox-field/CheckboxField.tsx
```diff
@@ -47,9 +47,9 @@ export const CheckboxField: React.FC<CheckboxFieldProps> = ({
             value={value}
             {...remainingProps}
           />
-        )}
-      />
-    </InlineFieldWrapper>
+        </InlineFieldWrapper>
+      )}
+    />
```
thomasdigby, line 52: Had to switch around the order of the `Controller` and `InlineFieldWrapper` to ensure that `Checkbox` is a direct child of `InlineFieldWrapper`

src/components/field-wrapper/FieldWrapper.mdx
```diff
@@ -7,16 +7,24 @@ category: Forms
...
 </InlineFieldWrapper>
 ```
+
+It can also be used alongside components like `Input`, `Select` or `Switch` when you may require some condensed inline fields.
+
+```tsx preview
+<InlineFieldWrapper direction="reverse" align="center" label="Toggle a field">
+  <Switch />
+</InlineFieldWrapper>
+```
```
andoulla, line 30: Can we add a small blurp about the variants here? just to make them more visible?

src/components/field-wrapper/InlineFieldWrapper.tsx
```diff
@@ -3,14 +3,34 @@ import * as React from 'react'
...
+const InlineLabel = styled(Label, {
+  display: 'flex',
+  fontWeight: 400,
+  maxWidth: 'max-content',
+  variants: {
+    align: {
+      baseline: { alignItems: 'baseline' },
+      center: { alignItems: 'center' }
+    },
+    direction: {
```
andoulla, line 20: how come we need a reverse variant?

andoulla (approved):

src/components/field-wrapper/InlineFieldWrapper.tsx
thomasdigby: This is to allow the label to sit on the left and the control on the right, this isn't used for checkbox or radio buttons but is necessary for our switch controls. The example is in the docs width the `Switch` and `InlineFieldWrapper`

PR #186, opened 2021-07-17
Feature: Calendar

thomasdigby:
Thought I would test `dayzed` as a potential candidate for our date picker, seems to work well, is a fairly simple implementation and the package is lightweight. There are also some good examples for selecting a date range etc. https://github.com/deseretdigital/dayzed/tree/master/docs/examples

I was imagining that we'd have a few components, this `Calendar` component which displays the month, allows selecting a date, date-range, moving between months etc. and a `DateInput` which combines that with an `Input` and `Popover` to allow users to select a date via the input.

Screenshot 2021-07-17 at 08 25 59


src/components/calendar/Calendar.mdx
```diff
@@ -0,0 +1,10 @@
+---
+title: Calendar
+component: Calendar
+description: This is the description
+category: Content
```
andoulla, line 5: Maybe this should be Form as it allows to input dates etc

src/components/calendar/Calendar.mdx
```diff
@@ -0,0 +1,10 @@
...
+---
+
+```jsx preview
+<Calendar />
+```
```
andoulla, line 10: Can we add the link to the Dayzed here like we do in other components with dependencies

src/components/calendar/Calendar.tsx
```diff
@@ -0,0 +1,133 @@
...
+      <Flex
+        css={{
+          justifyContent: 'space-between',
+          left: 0,
+          position: 'absolute',
+          right: 0,
+          top: 0
+        }}
+      >
+        <ActionIcon {...getBackProps({ calendars })} label="Previous month">
```
andoulla, line 71: We need to be able to override the labels in ActionIcon for translations. Maybe we can pass a copy object?

src/components/calendar/Calendar.tsx
```diff
@@ -0,0 +1,133 @@
...
+      true: { color: '$tonal400' }
+    }
+  }
+})
+
+type CalendarProps = DayzedInterface & {
+  css?: CSS
+}
+
+export const Calendar: React.FC<CalendarProps> = ({
```
andoulla, line 50: Do we need to be able to pass a locale? In the US the week starts on Sunday but EU is on Monday

andoulla:
Hey, could we merge this in? We might need it soon

thomasdigby:
@andoulla This is still a prototype at the moment, but a reasonably successful one; happy to hand over the reins to you or your pod if you need this component. I would like to retain the `Calendar`, `DateInput`/`DatePicker` format so we can keep the two main components separate.

src/components/calendar/Calendar.tsx
thomasdigby: Yeah that's a good shout, we can configure `firstDayOfWeek` for that specific issue. We'll have to test how specific locale dates are handled by dayzed

src/components/calendar/Calendar.mdx
thomasdigby: I wonder if `Calendar` shouldn't be exported by the component lib and instead is just a component used within `DateInput`. It's unlikely to be useful without user input so we can avoid documentation for this particular component for now

andoulla:
> @andoulla This is still a prototype at the moment, but a reasonably successful one; happy to hand over the reins to you or your pod if you need this component. I would like to retain the `Calendar`, `DateInput`/`DatePicker` format so we can keep the two main components separate.

We can finish it off if we need it (not sure yet). Could you please clarify what do you mean about 2 separate components? how would a DatePicker be different from a Calendar? We need a datepicker this is why I ask

thomasdigby:
> > @andoulla This is still a prototype at the moment, but a reasonably successful one; happy to hand over the reins to you or your pod if you need this component. I would like to retain the `Calendar`, `DateInput`/`DatePicker` format so we can keep the two main components separate.
> 
> We can finish it off if we need it (not sure yet). Could you please clarify what do you mean about 2 separate components? how would a DatePicker be different from a Calendar? We need a datepicker this is why I ask

So we would retain the current `Calendar` component as a way to display the calendar and bind events for moving between months and selecting dates, the `DatePicker` component would then combine that with an input, popover etc. and would be responsible for ensuring the date selection is propogated back to the input.

We may want to use the calendar view separately to the form usage in future so it makes sense to keep their concerns separate.

PR #187, opened 2021-07-19
Fix: export StyledButton to allow custom Button logic

Mhoog:
We want to use `next/link` to create links that use client-side routing:
https://nextjs.org/docs/api-reference/next/link

If you want to wrap that around `FunctionComponents` you need to pass the `ref` and the props `onClick` and `href` for it to work.

However, the `Button` component sets `onClick` to `undefined` if there is an `href`:
https://github.com/Atom-Learning/components/blob/main/src/components/button/Button.tsx#L218

This causes the client-side routing to fail and fallback to server side routing.

Exporting `StyledButton` would allow us to create our own `Button` component that does pass on the `onClick` in case there is an `href`, while still use all the styling defined here.


thomasdigby:
Do we need to add that export to the main components export as well? How do we choose which styled components we export?

hollg:
I agree with @thomasdigby that this requires a further change to `src/components/index.ts` to export the `StyledButton`. 

> How do we choose which styled components we export?
I'm not sure how to define the category of styled components that are useful to export, but I think this definitelys fits in it 😅. The `Button` adds logic that (as we have just learned) isn't always desirable, but is what we want as a default. This lets us compose other "smart" button components elsewhere, which I think seems like the right approach.

...unless we take the stance that all the logic should be included only at the highest level where it will always be applicable, and move it to the integration layer in `atom-core`? 🤷🏻‍♂️

Mhoog:
@thomasdigby, @hollg,

I didn't add the export to the `index.ts` files as a sort of way to say _"It's possible, but we don't support it"_.

Happy to add it to the index files if you feel different about that?

And to choose which components we export, I think it's only useful for components that have a `Styled...` variant that includes some styling _and_ have logic that forces / override certain props. I saw that the `Link` component had a similar structure, so also exported that one. All the others seem to be flexible enough?

thomasdigby:
@Mhoog Can we even access that export if it's not exported from our main exports file? Everythings bundled into a single file so you can't access it via the file path. Does rollup still include the export syntax in the bundle if it's not included within the main exports file?

thomasdigby:
Happy to include it as a undocumented feature if it works as you've written it

Mhoog:
> @Mhoog Can we even access that export if it's not exported from our main exports file? Everythings bundled into a single file so you can't access it via the file path. Does rollup still include the export syntax in the bundle if it's not included within the main exports file?

Yup, you were right there. Pushed another change that includes the correct exports.

thomasdigby (approved):

PR #188, opened 2021-07-21
feat: allow rendering custom components in MarkdownContent

Mhoog:
## Background
For the marketing website we want to render our own component instead of the default component for one specific Markdown type (`link`).

## Changes to `MarkdownContent`
- [x] Moved all components we render for the different Markdown types into their own files.
- [x] Created a map that maps the Markdown types to the different components.
- [x] Added a `customComponents` prop to `MarkdownContent`.
- [x] Removed the `handleDirectives` prop, as directives can now also be handled by the `customComponents` prop.
- [x] Updated the documentation for `MarkdownContent` to reflect the changes.
- [x] Updated the tests for `MarkdownContent` to reflect the changes.
- [x] Improved typing.

## Other changes
- [x] Set `alwaysTryTypes` to `true` in `.eslintrc`
- [x] Added the `Markdown*` components to `StackContent` to make sure there is space between the elements.


src/components/markdown-content/MarkdownContent.tsx
```diff
@@ -1,147 +1,89 @@
...
+  const generateNodeKey = (node: Content) => {
+    if (node.position?.start) {
+      const { line, column, offset } = node.position?.start
+      return `${node.type}${line}${column}${offset}`
     }
 
-    if (node.type === 'textDirective') {
-      return handleDirectives(node, handleNode)
-    }
+    return `${node.type}${+new Date()}`
```
thomasdigby, line 74: Is there any way we can avoid this `new Date()`? Do we have any other unique data that we can use? I assume that's what the combo of `type`, `line`, `column` and `offset` is doing with `node.position.start`

Mhoog: There is one type (that we don't use) that doesn't have `node.position.start` and doesn't have any other unique data.

This `new Date()` key will only be used for that type, all the others should use the `node.position.start` combo.

src/components/markdown-content/MarkdownContent.test.tsx
```diff
@@ -1,8 +1,10 @@
 import { render, screen } from '@testing-library/react'
 import { axe } from 'jest-axe'
+import { TextDirective } from 'mdast-util-directive'
```
thomasdigby, line 3: Could we export this type (and any others worth including?) so when the library is consumed they can use it without having to reference `mdast-util-directive`?

src/components/markdown-content/MarkdownContent.tsx
thomasdigby: Ok cool, might be worth adding a comment just above this with that info 👍

thomasdigby (approved):

hollg (approved):
👍🏻

src/components/markdown-content/MarkdownContent.test.tsx
Mhoog: Just tried to do that, but noticed that the types were not available anymore because we had to downgrade the packages (to allow CJS).

Had to remove typings for `node` on `MarkdownContentProps` and `handleNode` instead 😞

PR #191, opened 2021-07-26
feat: added href functionality with styling - ActionIcon

LeahB8:
- Added `href` prop capability to `ActionIcon`
- Generic and reusable in different usages - `atom-core` & `marketing-site`

### Screenshot of both uses of ActionIcon
Screenshot 2021-07-26 at 17 56 28

### DIFF
--- a/src/components/action-icon/ActionIcon.tsx
+++ b/src/components/action-icon/ActionIcon.tsx
@@ -128,8 +128,40 @@ const StyledButton = styled('button', {
   ]
 })
 
+const StyledForm = styled('form', { display: 'inline' })
+
 type ActionIconProps = React.ComponentProps<typeof StyledButton> & {
   label: string
+}
+
+type ActionIconButtonProps = React.ComponentProps<typeof StyledButton> & {
+  children: React.ReactNode
+  message: string
+}
+
+const ActionIconButton: React.FC<ActionIconButtonProps> = ({
+  children,
+  message,
+  ...restProps
+}) => {
+  return (
+    <StyledButton {...restProps}>
+      {React.Children.map(children, (child) => {
+        // TS needs this check for any following code to access child.type
+        // even with optional chaining
+        if (!React.isValidElement(child)) {
+          throw new Error(message)
+        }
+
+        invariant(
+          child.type === Icon,
+          `Children of type ${child?.type} aren't permitted. Only an ${Icon.displayName} component is allowed in ${ActionIcon.displayName}`
+        )
+
+        return child
+      })}
+    </StyledButton>
+  )
 }
 
 export const ActionIcon: React.FC<ActionIconProps> = React.forwardRef(
@@ -140,6 +172,7 @@ export const ActionIcon: React.FC<ActionIconProps> = React.forwardRef(
       appearance = 'subtle',
       size = 'md',
       label,
+      href,
       ...remainingProps
     },
     ref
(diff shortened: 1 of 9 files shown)


src/components/action-icon/ActionIcon.tsx
```diff
@@ -148,32 +181,32 @@ export const ActionIcon: React.FC<ActionIconProps> = React.forwardRef(
...
+      children,
+      message: INVALID_CHILDREN_MESSAGE,
+      ...remainingProps,
+      ...buttonProps
+    }
+
+    if (href) {
+      // Wrapped in a form to keep styles consistent with non-href ActionIcons
+      return (
+        <StyledForm action={href} method="post">
```
thomasdigby, line 203: Why do we need a form for this? 🤔

LeahB8: It was the only way I found to keep styles consistent across an `ActionIcon` with `href` prop and one without. Couldn't understand how it was getting styled differently when it was an `a` tag compared to a `button`. This is how it looked with an a tag with href prop and a button without         
Screenshot 2021-07-26 at 17 13 23

thomasdigby: Introducing a form to get some styles working correctly is really heavy handed and potentially inaccessible as well, a screenreader would announce a form instead of a link. There's clearly _something_ wrong here that isn't allowing the anchor to display correctly. Let me take a look.

LeahB8: Okay fair. I might have just reached my limit on css for one day. Happy to revisit this

src/components/action-icon/ActionIcon.tsx
```diff
@@ -128,8 +132,10 @@ const StyledButton = styled('button', {
   ]
 })
 
-type ActionIconProps = React.ComponentProps<typeof StyledButton> & {
+type ActionIconProps = React.ComponentPropsWithRef<typeof StyledButton> & {
```
andoulla, line 135: Correct me if I'm wrong but we might need to pass a ref, like when we need to wrap the action icon with the Tooltip.

thomasdigby: Yep this is what `ComponentWithRef` is doing, did you read it as `ComponentWithoutRef`?

andoulla: Yes, so sorry, sleep deprivation is getting to me.

andoulla (approved):

src/components/action-icon/ActionIcon.tsx
```diff
@@ -148,10 +156,14 @@ export const ActionIcon: React.FC<ActionIconProps> = React.forwardRef(
 
     invariant(React.Children.count(children) === 1, INVALID_CHILDREN_MESSAGE)
 
+    const optionalLinkProps = href
+      ? { as: 'a', href, onClick: undefined }
+      : ({ onClick, type: 'button' } as const)
```
hollg, line 161: What is `as const` doing here? I've never really understood it :D

src/components/action-icon/ActionIcon.tsx
```diff
@@ -128,8 +132,10 @@ const StyledButton = styled('button', {
   ]
 })
 
-type ActionIconProps = React.ComponentProps<typeof StyledButton> & {
+type ActionIconProps = React.ComponentPropsWithRef<typeof StyledButton> & {
   label: string
+  href?: string
```
hollg, line 137: It looks like the logic we want is that this component accepts _either_ an `href` or an `onClick`, but not both. I think we can use TS to declare that, with something like this:

```ts
type ActionIconProps = React.ComponentPropsWithRef<typeof StyledButton> & { label: string } & ( { href: string, onClick?: never } | { onClick: (...args: unknown[]) => void, href?: never } )
```

hollg: I think we should probably also document the logic. Your current implementation seems to overwrite any provided `onClick` with `undefined` if `href` is also provided, which I think is worth pointing out. 

Even with my suggested TS version, I think it would be worth a line in the documentation for this component.

src/components/action-icon/ActionIcon.tsx
thomasdigby: As far as I'm concerned this is a bit of a stop gap to get this update to work with TS before the Stitches crew sort out their types. `as const` essentially removes the chance that `onClick` or `type` will be anything other than what is declared here, so no potential `undefined` or falsey values, Stitches likes this. 👍 

I want us to overhaul the types in the entire lib when we have an updated beta from Stitches so hopefully this will be irrelevant soon.

src/components/action-icon/ActionIcon.tsx
thomasdigby: Yeah I skipped this due to time/laziness. I attempted to get @michalrozenek's `NavigatorActions` working here but couldn't get it to work, let me give it a second shot.

PR #195, opened 2021-07-30
Feat: table styling updates

andoulla:
Adding styling changes using the new brand colours:
- Changing the header colour and border-radius
- Colouring every other row for better readability
- Creating a larger row to keep internal controls/buttons clickable in smaller devices

Screenshot 2021-07-30 at 12 38 29

### DIFF
--- a/src/components/table/Table.mdx
+++ b/src/components/table/Table.mdx
@@ -5,10 +5,10 @@ description: The Table component displays a collection of data grouped into rows
 category: Layout
 ---
 
-The `Table` component structure mirrors that of a regular HTML table, with the smaller `Table.Body`, `Table.Cell`, `Table.Footer`, `Table.Header`, `Table.HeaderCell` and `Table Row` components corresponding to the `<tbody>`, `<td>`, `<tfoot>`, `<thead>`, `<th>` and `<tr>` tags, respectively. 
+The `Table` component structure mirrors that of a regular HTML table, with the smaller `Table.Body`, `Table.Cell`, `Table.Footer`, `Table.Header`, `Table.HeaderCell` and `Table Row` components corresponding to the `<tbody>`, `<td>`, `<tfoot>`, `<thead>`, `<th>` and `<tr>` tags, respectively.
 
 ```jsx preview
-<Table css={{ width: '500px', mb: '100px'}}>
+<Table css={{ width: '500px', mb: '100px' }}>
   <Table.Header>
     <Table.Row>
       <Table.HeaderCell>First Name</Table.HeaderCell>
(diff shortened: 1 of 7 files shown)


src/components/table/TableHeaderCell.tsx
```diff
@@ -1,13 +1,22 @@
 import { styled } from '~/stitches'
 
 export const TableHeaderCell = styled('th', {
-  bg: '$secondary',
+  bg: '$tertiary',
   color: 'white',
   fontFamily: '$body',
   fontWeight: 600,
+  lineHeight: '$fontSizes$lg',
```
thomasdigby, line 8: We'll want to declare line height as unitless instead of pixel value, I would match the `md` text sizes `1.5`. I'll create a ticket for sharing these styles as there's a bit of repitition in some of our other components as well.

src/components/table/Table.tsx
```diff
@@ -23,14 +23,29 @@ type TableSubComponents = {
...
+  variants: {
+    size: {
+      md: {
+        td: {
+          height: '$5'
+        }
+      },
+      lg: {
+        td: {
+          height: 'calc($5 + $0)'
```
thomasdigby, line 36: I'm a little uneasy applying styles through the HTML element even though we can be confident that the element itself is used, I guess the weigh up is either mapping through child elements to apply a props to the `Cell` component or asking the user to apply that prop themselves, neither of which are good options. Let's stick with this but might be worth considering how we can consistently apply styles for this type of component from a single prop.

src/components/table/Table.tsx
```diff
@@ -23,14 +23,29 @@ type TableSubComponents = {
 const StyledTable = styled('table', {
   width: '100%',
   fontFamily: '$sans',
-  fontSize: '$sm'
+  fontSize: '$sm',
+  variants: {
+    size: {
+      md: {
+        td: {
+          height: '$5'
```
thomasdigby, line 31: Can we change these to `minHeight` to ensure they work resposively and with varied content sizes?

andoulla: It is a fair point, I will check with Atanas about it as I think the intent is that we won't have varied row sizes. We want to fix them at those 2

src/components/table/TableCell.tsx
```diff
@@ -2,11 +2,16 @@ import { styled } from '~/stitches'
 
 export const TableCell = styled('td', {
   borderBottom: '1px solid $tonal200',
+  boxSizing: 'border-box',
   color: '$tonal600',
-  fontFamily: '$sans',
+  fontFamily: '$body',
+
   p: '$2 $3',
```
thomasdigby, line 9: We should tweak the vertical padding for the table cell so there's not too much space around the buttons in a row. I reckon we can go down to `$1` (4px)

thomasdigby: That means we can probably drop the `lg` table size to `$4` (40px) minHeight, and `$3` (32px) for `md` size

src/components/table/Table.mdx
```diff
@@ -64,27 +64,91 @@ Best practice: Due to rendering issues with some browsers, adding borders to `Ta
...
+      >
+        Rakim
+      </Table.Cell>
+      <Table.Cell
+        css={{ bg: '$tonal100', borderBottom: '1px solid $subjectEnglish' }}
+      >
+        Jackson
+      </Table.Cell>
+      <Table.Cell
+        css={{ bg: '$tonal100', borderBottom: '1px solid $subjectEnglish' }}
```
thomasdigby, line 78: Should we include these styles in the docs now we're adding specific styles for even rows?

src/components/table/Table.tsx
andoulla: Just checked with Atanas, he is happy with changing this. We can always revert it later if we need to go back to just 2 specific sizes

src/components/table/Table.mdx
andoulla: No you are right, I will remove

src/components/table/Table.tsx
thomasdigby: We'll need to have it `minHeight` otherwise it'll overflow the container, I get the need for consistent heights (which this will still achieve) but we can't disregard the potential for there to be longer content.

src/components/table/TableHeaderCell.tsx
andoulla: So I should just use 1.5 here instead of a tocken?

thomasdigby: Yep, `lineHeight: 1.5`

src/components/table/Table.mdx
```diff
@@ -64,27 +64,91 @@ Best practice: Due to rendering issues with some browsers, adding borders to `Ta
...
+        Footer 3
+      </Table.FooterCell>
+    </Table.Row>
   </Table.Footer>
 </Table>
 ```
+
+## Variants
+
+The `Table` component has 2 size variants that control the row height. The to available sizes are `md`(default) and `lg` which should be used when the table needs to show more than just text, ie: `Button(s)` and `ActionIcon(s)`. The increased height helps keep the elements clickable on smaller screensizes.
```
hollg, line 130:
```suggestion
The `Table` component has 2 size variants that control the row height. The two available sizes are `md`(default) and `lg` which should be used when the table needs to show more than just text, ie: `Button(s)` and `ActionIcon(s)`. The increased height helps keep the elements clickable on smaller screensizes.
```

src/components/table/Table.tsx
hollg: Could we not just reference the relevant React component here as the selector? Would that resolve your concern @thomasdigby ?

andoulla: I was trying to figure out a way of passing the size prop from `Table` to `TableCell` so we just need to pass it on the Table component and then it trickles down but I couldn't figure a clean readable way for it. 

I avoided doing add the variant on the `TableCell` as we don't want users to be able to have a small row and a large row. They have to be consistent.

hollg: Could all of that not still be achieved if we replace `td` on line 35 with a reference to the `TableCell` component?

src/components/table/Table.mdx
andoulla: thank you

src/components/table/Table.tsx
thomasdigby: @hollg I actually didn't explain myself properly sorry, it's less about the selector (although a bit about the selector) and more about the placement of those styles, as it breaks component/style encapsulation and overrides the component without the predictability of a prop. Ideally we'd co-locate this height style with the `TableCell` itself, but then you get into the issue that we're discussing alongside this of how to pass that along cleanly.

andoulla: Yes we can, sorry your message didn't come through before I posted my response

hollg: Haha, sorry both, I interrupted _and_ misunderstood what I interrupted!

andoulla: but I think I would still do @hollg suggestion as it is better. Doesn't solve the whole problem but it is an improvement

src/components/table/TableCell.tsx
andoulla: it looks a bit too narrow I think
image

andoulla: we would definitely have to overwrite it in the assessment tables

thomasdigby: Huh this doesn't look at all like I expected it to, let me quickly play with it in the sandbox!

andoulla: thanks

thomasdigby: So the padding is fine as `$2 $3`, we should change the height of `md` and `lg` to `$4` and `$5` respectively, if we're adding `Button` or `ActionIcon` within a table cell, the 16px of vertical padding along with the 32px height button/icon will totaly the 48px height for the large table. 40px will need to be the minimum height for the medium table as the padding and line-height total more than that anyway.

So essentially:

```ts
  variants: {
    size: {
      md: {
        td: { height: '$4' }
      },
      lg: {
        td: { height: '$5' }
      }
    }
  }
```

is the way to go.

Also, just FYI, `min-height` doesn't work on table cells (we learn something new every day!) and `height` acts like `min-height` in this context, so something like `height: '$5'` is totally fine for our table cell 👍

src/components/table/TableCell.tsx
```diff
@@ -2,11 +2,16 @@ import { styled } from '~/stitches'
 
 export const TableCell = styled('td', {
   borderBottom: '1px solid $tonal200',
+  boxSizing: 'border-box',
   color: '$tonal600',
-  fontFamily: '$sans',
+  fontFamily: '$body',
+
```
thomasdigby, line 8: We should add a `lineHeight: 1.5` here as well

src/components/table/TableCell.tsx
andoulla: I didn't know that min-height didn't work (how embarrassing). 

With the md size, would it be ok that it will look different depending on if it has buttons etc or not. I think this we what we were trying to avoid by setting the height in the first place. 

 the use case I'm thinking if we have a row table that might conditionally show a control after the data have been updated from the API. Which means the users will see the table grow. it will be so fast it might not be noticable

src/components/table/TableCell.tsx
andoulla: In the designs this is slightly bigger so I'm trying to calculate the right value

src/components/table/TableCell.tsx
thomasdigby: I think in that case it's on the user to choose `lg` to support that? We can't have all tables be large enough to render a button as it'll be too big for standard data

src/components/table/TableCell.tsx
thomasdigby: We should keep it consistent with other text so 1.5 unless it's really _really_ necessary

thomasdigby (approved):

src/components/table/Table.test.tsx
```diff
@@ -28,6 +28,32 @@ describe(`Table component`, () => {
     expect(container).toMatchSnapshot()
   })
 
+  it('renders with size set to log', async () => {
```
alexandrunemes, line 31: smal typo: 'renders with size set to _lg_'

alexandrunemes:
Approving this. Only thing I could catch was a small typo.

alexandrunemes (approved):

src/components/table/Table.test.tsx
andoulla: thanks, fixed

shaneakins (commented):
Looks good. Is the banding (alternate coloring of rows) the new standard?

andoulla:
> Looks good. Is the banding (alternate coloring of rows) the new standard?
Yes it is but it can always be overwritten if we need it gone

PR #196, opened 2021-08-02
feat: added disabled state to action icon

alexandrunemes:
Small change that adds some styling to the ActionIcon's disabled state, to let the user know the element is not clickable.
Screenshot 2021-08-02 at 15 53 08


src/components/action-icon/ActionIcon.tsx
```diff
@@ -42,6 +42,11 @@ const StyledButton = styled('button', {
   justifyContent: 'center',
   p: 'unset',
   transition: 'all 125ms ease-out',
+  '&[disabled], &[disabled]:hover, &[disabled]:focus': {
```
andoulla, line 45: Should we handle not allow the click function to be executed when the button is disabled?

andoulla: yes but it can return an href and that doesn't have a disabled state. Sorry should have been more specific in my original comment

alexandrunemes: I've added handling for passing href and disabled simultaneously.

src/components/action-icon/ActionIcon.tsx
```diff
@@ -179,6 +190,8 @@ export const ActionIcon = React.forwardRef<HTMLButtonElement, ActionIconProps>(
         appearance={appearance}
         size={size}
         ref={ref}
+        disabled={disabled}
+        data-testid="actionIcon"
```
thomasdigby, line 194: Could we remove this and test by role instead? Just so we don't pollute the output for consumers of the design system

src/components/action-icon/ActionIcon.tsx
```diff
@@ -167,7 +173,12 @@ export const ActionIcon = React.forwardRef<HTMLButtonElement, ActionIconProps>(
...
 
     const optionalLinkProps = href
-      ? ({ as: 'a', href, onClick: undefined } as const)
+      ? ({
+          as: 'a',
+          href: disabled ? null : href,
+          onClick: undefined,
+          'aria-disabled': !!disabled
+        } as const)
       : ({ type: 'button' } as const)
```
thomasdigby, line 182: We may need to add a conditional for `onClick` when it's a button as well, if the user clicks the `ActionIcon` when it's disabled would the `onClick` still fire? `<ActionIcon onClick={() => console.log('yolo')} disabled />`

alexandrunemes: no it wouldn't. I think it's native disabled button handling. I've tested this exact scenario.

thomasdigby (approved):

andoulla (approved):

PR #198, opened 2021-08-03
Adjust action icon disabled state colours

alexandrunemes:
Adjusts ActionIcon's disabled state colours.
Screenshot 2021-08-03 at 15 40 36

### DIFF
--- a/src/components/action-icon/ActionIcon.tsx
+++ b/src/components/action-icon/ActionIcon.tsx
@@ -14,7 +14,7 @@ const getSimpleVariant = (color: string) => ({
   '&:hover, &:focus': { color }
 })
 const getSolidVariant = (color: string) => ({
-  bg: '$tonal200',
+  bg: '$tonal100',
   color: '$tonal700',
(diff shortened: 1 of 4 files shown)


andoulla (approved):

thomasdigby:
Apologies, I know I said "I'm happy with whatever you decide to do", but I think people would struggle to know which icons here are disabled. One row is disabled, the other is not.

Screenshot 2021-08-03 at 13 58 40

thomasdigby:
I also think there's a bigger issue here about the variants for the action icon and the way we utilise our theme. Should we ever default to a grey icon/bg/border? It seems like if it should be clickable then it should be `$primary` coloured or one of the stateful colours?

andoulla:
> Apologies, I know I said "I'm happy with whatever you decide to do", but I think people would struggle to know which icons here are disabled. One row is disabled, the other is not.
> 
> Screenshot 2021-08-03 at 13 58 40

Could we have a chat with you and Javier to sort this out as we keep going back and forth?

PR #199, opened 2021-08-03
Action icon / Button variants

thomasdigby:
- Removed "subtle" variant
- Updated theme variants to always show a theme color rather than grey
- Updated disabled state to be consistent across appearances
- Added a white background to each variant to improve legibility on other background colors

Screenshot 2021-08-03 at 14-31-53 Component lib sandbox

### DIFF
--- a/src/components/action-icon/ActionIcon.tsx
+++ b/src/components/action-icon/ActionIcon.tsx
@@ -8,32 +8,34 @@ import { Override } from '~/utilities'
 
 import { Icon } from '../icon/Icon'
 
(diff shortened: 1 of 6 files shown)


andoulla:
I think we should replicate this for the actual `Button` too. it will keep this consistent

thomasdigby:
@andoulla 
Screenshot 2021-08-03 at 14 57 56

With buttons

alexandrunemes:
Loving this new look!

alexandrunemes (approved):

andoulla (approved):

PR #200, opened 2021-08-04
Checkbox fixes, Adding required prop to RadioField and Checkbox

Jdrazong:
- Fixed checkbox state handling
- Added required prop to RadioGroup and Checkbox
- Changed the logic to retrieve errors in components by using 'get' funciton from 'react-hook-form' (necessary to get errors in a dynamic form when using useFieldArray functionality)


src/components/checkbox-field/CheckboxField.tsx
```diff
@@ -1,49 +1,61 @@
 import * as React from 'react'
-import { useState } from 'react'
 import { Controller, useFormContext } from 'react-hook-form'
 
 import { Checkbox } from '~/components/checkbox'
 import { InlineFieldWrapper } from '~/components/field-wrapper'
 import { ValidationOptions } from '~/components/form'
 import { CSS } from '~/stitches'
+import { get } from 'react-hook-form'
```
hollg, line 8: This can be merged with the import on line `2`.

Jdrazong: fixed, thank you!

src/components/checkbox-field/CheckboxField.tsx
```diff
@@ -9,41 +8,53 @@ import { CSS } from '~/stitches'
...
   name,
+  required = false,
   validation,
   ...remainingProps
 }) => {
   const { control, errors } = useFormContext()
-  const [isChecked, setIsChecked] = useState(false)
 
-  const error = errors[name]
+  const error = get(errors, name)
```
hollg, line 34: Do you have a link to the docs on this `get` function? I can't find it 🤔

thomasdigby:
@Jdrazong Just a reminder to remove `WIP` from the title

src/components/checkbox-field/CheckboxField.tsx
Jdrazong: they have a separate component https://github.com/react-hook-form/error-message that is not included in the base library and the get is used there to retrieve errors properly => for useFieldArray the errors are in a field that is an array of objects so getting them by normal reflection errors[name] won't work. https://github.com/react-hook-form/react-hook-form/blob/master/src/utils/get.ts here is the get function in the utils folder in the 'react-hook-form' lib where you can see that it deconstructs the path if it is nested

src/components/radio-field/RadioGroupField.tsx
```diff
@@ -18,40 +17,51 @@ type RadioGroupFieldProps = {
...
-      )}
-    </Fieldset>
+    <FieldWrapper
+      fieldId={name}
+      label={label}
+      required={required}
+      css={css}
+      error={error?.message}
+    >
+      <Fieldset>
```
thomasdigby, line 47: We'll need to retain the `fieldset` and `legend` as it's necessary for the radio button markup, using `FieldWrapper` will render too many nested `label`s.

I assume we can still retain the fix you made and still revert to this:

```jsx
<Fieldset css={css}>
  <Label as="legend" css={{ p: 0, mb: '$3' }} required={required}>
    {label}
  </Label>
  <Controller>
    {/* etc. */}
  </Controller>
  {error?.message && (
    <ValidationError css={{ mt: '$1' }}>
      {error?.message}
    </ValidationError>
  )}
</Fieldset>
```

src/components/radio-field/RadioGroupField.tsx
```diff
@@ -18,40 +17,51 @@ type RadioGroupFieldProps = {
...
-          </RadioButtonGroup>
-        )}
-      />
-      {errors[name] && (
-        <ValidationError css={{ mt: '$1' }}>
-          {errors[name].message}
-        </ValidationError>
-      )}
-    </Fieldset>
+    <FieldWrapper
```
hollg, line 40: Could you explain the logic to switching using `FieldWrapper` here? It's a while since I've looked at these components so I could be getting something mixed up, but I the `FieldWrapper` is only intended to wrap one input/label pair. 

`fieldset` elements (which is what `<Fieldset>` renders) are described by `legend`s, not `label`s, and I think this change breaks that relationship. (See line `31` of this file before your change, we were rendering a `<Label as="legend>` for the `<Fieldset>`.

src/components/checkbox-field/CheckboxField.tsx
hollg: Interesting! Is `get` not documented anywhere? If it's not documented as being part of the public API then I'd be a little nervous about it disappearing with a future update that isn't necessarily listed as breaking

src/components/radio-field/RadioGroupField.tsx
Jdrazong: Fixed, thank you!

src/components/radio-field/RadioGroupField.tsx
Jdrazong: Fixed, thank you!

src/components/checkbox-field/CheckboxField.tsx
thomasdigby: It is mentioned here, but only in regards to `lodash` `get` https://react-hook-form.com/advanced-usage/#ErrorMessages

Jdrazong: To be on the safe side we could probably just write our own function and use lodash/get for the places I added get now just would need to copy the regex 'react-hook-form' uses to determine if a field is a nested one and use the lodash/get or errors[name] depending on the check. What do you guys think?

Jdrazong: One thing to note the way fields are named in dynamic array in v7 is also changed so that would need to be updated as well.

src/components/checkbox-field/CheckboxField.tsx
```diff
@@ -9,48 +8,55 @@ import { CSS } from '~/stitches'
...
-      rules={{ required: true }}
+      defaultValue={defaultChecked}
+      rules={validation}
       render={({ onChange, value, name: innerName }) => (
         <InlineFieldWrapper
           css={css}
           description={description}
-          error={error}
+          error={error?.message}
           label={label}
```
thomasdigby, line 49: Do we need to add `required` here? I can't see where we're using it

Jdrazong: Fixed, thank you!

src/components/checkbox-field/CheckboxField.tsx
thomasdigby: @Jdrazong @hollg I suggest we pin the current version (or latest in 6.*) of `react-hook-form` and use the `get` they provide. As it's an importable module, if we upgrade in future and it goes missing, it should be quickly flagged by the module resolver in Node as a missing import.

Jdrazong: @thomasdigby @hollg Just searched the discussions page on react-hook-form and found this => https://github.com/react-hook-form/react-hook-form/discussions/5930 The author recommends using get from lodash instead of utils function so probably we shoulnd't count on a continous support. Should I add a custom function with lodash/get for this one or should I pin the react-hook-form version?

Jdrazong: @thomasdigby @hollg Added a getErrors util in a new commit let me know if it's ok or if I should revert it.

src/utilities/functions/getErrors.ts
```diff
@@ -0,0 +1,5 @@
+import get from 'lodash.get'
+
+export const getErrors = <T extends unknown>(obj: T, path: string) => {
```
hollg, line 3: I think writing a custom function for this makes sense, but since one of the arguments will always be the same thing (the errors object), perhaps we could make this a hook, call `useFormContext` in it to get the errors object, and only require the field name to be passed in? (I also think it would be worth using a more specific paremeter name.)

I think this would also remove the need for `<T extends unknown>`

hollg: I would also suggest keeping this file inside the `components/form` directory, since it's so specific to that component

package.json
```diff
@@ -125,6 +125,7 @@
     "@stitches/react": "0.1.7",
     "@types/react": "^17.0.2",
     "invariant": "^2.2.4",
+    "lodash.get": "^4.4.2",
```
thomasdigby, line 128: I'd prefer to utilise a smaller module for this function, something like https://github.com/developit/dlv could work? It's 120b instead of lodash's 1.8kb

src/components/checkbox-field/CheckboxField.tsx
thomasdigby: This is the right way to go 👍

package.json
Jdrazong: @thomasdigby Just tried delve and it doesn't work for nested array unfortunately

src/utilities/functions/getErrors.ts
Jdrazong: @hollg Thank you! Added a hook useFieldError, let me know if it's ok

package.json
thomasdigby: Did you try with dot notation? You can still access array keys via dot notation `key.3.nestedKey`

Jdrazong: That would require parsing the name string differently I think since the docs for react-hook-form v6 are about using fieldArrayObject[index].fieldName vs v7 where fieldArrayObject.index.fieldName is used instead. Should I add parsing for the string and use delve?

src/components/form/useFieldError.ts
```diff
@@ -0,0 +1,25 @@
...
+    if (currentFieldError) {
+      setError(currentFieldError.message)
+    } else {
+      setError(undefined)
+    }
+  }, [currentFieldError])
+
+  return {
+    error
+  }
```
thomasdigby, line 25: Out of curiosity, could we skip a lot of the state setting and useEffect here? Or is it necessary to access the data after the initial render?

```tsx
export const useFieldError = (fieldName: string) => {
  const { errors } = useFormContext()
  const fieldPath = fieldName.split(/[,[\].]+?/).filter(Boolean)

  return {
    error: delve(errors, fieldPath)
  }
}
```

thomasdigby (approved):

PR #202, opened 2021-08-04
Fix: Action icon neutral theme

thomasdigby:
Any thoughts on the name here? We're going to need a "grey" icon with hover state for `PasswordInput` visibility toggle, Dialog Close buttons, etc. etc.

### DIFF
--- a/src/components/dialog/DialogContent.tsx
+++ b/src/components/dialog/DialogContent.tsx
@@ -4,7 +4,8 @@ import * as React from 'react'
 
 import { keyframes, styled } from '~/stitches'
 
-import { Icon } from '../icon'
+import { ActionIcon } from '../action-icon/ActionIcon'
+import { Icon } from '../icon/Icon'
 
 const contentOnScreen = 'translate3d(-50%, -50%, 0)'
 const contentOffScreen = 'translate3d(-50%, 50vh, 0)'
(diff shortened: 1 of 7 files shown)


andoulla:
> Any thoughts on the name here? We're going to need a "grey" icon with hover state for `PasswordInput` visibility toggle, Dialog Close buttons, etc. etc.

`Unremarkable` or `bland` could do? all jokes aside neutral sort of makes more sense as it pairs well with the rest of the names

src/components/dialog/DialogContent.tsx
```diff
@@ -93,9 +81,15 @@ export const DialogContent: React.FC<DialogContentProps> = ({
   <>
     <StyledDialogOverlay />
     <StyledDialogContent size={size} {...remainingProps}>
-      <StyledDialogClose>
+      <ActionIcon
+        as={Close}
+        css={{ position: 'absolute', right: '$4', top: '$4' }}
+        label="Close dialog"
```
andoulla, line 87: not a big fan of hardcoded values we can't overwrite? Shall we use this as a default but give the option to pass a translation down to it?

andoulla (approved):

src/components/dialog/DialogContent.tsx
thomasdigby: Ah yes I knew there was something you'd pick up but couldn't remember what it was 😆 Agree! I'll add a prop for this

andoulla: It is such a silly little thing but it eventually adds app. there is another instance of it too, I just did the first one

alexandrunemes (approved):

PR #212, opened 2021-08-12
Feature: Upgrade to Stitches 1.0

thomasdigby:
Upgrade to Stitches 1.0 https://github.com/modulz/stitches/releases/tag/v1.0.0 🍾 

- Handled all changes to the API
- Upgraded TS and related TS plugins, fixed any TS issues with the new version
- Patched the CSSOM (JSDOM layer) in our tests (so snapshots work mostly as expected)

Some side-effects:
- All tokens were previously prefixed with `sx` (this was a bug on the specific version we're currently on) so that prefix has now been removed in all custom properties
- All snapshots have been updated and styles are now nested under `@media` due to a quirk in how JSDOM handles the CSSOM
- Due to this CSSOM issue we can't test for styles in the DOM or use RTL methods like `toBeVisible()`, have updated some impacted tests with snapshots instead


src/components/button/__snapshots__/Button.test.tsx.snap
```diff
@@ -1,175 +1,160 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Button component Loading state renders a loading button 1`] = `
-@keyframes sxwho2l {
-  0% {
-    opacity: 0;
+@media  {
```
andoulla, line 4: how  come we no longer have the @keyframes into the snapshot? Do they still work for sure ?

andoulla: btw it is very confusing that everything is wrapped in @media. Is this a  permanent thing?

src/components/stack-content/StackContent.test.tsx
```diff
@@ -10,7 +10,7 @@ import { StackContent } from '.'
 
 describe('StackContent component', () => {
   it('applies styles correctly to children', async () => {
```
andoulla, line 12: tiny nitpick: can we also chance this description now that we don't check for styles? "renders a stack" or something similar should be more than enough

src/stitches.ts
```diff
@@ -1,81 +1,72 @@
...
+import { createStitches } from '@stitches/react'
 
-type CSSValue = number | string
-type CSSBlob = { [key: string]: CSSValue }
-
-// TODO: assess how intuitive the team finds these
 export const utils = {
-  p: () => (value: CSSValue): CSSBlob => ({
+  bg: (value: Stitches.PropertyValue<'backgroundColor'>) => ({
+    backgroundColor: value
```
andoulla, line 8: why switch this to only set the colour? Personally, I would expect a `bg` shorthand to work just like the `background` prop

src/stitches.ts
```diff
@@ -1,81 +1,72 @@
...
+
+  inset: (value: Stitches.ScaleValue<'space'> | number | string) => ({
+    top: value,
+    right: value,
+    bottom: value,
+    left: value
+  }),
+
+  size: (value: Stitches.ScaleValue<'size'> | number | string) => ({
+    height: value,
```
andoulla, line 19: is there a reason that the option to pass an array of values is removed?

src/components/button/__snapshots__/Button.test.tsx.snap
thomasdigby: Good catch on the keyframes, seems like they're missing from the snapshots.

Unfortunately we can't do much about this output (either missing keyframes or the wrapping `@media`) as the style insertion is controlled entirely by stitches; and even then we're having to patch our testing layer to make sure that it works with JSDOM.

I'll raise a bug on the repo when I'm back from 🌴

src/stitches.ts
thomasdigby: We almost exclusively use `background` for color in the main codebase (literally 1 out of ~280 is something other than a color (`linear-gradient`)), so for me this is more about defining a util that will be useful and typed appropriately for that use.

Let me see if `Stitches.PropertyValue<'background'>` returns similarly useful values

src/stitches.ts
thomasdigby: Just for clarity, I think declaring `height: 10px; width: 15px;` is clearer than trying to parse `size: [15px, 10px]`, and as we're currently not using it anywhere it's safe to remove.

src/stitches.ts
thomasdigby: It does!

PR #221, opened 2021-09-28
Feature: Date input

thomasdigby:
## Changes
- Introduces 2 new (user facing) components, `DateInput` and `DateField` for selecting a single date. Powered by this neat little package [`Dayzed`](https://github.com/deseretdigital/dayzed) which provides the data and functionality but leaves all rendering to us
- Allows the user to type in a date to an input, as well as opening a popover interface to select a date from a series of months. The video hopefully communicates how the component works better than a description will...
- Allows the developer to pass in an initial selected date, the format with which to parse the user inputted date, any further props can be used to configure the `dayzed` instance which controls the render of the calendar itself
- Updated `Popover` to not render the close button by default, added a prop to enable developers to add this if need be. Tightened up the styling of the popover itself.
- Removed the ability to pass `css` into `Popover.Root`

https://user-images.githubusercontent.com/3226804/137987203-62e2cf5c-f0a6-4f89-adbc-fa8d6f97f7f4.mov


shaneakins:
It's most likely an edge case, but if we're using the `sm` variant of the `InputField` (which has `sm` and `md` variants) and we also need the date picker in that field group, there's currently not a defined `sm` variant for it. I see you're passing on the remaining props; however, here's what I'm seeing when I currently try to use the `size` prop on the `DateInput` component

Screen Shot 2021-12-13 at 14 36 37

andoulla:
> It's most likely an edge case, but if we're using the `sm` variant of the `InputField` (which has `sm` and `md` variants) and we also need the date picker in that field group, there's currently not a defined `sm` variant for it. I see you're passing on the remaining props; however, here's what I'm seeing when I currently try to use the `size` prop on the `DateInput` component
> 
> Screen Shot 2021-12-13 at 14 36 37

I had something similar with the SearchInput, I would try and copy the logic from there.

src/components/calendar/Calendar.tsx
```diff
@@ -0,0 +1,135 @@
...
+  refDateSelected,
+  refDateToday,
+  firstDayOfWeek = 0,
+  monthNames = monthNamesShort,
+  weekdayNames = weekdayNamesShort,
+  ...remainingProps
+}) => {
+  const { calendars, getBackProps, getForwardProps, getDateProps } = useDayzed({
+    firstDayOfWeek,
+    showOutsideDays: true,
```
andoulla, line 56: what's showOutsideDays?

src/components/calendar/Calendar.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  if (!calendars.length) return null
+
+  return (
+    <Box
+      css={{ position: 'relative', width: 'min-content', mt: '-$1', ...css }}
+    >
+      <Flex css={{ position: 'absolute', top: 0, right: '-$1' }}>
+        <ActionIcon
+          label="Previous month"
```
andoulla, line 68: Wondering if we should provide a way of overriding these with a translated copy. Want to avoid the situation where everything else is translated for accessibility readers and the label isn't

src/components/popover/PopoverContent.tsx
```diff
@@ -16,10 +16,9 @@ import { Icon } from '../icon/Icon'
 const StyledContent = styled(Content, {
   bg: 'white',
   borderRadius: '$1',
-  boxShadow: '$2',
+  boxShadow: '$1',
   maxWidth: '90vw',
-  p: '$sizes$2',
-  pr: '$6',
+  p: '$sizes$1',
```
andoulla, line 21: How come we are altering the popover?

src/components/calendar/Calendar.tsx
kerryambt: showing the days outside of the current month that appear on the calendar grid, rather than just having that as whitespace.

src/components/calendar/Calendar.tsx
kerryambt: yeah good point, added :)

src/components/popover/PopoverContent.tsx
kerryambt: this was on the branch before I started on it, didn't notice :o but I think some of these changes are a little odd so I've brought them back more inline with what's currently on main.

src/components/calendar/Calendar.tsx
andoulla: could we please add this to the documentation. that it defaults to true

kerryambt: sure, will do :)

andoulla (approved):

src/components/calendar/Calendar.tsx
```diff
@@ -0,0 +1,137 @@
...
+          label={labels.next}
+          theme="neutral"
+          size="lg"
+          {...getForwardProps({ calendars })}
+        >
+          <Icon is={ChevronRight} />
+        </ActionIcon>
+      </Flex>
+      {calendars.map(({ month, year, weeks }) => (
+        <div key={`${month}${year}`}>
```
Mhoog, line 87: Should be fine to use `<Box />` here instead, right?

kerryambt: oh yeah :o I missed this, was there when I picked up the work. changed :)

src/components/calendar/Calendar.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+Date.now = jest.fn(() => Date.parse('2021-12-14'))
+
+const props = {
+  onDateSelected: jest.fn(),
+  date: new Date('12/25/21')
+}
+
+describe('Calendar component', () => {
+  it('renders', async () => {
+    const { container } = render(<Calendar {...props} />)
```
avirati, line 19: Suggestion - 
`const { container } = render(<Calendar {...props} />)` can be extracted into `beforeEach` (and removed from subsequent `it()`s), and references for `container` and `rerender` can be saved. Use `rerender` when you need to pass on extra props

kerryambt: appreciate the suggestion but I personally think this is over abstraction. I think each test should, unless the logic being repeated was HUGE, have its render obvious and self contained. happy to bow to an overall consensus but I'm not a fan of this level of abstraction, I think it removes readability. def appreciate the suggestion though, it's good to discuss things :)

avirati: Thanks for the explanation, appreciate it !! Please feel free to resolve :)

avirati (approved):

andoulla (approved):

src/components/date-field/DateField.tsx
```diff
@@ -0,0 +1,52 @@
...
+      fieldId={name}
+      label={label}
+      prompt={prompt}
+      required={Boolean(validation?.required)}
+    >
+      <DateInput
+        id={name}
+        name={name}
+        ref={ref}
+        {...(error && { state: 'error' })}
```
thomasdigby, line 45: There's an issue here, when selecting a date with the calendar popover, it doesn't clear the error state from the field. Not sure what to suggest as a solution, might need to dig into `react-hook-form`s error handling and figure out how to trigger a "re-validate" on command

kerryambt: https://user-images.githubusercontent.com/66493855/148233008-2aca59aa-94ce-4a98-8426-5038ebfa877b.mov

fixed this @thomasdigby :)

PR #222, opened 2021-10-01
Fix: Video prop override

avirati:
Issue : 
`<Video />` component had a fixed `100%` height and width. It was not possible to override that from outside

Fix : 
Moved `{...remainingProps}` towards the end to allow prop overriding

### DIFF
--- a/src/components/video/Video.tsx
+++ b/src/components/video/Video.tsx
@@ -36,12 +36,12 @@ export const Video: React.FC<VideoProps> = ({
       }}
     >
       <StyledVideo
-        {...remainingProps}
         role="figure"
         url={`https://player.vimeo.com/video/${id}`}
         height="100%"
         width="100%"
         css={{ position: 'absolute', top: 0, left: 0 }}
+        {...remainingProps}
       />
     </Box>
   </CSSWrapper>


hollg:
I think the intent with the original positioning of `{...remainingProps}` was to avoid the non-css props from being overridden. Could we maintain that behaviour while allowing you to override the CSS props you need to by spreading the incoming `css` prop into the end of the css prop passed into `StyledVideo`?

hollg:
@avirati has now pointed out that the `height` and `width` are being set as `HTML` attributes, not just in the `css` prop (d'oh!)

I think we probably _should_ allow the user to override all the `ReactPlayer` props because they're part of our component's API. Having a TS type that includes them but not doing anything with them seems pretty misleading now that I think about it.

@thomasdigby @andoulla do you know of any specific gotchas we need to avoid here, or reasons that we'd have decided not to tel the user override the `height` and `width`?

@avirati finally, it might be worth holding off on merging this until we hear from Jess about the colour of the play icon. If we do want to update the default then we could just do that in this one PR after you've got the code working in the app?

avirati:
@hollg No worries. I'll wait for things to get a bit more clear. Meanwhile, I will proceed with my original task

thomasdigby:
@hollg @avirati I think the `height` and `width` props are being set (and not overridden) as they're required to render the Video at the correct ratio. You can see that the parent container has a `height: 0` and `paddingTop` that reflects the required ratio of the video, if we allow overriding/removal of the height/width attributes then the Video will not fill the parent and it won't render as expected.

@avirati What's the original issue that prompted this fix?

avirati:
@thomasdigby I was trying to render a video and wanted a smaller size as per design specs. I found out that it was not possible to control the size of video player either via props, or via CSS properties. The props `height` and `width` are passed on to ReactPlayer which is what fixes the size of video player.

Reference : 
https://github.com/cookpete/react-player#props

thomasdigby:
@avirati Could you not pass in a width/max/min-width to the component? e.g. `<Video css={{ width: '100px' }} />`.

That way you should retain the correct ratio, while also controlling the width of the video

avirati:
@thomasdigby You are right !! I think I tried to override `height` and not `width`. My bad, should have checked both :(

avirati:
Closing the MR since this change is not needed to achieve the mentioned goal

PR #226, opened 2021-10-05
Feat/upgrade radix

hollg:
- [X] Upgrade all Radix packages to version `0.1.0` to solve a problem where competing versions of `@radix-ui/react-dismissable-layer` prevent pointer events from registering correctly
- [X] Replace all instances of `@radix-ui/react-slot` with the new `asChild` prop (https://www.radix-ui.com/docs/primitives/overview/styling#changing-the-rendered-element)
- [x] Fix failing tests

These upgrades deprecates the `as` prop, which we were using along with `Slot` to control how we rendered our trigger components for the `AlertDialog`, `Dialog` and `Popover`. We can now get the default element (a `<button>` in all three cases) by passing no props to the trigger components and rendering the relevant text directly as children, or we can use the new `asChild` prop to render one of our own components as the trigger.

```tsx
const OptionOne = () => <Popover.Trigger>click me</Popover.Trigger>

// the Button's onClick is automatically set by the Trigger
const OptionTwo = () => <Popover.Trigger asChild><Button>click me</Button></PopoverTrigger>
```


src/components/tooltip/TooltipTrigger.tsx
```diff
@@ -8,7 +7,7 @@ export const TooltipTrigger: React.FC<TooltipTriggerProps> = ({
   children,
   ...remainingProps
 }) => (
-  <Trigger as={Slot} {...remainingProps}>
+  <Trigger asChild {...remainingProps}>
     {children}
   </Trigger>
```
thomasdigby, line 12: Stalking before this is finished (sorry) but we could reduce this to `<Trigger {...props} />` as per the `AlertDialogTrigger`

hollg: I think we can actually remove these components entirely and just re-export the components from Radix directly, if we're happy to pass `asChild` from the app when we need to.

thomasdigby: Yeah lovely 👌 I wonder why we didn't do that in the first place?

hollg: I think it was to avoid needing to either re-export `Slot` or have `@radix/react-slot` as a dependency everywhere we use this package. But obviously we don't have that issue since they 🗑️ed it

src/components/checkbox-field/CheckboxField.tsx
```diff
@@ -51,9 +51,7 @@ export const CheckboxField: React.FC<CheckboxFieldProps> = ({
             defaultValue={defaultChecked ? CheckboxValue.ON : CheckboxValue.OFF}
             checked={value}
             name={innerName}
-            onCheckedChange={(event) => {
-              onChange(event.target.value !== CheckboxValue.ON)
-            }}
+            onCheckedChange={onChange}
```
hollg, line 54: The `onCheckedChange` API has been updated and now only has the new state passed as an argument. This is equivalent to what we want to pass to the `onChange` from `react-hook-form`, so this has been simplified.

src/components/alert-dialog/alert-context/AlertDialog.tsx
```diff
@@ -26,7 +26,6 @@ export const Alert: React.FC<AlertDialogContentProps> = ({
   <AlertDialog defaultOpen>
     <AlertDialog.Content
       onEscapeKeyDown={(e) => e.preventDefault()}
-      onPointerDownOutside={(e) => e.preventDefault()}
```
hollg, line 29: @thomasdigby this prop doesn't exist any more. Do you know what we were using it to do?

src/components/tooltip/TooltipTrigger.tsx
```diff
@@ -7,8 +6,4 @@ type TooltipTriggerProps = React.ComponentProps<typeof Trigger>
 export const TooltipTrigger: React.FC<TooltipTriggerProps> = ({
   children,
   ...remainingProps
-}) => (
-  <Trigger as={Slot} {...remainingProps}>
-    {children}
-  </Trigger>
-)
```
thomasdigby, line 14: We can probably kill this component as well if we replace its reference with `Trigger` directly from Radix

src/components/checkbox/Checkbox.tsx
```diff
@@ -6,6 +6,12 @@ import { styled } from '~/stitches'
 
 import { Icon } from '../icon'
 
+const StyledIndicator = styled(RadixCheckbox.Indicator, {
+  display: 'flex',
+  alignItems: 'center',
+  justifyContent: 'center'
+})
```
thomasdigby, line 13: You may be able to remove this exact CSS from `StyledCheckbox` as it's likely to not have any effect now

src/components/tooltip/TooltipTrigger.tsx
hollg: Great catch, thanks! Updated.

src/components/checkbox/Checkbox.tsx
thomasdigby: Huh, I can't see why it should aside from maybe the `display` property, but all good let's keep it 👍

thomasdigby (approved):
There are a few snapshots failing I think, otherwise good to go 👍

PR #232, opened 2021-10-11
chore: added coming soon proposal

andoulla:
Here is a proposal on how this list could look. We could add more fields like whether the designs are ready or the component exists, happy to hear suggestions.

Please comment and let me know of any components we currently know that need to be added, as this is just a small sample to see the structure

image


thomasdigby:
It would be great to have it in a format that we can consume within the documentation site and show in the menu, any thoughts on how we could do this?

PR #253, opened 2021-11-10
feat: Accordion Component

kerryambt:
Adding an Accordion component because it is needed for a ticket. The Accordion is taken from Radix and has been styled to fit the design system.

image

https://user-images.githubusercontent.com/66493855/141107043-b49e84bf-aaa7-4ee9-b851-b8afb6902025.mov

### DIFF
--- a/src/components/accordion/Accordion.tsx
+++ b/src/components/accordion/Accordion.tsx
@@ -1,0 +1,32 @@
+import { Root } from '@radix-ui/react-accordion'
+import { AccordionItem } from './AccordionItem'
+import { AccordionTrigger } from './AccordionTrigger'
+import { AccordionContent } from './AccordionContent'
+
+import React from 'react'
+
+import { styled } from '~/stitches'
+
+const StyledRoot = styled(Root, {
+  width: '100%'
+})
+
+type AccordionProps = React.ComponentProps<typeof StyledRoot> & {
+  type: 'single'
+}
+
+export const Accordion: React.FC<AccordionProps> & {
+  Item: typeof AccordionItem
+  Content: typeof AccordionContent
+  Trigger: typeof AccordionTrigger
+} = ({ type = 'single', children, ...remainingProps }) => (
+  <StyledRoot type={type} {...remainingProps}>
+    {children}
+  </StyledRoot>
+)
+
+Accordion.Item = AccordionItem
+Accordion.Content = AccordionContent
+Accordion.Trigger = AccordionTrigger
+
+Accordion.displayName = 'Accordion'
--- a/src/components/accordion/Accordion.tsx
+++ b/src/components/accordion/Accordion.tsx
@@ -11,6 +11,8 @@ const StyledRoot = styled(Root, {
   width: '100%'
 })
 
+// TODO: this isn't perfect, trying to find a fix from creators,
+// will hopefully fix in another PR in the future
 type AccordionProps = React.ComponentProps<typeof StyledRoot> & {
   type: 'single'
 }
--- a/src/components/accordion/AccordionContent.tsx
+++ b/src/components/accordion/AccordionContent.tsx
@@ -1,0 +1,36 @@
+import { Content } from '@radix-ui/react-accordion'
+import React from 'react'
+
+import { keyframes, styled } from '~/stitches'
+
+const open = keyframes({
+  from: { height: 0 },
+  to: { height: 'var(--radix-accordion-content-height)' }
+})
+
+const close = keyframes({
+  from: { height: 'var(--radix-accordion-content-height)' },
+  to: { height: 0 }
+})
+
+const StyledContent = styled(Content, {
+  border: 0,
+  width: '100%',
+  bg: '#fff',
+
+  '@allowMotion': {
+    '&[data-state="open"]': { animation: `${open} 300ms ease-out` },
+    '&[data-state="closed"]': { animation: `${close} 300ms ease-out` }
+  }
+})
+
+type AccordionContentProps = React.ComponentProps<typeof StyledContent>
+
+export const AccordionContent: React.FC<AccordionContentProps> = ({
+  children,
+  ...remainingProps
+}) => (
+  <StyledContent title="accordionContent" {...remainingProps}>
+    {children}
+  </StyledContent>
+)
--- a/src/components/accordion/AccordionItem.tsx
+++ b/src/components/accordion/AccordionItem.tsx
@@ -1,0 +1,32 @@
+import { Item } from '@radix-ui/react-accordion'
+import React from 'react'
+
+import { styled } from '~/stitches'
+
+const StyledItem = styled(Item, {
+  width: '100%',
+  overflow: 'hidden',
+
+  '&:not(:last-child)': {
+    mb: '$1'
+  },
+  '&[data-state="open"]': {
+    '& [name="accordionTrigger"]': { borderRadius: '$0 $0 0 0' },
+    '& [title="accordionContent"]': { borderRadius: '0 0 $0 $0' }
+  },
+  '&[data-state="closed"]': {
+    '& [name="accordionTrigger"]': { borderRadius: '$0' }
+  }
+})
+
+type AccordionItemProps = React.ComponentProps<typeof StyledItem>
+
+export const AccordionItem: React.FC<AccordionItemProps> = ({
+  children,
+  value,
+  ...remainingProps
+}) => (
+  <StyledItem value={value} {...remainingProps}>
+    {children}
+  </StyledItem>
+)
--- a/src/components/accordion/AccordionTrigger.tsx
+++ b/src/components/accordion/AccordionTrigger.tsx
@@ -89,7 +89,7 @@ const StyledTrigger = styled(Trigger, {
         '#fff',
         darken(0.1, '#fff'),
         darken(0.15, '#fff'),
-        '$000'
+        '#000'
       )
     }
   ]
@@ -98,18 +98,17 @@ const StyledTrigger = styled(Trigger, {
 type AccordionTriggerProps = Override<
   React.ComponentProps<typeof StyledTrigger>,
   Stitches.VariantProps<typeof StyledTrigger> & {
-    text?: string
     theme: string
   }
 >
 
 export const AccordionTrigger: React.FC<AccordionTriggerProps> = ({
-  text,
   theme = 'primaryDark',
+  children,
   ...remainingProps
 }) => (
   <StyledTrigger theme={theme} name="accordionTrigger" {...remainingProps}>
-    <Box>{text}</Box>
+    <Box>{children}</Box>
     <Icon is={ChevronDown} />
   </StyledTrigger>
 )
--- a/src/components/accordion/AccordionTrigger.tsx
+++ b/src/components/accordion/AccordionTrigger.tsx
@@ -1,0 +1,117 @@
+import { Trigger } from '@radix-ui/react-accordion'
+import React from 'react'
+import type * as Stitches from '@stitches/react'
+
+import { Box, Icon } from '~/index'
+import { ChevronDown } from '@atom-learning/icons'
+
+import { styled, theme } from '~/stitches'
+import { darken } from 'polished'
+import { Override } from '~/utilities'
+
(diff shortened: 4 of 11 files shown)


src/components/accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+const StyledItem = styled(Item, {
+  width: '100%',
+  overflow: 'hidden',
+
+  '&:not(:last-child)': {
+    mb: '$1'
+  },
+  '&[data-state="open"]': {
+    '& [name="accordionTrigger"]': { borderRadius: '$0 $0 0 0' },
```
shaneakins, line 14: You can reduce this to just `borderRadius: 0`

src/components/accordion/AccordionTrigger.tsx
```diff
@@ -0,0 +1,117 @@
...
+        darken(0.15, theme.colors.tonal600.value)
+      )
+    },
+    {
+      theme: 'light',
+      css: getTriggerVariant(
+        '#fff',
+        darken(0.1, '#fff'),
+        darken(0.15, '#fff'),
+        '$000'
```
shaneakins, line 92: Should this be `'#000'`?

src/components/accordion/AccordionTrigger.tsx
```diff
@@ -0,0 +1,117 @@
...
+type AccordionTriggerProps = Override<
+  React.ComponentProps<typeof StyledTrigger>,
+  Stitches.VariantProps<typeof StyledTrigger> & {
+    text?: string
+    theme: string
+  }
+>
+
+export const AccordionTrigger: React.FC<AccordionTriggerProps> = ({
+  text,
```
hollg, line 107: Would it be possible to render this component's `children` rather than specifying a `text` prop? I think that including a chevron here is the right thing to do, but if it's possible to keep that in place while allowing arbitrary `children` to be rendered, I think that flexibility would be nice & more in keeping with our component composition model.

src/components/accordion/Accordion.tsx
```diff
@@ -0,0 +1,32 @@
...
+import React from 'react'
+
+import { styled } from '~/stitches'
+
+const StyledRoot = styled(Root, {
+  width: '100%'
+})
+
+type AccordionProps = React.ComponentProps<typeof StyledRoot> & {
+  type: 'single'
```
hollg, line 15: What's the reason for adding `type` here? It's already on the prop type for `Root` (and by extension `StyledRoot`), isn't it?

src/components/accordion/AccordionTrigger.tsx
```diff
@@ -0,0 +1,117 @@
...
+  border: 0,
+  py: '$3',
+  px: '$4',
+  width: '100%',
+  display: 'flex',
+  justifyContent: 'space-between',
+  alignItems: 'center',
+  cursor: 'pointer',
+
+  '&[data-state="open"]': {
```
hollg, line 43: These animations a great idea. Could you post a screencap in the PR description so that we can see them in action?

src/components/accordion/AccordionContent.tsx
```diff
@@ -0,0 +1,36 @@
...
+  from: { height: 'var(--radix-accordion-content-height)' },
+  to: { height: 0 }
+})
+
+const StyledContent = styled(Content, {
+  border: 0,
+  width: '100%',
+  bg: '#fff',
+
+  '@allowMotion': {
```
hollg, line 21: 🔥

hollg (changes requested):
This is cool! Just a couple of v minor requests

src/components/accordion/AccordionItem.tsx
shaneakins: Ahh gotcha - thanks for clarifying

src/components/accordion/AccordionContent.tsx
kerryambt: this is a good thing right? XD

src/components/accordion/AccordionTrigger.tsx
kerryambt: will do :)

src/components/accordion/Accordion.tsx
kerryambt: because if you don't it throws an error. spent a good while on this and this seems to be the only way it'll not throw an error. it seems to be confused with the way accordion has multiple interfaces that are passed in dynamically based on what value you set for type. if you leave it out, you have the define the type, which I didn't want as it made using it a whole more fiddly. If you put this in you can specify single or multiple as a type and it works - if it's overriding something so typescript doesn't have a grump then it seems like the best solution imo.

src/components/accordion/AccordionTrigger.tsx
kerryambt: oops! yes. well spotted.

src/components/accordion/AccordionTrigger.tsx
kerryambt: if you think that's best then sure. I'll do that now.

src/components/accordion/Accordion.tsx
hollg: Weird! This line is specifying that is must be `'single'` though, isn't it? Whereas we want it to be _either_ `'single'`, `'multiple'`, or whatever else is valid.

kerryambt: isn't this line saying there must be a property called type, of type string, and single is the default? either way I thought initially what you thought and it doesn't work without throwing an error. it's a weird one.

hollg: I think you want `type: 'single' | 'multiple' | 'whateverElse'`

TypeScript is quite happy to accept `'single'` as a _type_, which only matches other strings with the exact same value. It's pretty weird imo. But with the `|` version (`|` is basically a logical "OR") and your default assignment in the component prop later on, I think you should get the result you want!

hollg: Also, at runtime all this TS has been compiled into JS, so all these TS-specific features like type definitions don't exist in the code at that point. One implication of this is that we can't use a type definition to specify a default value.

kerryambt: tried 'single' | 'multiple' and it doesn't work. setting it to 'single' allows you to still specify 'multiple' or 'single' on the component and it works. I know how that sounds, but, yeah. happy to show you on slack if you want :)

src/components/accordion/AccordionTrigger.tsx
thomasdigby: I'd agree with @hollg here, we want to be flexible with our composition model so allowing a user to add an additional `Icon` or `Divider` or any other inline component would be really beneficial 👍

andoulla:
For this to be usable in the admin, it might need to have the option to open a different section without closing the current open section

src/components/accordion/AccordionTrigger.tsx
```diff
@@ -0,0 +1,117 @@
...
+    theme: string
+  }
+>
+
+export const AccordionTrigger: React.FC<AccordionTriggerProps> = ({
+  text,
+  theme = 'primaryDark',
+  ...remainingProps
+}) => (
+  <StyledTrigger theme={theme} name="accordionTrigger" {...remainingProps}>
```
thomasdigby, line 110: Do we need this `name` prop? I can't find it in the [Radix docs](https://www.radix-ui.com/docs/primitives/components/accordion) and it's generic so assume the aim would be to overwrite it if it _was_ necessary?

src/components/accordion/AccordionTrigger.tsx
```diff
@@ -0,0 +1,117 @@
...
+    {
+      theme: 'light',
+      css: getTriggerVariant(
+        '#fff',
+        darken(0.1, '#fff'),
+        darken(0.15, '#fff'),
+        '$000'
+      )
+    }
+  ]
```
thomasdigby, line 95: We can avoid these `compoundVariants` altogether as we're not using a combination of variants to style this component https://stitches.dev/docs/variants#compound-variants

```ts
variants: {
  theme: {
    primary: getTriggerVariant('$primary', '$primaryMid', '$primaryDark'),
    primaryDark: getTriggerVariant(
      '$primaryDark',
      darken(0.1, theme.colors.primaryDark.value),
      darken(0.15, theme.colors.primaryDark.value)
    ),
    dark: getTriggerVariant(
      '$tonal600',
      darken(0.1, theme.colors.tonal600.value),
      darken(0.15, theme.colors.tonal600.value)
    ),
    light: getTriggerVariant(
      '#fff',
      darken(0.1, '#fff'),
      darken(0.15, '#fff'),
      '#000'
    )
  }
}
```

However, I also wonder if each of these themes are necessary? Do we have a need for `dark` or `primaryDark` at the moment? I have no issue with them if they are, but I'm really keen to avoid pre-optimising or introducing potential themes if they're not used, it makes it trickier to change the API later on

src/components/accordion/AccordionTrigger.tsx
```diff
@@ -0,0 +1,117 @@
...
+})
+
+type AccordionTriggerProps = Override<
+  React.ComponentProps<typeof StyledTrigger>,
+  Stitches.VariantProps<typeof StyledTrigger> & {
+    text?: string
+    theme: string
+  }
+>
```
thomasdigby, line 103: I know you had a discussion with @hollg on the types, but we _should_ be able to reduce this to `type AccordionTriggerProps = React.ComponentProps<typeof StyledTrigger>` as the `theme` will be made available from `typeof StyledTrigger` anyway, and I assume we're moving away from `text: string` as well

src/components/accordion/AccordionTrigger.tsx
```diff
@@ -0,0 +1,117 @@
...
+  theme = 'primaryDark',
+  ...remainingProps
+}) => (
+  <StyledTrigger theme={theme} name="accordionTrigger" {...remainingProps}>
+    <Box>{text}</Box>
+    <Icon is={ChevronDown} />
+  </StyledTrigger>
+)
+
+AccordionTrigger.toString = () => 'accordionTrigger'
```
thomasdigby, line 116: What's the reason for this `toString` addition here?

hollg (approved):

src/components/accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,32 @@
...
+    '& [title="accordionContent"]': { borderRadius: '0 0 $0 $0' }
+  },
+  '&[data-state="closed"]': {
```
thomasdigby, line 17: Can we move these styles to the components that use them? e.g.

```ts
const StyledTrigger = styled(Trigger, {
  '[data-state="closed"] &': { borderRadius: '$0' }
})
```

thomasdigby (commented):
A few comments but looking good overall 👍

src/components/accordion/AccordionTrigger.tsx
kerryambt: that should have been removed, sorry! removing now.

src/components/accordion/AccordionTrigger.tsx
kerryambt: this was due to how I had some CSS classes on a parent component and needed to reference a specific child, but I have now moved the CSS down so this is redundant and will be removed.

src/components/accordion/AccordionTrigger.tsx
kerryambt: this does indeed now work - thanks :)

src/components/accordion/AccordionItem.tsx
kerryambt: good point - done

src/components/accordion/AccordionTrigger.tsx
kerryambt: primary dark is what was requested by design, and I think light would be really helpful. should we then remove dark, and also remove primary leaving primary dark in place? I don't think design need a primary at the moment.

thomasdigby: Yeah I think that's worth doing, let's remove both primary and dark for now. Easy to add them in if we need em!

PR #254, opened 2021-11-10
Fix/accordion component fixes

kerryambt:
Making some changes to the Accordion component that were missed on the last PR.

### DIFF
--- a/src/components/accordion/Accordion.mdx
+++ b/src/components/accordion/Accordion.mdx
@@ -11,23 +11,21 @@ The `Accordion.Trigger` has been simplified to include a chevron icon. Generally
 
 Default styling has been applied to `Accordion.Trigger`, but `Accordion.Content` is an empty container without styling. Should only text be placed inside, it is highly advisable to apply spacing to align with the styling of the rest of the Accordion. This can be done with either the `css` property, or by placing any other components inside the `Accordion.Content`.
 
-NOTE: spacing should not be applied to the `Accordion.Content`, and should instead be applied to child components. Otherwise, the animation will not display correctly.
-
 You can Read more about the underlying UI component on the [Radix UI documentation site](https://radix-ui.com/primitives/docs/components/accordion).
 
 ```tsx preview
 <Accordion type="single" defaultValue="1">
   <Accordion.Item value="1">
     <Accordion.Trigger>Accordion Header 1</Accordion.Trigger>
-    <Accordion.Content>
-      <Text css={{ p: '$2', height: '100px' }}>Some content here</Text>
+    <Accordion.Content css={{ p: '$2', height: '100px' }}>
+      I'm styled on my container
     </Accordion.Content>
   </Accordion.Item>
   <Accordion.Item value="2">
     <Accordion.Trigger>Accordion Header 2</Accordion.Trigger>
     <Accordion.Content>
       <Text css={{ p: '$2', height: '150px' }}>
-        Slightly larger content here
+        I'm styled on a child component
       </Text>
     </Accordion.Content>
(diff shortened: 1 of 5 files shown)


src/components/accordion/AccordionTrigger.tsx
```diff
@@ -55,62 +55,30 @@ const StyledTrigger = styled(Trigger, {
...
-    },
-    {
-      theme: 'light',
-      css: getTriggerVariant(
+      ),
+      light: getTriggerVariant(
         '#fff',
         darken(0.1, '#fff'),
         darken(0.15, '#fff'),
         '#000'
```
thomasdigby, line 67: We should always use our `$tonal` values, rather than `black` or `#000`. `$tonal600` is our darkest and is used for foreground text so that would be a good replacement

thomasdigby:
@kerryambt I also noticed a slight jump when closing/opening the accordion content, can you share the video again in this PR? It might be related to the internal padding on `AccordionContent`, we may need to apply the padding another level down so the height can be animated without padding affecting it

src/components/accordion/AccordionTrigger.tsx
kerryambt: ah okay - thanks for clarifying :)

src/components/accordion/AccordionContent.tsx
```diff
@@ -29,8 +30,4 @@ type AccordionContentProps = React.ComponentProps<typeof StyledContent>
 export const AccordionContent: React.FC<AccordionContentProps> = ({
   children,
   ...remainingProps
-}) => (
-  <StyledContent title="accordionContent" {...remainingProps}>
-    {children}
-  </StyledContent>
-)
+}) => <StyledContent {...remainingProps}>{children}</StyledContent>
```
thomasdigby, line 33: I think this is actually a good chance to use our `CSSWrapper` utility, it'll apply styles if provided, and if not just render the children. That way we can still have the same `<AccordionContent css={{ p: '$2' }}>` model (which works quite well), but avoid the issue of applying padding to the main content wrapper.

```tsx
export const AccordionContent: React.FC<AccordionContentProps> = ({
  children,
  css,
  ...remainingProps
}) => (
  <StyledContent {...remainingProps}>
    <CSSWrapper css={css}>
      {children}
    </CSSWrapper>
  </StyledContent>
)
```

kerryambt: have seen this but hadn't used it to know how to use it - very useful! have changed over now :)

src/components/accordion/Accordion.mdx
```diff
@@ -24,7 +24,9 @@ You can Read more about the underlying UI component on the [Radix UI documentati
   <Accordion.Item value="2">
     <Accordion.Trigger>Accordion Header 2</Accordion.Trigger>
     <Accordion.Content>
-      <Text css={{ p: '$2', height: '150px' }}>I'm using child components</Text>
+      <Text css={{ p: '$2', height: '150px' }}>
```
thomasdigby, line 27: Can we apply these attributes (although probably not height) to `AccordionContent` now we've changed the API

thomasdigby (approved):

hollg:
There's a circular dependency in the Accordion components because `AccordionTrigger.tsx` imports something from an index file at too high a level (should be fixed if you import other components directly from their own directories).

src/components/accordion/Accordion.mdx
kerryambt: sure - done!

thomasdigby (approved):

hollg (approved):

PR #255, opened 2021-11-11
Patch: more table styles

hollg:
Updates table styles after some changes to designs and some resolved miscommunications.

Screenshot 2021-11-11 at 08 18 18

### DIFF
--- a/src/components/table/Table.tsx
+++ b/src/components/table/Table.tsx
@@ -31,16 +31,10 @@ const StyledTable = styled('table', {
       md: {
         [`${TableCell}, ${TableHeaderCell}, ${TableFooterCell}`]: {
           height: '$4'
-        },
-        [`${TableHeaderCell}`]: {
-          height: '$4'
         }
       },
       lg: {
         [`${TableCell}, ${TableHeaderCell}, ${TableFooterCell}`]: {
-          height: '$5'
-        },
-        [`${TableHeaderCell}`]: {
           height: '$5'
         }
       }
--- a/src/components/table/Table.tsx
+++ b/src/components/table/Table.tsx
@@ -22,8 +22,6 @@ type TableSubComponents = {
 
 const StyledTable = styled('table', {
   borderCollapse: 'separate',
-  borderTopLeftRadius: '$0',
-  borderTopRightRadius: '$0',
   borderSpacing: 0,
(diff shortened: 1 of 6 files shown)


thomasdigby:
Remember to commit using the semantic message style, none of these commits will be picked up by semantic release and it won't publish a new version. We can still merge and use @kerryambt's update to publish your changes

src/components/table/Table.tsx
```diff
@@ -21,22 +21,25 @@ type TableSubComponents = {
...
   variants: {
     size: {
       md: {
-        [`${TableCell}`]: {
+        [`${TableCell}, ${TableHeaderCell}`]: {
           height: '$4'
         },
         [`${TableHeaderCell}`]: {
           height: '$4'
         }
```
thomasdigby, line 36: We can remove this extra selector/style here

thomasdigby (approved):

hollg:
> Remember to commit using the semantic message style, none of these commits will be picked up by semantic release and it won't publish a new version. We can still merge and use @kerryambt's update to publish your changes

Really? I started a commit with `patch` but the precommit hook stopped me because `patch` isn't a valid commit type. It gave a list of valid types which included `style` so that's what I used. Is that precommit hook not aligned with what the semantic-release package?

hollg:
@thomasdigby I've added an empty `fix` commit so I think it should get picked up now

Mhoog (approved):

thomasdigby:
> > Remember to commit using the semantic message style, none of these commits will be picked up by semantic release and it won't publish a new version. We can still merge and use @kerryambt's update to publish your changes
> 
> 
> 
> Really? I started a commit with `patch` but the precommit hook stopped me because `patch` isn't a valid commit type. It gave a list of valid types which included `style` so that's what I used. Is that precommit hook not aligned with what the semantic-release package?

Ah apologies, might have jumped the gun there. I would double check that it's aligned with semantic release so we're sure that it'll trigger a release correctly, I'm not sure what the overlap is.

andoulla:
> @thomasdigby I've added an empty `fix` commit so I think it should get picked up now

We have a commit linter that shouldn't allow commits that don't follow the semantic format. I remember setting it up and it used to prevent most mistakes.

PR #257, opened 2021-11-15
Feature: Slider

thomasdigby:
Screenshot 2021-11-15 at 22 18 57


src/components/slider-field/SliderField.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+describe('SliderField component', () => {
+  it('renders', async () => {
+    const { container } = render(<ExampleSliderField />)
+
+    expect(container).toMatchSnapshot()
+  })
+
+  //TODO: figure out how to pass aria-label properly so that there are no a11y issues
+  it.skip('has no programmatically detectable a11y issues', async () => {
```
andoulla, line 22: what are the issues that ally has reported

src/components/slider/Slider.test.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+describe('Slider component', () => {
+  it('renders', async () => {
+    const { container } = render(<Slider defaultValue={[50]} />)
+
+    expect(container).toMatchSnapshot()
+  })
+
+  //TODO: figure out how to pass aria-label properly so that there are no a11y issues
+  it.skip('has no programmatically detectable a11y issues', async () => {
```
andoulla, line 15: If the error is that it needs a label, we can wrap this component with a Label (there should be a couple of examples like that

src/components/slider/Slider.tsx
```diff
@@ -0,0 +1,89 @@
...
+  '&[data-orientation="horizontal"]': {
+    height: '$0'
+  },
+  '&[data-orientation="vertical"]': {
+    flexDirection: 'column',
+    width: '$0'
+  },
+  '&[data-disabled]': { cursor: 'not-allowed' },
+  variants: {
+    border: {
```
thomasdigby, line 30: I wouldn't take my initial styling work too seriously here, is there a need for both bordered and solid variants of the slider? I'd stick to a simple, single design if we can, do we have a design reference for it in Figma?

src/components/slider/Slider.test.tsx
kerryambt: yeah I tried that as I had a look at some other examples here. for some reason a11y wants the label to be on the moveable part of the slider? I could only solve it by passing label as a prop and putting it onto the SliderThumb, which didn't seem correct :o

src/components/slider-field/SliderField.test.tsx
kerryambt: as in comment below, error is that it needs a label. didn't seem right to drill label down when it seemed like that wasn't the correct place for it anyway. figured flagging it as a tag would get some eyes on it at this stage :)

src/components/slider/Slider.test.tsx
thomasdigby: Yeah, I remember this issue, the slider does adhere to the W3C accessibility guidelines so I'm not sure what to recommend with this test... https://www.radix-ui.com/docs/primitives/components/slider#accessibility

src/components/slider/Slider.tsx
```diff
@@ -0,0 +1,89 @@
...
+      border={border}
+      defaultValue={defaultValue}
+      value={value}
+      {...remainingProps}
+    >
+      <StyledTrack>
+        <StyledRange />
+      </StyledTrack>
+      {thumbs?.length && thumbs.map((_, i) => <StyledThumb key={i} />)}
+    </StyledSlider>
```
thomasdigby, line 85: Screenshot 2021-12-06 at 20 08 47

How do you think we should handle these inline value markers? I'd be keen to include them as part of the slider/slider-field but we need to understand how to space them, link them to slider values, e.g. steps

src/components/slider/Slider.tsx
thomasdigby: Yeah definitely worth checking in with design for this, I suspect that varying background colours might be preferable instead. The `border: 1px inset` thing renders pretty inconsistently cross-browser anyway and there's no other way to run a border only around the part of the track that isn't selected...

src/components/slider/index.ts
```diff
@@ -0,0 +1,2 @@
+export { Slider } from './Slider'
+export type { SliderProps } from './Slider'
```
thomasdigby, line 2: It would be great to be less explicit about the exports here, `export * from './Slider'` will handle all exports from `'./Slider'`

src/components/slider/Slider.test.tsx
kerryambt: yeah me either :o

src/components/slider/Slider.tsx
kerryambt: just waiting to hear back from design but I've tested with `tonal200` and it looks a lot nicer than the border.

kerryambt: image

src/components/slider/index.ts
kerryambt: good point!

src/components/slider/Slider.tsx
kerryambt: my initial thought is passing an array of objects to the slider something like:
```
[
  { value: 0, label: 'min' }
  { value: 50, label: 'half' },
  { value: 100, label: 'max' }
]
```

Then if the array isn't empty render a box under the slider and position these as absolutely positioned text components, using the value as a percentage?

I guess following that you could add in a conditional bold when the value on the slider is the value selected. Furthermore we could have an optional 'snap' where you pass a number (or it goes to a default of say 5) and if the value of the slider is +/- snap value from a point in the array, you make the value be that number instead? 

I'll have a play around and see what I can achieve :)

src/components/slider/Slider.tsx
thomasdigby: Yep, that looks good, although I'd potentially look at reducing the height of the track, maybe to `$space$1` (0.25rem). It could help to tighten it up a bit, especially when used alongside other form elements. You could test this in the components view of the sandbox, which I think you can access from `http://localhost:1234/components`

src/components/slider/Slider.tsx
thomasdigby: Yep, I would open up for feedback on what API might be best here, it would need to work along the sliders `step` prop (https://www.radix-ui.com/docs/primitives/components/slider#root).

You should also consider how the spacing would work when the first and last items are aligned with the limits of the track as opposed to the actual value itself... not an easy one and might be worth seeing what design think as well!

kerryambt: image

image

https://user-images.githubusercontent.com/66493855/145034088-824a7f45-48d3-46ec-bbf8-ef4bf49d06d2.mov

So I came up with this. It works with the step property to snap where needed, spaces properly, and aligns the beginning and end labels correctly. Also correctly accounts for min and max being different from the default 0-100. Let me know what you think :) 

I'm waiting to hear back from design but I'll drop them some screenshots of this and get an opinion there too.

src/components/slider/Slider.tsx
kerryambt: yeah that does look a little cleaner. changed now :)
image

src/components/slider/Slider.tsx
thomasdigby: Woah nice, that's looking really great. Let me dig into it and feedback now

src/components/slider/Slider.tsx
```diff
@@ -0,0 +1,141 @@
...
+                  max
+                )}%, 0)`,
+                color: '$tonal300'
+              }}
+            >
+              {step.label}
+            </Text>
+          ))}
+        </Box>
+      )}
```
thomasdigby, line 136: I'd look at breaking this out into its own file, `SliderSteps.tsx` or similar. That way you can encapsulate the logic and rendering for this particular part of the slider into a single view.

src/components/slider/Slider.tsx
```diff
@@ -0,0 +1,141 @@
...
+              key={`sliderStep${step.label}`}
+              css={{
+                position: 'absolute',
+                top: '$3',
+                left: `${getPercentValue(step.value, min, max)}%`,
+                transform: `translate(-${getTransformValue(
+                  step.value,
+                  min,
+                  max
+                )}%, 0)`,
```
thomasdigby, line 128: I'd change a few things here to optimise the CSS.

- For these dynamic properties, I would add them as `style` instead of `css`. That way we avoid creating a new class for each new label, inline style is perfect for these properties.
- We should also try to use the container to position the elements vertically, we can lose `top` completely if we add an `mt` to the container instead.
- I think we could also use the `step.value` as the key here, it's more likely to be unique than the label (although it's unlikely we'll ever run into this)
- We should render the `Text` as a `span` as `p` isn't suitable for this text 

```tsx
<Text
  as="span"
  key={step.value}
  style={{
    left: `${getPercentValue(step.value, min, max)}%`,
    transform: `translateX(-${getTransformValue(step.value, min, max)}%)`
  }}
  css={{ position: 'absolute', color: '$tonal300' }}
>
  {step.label}
</Text>
```

src/components/slider/Slider.tsx
```diff
@@ -0,0 +1,141 @@
...
+        {...remainingProps}
+      >
+        <StyledTrack>
+          <StyledRange />
+        </StyledTrack>
+        {thumbs?.length && thumbs.map((_, i) => <StyledThumb key={i} />)}
+      </StyledSlider>
+
+      {steps?.length > 0 && (
+        <Box css={{ width: '100%', position: 'relative', mb: '$6' }}>
```
thomasdigby, line 116: As I mentioned in the previous comment, we could tweak this to position it relative to the `Slider` rather than relying on the `top` value for each label. For example, (I pulled this out into its own styled component), by using an `mt` we can position all labels at once, if we also use `height` (as position absolute takes each label out of the flow) we can also account for the height relative to other sibling components.

```tsx
export const SliderSteps = styled('div', {
  height: '$space$3',
  mt: '$2',
  position: 'relative',
  width: '100%'
})
```

Screenshot 2021-12-07 at 20 14 50

src/components/slider/Slider.tsx
```diff
@@ -0,0 +1,141 @@
...
+  value,
+  defaultValue,
+  min = 0,
+  max = 100,
+  theme = 'dark',
+  ...remainingProps
+}) => {
+  const thumbs = value || defaultValue
+  return (
+    <>
```
thomasdigby, line 100: If we change this wrapping element to `CSSWrapper` then we can pass `css` into the component to handle spacing contextually. e.g. `<Slider css={{ mb: '$4' }} />`

src/components/slider/Slider.tsx
```diff
@@ -0,0 +1,141 @@
...
+})
+
+const StyledSlider = styled(Root, {
+  alignItems: 'center',
+  display: 'flex',
+  position: 'relative',
+  touchAction: 'none',
+  userSelect: 'none',
+  '&[data-orientation="horizontal"]': {
+    height: '$0'
```
thomasdigby, line 24: If we bump this up to `$1` then it covers the height of the entire slider, which will be beneficial if ever need to add external spacing. It might also be worth adding a `cursor: 'pointer'` on the `StyledSlider` as this whole wrapper is interactive, it also helps with the UX when selecting a value.

Screenshot 2021-12-07 at 20 19 38

src/components/slider/Slider.mdx
```diff
@@ -0,0 +1,62 @@
...
+The steps work well with the built in `step` property, which defaults to 1 and changes the size of each movement. For example, this would limit the slider to three values only.
+
+```tsx preview
+const steps = [
+    { value: 10, label: "min" },
+    { value: 15, label: "mid" },
+    { value: 20, label: "max" }
+]
+
+<Slider steps={steps} step={5} min={10} max={20} defaultValue={[15]} />
```
thomasdigby, line 53:
```tsx
<Slider
  steps={[
    { value: 10, label: "min" },
    { value: 15, label: "mid" },
    { value: 20, label: "max" }
  ]}
  step={5}
  min={10}
  max={20}
  defaultValue={[15]}
/>
```

We'll need to pass these values directly in the component in our code snippets as the plugin rendering them only expects JSX components

src/components/slider/Slider.tsx
```diff
@@ -0,0 +1,141 @@
...
+    flexDirection: 'column',
+    width: '$0'
+  },
+  '&[data-disabled]': { cursor: 'not-allowed' },
+  variants: {
+    theme: {
+      light: {
+        [`${StyledTrack}`]: {
+          bg: '#fff',
+          '&[data-disabled]': { bg: '$tonal100' }
```
thomasdigby, line 36: We can include this in the root `&[data-disabled]` selector as it's duplicated across both themes

src/components/slider/Slider.tsx
```diff
@@ -0,0 +1,141 @@
...
+        }
+      },
+      dark: {
```
thomasdigby, line 39: I'm keen for us to look at our theme names holistically across the library, but in the short-term `dark` doesn't sit right with me for some reason, I'm wondering if we could use `tonal` or `neutral` or something like that? Both of those suggestions aren't _great_ but I think `dark` implies something that it's not and may get confused with "dark mode" where the entire colour scheme of the component gets flipped

thomasdigby (commented):
Really like the slider step label solution, just need some refinement then I'm happy to get this in 🙌

src/components/slider/Slider.tsx
kerryambt: yeah I wasn't sold on dark as I agree it makes one think of dark mode. I've gone with tonal but I agree a holistic conversation to come up with some conventions would be helpful :)

src/components/slider/Slider.tsx
kerryambt: I feel like I tried this but it didn't didn't work, must have done something wrong as it does now. changed.

src/components/slider/Slider.mdx
kerryambt: ah right, that's good to know, thanks

src/components/slider/Slider.tsx
kerryambt: yeah good point :)

src/components/slider/Slider.tsx
kerryambt: this seemed to be working before, but it seems better to use `CSSWrapper` than a JSX fragment so I'll change this :)

src/components/slider/Slider.tsx
kerryambt: yeah this is a better way of handling this. changed.

src/components/slider/Slider.tsx
kerryambt: thankyou for the explanation `style` vs `css` here, wasn't entirely sure what the difference was as I'd used style a lot in previous work. changed :)

src/components/slider/Slider.tsx
kerryambt: much cleaner this way, good point

src/components/slider/SliderSteps.tsx
```diff
@@ -0,0 +1,70 @@
...
+    value: number,
+    min: number,
+    max: number
+  ): number => {
+    const percentage = getPercentValue(value, min, max)
+
+    if (percentage <= 10) return 0
+    if (percentage >= 90) return 100
+    return 50
+  }
```
thomasdigby, line 42: I would pull these functions out of the component so we're not redeclaring them with each render. They have separate arguments anyway so you should be able to move them out and above `SliderSteps` without issue

src/components/slider/Slider.test.tsx
```diff
@@ -0,0 +1,20 @@
...
+    const { container } = render(<Slider defaultValue={[50]} />)
+
+    expect(await axe(container)).toHaveNoViolations()
+  })
+})
```
thomasdigby, line 20: Can we add a test with these slider steps?

src/components/slider/Slider.test.tsx
thomasdigby: Let's keep `it.skip()` here and add a comment above about why we're skipping

thomasdigby:
Looks good to me aside from a few remaining bits, let's get some other reviews and then merge 👍

src/components/slider/Slider.test.tsx
thomasdigby: Snapshot at least, so we can see and compare the output HTML

src/components/slider/Slider.test.tsx
kerryambt: will do

src/components/slider/Slider.test.tsx
kerryambt: done :)

src/components/slider/SliderSteps.tsx
kerryambt: yeah good point, done :)

thomasdigby (commented):
I can't approve as I committed code as part of this PR, but good to go from my point of view

andoulla (approved):

Mhoog (approved):

PR #259, opened 2021-11-18
Fix: Tabs mobile view

avirati:
### Tasks Undertaken
1. Based on total width of tab container and screen size, show `<` & `>` buttons to allow easy scrolling of tabs.
2. Scroll amount is at 25% of tab container width
3. Conditionally show / hide scroll buttons
4. Bring the components up to new design - supporting two variations

```typescript
<Tabs appearance="light"></Tabs>
<Tabs appearance="dark"></Tabs>
```

#### Desktop
https://user-images.githubusercontent.com/1936119/142402905-7f749c94-1704-4724-b0a5-5ae5db754703.mov

#### Mobile
https://user-images.githubusercontent.com/1936119/142402968-9b3edec9-949a-44e0-a218-59de1c583e55.mov

### DIFF
--- a/package.json
+++ b/package.json
@@ -148,6 +148,7 @@
     "react-hook-form-persist": "^2.0.0",
     "react-hot-toast": "^1.0.2",
     "react-player": "^2.9.0",
+    "throttle-debounce": "^3.0.1",
     "uid": "^2.0.0"
   },
   "husky": {
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "eslint-plugin-simple-import-sort": "^7.0.0",
     "glob": "^7.1.6",
     "husky": "^4.3.8",
-    "jest": "^26.6.3",
+    "jest": "^27.0.6",
     "jest-axe": "^4.1.0",
     "jest-dom": "^4.0.0",
     "lint-staged": "^10.5.3",
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "eslint-plugin-simple-import-sort": "^7.0.0",
     "glob": "^7.1.6",
     "husky": "^4.3.8",
-    "jest": "^27.0.6",
+    "jest": "^26.6.3",
     "jest-axe": "^4.1.0",
     "jest-dom": "^4.0.0",
     "lint-staged": "^10.5.3",
--- a/src/components/tabs/TabTrigger.tsx
+++ b/src/components/tabs/TabTrigger.tsx
@@ -19,7 +19,8 @@ const StyledTabTrigger = styled(Trigger, {
         bg: 'white',
         '&[data-state="active"]': {
           color: '$primary',
-          textShadow: '0px 0px 1px currentColor',
+          fontWeight: 600,
+          letterSpacing: '-0.005em',
           boxShadow: 'inset 0 -2px 0 0 currentColor'
         },
         '&[data-state="inactive"]': {
@@ -42,7 +43,8 @@ const StyledTabTrigger = styled(Trigger, {
       dark: {
         color: 'white',
         '&[data-state="active"]': {
-          textShadow: '0px 0px 1px currentColor',
+          fontWeight: 600,
+          letterSpacing: '-0.005em',
           boxShadow: 'inset 0 -2px 0 0 currentColor'
         },
         '&:not([data-disabled]):hover': {
--- a/src/components/tabs/TabTrigger.tsx
+++ b/src/components/tabs/TabTrigger.tsx
@@ -24,18 +24,14 @@ const StyledTabTrigger = styled(Trigger, {
         '&[data-state="inactive"]': {
           color: '$tonal500'
         },
-        '&:hover': {
+        '&:not([data-disabled]):hover': {
           color: '$primary',
           bg: opacify(-0.9, theme.colors.primary.value)
         },
-        '&:active': {
+        '&:not([data-disabled]):active': {
           color: '$primary',
           bg: opacify(-0.8, theme.colors.primary.value),
           boxShadow: 'none'
-        },
-        '&:focus': {
-          color: '$primary',
-          boxShadow: 'inset 0 0 0 2px currentColor'
         },
         '&[data-disabled],&[data-disabled]:hover': {
           color: '$tonal200',
@@ -48,10 +44,10 @@ const StyledTabTrigger = styled(Trigger, {
           textShadow: '0px 0px 1px currentColor',
           boxShadow: 'inset 0 -2px 0 0 currentColor'
         },
-        '&:hover': {
+        '&:not([data-disabled]):hover': {
           bg: opacify(-0.8, 'white')
         },
-        '&:active': {
+        '&:not([data-disabled]):active': {
           bg: opacify(-0.7, 'white'),
           boxShadow: 'none'
         },
--- a/src/components/tabs/TabTrigger.tsx
+++ b/src/components/tabs/TabTrigger.tsx
@@ -10,21 +10,29 @@ const StyledTabTrigger = styled(Trigger, {
   fontFamily: '$body',
   p: '$4',
   userSelect: 'none',
-  transition: '0.3s'
+  transition: '0.3s',
+  variants: {
+    theme: {
+      light: {},
+      dark: {}
+    }
+  }
 })
 
-type TabTriggerProps = React.ComponentProps<typeof StyledTabTrigger> & {
+interface TabTriggerProps
+  extends React.ComponentProps<typeof StyledTabTrigger> {
   value: string
   disabled?: boolean
 }
 
 export const TabTrigger: React.FC<TabTriggerProps> = ({
   children,
+  theme,
   disabled = false,
   css,
   ...otherProps
 }) => (
-  <StyledTabTrigger css={css} disabled={disabled} {...otherProps}>
+  <StyledTabTrigger css={css} disabled={disabled} theme={theme} {...otherProps}>
     {children}
   </StyledTabTrigger>
 )
--- a/src/components/tabs/TabTrigger.tsx
+++ b/src/components/tabs/TabTrigger.tsx
@@ -1,7 +1,8 @@
 import { Trigger } from '@radix-ui/react-tabs'
 import * as React from 'react'
+import { opacify } from 'polished'
 
-import { styled } from '~/stitches'
+import { styled, theme } from '~/stitches'
 
 const StyledTabTrigger = styled(Trigger, {
   textTransform: 'uppercase',
@@ -13,8 +14,55 @@ const StyledTabTrigger = styled(Trigger, {
   transition: '0.3s',
   variants: {
     theme: {
-      light: {},
-      dark: {}
+      light: {
+        bg: 'white',
+        '&[data-state="active"]': {
+          color: '$primary',
+          textShadow: '0px 0px 1px currentColor',
+          boxShadow: 'inset 0 -2px 0 0 currentColor'
+        },
+        '&[data-state="inactive"]': {
+          color: '$tonal500'
+        },
+        '&:hover': {
+          color: '$primary',
+          bg: opacify(-0.9, theme.colors.primary.value)
+        },
+        '&:active': {
+          color: '$primary',
+          bg: opacify(-0.8, theme.colors.primary.value),
+          boxShadow: 'none'
+        },
+        '&:focus': {
+          color: '$primary',
+          boxShadow: 'inset 0 0 0 2px currentColor'
+        },
+        '&[data-disabled],&[data-disabled]:hover': {
+          color: '$tonal200',
+          cursor: 'default'
+        }
+      },
+      dark: {
+        color: 'white',
+        '&[data-state="active"]': {
+          textShadow: '0px 0px 1px currentColor',
+          boxShadow: 'inset 0 -2px 0 0 currentColor'
+        },
+        '&:hover': {
+          bg: opacify(-0.8, 'white')
+        },
+        '&:active': {
+          bg: opacify(-0.7, 'white'),
+          boxShadow: 'none'
+        },
+        '&:focus': {
+          boxShadow: 'inset 0 0 0 2px currentColor'
+        },
+        '&[data-disabled],&[data-disabled]:hover': {
+          color: '$tonal200',
+          cursor: 'default'
+        }
+      }
     }
   }
 })
--- a/src/components/tabs/TabTrigger.tsx
+++ b/src/components/tabs/TabTrigger.tsx
@@ -4,24 +4,13 @@ import * as React from 'react'
 import { styled } from '~/stitches'
 
 const StyledTabTrigger = styled(Trigger, {
-  color: '$secondary',
+  textTransform: 'uppercase',
   cursor: 'pointer',
   flexShrink: 0,
   fontFamily: '$body',
   p: '$4',
   userSelect: 'none',
-  '&:hover': {
-    color: '$primary'
-  },
-  '&[data-disabled],&[data-disabled]:hover': {
-    color: '$tonal300',
-    cursor: 'default'
-  },
-  '&[data-state="active"]': {
-    color: '$primary',
-    fontWeight: 'bold',
-    boxShadow: 'inset 0 -3px 0 0 currentColor'
-  }
+  transition: '0.5s'
 })
 
 type TabTriggerProps = React.ComponentProps<typeof StyledTabTrigger> & {
--- a/src/components/tabs/TabTrigger.tsx
+++ b/src/components/tabs/TabTrigger.tsx
@@ -71,10 +71,9 @@ export const TabTrigger: React.FC<TabTriggerProps> = ({
   children,
   theme,
   disabled = false,
-  css,
   ...otherProps
 }) => (
-  <StyledTabTrigger css={css} disabled={disabled} theme={theme} {...otherProps}>
+  <StyledTabTrigger disabled={disabled} theme={theme} {...otherProps}>
     {children}
   </StyledTabTrigger>
 )
--- a/src/components/tabs/TabTrigger.tsx
+++ b/src/components/tabs/TabTrigger.tsx
@@ -10,7 +10,7 @@ const StyledTabTrigger = styled(Trigger, {
   fontFamily: '$body',
   p: '$4',
   userSelect: 'none',
-  transition: '0.5s'
+  transition: '0.3s'
 })
 
 type TabTriggerProps = React.ComponentProps<typeof StyledTabTrigger> & {
--- a/src/components/tabs/TabTrigger.tsx
+++ b/src/components/tabs/TabTrigger.tsx
@@ -35,7 +35,7 @@ const StyledTabTrigger = styled(Trigger, {
         },
         '&[data-disabled],&[data-disabled]:hover': {
           color: '$tonal200',
-          cursor: 'default'
+          cursor: 'not-allowed'
         }
       },
       dark: {
@@ -51,12 +51,9 @@ const StyledTabTrigger = styled(Trigger, {
           bg: opacify(-0.7, 'white'),
           boxShadow: 'none'
         },
-        '&:focus': {
-          boxShadow: 'inset 0 0 0 2px currentColor'
-        },
         '&[data-disabled],&[data-disabled]:hover': {
           color: '$tonal200',
-          cursor: 'default'
+          cursor: 'not-allowed'
         }
       }
     }
--- a/src/components/tabs/TabTrigger.tsx
+++ b/src/components/tabs/TabTrigger.tsx
@@ -10,6 +10,7 @@ const StyledTabTrigger = styled(Trigger, {
   flexShrink: 0,
   fontFamily: '$body',
   p: '$4',
+  height: '$5',
   userSelect: 'none',
   transition: '0.3s',
   variants: {
--- a/src/components/tabs/Tabs.mdx
+++ b/src/components/tabs/Tabs.mdx
@@ -33,11 +33,11 @@ It takes a `defaultValue` prop that should match one of the triggers' `value` pr
 <Tabs defaultValue="tab1">
 ```
 
-It also takes a `appearance` prop that should either be "light" or "dark".
+It also takes a `theme` prop that should either be "light" or "dark".
 
 ```tsx
-<Tabs appearance="light"></Tabs>
-<Tabs appearance="dark"></Tabs>
+<Tabs theme="light"></Tabs>
+<Tabs theme="dark"></Tabs>
 ```
 
 ## Tabs.TriggerList
--- a/src/components/tabs/Tabs.mdx
+++ b/src/components/tabs/Tabs.mdx
@@ -27,10 +27,17 @@ The component provides a set of default styles, which can be overwritten by usin
 </Tabs>
 ```
 
-It also takes a `defaultValue` prop that should match one of the triggers' `value` prop.
+It takes a `defaultValue` prop that should match one of the triggers' `value` prop.
 
 ```tsx
 <Tabs defaultValue="tab1">
+```
+
+It also takes a `appearance` prop that should either be "light" or "dark".
+
+```tsx
+<Tabs appearance="light"></Tabs>
+<Tabs appearance="dark"></Tabs>
 ```
 
 ## Tabs.TriggerList
--- a/src/components/tabs/Tabs.mdx
+++ b/src/components/tabs/Tabs.mdx
@@ -36,6 +36,7 @@ It also takes a `defaultValue` prop that should match one of the triggers' `valu
 ## Tabs.TriggerList
 
 The `Tabs.TriggerList` component simply holds the individual `Tabs.Trigger` components. It can also get custom styling via the `css` prop.
+`Tabs.TriggerList` will automatically show `<` & `>` buttons in case the content is overshooting the available space.
 
 ## Tabs.Trigger
 
--- a/src/components/tabs/Tabs.test.tsx
+++ b/src/components/tabs/Tabs.test.tsx
@@ -19,9 +19,32 @@ const TabsTest = ({ defaultValue = 'tab1' }) => (
   </IdProvider>
 )
 
+const MobileTabsTest = ({ defaultValue = 'tab1' }) => (
+  <IdProvider>
+    <Tabs defaultValue={defaultValue}>
+      <Tabs.TriggerList enableTabScrolling>
+        <Tabs.Trigger value="tab1">
+          Trigger 1 which is going to be long
+        </Tabs.Trigger>
+        <Tabs.Trigger value="tab2">
+          Trigger 2 which is going to be even longer
+        </Tabs.Trigger>
+      </Tabs.TriggerList>
+      <Tabs.Content value="tab1">Important content for tab 1</Tabs.Content>
+      <Tabs.Content value="tab2">Important content for tab 2</Tabs.Content>
+    </Tabs>
+  </IdProvider>
+)
+
 describe('Tabs component', () => {
   it('renders', async () => {
     const { container } = render(<TabsTest />)
+
+    expect(container).toMatchSnapshot()
+  })
+
+  it('renders mobile view', async () => {
+    const { container } = render(<MobileTabsTest />)
 
     expect(container).toMatchSnapshot()
   })
--- a/src/components/tabs/Tabs.tsx
+++ b/src/components/tabs/Tabs.tsx
@@ -15,7 +15,7 @@ const StyledRoot = styled(Root, {
   pt: '$4',
   px: '$2',
   variants: {
-    appearance: {
+    theme: {
       light: {
         'div[role="tab"]': {
           bg: 'white',
@@ -108,8 +108,8 @@ export const Tabs: React.FC<TabsProps> & {
   TriggerList: typeof TriggerListWrapper
   Trigger: typeof TabTrigger
   Content: typeof StyledTabContent
-} = ({ appearance = 'light', ...remainingProps }) => (
-  <StyledRoot appearance={appearance} {...remainingProps} />
+} = ({ theme = 'light', ...remainingProps }) => (
+  <StyledRoot theme={theme} {...remainingProps} />
 )
 
 Tabs.TriggerList = TriggerListWrapper
--- a/src/components/tabs/Tabs.tsx
+++ b/src/components/tabs/Tabs.tsx
@@ -54,7 +54,7 @@ const StyledRoot = styled(Root, {
           borderBottom: '1px solid $primaryDark'
         },
         'button[role="scrollbar"]': {
-          bg: 'white !important',
+          bg: 'white',
           opacity: 0.8
         }
       },
@@ -90,10 +90,10 @@ const StyledRoot = styled(Root, {
         'div[role="tablist"]': {
           borderBottom: '1px solid white'
         },
-        'button[role="scrollbar"]': {
-          bg: '$primaryDark !important',
+        'button[role="scrollbar"], button[role="scrollbar"]:focus': {
+          bg: '$primaryDark',
           opacity: 0.8,
-          color: 'white !important'
+          color: 'white'
         }
       }
     }
--- a/src/components/tabs/Tabs.tsx
+++ b/src/components/tabs/Tabs.tsx
@@ -12,7 +12,6 @@ type TabsProps = React.ComponentProps<typeof StyledRoot>
 const StyledRoot = styled(Root, {
   display: 'flex',
   flexDirection: 'column',
-  pt: '$4',
   variants: {
     theme: {
       light: {
--- a/src/components/tabs/Tabs.tsx
+++ b/src/components/tabs/Tabs.tsx
@@ -1,148 +1,115 @@
-import { Content, List, Root } from '@radix-ui/react-tabs'
-import React, { useState, useEffect, useRef, useCallback } from 'react'
-import { ChevronLeft, ChevronRight } from '@atom-learning/icons'
+import { Content, Root } from '@radix-ui/react-tabs'
+import React from 'react'
+import { opacify } from 'polished'
 
-import { styled } from '~/stitches'
-import { ActionIcon } from '~/components/action-icon'
-import { Flex } from '~/components/flex'
-import { Icon } from '~/components/icon'
+import { styled, theme } from '~/stitches'
 
 import { TabTrigger } from './TabTrigger'
+import { TriggerListWrapper } from './TabsTriggerList'
 
 type TabsProps = React.ComponentProps<typeof StyledRoot>
 
 const StyledRoot = styled(Root, {
   display: 'flex',
-  flexDirection: 'column'
+  flexDirection: 'column',
+  pt: '$4',
+  px: '$2',
+  variants: {
+    appearance: {
+      light: {
+        'div[role="tab"]': {
+          bg: 'white',
+          '&[data-state="active"]': {
+            color: '$primary',
+            fontWeight: 600,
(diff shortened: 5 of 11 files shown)


src/components/tabs/TabsTriggerList.tsx
```diff
@@ -0,0 +1,121 @@
...
+}) => {
+  const triggerListRef = useRef<HTMLDivElement>(null)
+  const [showLeftScroller, setShowLeftScroller] = useState<boolean>(false)
+  const [showRightScroller, setShowRightScroller] = useState<boolean>(false)
+
+  const scrollTriggerListTo = useCallback((direction: 'left' | 'right') => {
+    const triggerList = triggerListRef.current
+    if (triggerList) {
+      const { scrollWidth, scrollLeft } = triggerList
+      const scrollAmount = Math.round(scrollWidth / 4) // 25% of whole scroll width
```
hollg, line 37: How was `25%` decided on? I'm just wondering whether it will be possible that on some screen widths, not all tabs would be accessible because 25% would scroll you too far past them

src/components/tabs/TabsTriggerList.tsx
```diff
@@ -0,0 +1,121 @@
...
+      if (direction === 'right') {
```
hollg, line 38: Could this two-level nested `if` be avoided if we did `if (direction === 'right' && triggerList.scrollLeft + scrollAmount <= scrollWidth)` here?

src/components/tabs/TabsTriggerList.tsx
avirati: That's a great point. Here is the slack discussion [thread](https://atom-learning.slack.com/archives/G01G926F9DM/p1637147852018100?thread_ts=1637147252.013600&cid=G01G926F9DM)

src/components/tabs/TabsTriggerList.tsx
avirati: I think not, because we also have an else block inside, which requires `triggerList` to be truthy. Let me know if I am misunderstanding something

src/components/tabs/TabsTriggerList.tsx
```diff
@@ -0,0 +1,121 @@
...
+        if (triggerList.scrollLeft + scrollAmount <= scrollWidth) {
+          triggerList.scroll({
+            left: triggerList.scrollLeft + scrollAmount,
+            behavior: 'smooth'
+          })
+        }
+      } else {
+        triggerList.scroll({
+          left: scrollLeft - scrollAmount > 0 ? scrollLeft - scrollAmount : 0,
```
hollg, line 47: If we have two conditional branches where we call the same function (`triggerList.scroll`) and just pass different values in, I'd be keen to simplify the code so that there's only one call. E.g.

```
const leftValue = someCondition ? someAmount : someOtherAmount
triggerList.scroll({ left: leftValue, behaviour: 'smooth' })
```

Maybe we'd need to use `if` rather than a ternary, to avoid lots of ternary nesting, but I think that still ends up a little easier to follow.

src/components/tabs/TabTrigger.tsx
```diff
@@ -4,24 +4,13 @@ import * as React from 'react'
...
-  '&[data-disabled],&[data-disabled]:hover': {
-    color: '$tonal300',
-    cursor: 'default'
-  },
-  '&[data-state="active"]': {
-    color: '$primary',
-    fontWeight: 'bold',
-    boxShadow: 'inset 0 -3px 0 0 currentColor'
-  }
+  transition: '0.5s'
```
hollg, line 13: Half a second feels quite long to me — could we try something shorter?

src/components/tabs/Tabs.tsx
```diff
@@ -1,36 +1,118 @@
...
 const StyledRoot = styled(Root, {
   display: 'flex',
-  flexDirection: 'column'
+  flexDirection: 'column',
+  pt: '$4',
+  px: '$2',
+  variants: {
+    appearance: {
+      light: {
+        'div[role="tab"]': {
```
thomasdigby, line 20: I understand why you've brought these styles up to the `Tab` component, but I'd rather pass the `appearance` down as a prop to correctly co-locate and encapsulate the styles within each sub-component. e.g.

```tsx
export const Tabs: React.FC<TabsProps> & {
  TriggerList: typeof TriggerListWrapper
  Trigger: typeof TabTrigger
  Content: typeof StyledTabContent
} = ({ appearance = 'light', children, ...remainingProps }) => (
  <StyledRoot appearance={appearance} {...remainingProps}>
    {React.Children.map(children, (child) => {
      // you may need something similar within `TriggerListWrapper` to handle `TabTrigger`
      if (child.type === TriggerList || child.type === StyledTabContent) {
        return React.cloneElement(child, { appearance })
      }
    )}
  </StyledRoot>
)
```

src/components/tabs/Tabs.tsx
```diff
@@ -1,36 +1,118 @@
...
+type TabsProps = React.ComponentProps<typeof StyledRoot>
 
 const StyledRoot = styled(Root, {
   display: 'flex',
-  flexDirection: 'column'
+  flexDirection: 'column',
+  pt: '$4',
+  px: '$2',
+  variants: {
+    appearance: {
```
thomasdigby, line 18: We should use `theme` as the variant name here, for consistency mainly.

src/components/tabs/Tabs.tsx
```diff
@@ -1,36 +1,118 @@
...
+import { TriggerListWrapper } from './TabsTriggerList'
+
+type TabsProps = React.ComponentProps<typeof StyledRoot>
 
 const StyledRoot = styled(Root, {
   display: 'flex',
-  flexDirection: 'column'
+  flexDirection: 'column',
+  pt: '$4',
+  px: '$2',
```
thomasdigby, line 16: What's the reason for the extra padding on the container here?

src/components/tabs/Tabs.tsx
thomasdigby: I have more comments on the styles themselves but I'd prefer to add those when we've moved them to the component files 👍

src/components/tabs/Tabs.tsx
```diff
@@ -1,36 +1,118 @@
...
+        'div[role="tabpanel"]': {
+          bg: 'white',
+          color: '$textForeground',
+          fontFamily: '$body'
+        },
+        'div[role="tablist"]': {
+          borderBottom: '1px solid $primaryDark'
+        },
+        'button[role="scrollbar"]': {
+          bg: 'white !important',
```
andoulla, line 57: How come we need the important(s) here? I just had to ask

src/components/tabs/TabsTriggerList.tsx
```diff
@@ -0,0 +1,121 @@
...
+import { Flex } from '~/components/flex'
+import { Icon } from '~/components/icon'
+
+type ListProps = React.ComponentProps<typeof StyledTriggerList> & {
+  enableTabScrolling?: boolean
+}
+
+const StyledActionIcon = styled(ActionIcon, {
+  position: 'absolute',
+  top: 'calc(50% - 20px)'
```
andoulla, line 16: I'm not sure I follow this, why -20?

src/components/tabs/TabsTriggerList.tsx
```diff
@@ -0,0 +1,121 @@
...
+import { styled } from '~/stitches'
+import { ActionIcon } from '~/components/action-icon'
+import { Flex } from '~/components/flex'
+import { Icon } from '~/components/icon'
+
+type ListProps = React.ComponentProps<typeof StyledTriggerList> & {
+  enableTabScrolling?: boolean
+}
+
+const StyledActionIcon = styled(ActionIcon, {
```
andoulla, line 14: I would call them `StyledChevronIcon` as it is very explicit that they are (similar to the carousel component)

src/components/tabs/Tabs.tsx
thomasdigby: If you need a reference for our `React.Children` usage in the design system, worth looking at https://github.com/Atom-Learning/components/blob/main/src/components/action-icon/ActionIcon.tsx

src/components/tabs/Tabs.tsx
```diff
@@ -1,36 +1,118 @@
...
+            color: '$primary',
+            fontWeight: 600,
+            boxShadow: 'inset 0 -2px 0 0 currentColor'
+          },
+          '&[data-state="inactive"]': {
+            color: '$tonal500'
+          },
+          '&:hover': {
+            color: '$primary',
+            bg: opacify(-0.9, theme.colors.primary.value)
```
andoulla, line 32: This is just for my benefit but where is this function coming from. When I google it I only get the SASS function for it

src/components/tabs/TabTrigger.tsx
avirati: Sure, this is how it looks with 0.3s

https://user-images.githubusercontent.com/1936119/142572164-6ac39fe5-257e-4aae-97d1-648f98d1e6d9.mov

src/components/tabs/Tabs.tsx
avirati: Designs indicate slight padding on the left. Kept it on x axis to keep it consistent
https://www.figma.com/file/WLQxMluFW4OnaxXL03Gygw/Critique-Design?node-id=430%3A579

src/components/tabs/Tabs.mdx
```diff
@@ -27,15 +27,23 @@ The component provides a set of default styles, which can be overwritten by usin
...
 ```
 
-It also takes a `defaultValue` prop that should match one of the triggers' `value` prop.
+It takes a `defaultValue` prop that should match one of the triggers' `value` prop.
 
 ```tsx
 <Tabs defaultValue="tab1">
 ```
 
+It also takes a `appearance` prop that should either be "light" or "dark".
```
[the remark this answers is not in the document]
avirati: That's up to the user to decide, I think. But I am open to further discussions

src/components/tabs/TabsTriggerList.tsx
avirati: The button that gets rendered has a size of 40px. This intends to horizontally align an absolute positioned element. Interested in knowing if there is a better way to handle this

src/components/tabs/TabsTriggerList.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+      const shouldShowScroller = scrollWidth > offsetWidth
+      if (shouldShowScroller) {
+        setShowRightScroller(true)
+      }
+    }
+  }, [])
+
+  const showScroller = showLeftScroller || showRightScroller
+
```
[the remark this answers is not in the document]
avirati: Great, thanks ! Have made it so

src/components/tabs/Tabs.tsx
avirati: Coming from `polished`

```typescript
import { opacify } from 'polished'
```

This was suggested by @thomasdigby and previously used (and probably introduced) by @Mhoog in some places

src/components/tabs/Tabs.tsx
avirati: We don't, had put them for quick testing and forgot afterwards. Rectified. Apologies !

src/components/tabs/Tabs.tsx
avirati: Makes complete sense ! Thanks. On it 👍

src/components/tabs/TabTrigger.tsx
hollg: Cool! I think this looks good but @thomasdigby usually has opinions on this so it might be worth waiting for his approval on it 😄

src/components/tabs/TabsTriggerList.tsx
hollg: Sorry Avi I'm not sure I follow. Maybe my first comment wasn't clear! The `else` isn't nested within the `if (direction === 'right')` conditional block so I don't see changing that condition should affect it.

Does logic like this not work?

```
if (triggerList) {
  const { scrollWidth, scrollLeft } = triggerList
  const scrollAmount = Math.round(scrollWidth / 4) // 25% of whole scroll width
  
  if (direction === 'right' && triggerList.scrollLeft + scrollAmount <= scrollWidth) {
     triggerList.scroll({
            left: triggerList.scrollLeft + scrollAmount,
            behavior: 'smooth'
          })
  } else {
    triggerList.scroll({
      left: scrollLeft - scrollAmount > 0 ? scrollLeft - scrollAmount : 0,
      behavior: 'smooth'
    })
  }
} 
```

package.json
```diff
@@ -93,7 +93,7 @@
     "eslint-plugin-simple-import-sort": "^7.0.0",
     "glob": "^7.1.6",
     "husky": "^4.3.8",
-    "jest": "^27.0.6",
+    "jest": "^26.6.3",
```
thomasdigby, line 96: What's the reason for this downgrade?

src/components/tabs/TabsTriggerList.tsx
avirati: That's an old comment @hollg 
It looks like this now

```typescript
if (triggerList) {
      const { scrollWidth, scrollLeft } = triggerList
      const scrollAmount = Math.round(scrollWidth / 4) // 25% of whole scroll width
      let left = scrollLeft
      if (direction === 'right') {
        const newScrollAmount = triggerList.scrollLeft + scrollAmount
        left = newScrollAmount <= scrollWidth ? newScrollAmount : scrollLeft
      } else {
        const newScrollAmount = scrollLeft - scrollAmount
        left = newScrollAmount > 0 ? newScrollAmount : 0
      }

      triggerList.scroll({
        left,
        behavior: 'smooth'
      })
...
}
```

src/components/tabs/Tabs.tsx
thomasdigby: I think that may be incidental, these values seem to represent spacing between the tab container and other sibling elements, so I would remove them and if they're required allow the user to add them when needed.

package.json
avirati: Had to downgrade to fix this error (while trying to locally test things, didn't succeed :()
https://atom-learning.slack.com/archives/C02EZ3H6P1U/p1637300430197000?thread_ts=1637232841.193900&cid=C02EZ3H6P1U

src/components/tabs/TabsTriggerList.tsx
hollg: Apologies! New version looks great 👍🏻

src/components/tabs/Tabs.tsx
avirati: Roger that, on it 👍

src/components/tabs/TabsTriggerList.tsx
avirati: @thomasdigby need your input here

src/components/tabs/Tabs.mdx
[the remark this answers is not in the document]
andoulla: Talking to jessica they had specific use cases for when to use the dark one but happy to come back and add more docs when things are finalised

andoulla (approved):

src/components/tabs/TabsTriggerList.tsx
avirati: Changed to calc(50% - ($4 + $1)), thanks @andoulla for showing me the this !

src/components/tabs/Tabs.tsx
```diff
@@ -1,36 +1,63 @@
...
 })
 
 const StyledTabContent = styled(Content, {
-  flexGrow: 1
+  flexGrow: 1,
+  variants: {
+    theme: {
+      light: {
+        color: '$textForeground',
+        fontFamily: '$body'
```
thomasdigby, line 37: We can move this to default styles as it doesn't change per variant

src/components/tabs/utils.ts
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+
+export const passPropsToChildren = (
+  children,
+  props: Record<string, any>,
+  allowedChildNodeTypes: React.FunctionComponent[] = []
+) => {
+  return React.Children.map(children, (child) => {
+    if (allowedChildNodeTypes.includes(child?.type)) {
```
thomasdigby, line 9: We may need to add a check for a valid react element here, TS has thrown an error in the past when I haven't included this

```ts
if (!React.isValidElement(child)) {
  throw new Error()
}
```

src/components/tabs/TabsTriggerList.tsx
```diff
@@ -0,0 +1,166 @@
...
+      </StyledTriggerList>
+    )
+  }
+
+  return (
+    <Flex css={{ position: 'relative' }}>
+      {showLeftScroller && (
+        <StyledChevronIcon
+          size="lg"
+          role="scrollbar"
```
thomasdigby, line 130: Is `role="scrollbar"` the correct role for this component? Reading on the W3 spec it seems like this is specifically for a scrollbar and not the controls for the scrollbar https://www.w3.org/TR/wai-aria-1.1/#scrollbar

src/components/tabs/TabsTriggerList.tsx
thomasdigby: I actually am not super clear on Gary's point.

25% is relative to the width of the container for the tab trigger list, not the total scrollable length of the trigger list. So for example, if we have two tabs at 100% and 50% of the width of the container and we scroll along at 25% of the container width, the first tab will have 75% in view, and the second will be 25% in view. No matter the size of the tab, with each scroll left/right you're only moving the tab 25% of the container width in either direction.

I might not have described that well, please let me know if I haven't

hollg: I had definitely misunderstood what that `25%` represents, will resolve this thread. Thanks Tom!

thomasdigby: Oh, I've just realised what you mean, we should be using `clientWidth` instead of `scrollWidth` here, to ensure that we're scrolling 25% of the container width, not of the content width.

And with that in mind, I wonder whether 33% (or even 50%) might work better at smaller viewport sizes

src/components/tabs/TabsTriggerList.tsx
```diff
@@ -0,0 +1,166 @@
...
+          setShowLeftScroller(false)
+          setShowRightScroller(true)
+        } else if (scrollLeft + offsetWidth === scrollWidth) {
+          setShowRightScroller(false)
+          setShowLeftScroller(true)
+        } else {
+          setShowLeftScroller(true)
+          setShowRightScroller(true)
+        }
+      }, 300)
```
thomasdigby, line 98: Is this value consistent cross-browser?

src/components/tabs/TabsTriggerList.tsx
```diff
@@ -0,0 +1,166 @@
...
+            '&::-webkit-scrollbar': { display: 'none' }
+          }}
+        >
+          {passPropsToChildren(children, { theme }, [TabTrigger])}
+        </StyledTriggerList>
+        {showRightScroller && (
+          <StyledChevronIcon
+            size="lg"
+            role="scrollbar"
+            label="scroll-right"
```
thomasdigby, line 149: Can we treat this label as human readable text? e.g.

```suggestion
            label="Scroll right"
```

src/components/tabs/TabsTriggerList.tsx
```diff
@@ -0,0 +1,166 @@
...
+        )}
+        <StyledTriggerList
+          {...rest}
+          ref={triggerListRef}
+          theme={theme}
+          css={{
+            ...css,
+            width: '100%',
+            overflowX: 'auto',
+            '&::-webkit-scrollbar': { display: 'none' }
```
thomasdigby, line 140: Do we need any other properties to hide/remove the scrollbar cross-browser?

src/components/tabs/TabsTriggerList.tsx
```diff
@@ -0,0 +1,166 @@
...
+    const triggerList = triggerListRef.current
+    if (triggerList) {
+      const { offsetWidth, scrollWidth } = triggerList
+
+      const shouldShowScroller = scrollWidth > offsetWidth
+      if (shouldShowScroller) {
+        setShowRightScroller(true)
+      }
+    }
+  }, [])
```
thomasdigby, line 112: We should look to add this functionality on resize as well

src/components/tabs/TabsTriggerList.tsx
```diff
@@ -0,0 +1,166 @@
...
+interface ListProps extends React.ComponentProps<typeof StyledTriggerList> {
+  enableTabScrolling?: boolean
+}
+
+const fadedWhite = opacify(-0.2, 'white')
+const fadedPrimaryDark = opacify(-0.2, theme.colors.primaryDark.value)
+
+const StyledChevronIcon = styled(ActionIcon, {
+  position: 'absolute',
+  top: 'calc(50% - ($4 + $1))',
```
thomasdigby, line 21: If we want the `ActionIcon` to be the height of the trigger list, we should be able to set `top: 0` right? Does the `ActionIcon` not match the height at the moment?

src/components/tabs/TabsTriggerList.tsx
thomasdigby: Also, we can just add these CSS properties to the `StyledTriggerList` itself, rather than passing them down through `css={}`

src/components/tabs/TabsTriggerList.tsx
```diff
@@ -0,0 +1,166 @@
...
+  variants: {
+    theme: {
+      light: {
+        '&[role="scrollbar"], &[role="scrollbar"]:not(:disabled):hover, &[role="scrollbar"]:not(:disabled):focus, &[role="scrollbar"]:not(:disabled):active': {
```
thomasdigby, line 25: Are we overriding the default theme of the `ActionIcon` here? Is that why we're adding the extra specificity? It would be great if we were able to write something like this to handle that and simplify the selectors:

```ts
const StyledChevronIcon = styled(ActionIcon, {
  color: 'currentColor',
  position: 'absolute',
  top: 'calc(50% - ($4 + $1))',
  variants: {
    theme: {
      light: {
        '&, &:hover, &:focus, &:active': { bg: fadedWhite }
      },
      dark: {
        '&, &:hover, &:focus, &:active': { bg: fadedPrimaryDark }
      }
    }
  }
})
```

src/components/tabs/TabsTriggerList.tsx
```diff
@@ -0,0 +1,166 @@
...
+          >
+            <Icon is={ChevronRight} size="lg" />
+          </StyledChevronIcon>
+        )}
+      </Flex>
+    )
+  }
+
+  return (
+    <StyledTriggerList css={css} theme={theme} {...rest} ref={triggerListRef}>
```
thomasdigby, line 162: We should be able to pass `css` as part of `{...rest}` so we can remove it here and in our props

src/components/tabs/TabTrigger.tsx
thomasdigby: I have _opinions!?_ I think this looks good 👍

src/components/tabs/TabTrigger.tsx
```diff
@@ -1,41 +1,82 @@
...
-    fontWeight: 'bold',
-    boxShadow: 'inset 0 -3px 0 0 currentColor'
+  transition: '0.3s',
+  variants: {
+    theme: {
+      light: {
+        bg: 'white',
+        '&[data-state="active"]': {
+          color: '$primary',
+          textShadow: '0px 0px 1px currentColor',
```
thomasdigby, line 21: Are we using this in lieu of `fontWeight: 'bold'`? I think we need to set every tab as either `bold` or `normal` so we can avoid the pixel jump when selecting a tab

thomasdigby:
Added a bunch of comments, mostly minor changes, but overall it's looking great. Excited to get this in! 👍

src/components/tabs/TabsTriggerList.tsx
avirati: I see, should it just be "button" then ?

src/components/tabs/TabsTriggerList.tsx
avirati: It is, since other browsers do not have a smooth animation inbuilt, it's almost immediate, snappy. Which makes me wonder, are we okay with that ?

avirati: Do you think `500` would be a better alternate ?

src/components/tabs/TabsTriggerList.tsx
avirati: Added `scrollbarWidth: 'none'` for firefox to hide the scrollbar.
Moved the css properties above to `StyledTriggerList` ✔️

src/components/tabs/TabsTriggerList.tsx
avirati: ActionIcon is 40px in size whereas the tabs are 49px in height. Do you think they should have / could be matched ?

src/components/tabs/TabsTriggerList.tsx
avirati: My initial hunch was this but I think the additional specificity is required to properly override css pseudo states

src/components/tabs/TabTrigger.tsx
avirati: Yes, I used textShadow which was inspired from a previous implementation of tabs in our code. I also understand what you are trying to imply here, that we set aside a fixed sized text so that the jump does not happen, however while implementing this, I still see the jump. Perhaps you could help me out here with some more examples / reading material. Interested to know how this can be solved

thomasdigby: Huh, I don't know why it would jump if each tab is `bold`, let's chat about this in person when you're available. We should avoid using `textShadow` for bolding anyway as it'll render a "faux bold" style that won't match with the actual bold weight that the typeface provides. This article is almost 10 years old (!) but it still holds true https://alistapart.com/article/say-no-to-faux-bold/

src/components/tabs/TabTrigger.tsx
```diff
@@ -1,41 +1,79 @@
...
 
 export const TabTrigger: React.FC<TabTriggerProps> = ({
   children,
+  theme,
   disabled = false,
   css,
   ...otherProps
 }) => (
-  <StyledTabTrigger css={css} disabled={disabled} {...otherProps}>
+  <StyledTabTrigger css={css} disabled={disabled} theme={theme} {...otherProps}>
```
thomasdigby, line 76:
```suggestion
  <StyledTabTrigger disabled={disabled} theme={theme} {...otherProps}>
```

src/components/tabs/utils.ts
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+
+export const passPropsToChildren = (
+  children,
+  props: Record<string, any>,
+  allowedChildNodeTypes: any[] = []
```
thomasdigby, line 6: Could we tighten up this type?

```suggestion
  allowedChildNodeTypes: ReactElement[] = []
```

src/components/tabs/TabsTriggerList.tsx
thomasdigby: @avirati Can you confirm that we don't need to use `clientWidth` here?

src/components/tabs/TabsTriggerList.tsx
thomasdigby: The tabs should have a set height, I'm surprised they haven't got one. I'd use this opportunity to set a height on tabs so we can match with the `ActionIcon`. `$sizes$5` is 48px which should work well. You may need to set a height here using `size: $5` as well.

src/components/tabs/Tabs.tsx
```diff
@@ -1,36 +1,61 @@
-import { Content, List, Root } from '@radix-ui/react-tabs'
-import * as React from 'react'
+import { Content, Root } from '@radix-ui/react-tabs'
+import React from 'react'
```
thomasdigby, line 2: Why did we switch to the interop default import here?

src/components/tabs/TabsTriggerList.tsx
```diff
@@ -0,0 +1,176 @@
...
+  }
+})
+
+const StyledTriggerList = styled(List, {
+  flexShrink: 0,
+  display: 'flex',
+  width: '100%',
+  overflowX: 'auto',
+  '&::-webkit-scrollbar': { display: 'none' },
+  scrollbarWidth: 'none',
```
thomasdigby, line 47: It might be a nice improvement to add these properties to a custom utility in our stitches config, e.g.

```ts
scrollbar: 'none'
```

```ts
scrollbar: () => ({
  '&::-webkit-scrollbar': { display: 'none' },
  scrollbarWidth: 'none'
})
```

Maybe one for the backlog rather than this PR

src/components/tabs/TabsTriggerList.tsx
```diff
@@ -0,0 +1,176 @@
...
+            <Icon is={ChevronLeft} size="lg" />
+          </StyledChevronIcon>
+        )}
+        <StyledTriggerList {...rest} ref={triggerListRef} theme={theme}>
+          {passPropsToChildren(children, { theme }, [TabTrigger])}
+        </StyledTriggerList>
+        {showRightScroller && (
+          <StyledChevronIcon
+            size="lg"
+            role="scrollbar"
```
thomasdigby, line 158: Can't reply to the original comment, but we should just remove this `role`, the `ActionIcon` component already renders a button

src/components/tabs/Tabs.test.tsx
```diff
@@ -19,13 +19,36 @@ const TabsTest = ({ defaultValue = 'tab1' }) => (
...
   it('renders', async () => {
     const { container } = render(<TabsTest />)
 
     expect(container).toMatchSnapshot()
   })
 
+  it('renders mobile view', async () => {
+    const { container } = render(<MobileTabsTest />)
+
+    expect(container).toMatchSnapshot()
```
thomasdigby, line 49: Does this render the scrolling buttons? I couldn't see anything matching them in the snapshot

thomasdigby (changes requested):
Apologies for another flurry of comments, I just want to make sure that we get this right. I think this should be good once we work through the last few bits 👍

src/components/tabs/TabsTriggerList.tsx
avirati: @thomasdigby @hollg , What I had in mind was, let's say we have 20 tabs in the container with very small width, 25% of `scrollWidth` would probably scroll 5 items at once. `clientWidth` would certainly solve that problem to some extent, but I wouldn't want to increase the percentage.

Ideally, I would want it to snap (and select / focus) one tab at a time, per scroll click - think of google sheets scrolling, it snaps to one cell at a time. This can be handled in a separate ticket, though

src/components/tabs/TabsTriggerList.tsx
avirati: Set a fixed height to Tab.
Added a new variation to `size` prop of `ActionIcon` of `xl`, which equates to `$5`

src/components/tabs/utils.ts
avirati: TS was throwing errors, using `JSXElementConstructor<any>[]` instead.
`allowedChildNodeTypes: JSXElementConstructor<any>[] = []`

src/components/tabs/Tabs.tsx
avirati: I think, it was unintentional. Will revert 👍

src/components/tabs/TabsTriggerList.tsx
avirati: Issue created and assigned to self
https://github.com/Atom-Learning/components/issues/263

src/components/tabs/TabsTriggerList.tsx
avirati: Alright, but to achieve the needed specificity, would `!important` be acceptable ?

```typescript
const StyledChevronIcon = styled(ActionIcon, {
  position: 'absolute',
  variants: {
    theme: {
      light: {
        '&, &:not(:disabled):hover, &:not(:disabled):focus, &:not(:disabled):active': {
          bg: fadedWhite,
          color: 'currentColor'
        }
      },
      dark: {
        '&, &:not(:disabled):hover, &:not(:disabled):focus, &:not(:disabled):active': {
          bg: fadedPrimaryDark,
          color: 'currentColor !important'
        }
      }
    }
  }
})
```

src/components/tabs/Tabs.test.tsx
avirati: Strange, I'll investigate.

src/components/tabs/TabsTriggerList.tsx
avirati: Resolving this since a parallel thread is open on the same

src/components/tabs/Tabs.test.tsx
avirati: was able to circumvent that by setting `display` on buttons, instead of conditionally rendering them

avirati: Will try to solve this with opacity + visibility + pointer-events: none

src/components/tabs/TabsTriggerList.tsx
avirati: Post discussion, it was decided that we would enhance it as per this feature request https://github.com/Atom-Learning/components/issues/264

Resolving this

src/components/tabs/TabTrigger.tsx
avirati: Post discussion, resolving this by setting fontWeight and negative letter spacing

src/components/tabs/TabsTriggerList.tsx
avirati: Post discussion, `!important` seems unavoidable.

```typescript
const StyledChevronIcon = styled(ActionIcon, {
  position: 'absolute',
  variants: {
    theme: {
      light: {
        bg: `${fadedWhite} !important`
      },
      dark: {
        bg: `${fadedPrimaryDark} !important`,
        color: 'currentColor !important'
      }
    }
  }
})
```

src/components/tabs/Tabs.tsx
```diff
@@ -1,36 +1,61 @@
...
+import { TriggerListWrapper } from './TabsTriggerList'
+import { passPropsToChildren } from './utils'
+
+type TabsProps = React.ComponentProps<typeof StyledRoot>
 
 const StyledRoot = styled(Root, {
   display: 'flex',
-  flexDirection: 'column'
+  flexDirection: 'column',
+  pt: '$4',
```
thomasdigby, line 15: I'm still not clear as to why we need this `padding-top` here, can we remove it and apply it through `css` in core where it's required?

src/components/tabs/TabsTriggerList.tsx
```diff
@@ -0,0 +1,178 @@
...
+      },
+      dark: {
+        bg: `${fadedPrimaryDark} !important`,
+        color: 'currentColor !important'
+      }
+    },
+    visible: {
+      true: {
+        opacity: 1,
+        visibility: 'visible',
```
thomasdigby, line 35: We need to add a `transition: all 125ms` to the base styles here to allow the icon to fade in and out

thomasdigby (commented):
Awesome, just two final bits then we're good to go

src/components/tabs/Tabs.tsx
avirati: Added it because the design had it, but yes I get your point, this should come from core

thomasdigby (approved):
Excellent work!

thomasdigby (approved):

PR #261, opened 2021-11-22
feat: adding new table variants

kerryambt:
Added a few options to the table as they were needed for use with Accordions, and will likely be useful in the future elsewhere.

- `<Table applyRadius="false">` stops borderRadius being applied, useful if your container has a radius and multiple radiuses look odd. Default is `true`.
- `<Table.Header theme="primary">` changes the colour of the header rows. Options are `primary` and `primaryDark`. Default is `primaryDark`.

image

### DIFF
--- a/src/components/table/Table.mdx
+++ b/src/components/table/Table.mdx
@@ -123,10 +123,10 @@ The `Table` component has 2 size variants that control the row height. The two a
 </Table>
 ```
 
-The `Table` component by default renders with a slight border radius. If you want to remove this, for example if you are rendering the `Table` inside another component that has border radius, you can remove this by setting `applyRadius="false"`.
+The `Table` component by default renders with a slight border radius. If you want to remove this, for example if you are rendering the `Table` inside another component that has border radius, you can remove this by setting `corners="square"`. The default is `corners="round"`.
 
 ```tsx
-<Table applyRadius="false">...</Table>
+<Table corners="square">...</Table>
 ```
 
 `Table.Header` accepts a theme prop, the current available options are `primary` and `primaryDark`. The default is `primaryDark`.
--- a/src/components/table/Table.mdx
+++ b/src/components/table/Table.mdx
@@ -122,3 +122,17 @@ The `Table` component has 2 size variants that control the row height. The two a
   </Table.Body>
 </Table>
 ```
+
+The `Table` component by default renders with a slight border radius. If you want to remove this, for example if you are rendering the `Table` inside another component that has border radius, you can remove this by setting `applyRadius="false"`.
+
+```tsx
+<Table applyRadius="false">...</Table>
+```
+
+`Table.Header` accepts a theme prop, the current available options are `primary` and `primaryDark`. The default is `primaryDark`.
+
+```tsx
+<Table>
+  <Table.Header theme="primary">...</Table.Header>
+</Table>
+```
--- a/src/components/table/Table.test.tsx
+++ b/src/components/table/Table.test.tsx
@@ -7,7 +7,7 @@ import { Table } from './Table'
 describe(`Table component`, () => {
   it('renders', async () => {
     const { container } = await render(
-      <Table css={{ height: '100px', width: '400px', color: '$primary500' }}>
+      <Table css={{ height: '100px', width: '400px' }}>
         <Table.Header>
           <Table.Row>
             <Table.HeaderCell>Column A</Table.HeaderCell>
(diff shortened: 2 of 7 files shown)


src/components/table/Table.test.tsx
```diff
@@ -54,6 +54,55 @@ describe(`Table component`, () => {
     expect(container).toMatchSnapshot()
   })
 
+  it('renders with no radius', async () => {
+    const { container } = await render(
+      <Table
+        applyRadius="false"
+        css={{ height: '100px', width: '400px', color: '$primary500' }}
```
andoulla, line 61: if the table now has a default and a variant I don't think we should pass `color: '$primary500'` what do you think?

src/components/table/Table.tsx
```diff
@@ -46,8 +58,11 @@ type TableProps = React.ComponentProps<typeof StyledTable>
 
 export const Table: React.FC<TableProps> & TableSubComponents = ({
   size = 'md',
+  applyRadius = 'true',
```
andoulla, line 61: why is true here a string?

hollg:
Does `<Table css={{borderRadius: 'none'}}/>` not work? 

I'd be keen to avoid boolean props like this where possible because once you start adding them it's easy to end up with components that aren't actually that flexible any more.

hollg:
Oh, sorry, I think I see the issue: there `border-radius` is applied to some header cells too. We could still use the `css` prop in each case, although with a more complicated value. 

If we want to avoid that then I'd recommend renaming the variant to something descriptive rather than imperative. E.g. `corners`: `soft` or `square`. That makes it more in keeping with the convention established by our other themes.

src/components/table/Table.tsx
kerryambt: this was a typo, but I have changed it as per Gary's comments below.

src/components/table/Table.test.tsx
kerryambt: sorry, missed that! removed.

kerryambt:
@hollg I think it would be helpful to have as a property rather than using css, mostly because the css required is pretty clunky given a simple `borderRadius: 'none'` doesn't work.

Really good point on the naming convention though. I was staring at true being what it was and it didn't look right but somehow didn't think to rename. Have now set to `corners="rounded"` or `corners="square"`.

src/components/table/Table.tsx
```diff
@@ -38,6 +38,18 @@ const StyledTable = styled('table', {
...
+      round: {},
+      square: {
+        [`${TableHeaderCell}, ${TableRow}`]: {
+          borderRadius: '0',
+          '&:last-child': {
+            'td:first-child': { borderBottomLeftRadius: '0' },
+            'td:last-child': { borderBottomRightRadius: '0' }
+          }
+        }
+      }
```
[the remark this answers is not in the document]
thomasdigby: I think there's a further conversation to be had about the consistency with which we apply props that affect nested components, in other cases we've used `React.Children` to pass those props directly down to children, allowing the styles to stay co-located with the component. I'm happy with these styles for now but I think it'll be beneficial to bring them all in line and document the preferred method 👍

src/components/table/TableHeader.tsx
```diff
@@ -1,5 +1,28 @@
...
+const StyledTableHeader = styled('thead', {
+  variants: {
+    theme: {
+      primary: {
+        '& th': {
+          bg: '$primary'
+        }
+      },
+      primaryDark: {
+        '& th': {
```
thomasdigby, line 13: Can we target the `th` by the same `[${TableHeaderCell}]` convention we're using elsewhere?

thomasdigby (changes requested):
Just a few minor changes I reckon, looking good though! Hat tip @hollg for the improved `corners` API suggestion

src/components/table/TableHeader.tsx
kerryambt: good point - done :)

src/components/table/Table.mdx
```diff
@@ -122,3 +122,17 @@ The `Table` component has 2 size variants that control the row height. The two a
   </Table.Body>
 </Table>
 ```
+
+The `Table` component by default renders with a slight border radius. If you want to remove this, for example if you are rendering the `Table` inside another component that has border radius, you can remove this by setting `corners="square"`. The default is `corners="round"`.
+
+```tsx
+<Table corners="square">...</Table>
```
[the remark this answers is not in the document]
kerryambt: thanks! me too

src/components/table/Table.tsx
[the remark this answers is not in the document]
kerryambt: these are both really interesting points. I think a frontend team chat would be beneficial to discuss our overall composition for components here, and then documentation when it's done. I've not personally used `React.Children` before but to me the idea of passing props down the tree makes more sense because the props are being used where they need to be. I'm happy with a little bit of potential prop drilling for the sake of easier to read code when we're composing components (though generally drilling isn't great). 

For now I'm going to apply your suggestions, I think there's merit however for a conversation and then maybe a task (perhaps as tech debt) to update components to whatever system we decide. It makes sense to me to do that all at once so this can stay as is for now.

thomasdigby (approved):

PR #266, opened 2021-11-25
fix: css as per designs + SSR build

avirati:
1. Removed background from tabs container
2. Use `window` inside of `useEffect` to fix SSR build

### DIFF
--- a/src/components/tabs/TabsTriggerList.tsx
+++ b/src/components/tabs/TabsTriggerList.tsx
@@ -113,9 +113,9 @@ export const TriggerListWrapper: React.FC<ListProps> = ({
   }, [])
 
   useEffect(() => {
-    const onResize = debounce(() => {
+    const onResize = debounce(500, () => {
       setScreenWidth(window.innerWidth)
-    }, 500)
+    })
     window.addEventListener('resize', onResize)
 
     return () => {
--- a/src/components/tabs/TabsTriggerList.tsx
+++ b/src/components/tabs/TabsTriggerList.tsx
@@ -58,6 +58,7 @@ const StyledTriggerList = styled(List, {
         borderBottom: '1px solid $tonal300'
       },
       dark: {
+        bg: '$primaryDark',
         borderBottom: '1px solid $tonal200'
       }
     }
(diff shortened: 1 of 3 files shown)


src/components/tabs/TabsTriggerList.tsx
```diff
@@ -73,7 +74,7 @@ export const TriggerListWrapper: React.FC<ListProps> = ({
   const triggerListRef = useRef<HTMLDivElement>(null)
   const [showLeftScroller, setShowLeftScroller] = useState<boolean>(false)
   const [showRightScroller, setShowRightScroller] = useState<boolean>(false)
-  const [screenWidth, setScreenWidth] = useState<number>(window.innerWidth)
+  const [screenWidth, setScreenWidth] = useState<number>(0)
```
thomasdigby, line 77: Do we need to update this on initial render? I can't see where that would happen otherwise

avirati: Don't think we do, can be left undefined 👍

avirati: Updated to 
`const [screenWidth, setScreenWidth] = useState<number>()`

src/components/tabs/TabsTriggerList.tsx
```diff
@@ -73,7 +74,7 @@ export const TriggerListWrapper: React.FC<ListProps> = ({
   const triggerListRef = useRef<HTMLDivElement>(null)
   const [showLeftScroller, setShowLeftScroller] = useState<boolean>(false)
   const [showRightScroller, setShowRightScroller] = useState<boolean>(false)
-  const [screenWidth, setScreenWidth] = useState<number>(window.innerWidth)
+  const [screenWidth, setScreenWidth] = useState<number>()
```
thomasdigby, line 77: Sorry, I wasn't clear enough in my comment, where are we setting `screenWidth` on initial render so we know whether to render the action icons or not? If we load this view on a narrow viewport, will the action icons show correctly, or do you need to resize the browser to trigger it?

avirati: There is no manual resizing involved. This `useEffect` which listens to change on `screenWidth`, which in turn shows / hides the scroll button action icons

```typescript
useEffect(() => {
    const triggerList = triggerListRef.current
    if (triggerList) {
      const { offsetWidth, scrollWidth } = triggerList

      const shouldShowScroller = scrollWidth > offsetWidth

      // Scroll to left on resize, hide left scroll button and then decide if right scroll button should be visible or not
      triggerList.scroll?.({
        left: 0
      })
      setShowLeftScroller(false)
      setShowRightScroller(shouldShowScroller)
    }
  }, [screenWidth])
```

thomasdigby: Ah got it thanks, I misinterpreted when this `useEffect` would run. Looks all good to me 👍

thomasdigby (approved):

hollg (approved):

PR #267, opened 2021-11-26
feat: notification badge component

kerryambt:
image 
image

```tsx
<NotificationBadge value={3}>
      <ActionIcon appearance="outline" size="lg" isRounded>
        <Icon is={Controls} />
      </ActionIcon>
 </NotificationBadge>
```
A simple wrapper for any content - though typically ActionIcons - that will display some information as a notification overlay. This could be, for example, the number of filters selected from a button that selects filters.

Styling options include themes and appearance, however generally I'm trying to keep this lightweight.


src/components/notification-badge/NotificationBadge.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { Box } from '../box'
+
+import { styled } from '~/stitches'
+
+const StyledWrapper = styled(Box, {
+  position: 'relative'
+})
+
+const StyledBadge = styled(Box, {
+  position: 'absolute',
```
andoulla, line 12: Do we need to set the font family for the text?

andoulla (commented):
Just 2 minor comments, nice little component btw

src/components/notification-badge/NotificationBadge.tsx
kerryambt: I don't think so, we want DS stuff to be unassuming and this should just inherit font family from wherever it's used. the sandbox doesn't apply font family by default but I'm fairly certain in the app it would. sorry if I'm missing something though :)

src/components/notification-badge/NotificationBadge.mdx
```diff
@@ -0,0 +1,26 @@
...
+    <Icon is={Controls} />
+  </ActionIcon>
+</NotificationBadge>
+```
+
+## Value
+
+Whilst you would normally want to display a number, the `value` prop also supports strings.
+
+## Themes
```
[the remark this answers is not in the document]
kerryambt: really good point - will change now :)

src/components/notification-badge/NotificationBadge.tsx
andoulla: I'm not 100%, in places like inputs and buttons and labels we do set the `fontFamily: $body`. let's have a chat with the team on Monday.

thomasdigby: We'll want to use the `$body` font-family here as we're making a specific design choice about what font will render best in this context. The DS needs to be flexible but we also need to ensure consistency and apply lessons learned about our user's interaction with the UI, there are places where being opinionated is useful, even if this particular choice is a minor one.

src/components/notification-badge/NotificationBadge.tsx
```diff
@@ -0,0 +1,59 @@
...
+      },
+      neutral: {
+        bg: '#fff'
+      }
+    },
+    appearance: {
+      simple: {},
+      outline: {
+        border: '1px solid currentColor'
+      }
```
thomasdigby, line 38: I would simplify and reduce all of our theme and appearance options into the small subset we need for use in our app. That may just be `bg: $primary, color: 'white'` for now, which is great. If we need others then we should add them when they're needed to avoid bloating the DS with unused options.

src/components/notification-badge/NotificationBadge.tsx
```diff
@@ -0,0 +1,59 @@
...
+})
+
+const StyledBadge = styled(Box, {
+  position: 'absolute',
+  top: '-$2',
+  right: '-$2',
+  borderRadius: '$round',
+  p: '$1',
+  height: '$2',
+  width: '$2',
```
thomasdigby, line 18: Does this handle numbers like `888`? We may need this to be a `min-width` and some padding to ensure it maintains good spacing

src/components/notification-badge/NotificationBadge.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { styled } from '~/stitches'
+
+const StyledWrapper = styled(Box, {
+  position: 'relative'
+})
+
+const StyledBadge = styled(Box, {
+  position: 'absolute',
+  top: '-$2',
+  right: '-$2',
```
thomasdigby, line 14: I can imagine this may need to be configurable based on the different child components, e.g. `<ActionIcon isRounded>` may work differently than a `<Button>`. Can we try it with a few different components and see what's needed?

src/components/notification-badge/NotificationBadge.tsx
```diff
@@ -0,0 +1,59 @@
...
+}
+
+export const NotificationBadge: React.FC<NotificationBadgeProps> = ({
+  theme = 'primary',
+  appearance = 'simple',
+  value,
+  children
+}) => (
+  <StyledWrapper>
+    <StyledBadge appearance={appearance} theme={theme}>
```
thomasdigby, line 54: Is it worth using `role="status"` here? https://www.w3.org/TR/wai-aria-1.1/#status

src/components/notification-badge/NotificationBadge.tsx
kerryambt: thanks for the explanation there - changed now :)

src/components/notification-badge/NotificationBadge.tsx
andoulla: from experience going to 3 digit numbers, the circle becomes a bit of an oval if we want to keep the height the same. A pattern I have seen is to accommodate 2 digits and then show a + for anything over 100

src/components/notification-badge/NotificationBadge.tsx
kerryambt: so following this comment and your other comment about differing sizes of content I changed the positioning CSS to be percentage based and make sure the offset is always 50% of the badge's width. As such it works really well on all kinds of things:
image

src/components/notification-badge/NotificationBadge.tsx
kerryambt: good point, have fixed now :)

src/components/notification-badge/NotificationBadge.tsx
kerryambt: yeah seems a good opportunity to do so - will do :)

andoulla (approved):

src/components/notification-badge/NotificationBadge.mdx
```diff
@@ -0,0 +1,39 @@
+---
+title: NotificationBadge
```
thomasdigby, line 2:
```suggestion
title: Notification Badge
```

src/components/notification-badge/NotificationBadge.mdx
```diff
@@ -0,0 +1,39 @@
...
+component: NotificationBadge
+description: A badge that wraps content and displays a notification
+category: Layout
```
thomasdigby, line 5:
```suggestion
category: Feedback
```

I think this is a more appropriate category

src/components/notification-badge/NotificationBadge.mdx
```diff
@@ -0,0 +1,39 @@
...
+---
+
+The NotificationBadge component is a lightweight wrapper for content such as ActionIcons, but can be wrapped around anything.
```
thomasdigby, line 8:
```suggestion
The `NotificationBadge` component is a lightweight wrapper for content such as an `ActionIcon` but can be wrapped around any block-level element.
```

src/components/notification-badge/NotificationBadge.tsx
thomasdigby: Could we use the `primary` theme in the core app? I wonder if the `dark` option might not be as useful in the long run; we'll want this notification badge to be bright and clear to the user, so a neutral/tonal version may not "pop" as much. I could also imagine `$brandRed` being used, but we can cross that bridge when we need it!

src/components/notification-badge/NotificationBadge.tsx
```diff
@@ -0,0 +1,51 @@
...
+const StyledWrapper = styled(Box, {
+  position: 'relative'
+})
+
+const StyledBadge = styled(Box, {
+  fontFamily: '$body',
+  position: 'absolute',
+  top: '-$2',
+  left: '100%',
+  transform: 'translateX(-50%)',
```
thomasdigby, line 16: Could we use the transform for both `top` and `right` positioning?

```ts
top: 0,
right: 0,
transform: 'translate(-50%, -50%)'
```

Is really clear to me how the element is positioned

src/components/notification-badge/NotificationBadge.mdx
```diff
@@ -0,0 +1,39 @@
...
+```tsx preview
+<NotificationBadge value={1}>
+  <ActionIcon appearance="outline" size="lg" isRounded>
+    <Icon is={Controls} />
+  </ActionIcon>
+</NotificationBadge>
+<NotificationBadge theme="dark" value={2}>
+  <ActionIcon appearance="outline" size="lg" isRounded>
+    <Icon is={Controls} />
+  </ActionIcon>
```
thomasdigby, line 37: I would use the documentation as a good opportunity to show different combinations, maybe we could render a `Button` here?

thomasdigby: Oh and different values as well, e.g. `value={64}`

thomasdigby (commented):
This is looking great, weirdly excited about getting a notification count component in 😆

thomasdigby:
Oh one last thing that I forgot to mention, I reckon a `fontSize: '$sm'` might suit the size/layout a little better, reducing it to `14px`

src/components/notification-badge/NotificationBadge.mdx
kerryambt: oops - changed :)

src/components/notification-badge/NotificationBadge.mdx
kerryambt: yeah definitely, good catch

src/components/notification-badge/NotificationBadge.mdx
kerryambt: again nice catch, good to be specific :)

src/components/notification-badge/NotificationBadge.tsx
kerryambt: I also initially made a red one but heard your voice in my head telling me no XD I think you raise a good point about it popping, I'll stick with primary for now. if design don't like it we can add in others :)

src/components/notification-badge/NotificationBadge.tsx
kerryambt: yeah this does make more sense and makes it more universal

src/components/notification-badge/NotificationBadge.mdx
kerryambt: good idea :)

src/components/notification-badge/NotificationBadge.tsx
```diff
@@ -0,0 +1,42 @@
...
+type NotificationBadgeProps = React.ComponentProps<typeof StyledBadge> & {
+  value: number | string
+}
+
+export const NotificationBadge: React.FC<NotificationBadgeProps> = ({
+  value,
+  children
+}) => (
+  <StyledWrapper>
+    <StyledBadge role="status">{value}</StyledBadge>
```
thomasdigby, line 39: We need to check for `value` before rendering and maybe also ensure that we don't render value if it is `0`, currently we're displaying an empty circle without a number.

Screenshot 2021-12-02 at 14 44 22

src/components/notification-badge/NotificationBadge.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+const StyledBadge = styled(Flex, {
+  fontFamily: '$body',
+  fontSize: '$sm',
+  flexDirection: 'column',
+  justifyContent: 'center',
+  position: 'absolute',
+  top: '0',
+  right: '0',
+  transform: 'translate(50%, -50%)',
```
thomasdigby, line 20: We may need to update this to a set value to avoid larger numbers floating too far to the right, if we update it to `transform: translate(calc($sizes$2 / 2), -50%)` then it'll sit nicely

Screenshot 2021-12-02 at 14 48 01

src/components/notification-badge/NotificationBadge.tsx
```diff
@@ -0,0 +1,42 @@
...
+  borderRadius: '$round',
+  p: '$1',
```
thomasdigby, line 22: For the larger numbers, even just double digits we might want to add a touch more padding to the container. It might be worth adding a variant for spacing(?) as compact/comfortable, or set `p: $2` for all (if you're happy with the single-digit view being _not quite a circle_)

Screenshot 2021-12-02 at 14 50 31

src/components/notification-badge/NotificationBadge.tsx
kerryambt: didn't think of this :o done

src/components/notification-badge/NotificationBadge.tsx
kerryambt: good point!

src/components/notification-badge/NotificationBadge.tsx
kerryambt: I think for simplicity sake and how little you can tell it's not a circle with `p: $2`, I'll just go with that :)

src/components/notification-badge/NotificationBadge.tsx
```diff
@@ -0,0 +1,42 @@
...
+type NotificationBadgeProps = React.ComponentProps<typeof StyledBadge> & {
+  value: number | string
+}
+
+export const NotificationBadge: React.FC<NotificationBadgeProps> = ({
+  value,
+  children
+}) => (
+  <StyledWrapper>
+    {value ? <StyledBadge role="status">{value}</StyledBadge> : ''}
```
thomasdigby, line 39:
```suggestion
    {!!value && <StyledBadge role="status">{value}</StyledBadge>}
```

Could we convert the value to a boolean for this conditional? It should prevent the "render a 0" problem and might be a better solution than rendering an empty string 👍

kerryambt: yeah that makes more sense. I always forget about `!!`. will fix now :)

thomasdigby (approved):
Awesome!

PR #269, opened 2021-11-29
feat: adding optional uppercase to tabs

kerryambt:
Tab triggers were recently made uppercase by default, but design have said that this should be optional as we do not want all tabs to be uppercase. Simple change to allow an uppercase prop.

`<Tabs.TriggerList uppercase>...</Tabs.TriggerList>`

Uppercase:
image
Property not set:
image

### DIFF
--- a/src/components/tabs/Tabs.mdx
+++ b/src/components/tabs/Tabs.mdx
@@ -72,7 +72,7 @@ A `Tabs.Trigger` component can take a `disabled` prop, which would make it unsel
 
 ## Styling the `Tabs.TriggerList`
 
-The trigger list accepts an appearance property to apply uppercase to all tabs. Simply pass `appearance='uppercase'`. The default value is `appearance='default'`.
+The trigger list accepts an appearance property to apply uppercase to all tabs. Simply pass `appearance='uppercase'`.
 
 ```tsx preview
 <Tabs>
--- a/src/components/tabs/Tabs.mdx
+++ b/src/components/tabs/Tabs.mdx
@@ -72,11 +72,11 @@ A `Tabs.Trigger` component can take a `disabled` prop, which would make it unsel
 
 ## Styling the `Tabs.TriggerList`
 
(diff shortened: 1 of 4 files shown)


andoulla (approved):

src/components/tabs/Tabs.mdx
```diff
@@ -70,6 +70,21 @@ A `Tabs.Trigger` component can take a `disabled` prop, which would make it unsel
 </Tabs>
 ```
 
+## Styling the `Tabs.TriggerList`
+
+The trigger list accepts a property to apply uppercase to all tabs. Simply pass `uppercase`. Omiting this property will result in text rendered as it is typed.
+
+```tsx preview
+<Tabs>
+  <Tabs.TriggerList uppercase>
```
thomasdigby, line 79: My worry with this prop name is that it will open the floodgates for many more single-use props for particular design changes. I would like to use something similar to `appearance` instead so we can name and group these changes. `<Tabs.TriggerList appearance="alternate">` maybe. I'd like to get some more thoughts from design as well as to how they want to categorise and name this change

kerryambt: design didn't seem to mind too strongly either way but thought `appearance="uppercase"` seemed like a good idea so I've gone with that :)

src/components/tabs/TabsTriggerList.tsx
```diff
@@ -61,13 +61,20 @@ const StyledTriggerList = styled(List, {
         bg: '$primaryDark',
         borderBottom: '1px solid $tonal200'
       }
+    },
+    appearance: {
+      uppercase: {
+        textTransform: 'uppercase'
+      },
+      default: {}
```
thomasdigby, line 69: I don't think we need this default option, having a single option and not providing a default is fine 👍

thomasdigby (commented):
Just that minor point r.e. the default `appearance` option, otherwise all good!

src/components/tabs/TabsTriggerList.tsx
kerryambt: ah right, wasn't sure on this -thanks :)

thomasdigby (approved):

PR #270, opened 2021-11-30
fix: don't wrap table heading

hollg:
Prevent table column headers from wrapping onto a new line. 

Before:
Screenshot 2021-11-30 at 08 42 24
After:
Screenshot 2021-11-30 at 08 42 18

### DIFF
--- a/src/components/table/TableHeaderCell.tsx
+++ b/src/components/table/TableHeaderCell.tsx
@@ -7,7 +7,8 @@ export const TableHeaderCell = styled('th', {
   lineHeight: 1.5,
   p: '$2 $3',
   textAlign: 'left',
-  verticalAlign: 'middle'
+  verticalAlign: 'middle',
+  whiteSpace: 'nowrap'
 })
 
 TableHeaderCell.displayName = 'TableHeaderCell'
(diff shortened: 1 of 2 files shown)


kerryambt (approved):

thomasdigby:
@hollg Is the narrow viewport solution to this using the overflow container? Or are we deferring the ability to handle how it renders across viewport sizes to the user?

hollg:
@thomasdigby 
> @hollg Is the narrow viewport solution to this using the overflow container? Or are we deferring the ability to handle how it renders across viewport sizes to the user?

I think you're touching on the issue that made me hesitant to put this in the library originally. The overflow container is the way that we'll handle the issues created by narrow view ports in our app, so maybe this styling should be added by the overflow container itself?

thomasdigby:
> @thomasdigby
> 
> > @hollg Is the narrow viewport solution to this using the overflow container? Or are we deferring the ability to handle how it renders across viewport sizes to the user?
> 
> I think you're touching on the issue that made me hesitant to put this in the library originally. The overflow container is the way that we'll handle the issues created by narrow viewports in our app, so maybe this styling should be added by the overflow container itself?

Yep I think you're right, unless we're providing an option with the design system to handle this, it feels a bit too strict.

Issue #244, opened 2021-11-01
Feature request: Search Input


andoulla:
## Component Name

 Search Input

## Description(\*)

- Change the alignment from left to right
- Change icon colour

## Designs and UX wireframes(\*)

![image](https://user-images.githubusercontent.com/5703687/139695654-d9019ae2-1f67-4c94-bfe1-65f3c416a58a.png)

https://www.figma.com/file/lXBJN5xPUgN6oUTcfSP01j/Atom-Assessments?node-id=3358%3A21805

andoulla:
@thomasdigby Do you think we should consider making the alignment a prop so people have more flexibility?

Or should we remain strict to enforce consistency

thomasdigby:
@andoulla I think we should enforce some consistency here, my only issue with aligning to the right is that users may assume that it's a clickable button, "click here to search" sort of thing. We have that with the password visibility icon and the upcoming date input as well, that was the initial reason for aligning to the left.

PR #271, opened 2021-11-30
SearchInput updates

andoulla:
**Tasks:**
- [x] Move the search icon to the right side of the input
- [x] Show a clear icon when the user has typed into the input
- [x] Clear input on button clean

**Ticket:**
 https://www.pivotaltracker.com/story/show/180313561

**Recording:**
https://recordit.co/simbbNVZhG

Closes https://github.com/Atom-Learning/components/issues/244

### DIFF
--- a/src/components/search-input/SearchInput.tsx
+++ b/src/components/search-input/SearchInput.tsx
@@ -71,7 +71,13 @@ export const SearchInput: React.FC<SearchInputProps> = ({
 
   const getIcon = () => {
     if (activeIcon === INPUT_ICON.SEARCH)
-      return <StyledIcon is={Search} size={size} />
+      return (
+        <StyledIcon
+          is={Search}
+          size={size}
+          css={{ size: size == 'sm' ? '$1' : 20 }}
+        />
+      )
 
     return (
       <ActionIcon
@@ -94,6 +100,7 @@ export const SearchInput: React.FC<SearchInputProps> = ({
         {...remainingProps}
         value={inputValue}
         onChange={handleOnChange}
+        css={{ pr: size === 'sm' ? '$5' : '$6' }}
       />
       {getIcon()}
     </Box>
--- a/src/components/search-input/SearchInput.tsx
+++ b/src/components/search-input/SearchInput.tsx
@@ -1,36 +1,24 @@
 import { Close, Search } from '@atom-learning/icons'
 import * as React from 'react'
 
-import { CSS, styled } from '~/stitches'
-
-import { Box } from '../box/Box'
-import { Icon } from '../icon/Icon'
-import { Input } from '../input/Input'
+import { ActionIcon } from '~/components/action-icon'
+import { Box } from '~/components/box/'
+import { Icon } from '~/components/icon/'
+import { Input } from '~/components/input/'
+import { CSS } from '~/stitches'
 
 type SearchInputProps = React.ComponentProps<typeof Input> & {
   size: 'sm' | 'md'
   css?: CSS
   value?: string | number
+  clearText?: string
 }
-
-const StyledButton = styled('button', {
-  alignItems: 'center',
-  appearance: 'none',
-  bg: 'white',
-  border: 'unset',
-  borderRadius: '$0',
-  boxSizing: 'border-box',
-  cursor: 'pointer',
-  display: 'flex',
-  justifyContent: 'center',
-  p: 'unset',
-  transition: 'all 100ms ease-out'
-})
 
 export const SearchInput: React.FC<SearchInputProps> = ({
   size,
   css,
   value,
+  clearText = 'Clear',
   ...remainingProps
 }) => {
   const [inputValue, setInputValue] = React.useState<string | number>(
@@ -51,29 +39,31 @@ export const SearchInput: React.FC<SearchInputProps> = ({
   }
 
   const getIcon = () => {
-    const css: CSS = {
-      color: '$tonal300',
-      position: 'absolute',
-      size: size === 'sm' ? '$1' : 20,
-      top: size === 'sm' ? '$2' : 10,
-      right: size === 'sm' ? '$2' : 10
-    }
-
     if (!isClearVisible)
       return (
         <Icon
           is={Search}
           css={{
-            ...css,
+            color: '$tonal300',
+            position: 'absolute',
+            size: size === 'sm' ? '$1' : 20,
+            top: size === 'sm' ? '$2' : 10,
+            right: size === 'sm' ? '$2' : 10,
             pointerEvents: 'none'
           }}
         />
       )
 
     return (
-      <StyledButton onClick={handleClear} css={css}>
+      <ActionIcon
+        label={clearText}
+        appearance="simple"
+        theme="neutral"
+        css={{ bottom: 4, position: 'absolute', right: 0 }}
+        onClick={handleClear}
+      >
         <Icon is={Close} />
-      </StyledButton>
+      </ActionIcon>
     )
   }
 
--- a/src/components/search-input/SearchInput.tsx
+++ b/src/components/search-input/SearchInput.tsx
@@ -1,7 +1,7 @@
-import { Search } from '@atom-learning/icons'
+import { Close, Search } from '@atom-learning/icons'
 import * as React from 'react'
 
-import { CSS } from '~/stitches'
+import { CSS, styled } from '~/stitches'
 
 import { Box } from '../box/Box'
 import { Icon } from '../icon/Icon'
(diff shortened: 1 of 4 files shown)


src/components/search-input/SearchInput.tsx
```diff
@@ -10,34 +10,85 @@ import { Input } from '../input/Input'
...
+          is={Search}
+          css={{
+            ...css,
+            pointerEvents: 'none'
+          }}
+        />
+      )
+
+    return (
+      <StyledButton onClick={handleClear} css={css}>
```
Mhoog, line 74: Would it be possible to use `ActionIcon` here?

Also, looking at your recording, it looks like there is no `cursor: pointer` or hover effect on the input clear button. Maybe that is just the recording, but would be good to have a look at that 🙂

src/components/search-input/SearchInput.tsx
```diff
@@ -10,34 +10,85 @@ import { Input } from '../input/Input'
...
+        <Icon is={Close} />
+      </StyledButton>
+    )
+  }
+
+  return (
+    <Box css={{ position: 'relative', ...(css as CSS) }}>
```
Mhoog, line 81: Is there a reason to not type the `css` const above instead of typecasting?

```ts
const css: CSS = {
  ...
}
```

hollg:
This looks cool! The one change I'd suggest is to use an enum for the state that determines which icon is shown, rather than a boolean. Using a boolean two represent binary state often adds some confusing in my experience, for example with this statement in `getIcon`:

```tsx
if (!isClearVisible)
      return (
        <Icon
          is={Search}
          css={{
            ...css,
            pointerEvents: 'none'
          }}
        />
      )
```

It's not immediately clear on reading that how `isClearVisible` relates to that return value. If we use an enum value in the state, I think things become clearer:

```tsx
if (icon === ICON.SEARCH)
      return (
        <Icon
          is={Search}
          css={{
            ...css,
            pointerEvents: 'none'
          }}
        />
      )
```

This might not be as relevant here, but this would also make it simpler to add a new state in the future that needs a new icon.

src/components/search-input/SearchInput.tsx
andoulla: I debated in my head about using the ActionIcon but that renders a tooltip on hover and we didn't have that requirement for this. I can check with the team to see if it is a dealbreaker and if not switch it.

The cursor is there but when I try to record it it just doesn't catch  it

src/components/search-input/SearchInput.tsx
```diff
@@ -10,34 +10,85 @@ import { Input } from '../input/Input'
...
+  const handleOnChange = (event) => {
+    setInputValue(event.target.value)
+    setIsClearVisible(event.target.value.toString().length > 0)
+  }
+
+  const getIcon = () => {
+    const css = {
+      color: '$tonal300',
+      position: 'absolute',
+      size: size === 'sm' ? '$1' : 20,
```
Mhoog, line 57: Might be cleaner to use a Stitches variant and pass `size` onto this so that you can add the conditional styles in the variant.

src/components/search-input/SearchInput.tsx
Mhoog: Never mind! Just saw that the constant created above is not the `css` you're passing in here 😬 

Is typecasting still needed though? You're already setting the type of `css` to `CSS` in `searchInputProps`.

andoulla: It is not needed, it was simply there and I forgot to remove it. It is gone now

src/components/search-input/SearchInput.tsx
thomasdigby: The `ActionIcon` component shouldn't render a tooltip 🤔 Did you test with one?

src/components/search-input/SearchInput.tsx
```diff
@@ -10,34 +10,85 @@ import { Input } from '../input/Input'
...
+  }
+
+  return (
+    <Box css={{ position: 'relative', ...(css as CSS) }}>
       <Input
         size={size}
-        css={{ pl: size === 'sm' ? '$5' : '$6' }}
+        css={{ pr: size === 'sm' ? '$5' : '$6' }}
         {...remainingProps}
-        type="search"
```
thomasdigby, line 37: Is there a reason we're removing `type="search"` here?

src/components/search-input/SearchInput.tsx
```diff
@@ -10,34 +10,85 @@ import { Input } from '../input/Input'
...
-          top: size === 'sm' ? '$2' : 10,
-          left: size === 'sm' ? '$2' : 10
-        }}
-      />
+}) => {
+  const [inputValue, setInputValue] = React.useState<string | number>(
+    value || ''
+  )
+  const [isClearVisible, setIsClearVisible] = React.useState<boolean>(
+    (value && value?.toString().length > 0) || false
```
thomasdigby, line 40: Out of curiosity, when would `value` alone not suffice here?

andoulla: TS throws an error that value could be undefined. I didn't know how to overcome that without making the field mandatory.

src/components/search-input/SearchInput.tsx
andoulla: on some browsers, it adds a browser-specific x
image
I opted in removing it so we always show ours

src/components/search-input/SearchInput.tsx
andoulla: Yes you are both right, I got confused with the tooltipActionIcon we have in core. I have replaced this

andoulla:
> This looks cool! The one change I'd suggest is to use an enum for the state that determines which icon is shown, rather than a boolean. Using a boolean two represent binary state often adds some confusion in my experience, for example with this statement in `getIcon`:
> 
> ```tsx
> if (!isClearVisible)
>       return (
>         <Icon
>           is={Search}
>           css={{
>             ...css,
>             pointerEvents: 'none'
>           }}
>         />
>       )
> ```
> 
> It's not immediately clear on reading that how `isClearVisible` relates to that return value. If we use an enum value in the state, I think things become clearer:
> 
> ```tsx
> if (icon === ICON.SEARCH)
>       return (
>         <Icon
>           is={Search}
>           css={{
>             ...css,
>             pointerEvents: 'none'
>           }}
>         />
>       )
> ```
> 
> This might not be as relevant here, but this would also make it simpler to add a new state in the future that needs a new icon.

I'm tempted to not do this change as this situation is very binary (either we show the clear or not) Also, I can't come up with any use cases where we would want to show a different state. I'm thinking we can cross that bridge if we get to it.

I hope that is ok

hollg:
@andoulla I agree that it's binary, but will still strongly recommend the change. 
Booleans only capture one piece of data, but more than one is relevant. The boolean can only capture "is the clear icon visible?". It can't capture any of the following:
- whether the search icon is visible 
- the fact that there are two types of icon
- the fact that the two types of icon are mutually exclusive

All of that information has to be picked up from reading the surrounding code. Using enums for these cases is a good example of self-documenting code; all that extra information is baked right into the source code with no room for (mis)interpretation.

src/components/search-input/SearchInput.tsx
andoulla: brilliant thank you, this section was bothering me

andoulla:
> @andoulla I agree that it's binary, but will still strongly recommend the change. Booleans only capture one piece of data, but more than one piece of data is relevant. The boolean can only capture "is the clear icon visible?". It can't capture any of the following:
> 
> * whether the search icon is visible
> * the fact that there are two types of icon
> * the fact that the two types of icon are mutually exclusive
> 
> All of that information has to be picked up from reading the surrounding code. Using enums for these cases is a good example of self-documenting code; all that extra information is baked right into the source code with no room for (mis)interpretation.

Ok, I will go with the teams' consensus as it is not that much work anyway. I'm pushing this change now if you and @Mhoog  what to have a look

src/components/search-input/SearchInput.tsx
thomasdigby: `type="search"` provides information to the browser that's worth retaining, and users get additional benefits like a specific keyboard to show on mobile when searching. I would instead look at removing that extra close icon with some CSS, a quick search threw up this:

```css
input[type="search"]::-webkit-search-decoration,
input[type="search"]::-webkit-search-cancel-button,
input[type="search"]::-webkit-search-results-button,
input[type="search"]::-webkit-search-results-decoration { display: none; }
```

andoulla: actually, that worked like a charm. Pushing the change now

src/components/search-input/SearchInput.tsx
```diff
@@ -1,43 +1,109 @@
...
+
+  const handleClear = () => {
+    setInputValue('')
+    setIsClearVisible(INPUT_ICON.SEARCH)
+  }
+
+  const handleOnChange = (event) => {
+    setInputValue(event.target.value)
+    setIsClearVisible(
+      event.target.value.toString().length > 0
```
thomasdigby, line 72: I don't think we need to convert this `toString()` here, it'll always be a string. If TS is throwing an error, we might need to tell it what `event` is:

```ts
const handleOnChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```

src/components/search-input/SearchInput.tsx
```diff
@@ -1,43 +1,109 @@
...
-          top: size === 'sm' ? '$2' : 10,
-          left: size === 'sm' ? '$2' : 10
-        }}
-      />
-      <Input
+}) => {
+  const [inputValue, setInputValue] = React.useState<string | number>(
+    value || ''
+  )
+  const [isClearVisible, setIsClearVisible] = React.useState<INPUT_ICON>(
```
thomasdigby, line 60: This state value doesn't make any sense now we've changed the value, `iconToRender`, `activeIcon`, or similar would be more relevant

src/components/search-input/SearchInput.tsx
```diff
@@ -1,43 +1,109 @@
...
+  color: '$tonal300',
+  position: 'absolute',
+  pointerEvents: 'none',
+  variants: {
+    size: {
+      sm: {
+        top: '$2',
+        right: '$2',
+        size: '$1',
+        ml: '$5'
```
thomasdigby, line 32: I'm not sure `margin-left` achieves the same thing as the `padding-left` did previously on the input. This is absolutely positioned, which means margin won't apply to any sibling elements. The padding value ensured that the text didn't overlap the absolutely positioned icon, it's likely we'll need to restore that property on the input and remove this here.

src/components/search-input/SearchInput.tsx
```diff
@@ -1,43 +1,109 @@
...
+  }
+
+  const getIcon = () => {
+    if (isClearVisible === INPUT_ICON.SEARCH)
+      return <StyledIcon is={Search} size={size} />
+
+    return (
+      <ActionIcon
+        label={clearText}
+        appearance="simple"
```
thomasdigby, line 85: This is the default, we can remove it

src/components/search-input/SearchInput.tsx
```diff
@@ -1,43 +1,109 @@
...
+        theme="neutral"
+        css={{ bottom: 4, position: 'absolute', right: 0 }}
```
thomasdigby, line 87: Do we not need to apply the same sizing and positioning as the `StyledIcon`? Or at least look at changing the `size` prop on `ActionIcon` to match the input size.

I think `size="lg"` should match the `md` size input and `size="md"` should match the `sm` size input, we'll also be able to position it `top:0, right: 0` and have it fill the height required

src/components/search-input/SearchInput.tsx
andoulla: This is just for anyone following this thread. The above type still complains if `.length> 0` is used as the input could take a number as input. 

However, in this situation, changing the type of the component to only allow strings as a value
```
type SearchInputProps = React.ComponentProps<typeof Input> & {
  size: 'sm' | 'md'
  css?: CSS
  value?: string
  clearText?: string
}
```
and removing the length part of the condition allows the functionality to work just fine and it doesn't have any TS errors

src/components/search-input/SearchInput.tsx
andoulla: This is what I was missing. I couldn't figure out why they wouldn't align because I didn't realise the size difference between `ActionIcon` and`Input`.

Matching the size solves the positioning
image

src/components/search-input/SearchInput.tsx
```diff
@@ -1,43 +1,103 @@
...
+      md: {
+        top: 10,
+        right: 10,
+        size: 20
+      }
+    }
+  }
+})
+
+const StyledSearchInput = styled(Input, {
```
thomasdigby, line 42: We need to apply the padding that we removed from the `Input`, without it, the text will overlap the icons:

### Before
Screenshot 2021-12-03 at 09 06 02

### After

https://user-images.githubusercontent.com/3226804/144575796-d942379e-01ac-459d-99a1-9db58890d44d.mov

src/components/search-input/SearchInput.tsx
```diff
@@ -1,43 +1,103 @@
...
+  }
+
+  const handleOnChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setInputValue(event.target.value)
+    setActiveIcon(event.target.value ? INPUT_ICON.CLEAR : INPUT_ICON.SEARCH)
+  }
+
+  const getIcon = () => {
+    if (activeIcon === INPUT_ICON.SEARCH)
+      return <StyledIcon is={Search} size={size} />
```
thomasdigby, line 74: We may need to add some overriding `css={}` here as the `size` variants from `StyledIcon` is clashing with the size variants from `Icon`.

Screenshot 2021-12-03 at 09 13 00
Screenshot 2021-12-03 at 09 13 07

This is likely a bug with stitches but for now, we can solve it by passing size in `css`

```tsx
<StyledIcon
  is={Search}
  size={size}
  css={{ size: size == 'sm' ? '$1' : 20 }}
/>
```

thomasdigby (commented):
There's two minor styling issues that we need to fix, otherwise I think this is looking good 👍

src/components/search-input/SearchInput.tsx
andoulla: it looks like it is rendering the Icon props after the StyledIcon props. I have added this. thanks for the workaround

thomasdigby (approved):

alexandrunemes (approved):

PR #275, opened 2021-12-13
feat: slider value label

kerryambt:
I needed a slider value label for work on the main site and realised that a slider is the only form component that doesn't explicitly state what its value is. I believe this isn't great as all inputs should probably inform the user of what input they're making. So I played around a bit and made this.

image
image
image
image

Works with the slider steps, has custom empty state, supports custom value placement and pluralisation. There's also simple use case as well with default values so you can simply pass `showValue` to the slider and it will work.

### DIFF
--- a/src/components/slider-field/SliderField.mdx
+++ b/src/components/slider-field/SliderField.mdx
@@ -5,7 +5,7 @@ description: Combines a Slider with a label and form integration.
 category: Form fields
 ---
 
-`SliderField` renders a `Slider` with a label, and easy implementation with `Form`.
+`SliderField` renders a `Slider` with a label, and easy implementation with `Form`. It also renders a `Slider.Value` label underneath the slider.
 
 In it's most simple form, a slider allows one value to be set between two given values (default 0 and 100).
 
@@ -14,5 +14,20 @@ Please note: the `value` or `defaultValue` passed in should always be an array.
 ```tsx preview
 <Form>
   <SliderField name="slider" label="Select a value" defaultValue={[50]} />
+</Form>
+```
+
+## Value Labels
+
+The `Slider.Value` output label is displayed by default with the `SliderField`. Should you wish to remove this, you can pass `showValue={false}`.
+
+```tsx preview
+<Form>
+  <SliderField
+    name="slider"
+    label="Select a value"
+    defaultValue={[50]}
+    showValue={false}
+  />
 </Form>
 ```
--- a/src/components/slider-field/SliderField.mdx
+++ b/src/components/slider-field/SliderField.mdx
@@ -5,7 +5,7 @@ description: Combines a Slider with a label and form integration.
 category: Form fields
 ---
 
-`SliderField` renders a `Slider` with a label, and easy implementation with `Form`. It also renders a `Slider.Value` label underneath the slider.
+`SliderField` renders a `Slider` with a label, and easy implementation with `Form`. It also renders a `Slider.Value` label underneath the slider, and a `Slider.Steps` component should steps be passed in.
 
 In it's most simple form, a slider allows one value to be set between two given values (default 0 and 100).
 
@@ -17,17 +17,19 @@ Please note: the `value` or `defaultValue` passed in should always be an array.
 </Form>
 ```
 
-## Value Labels
-
-The `Slider.Value` output label is displayed by default with the `SliderField`. Should you wish to remove this, you can pass `showValue={false}`.
+With optional `Slider.Steps`:
 
 ```tsx preview
 <Form>
   <SliderField
     name="slider"
-    label="Select a value"
+    label="Select a value (now with steps)"
     defaultValue={[50]}
-    showValue={false}
+    steps={[
+      { value: 0, label: 'min' },
+      { value: 50, label: 'mid' },
+      { value: 100, label: 'max' }
+    ]}
   />
 </Form>
 ```
--- a/src/components/slider-field/SliderField.tsx
+++ b/src/components/slider-field/SliderField.tsx
@@ -22,8 +22,6 @@ export const SliderField: React.FC<SliderFieldProps> = ({
   name,
   defaultValue,
   outputLabel,
-  emptyData,
-  item,
   min = 0,
   max = 100,
   steps = [],
@@ -50,8 +48,6 @@ export const SliderField: React.FC<SliderFieldProps> = ({
             <Slider.Value
               value={value || defaultValue}
               outputLabel={outputLabel}
-              emptyData={emptyData}
-              item={item}
             />
           </Slider>
         )}
--- a/src/components/slider-field/SliderField.tsx
+++ b/src/components/slider-field/SliderField.tsx
@@ -3,20 +3,27 @@ import { Controller, useFormContext } from 'react-hook-form'
 
 import { FieldWrapper } from '~/components/field-wrapper'
 import { Slider, SliderProps } from '~/components/slider'
+import { SliderValueType } from '../slider/SliderValue'
 import type { CSS } from '~/stitches'
 
-type SliderFieldProps = SliderProps & {
-  css?: CSS
-  label: string
-  name: string
-  defaultValue: number[]
-}
+type SliderFieldProps = SliderProps &
+  SliderValueType & {
+    css?: CSS
+    label: string
+    name: string
+    defaultValue: number[]
+    showValue?: boolean
+  }
 
 export const SliderField: React.FC<SliderFieldProps> = ({
   css,
   label,
   name,
   defaultValue,
+  outputLabel,
+  emptyData,
+  item,
+  showValue = true,
   ...remainingProps
 }) => {
   const { control } = useFormContext()
@@ -28,13 +35,23 @@ export const SliderField: React.FC<SliderFieldProps> = ({
         name={name}
         defaultValue={defaultValue}
         render={({ onChange, value }) => (
-          <Slider
-            name={name}
-            defaultValue={defaultValue}
-            onValueChange={onChange}
-            value={value}
-            {...remainingProps}
-          />
+          <>
+            <Slider
+              name={name}
+              defaultValue={defaultValue}
+              onValueChange={onChange}
+              value={value}
+              {...remainingProps}
+            />
+            {showValue && (
+              <Slider.Value
+                value={value || defaultValue}
+                outputLabel={outputLabel}
+                emptyData={emptyData}
+                item={item}
+              />
+            )}
+          </>
         )}
       />
     </FieldWrapper>
--- a/src/components/slider-field/SliderField.tsx
+++ b/src/components/slider-field/SliderField.tsx
@@ -3,16 +3,17 @@ import { Controller, useFormContext } from 'react-hook-form'
 
 import { FieldWrapper } from '~/components/field-wrapper'
 import { Slider, SliderProps } from '~/components/slider'
+import { SliderStepsType } from '~/components/slider/SliderSteps'
 import { SliderValueType } from '../slider/SliderValue'
 import type { CSS } from '~/stitches'
 
 type SliderFieldProps = SliderProps &
+  SliderStepsType &
   SliderValueType & {
     css?: CSS
     label: string
     name: string
     defaultValue: number[]
-    showValue?: boolean
   }
 
 export const SliderField: React.FC<SliderFieldProps> = ({
@@ -23,7 +24,9 @@ export const SliderField: React.FC<SliderFieldProps> = ({
   outputLabel,
   emptyData,
   item,
-  showValue = true,
+  min = 0,
+  max = 100,
+  steps = [],
   ...remainingProps
 }) => {
   const { control } = useFormContext()
@@ -35,23 +38,22 @@ export const SliderField: React.FC<SliderFieldProps> = ({
         name={name}
         defaultValue={defaultValue}
         render={({ onChange, value }) => (
-          <>
-            <Slider
-              name={name}
-              defaultValue={defaultValue}
-              onValueChange={onChange}
-              value={value}
-              {...remainingProps}
+          <Slider
+            name={name}
+            defaultValue={defaultValue}
+            onValueChange={onChange}
+            value={value}
+            {...remainingProps}
+          >
+            <Slider.Steps min={min} max={max} steps={steps} />
+
+            <Slider.Value
+              value={value || defaultValue}
+              outputLabel={outputLabel}
+              emptyData={emptyData}
+              item={item}
             />
-            {showValue && (
-              <Slider.Value
-                value={value || defaultValue}
-                outputLabel={outputLabel}
-                emptyData={emptyData}
-                item={item}
-              />
-            )}
-          </>
+          </Slider>
         )}
       />
     </FieldWrapper>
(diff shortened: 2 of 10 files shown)


src/components/slider/Slider.mdx
```diff
@@ -60,6 +60,31 @@ The steps work well with the built in `step` property, which defaults to 1 and c
...
 ```
 
+## Value output
+
+The slider can be set to display a label below the track displaying the current value selected. To do this, pass in the prop `showValue`.
+
+NOTE: this will only work if there is only one value selected. In order to not overly complicate the output, it is recommended to indicate multiple values using `steps` (see above). Should you pass in this prop with multiple values, the value output will not be rendered.
+
+```tsx preview
+<Slider defaultValue={[50]} showValue />
```
shaneakins, line 70: How come you're using an array here for the default value?

src/components/slider/Slider.mdx
```diff
@@ -60,6 +60,31 @@ The steps work well with the built in `step` property, which defaults to 1 and c
...
+```
+
+By default this simply outputs the value, however this can be customised in a number of ways. `outputLabel` can be passed, which is a string that must contain the `$VALUE`. This will be replaced with the value when it renders. A string is used so that the value can be placed wherever it needs to be in the output sentence, such as in the example below.
+
+The `outputLabel` also will turn the string `$S` into conditional plurals, i.e. 's' will be output if the value is greater than 1. This is also shown in the example review.
```
shaneakins, line 75: What happens if the plural is nonstandard, e.g., _child/children, pony/ponies, sheep/sheep_? Maybe we should have a prop to indicate what the plural form should be?

src/components/slider/SliderSteps.tsx
```diff
@@ -1,11 +1,12 @@
 import React from 'react'
```
shaneakins, line 1: In TSX files, this should be `import * as React from 'react' `

src/components/slider/SliderValue.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react'
```
shaneakins, line 1: In TSX files, this should be `import * as React from 'react' `

src/components/slider/Slider.mdx
kerryambt: values for the slider always have to be an array, even if there's only one. the slider extends the radix slider primitive and that's the format that uses. I know it reads a bit weird but it allows for multiple thumbs on the slider.

shaneakins: Ahh good to know - thx!

src/components/slider/SliderSteps.tsx
kerryambt: did not know this but it makes sense - thanks for pointing this out :)

src/components/slider/Slider.mdx
kerryambt: fixed! great example of an edge case that's easily missed :)

src/components/slider/SliderValue.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export const SliderValue: React.FC<SliderValueProps> = ({
+  value = [],
+  outputLabel = 'Current value is $VALUE',
+  emptyData,
+  item = { single: '', plural: '' }
+}) => {
+  if (value.length > 1) return null
+
+  const outputValue = value.length > 0 ? value[0] : ''
```
Mhoog, line 24: Is there a reason why you'd want to return `null` in case there is more than 1 value, and `''` when there are 0 (or fewer) values? 

Would it be possible to combine these two?

```ts
if (value.length !== 1) return null
const outputValue = value[0] 
```

I think that would be easier to understand.

kerryambt: made these changes at different times and it slipped my attention that they could be combined - thanks, I'll change now :)

hollg:
I think providing a standardised way of displaying this value is 💯 and the logic in this new component seems like quite an elegant solution (it seems to me like it should work with other languages via our normal `i18n` utils, which is a big plus!), but I have concerns about API, which is decidedly _not_ using component composition. 

The fact that the current implementation adds 4 new optional props to `Slider` but all of them get passed straight to a nested component is a code smell in my opinion; it highlights the fact that those values aren't relevant to `Slider`'s intended single responsibility.

I'd much prefer that we expose this functionality as a separate component following the pattern established with our `Table`, `Dialog`, etc — something like `Slider.Value`.

I imagine that we will normally want to display the value when we use a `Slider`, but to me this feels like the situation we have with `Input`s and `Labels` — we always want the latter with the former, but we've been very disciplined about building them separately so that they can be composed in different ways. This allows us to provide components like `TextField`, which composes them together. I see that we already have a `SliderField`; could we achieve something similar here by including the value display in `SliderField` but not `Slider` itself?

src/components/slider/Slider.test.tsx
```diff
@@ -33,3 +37,31 @@ describe('Slider component', () => {
...
+  it('renders with a custom value label', async () => {
+    render(<Slider.Value value={[50]} outputLabel={outputLabel} />)
+
+    expect(await screen.findByText('50 test items')).toBeVisible()
+  })
+
+  it('renders with a custom empty value label', async () => {
+    render(<Slider.Value value={[0]} emptyData={emptyData} />)
+
+    expect(await screen.findByText('No test items')).toBeVisible()
```
hollg, line 57: I think this very tiny change would make the relationship between the input and expected output a little more explicit:
```suggestion
    expect(await screen.findByText(emptyData.label)).toBeVisible()
```

kerryambt: oh yeah, that makes a lot more sense! done.

Mhoog (approved):

src/components/slider-field/SliderField.tsx
```diff
@@ -3,20 +3,27 @@ import { Controller, useFormContext } from 'react-hook-form'
...
+    showValue?: boolean
+  }
 
 export const SliderField: React.FC<SliderFieldProps> = ({
   css,
   label,
   name,
   defaultValue,
+  outputLabel,
+  emptyData,
```
andoulla, line 24: what does emptyData is used for?

src/components/slider/Slider.tsx
```diff
@@ -91,12 +94,14 @@ export const Slider: React.FC<SliderProps> = ({
         <StyledTrack>
           <StyledRange />
         </StyledTrack>
-        {thumbs?.length && thumbs.map((_, i) => <StyledThumb key={i} />)}
+        {values?.length && values.map((_, i) => <StyledThumb key={i} />)}
```
andoulla, line 97: Doesn't this `key={i}` throw an error? React doesn't like using index as a key as it causes rendering issues

kerryambt: oddly no :o that's a good point. maybe the sandbox isn't running strict mode? at least it should be a warning. I've changed it anyway :)

src/components/slider-field/SliderField.mdx
```diff
@@ -16,3 +16,18 @@ Please note: the `value` or `defaultValue` passed in should always be an array.
...
+
+The `Slider.Value` output label is displayed by default with the `SliderField`. Should you wish to remove this, you can pass `showValue={false}`.
+
+```tsx preview
+<Form>
+  <SliderField
+    name="slider"
+    label="Select a value"
+    defaultValue={[50]}
+    showValue={false}
```
thomasdigby, line 30: I wonder if we should be strict about this as a "field" component and make an opinionated decision about the best UX for the slider in a form. Should we remove this conditional option and always render the value? Do you have any use-cases where we need to hide it?

thomasdigby (commented):
Thanks for working on this, definitely required for our `SliderField` and a neat solution for some of the labelling issues. A few minor points that may change/simplify the API a bit so would be good to hear your thoughts on my suggestions

src/components/slider/SliderValue.tsx
```diff
@@ -0,0 +1,35 @@
...
+  emptyData,
+  item = { single: '', plural: '' }
+}) => {
+  if (value.length !== 1) return null
+
+  const outputValue = value[0]
+
+  const outputString = outputLabel
+    .replace('$VALUE', outputValue.toString())
+    .replace('$ITEM', outputValue === 1 ? item?.single : item?.plural)
```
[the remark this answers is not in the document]
kerryambt: this sounds interesting but I'm not sure I understand how this fits in with the way the component works at the moment. Surely if you have `<SliderField {...props}/>` you can't just pass in something like `<SliderField label={(value) => 'value is ${value}'} {...rest}/>` (with template literal used obvs) and then do something like `const outputLabel = label()`, becausde it'd return `value is undefined`. Plus there's probably some added complexity with the value having to be in an array. SliderField doesn't know what value is unless we expect the use case of this to be people using a custom render of Form and watch the value or something, but that's clunky imo.

unless I'm missing something in which case please prod me in the right direction. I hadn't considered this direction and it's probably better so I'm keen to get it working if it's possible :)

src/components/slider-field/SliderField.mdx
kerryambt: my gut says 'you probably always want it' and I can't think of a reason not to want it, but perhaps there is one somewhere? I feel like making it an option is clinging to unneeded complexity now that I think of it. I'm going to make it always on :)

src/components/slider-field/SliderField.tsx
```diff
@@ -28,13 +35,23 @@ export const SliderField: React.FC<SliderFieldProps> = ({
...
+            />
+            {showValue && (
+              <Slider.Value
+                value={value || defaultValue}
+                outputLabel={outputLabel}
+                emptyData={emptyData}
+                item={item}
+              />
+            )}
+          </>
```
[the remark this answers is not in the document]
kerryambt: I think changing this and the SliderSteps to be `Slider.Steps` works really well. done :)

src/components/slider/SliderValue.tsx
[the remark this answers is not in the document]
thomasdigby: Do you have time today to chat about this? Think it's worth discussing in person just so we're on the same page

src/components/slider/SliderValue.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+type SliderValueProps = SliderValueType & {
+  value?: number[]
+}
+
+export const SliderValue: React.FC<SliderValueProps> = ({
+  value = [],
+  outputLabel = (value) => `Current value is ${value}`
+}) => {
+  if (value.length !== 1) return null
```
thomasdigby, line 17: How would we label multiple values? Should we look to return `value` regardless of the values within the array? Then you could render "10, 50 selected" e.g.

```tsx
<Slider defaultValue={[10, 50]}>
  <Slider.Value outputLabel={(value) => `${value.join(', ')} selected`} />
</Slider>
```

src/components/slider/SliderValue.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react'
+
+import { Text } from '../text'
+
+export type SliderValueType = {
+  outputLabel?: (value) => string
```
thomasdigby, line 6:
```suggestion
  outputLabel?: (value: number) => string
```

src/components/slider/Slider.tsx
```diff
@@ -64,20 +65,24 @@ const StyledThumb = styled(Thumb, {
   '&[data-disabled]': { bg: '$tonal200', cursor: 'not-allowed' }
 })
 
-export type SliderProps = React.ComponentProps<typeof StyledSlider> &
-  SliderStepsType
+export type SliderPointType = { label: string; value: number }
```
thomasdigby, line 68: Is this still used? I can't find it referenced in this PR

src/components/slider/Slider.mdx
```diff
@@ -1,13 +1,13 @@
...
-component: Slider
+component: Slider, Slider.Value, Slider.Steps
 description: The slider component implements the Slider component from Radix, with default styling and the CSS prop.
 category: Form primitives
 ---
 
 The `Slider` is a simple component that renders a form slider.
 
-The `Slider` component should be rendered with a label for accesibility reasons. Rather than using the `Slider` component directly, it is best to use the `SliderField` component, which provides a label and easily integrates with the `Form` component.
+The `Slider` component should be rendered with a label for accesibility reasons. Rather than using the `Slider` component directly, it is best to use the `SliderField` component, which provides a label and easily integrates with the `Form` component. It also adds the `Slider.Value` by default (with options to remove) in order to create a unified experience without bloating the individual component.
```
thomasdigby, line 10:
```suggestion
The `Slider` component should be rendered with a label for accesibility reasons. Rather than using the `Slider` component directly, it is best to use the `SliderField` component, which provides a field label, a value label, and easily integrates with the `Form` component.
```

Just a thought, to simplify a touch

src/components/slider/Slider.mdx
```diff
@@ -25,38 +25,75 @@ Should you wish to have more than one control on the slider, you can pass those
...
+  />
+</Slider>
 ```
 
-These values should be limited to the `min` and `max` values that can be passed in to the slider object, as it is not limited to a 0-100 range.
+The component requires min and max values, these should be the same as the optional values that are passed to the Slider component.
+
+`Slider.Steps` work well with the built in `step` property, which defaults to 1 and changes the size of each movement. For example, this would limit the slider to three values only.
+
+```tsx preview
```
thomasdigby, line 54:
```suggestion
```tsx
```

Let's not add too many rendered previews to the docs, a code example here is fine 👍

src/components/slider/Slider.mdx
```diff
@@ -25,38 +25,75 @@ Should you wish to have more than one control on the slider, you can pass those
...
+
+A separate component exists to output the value from a slider. This is most commonly used within the `SliderField`, but is here in case you need to compose your own complex component.
+
+NOTE: this will only work if there is only one value selected. In order to not overly complicate the output, it is recommended to indicate multiple values using `steps` (see above). If more than one value is passed in, the component will not render.
+
+```tsx preview
+<Slider.Value value={[50]} />
+```
+
+By default this simply outputs the value, however this can be customised in a number of ways. The property `outputLabel` accepts a callback function that takes a value and returns that in a string.
```
thomasdigby, line 78:
```suggestion
By default this simply outputs the value, however, this can be customised in a number of ways. The property `outputLabel` accepts a function that passes the value selected in the slider and expects a string returned for the label.
```

src/components/slider/Slider.mdx
```diff
@@ -25,38 +25,75 @@ Should you wish to have more than one control on the slider, you can pass those
...
+
+```tsx preview
+<Slider.Value
+  value={[50]}
+  outputLabel={(value) => `Current number of geese is ${value}`)}
+/>
+```
```
thomasdigby, line 85: I'd remove this example and rely on the following code block to communicate the use of value and pluralisation.

src/components/slider/Slider.mdx
```diff
@@ -25,38 +25,75 @@ Should you wish to have more than one control on the slider, you can pass those
...
 
-The steps work well with the built in `step` property, which defaults to 1 and changes the size of each movement. For example, this would limit the slider to three values only.
+You can also use this is in more complex ways, for example to set empty states and pluralisation, like below.
 
 ```tsx preview
```
thomasdigby, line 89: I don't think a preview is useful here as the value can't change, so you won't be able to see the pluralisation

src/components/slider/SliderValue.tsx
kerryambt: yeah good point on join - done :)

src/components/slider/SliderValue.tsx
kerryambt: given the join this is now `(value: number | number[]) => string` :)

src/components/slider/Slider.tsx
kerryambt: it's only used in one place so I'm gonna put it into that relevant file and remove the export

thomasdigby (approved):
Nice. 🚀

PR #276, opened 2021-12-14
feat: filter tabs component

kerryambt:
It was pointed out that using tabs without using tabs.content is not okay from an accessibility perspective, but I needed to do that when fetching data from an API. I made a component in generic components to mimic tab.triggerlists, but have since realised I need it in a bunch of places and that it makes sense to have it here. So, here is the FilterTabs component.

image

It borrows things from tabs, such as the scroll functionality, but otherwise stays as simple as possible.


andoulla (commented):
I'm just thinking out loud but is this something that belongs in the DS library? It feels very specific to the way we do filtering. Shouldn't this be part of the `_design_system` folder instead?

src/components/filter-tabs/FilterTabs.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { axe } from 'jest-axe'
+
+import { FilterTabs } from './'
+
+const filters = ['tab1', 'tab2', 'tab3']
+
+describe('FilterTabs component', () => {
+  it('renders', async () => {
+    const { container } = await render(
+      <FilterTabs filters={filters} value="tab1" onValueChange={() => null} />
```
andoulla, line 12: null works just find but I think we have been using jest.fn() in most places

src/components/filter-tabs/FilterTabs.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+    )
```
andoulla, line 13: This is a nitpick, but defining a const for all the props makes it simpler to maintain the tests. in the case of adding more props to the component we only need to add it to the const.

```
const props = {
filters= [tab1, tab2, tab3]
value = tab1
onValueChange={jest.fn()}
}

it('renders', async () => {
    const { container } = await render(
      <FilterTabs {...props} />

```

We can always overwrite any value that needs customisation

src/components/filter-tabs/FilterTabs.tsx
```diff
@@ -0,0 +1,167 @@
...
+const FilterTabsItem = styled(Flex, {
+  fontFamily: '$body',
+  p: '$4',
+  height: '$5',
+  flexShrink: '0',
+  '&:hover': {
+    bg: `${fadedPrimaryDark}`
+  },
+  '&[data-disabled=true]': {
+    bg: 'unset',
```
andoulla, line 42: Maybe disabled could be another variant?

src/components/filter-tabs/FilterTabs.tsx
```diff
@@ -0,0 +1,167 @@
...
+      tabsList.scroll?.({
+        left: 0
+      })
+      setShowLeftScroller(false)
+      setShowRightScroller(shouldShowScroller)
+    }
+  }, [screenWidth])
+
+  return (
+    <FilterTabsOuter>
```
andoulla, line 135: Maybe this could be `FilterTabsContainer` and the container could be `FilterTabsList` as it has  a list of Items

andoulla: Generally I would use container for the outmost containing div

src/components/filter-tabs/FilterTabs.test.tsx
kerryambt: I think it's a very sensible nitpick if it is a nitpick! I'll change now :)

src/components/filter-tabs/FilterTabs.tsx
kerryambt: I wasn't sure which was best, disabled didn't seem like a variant in the strictest sense so I went with a data attribute. not sure tbh :o

src/components/filter-tabs/FilterTabs.tsx
kerryambt: yeah that's a better name - will change now

src/components/filter-tabs/FilterTabs.test.tsx
kerryambt: noted! thanks :)

src/components/filter-tabs/FilterTabs.tsx
andoulla: I went through a few components and this is how we do it so lets ignore this comment

src/components/filter-tabs/FilterTabs.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+import { FilterTabsButton } from './FilterTabsButton'
+
+const fadedPrimaryDark = opacify(-0.9, theme.colors.primary.value)
+
+type FilterTabsType = {
+  filters: string[]
+  value?: string
+  defaultValue?: string
+  onValueChange: (value) => void
```
avirati, line 17: `value` is loosely typed here.

src/components/filter-tabs/FilterTabs.tsx
```diff
@@ -0,0 +1,167 @@
...
+    selected: {
+      true: {
+        fontWeight: '600',
+        color: '$primary'
+      }
+    }
+  }
+})
+
+export const FilterTabs: React.FC<FilterTabsType> &
```
avirati, line 56: Some of the code is identical to `src/components/tabs/TabsTriggerList.tsx`
Do we have any possibility of reusing that code ?

src/components/filter-tabs/FilterTabsButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+  '&:hover': { bg: 'rgba(255, 255, 255, 0.8) !important' },
+  position: 'absolute',
+  zIndex: '2',
+  top: '0'
+})
+
+type FilterTabsButtonProps = {
+  side: string
+  visible: boolean
+  onClick: (direction) => void
```
avirati, line 19:
```suggestion
  onClick: (direction: string) => void
```

src/components/filter-tabs/FilterTabsButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+const FilterTabsIconContainer = styled(ActionIcon, {
+  bg: 'rgba(255, 255, 255, 0.6) !important',
+  '&:hover': { bg: 'rgba(255, 255, 255, 0.8) !important' },
+  position: 'absolute',
+  zIndex: '2',
+  top: '0'
+})
+
+type FilterTabsButtonProps = {
+  side: string
```
avirati, line 17: Would prefer `side` to come from `Enum`
```typescript
Enum Side {
  LEFT = 'left',
  RIGHT = 'right'
}

type FilterTabsButtonProps = {
  side: Side
  ...
  
  ```

File: src/components/filter-tabs/FilterTabsButton.tsx (hunk 7 above)
Comment (avirati) [reply src/components/filter-tabs/FilterTabsButton.tsx:19]: Or 
`onClick: (direction: Side) => void`, if you choose to go in that direction

File: src/components/filter-tabs/FilterTabs.tsx (hunk 5 above)
Comment (kerryambt) [reply src/components/filter-tabs/FilterTabs.tsx:17]: oops! good catch

File: src/components/filter-tabs/FilterTabsButton.tsx (hunk 8 above)
Comment (kerryambt) [reply src/components/filter-tabs/FilterTabsButton.tsx:17]: yeah this is much cleaner, good idea :)

File: src/components/filter-tabs/FilterTabsButton.tsx (hunk 7 above)
Comment (kerryambt) [reply src/components/filter-tabs/FilterTabsButton.tsx:19]: have gone with the Side idea, but yeah good point again :)

File: src/components/filter-tabs/FilterTabs.tsx (hunk 6 above)
Comment (kerryambt) [reply src/components/filter-tabs/FilterTabs.tsx:56]: I don't know if we can. the only possible code I can see is the scroll function, but that's a callback and needs to be a callback for the event listeners to work properly. and if it's a callback it needs to be in a component, so it can't be exported. do you have any other ideas? happy to reuse if possible, maybe I'm missing something :)

Comment (avirati) [reply src/components/filter-tabs/FilterTabs.tsx:56]: Thanks for the explanation, I see your point. Alright, happy to approve !

**Review by avirati — APPROVED** [review]

File: src/components/filter-tabs/FilterTabs.tsx [hunk 9]
```diff
@@ -0,0 +1,167 @@
...
+    }
+  }, [screenWidth])
+
+  return (
+    <FilterTabsContainer>
+      <FilterTabsList {...remainingProps} ref={filterTabsRef}>
+        {filters.map((filter) => (
+          <FilterTabsItem
+            selected={filter === (value || currentValue)}
+            onClick={() => {
```
Mhoog, line 140: I think this would be a good candidate to extract into a separate `handleTabClick` function.

And, very minor, but we usually try to use an early return where possible:

```ts
const handleTabClick = (filter) => {
  if(disabled) return

  onValueChange(filter)
  setCurrentValue(filter)
}
```

src/components/filter-tabs/FilterTabs.tsx
```diff
@@ -0,0 +1,167 @@
...
+          setShowLeftScroller(true)
+        } else {
+          setShowLeftScroller(true)
+          setShowRightScroller(true)
+        }
+      }, 500)
+    }
+  }, [])
+
+  useEffect(() => {
```
Mhoog, line 106: I think it would make the component easier to read if this was extracted into a `useScreenSize` hook, so that you could use:

```ts
const { screenWidth, screenHeight } = useScreenSize()
```

src/components/filter-tabs/FilterTabs.tsx
kerryambt: good point! done :)

thomasdigby:
@kerryambt I'd agree with @andoulla here, I don't think this should live within the components library. I wouldn't necessarily be against a "Filter" component of sorts that allows a single selection from a list of filterable terms, but I'd prefer for that to be unique and distinct from the tabs interface. There's a lot of expected behaviour baked into the tabs design that I don't think is right to define here as a "Filter"

PR #277, opened 2021-12-15
feat: make it possible to show / hide close button in modal

avirati:
Currently, it is not possible to manipulate the close button in the modal. Adding a prop to `DialogContent` to make it possible. Snapshots in PR for reference

### DIFF
--- a/src/components/dialog/DialogContent.tsx
+++ b/src/components/dialog/DialogContent.tsx
@@ -69,20 +69,20 @@ const StyledDialogContent = styled(Content, {
 
 type DialogContentProps = React.ComponentProps<typeof StyledDialogContent> & {
   closeDialogText?: string
-  showCloseIcon?: boolean
+  showCloseButton?: boolean
 }
 
 export const DialogContent: React.FC<DialogContentProps> = ({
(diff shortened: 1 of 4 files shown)


hollg:
Could we add some documentation to the effect that any modal without a close icon _must_ have an action button that closes the modal? We need something that screen readers can see.

thomasdigby (approved):

src/components/dialog/DialogContent.tsx
```diff
@@ -69,26 +69,30 @@ const StyledDialogContent = styled(Content, {
...
-      <ActionIcon
-        as={Close}
-        css={{ position: 'absolute', right: '$4', top: '$4' }}
-        label={closeDialogText}
-        size="lg"
-        theme="neutral"
-      >
-        <Icon is={CloseIcon} />
-      </ActionIcon>
+      {showCloseIcon && (
```
thomasdigby, line 85: To align with the upcoming change to `Popover` within the `DateInput` PR (#221), could we change this prop name to `showCloseButton`?

thomasdigby (approved):

PR #281, opened 2022-01-11
feat: remove :before and :after for nested Text components

Mhoog:
## Issue
When nesting `<Text />` elements, it would always render the nested component on a new line, including spacing set on the `<Text />` element. This is usually not what you want when you nest `<Text />` elements, as the most common cases are with using inline HTML elements such as `span`, `strong`, `em`, etc.

### For code
```tsx
<Text>This is text, <Text as="strong">and more text</Text></Text>
```

### Before
image

### After
image

### DIFF
--- a/src/components/calendar/__snapshots__/Calendar.test.tsx.snap
+++ b/src/components/calendar/__snapshots__/Calendar.test.tsx.snap
@@ -43,11 +43,16 @@ exports[`Calendar component renders 1`] = `
     grid-gap: var(--space-1) var(--space-2);
   }
 
-  .c-dbleUi {
+  .c-fIXJvv {
     color: var(--colors-tonal600);
     font-family: var(--fonts-body);
     font-weight: 400;
     margin: 0;
+  }
+
+  .c-fIXJvv > .c-fIXJvv:before,
+  .c-fIXJvv > .c-fIXJvv:after {
+    display: none;
   }
 
   .c-epywoc {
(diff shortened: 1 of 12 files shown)


andoulla:
Does this break the capsize functionality?

hollg:
Just for my own understanding, is the benefit of _not_ doing `<Text>This is text, <strong>and more text</strong></Text>` just ensuring that we're using the interfaces provided by this library, or is there some functional benefit too?

Mhoog:
> Does this break the capsize functionality?

@andoulla No, the parent element still has the capsize functionality:
image

Mhoog:
> Just for my own understanding, is the benefit of _not_ doing `<Text>This is text, \<strong\>and more text\</strong\></Text>` just ensuring that we're using the interfaces provided by this library, or is there some functional benefit too?

@hollg, correct. The main benefit of using the `Text` component instead of `<strong />` is that it is then easier to style the inner text. If you'd want to style the `<strong>`, you'd first have to create a `const StyledStrong = styled('strong')`.

hollg (approved):

hollg:
> > Just for my own understanding, is the benefit of _not_ doing `<Text>This is text, \<strong\>and more text\</strong\></Text>` just ensuring that we're using the interfaces provided by this library, or is there some functional benefit too?
> 
> @hollg, correct. The main benefit of using the `Text` component instead of `<strong />` is that it is then easier to style the inner text. If you'd want to style the `\<strong\>`, you'd first have to create a `const StyledStrong = styled('strong')`.

Makes sense to me, thanks!

andoulla (approved):

PR #282, opened 2022-01-12
feat: tabs trigger list flex container

kerryambt:
### Description

I needed a way to make tab trigger lists work with flex, so that we could have items at the end of a tab trigger list (or start) that remain there, and do not affect the area for scrollable tabs. (i.e. the buttons aren't overlapping the item you've placed at the end).

This was almost entirely possible by adding CSS, but the wrapper you would need to add to allow it would mean theme properties weren't passed down properly to trigger lists or trigger items. So I made a simple wrapper component that handles the flex and passes the properties down to its children.

```tsx
<Tabs>
  <Tabs.TriggerListFlexContainer>
    <Tabs.TriggerList>
      <Tabs.Trigger></Tabs.Trigger>
      <Tabs.Trigger></Tabs.Trigger>
      <Tabs.Trigger></Tabs.Trigger>
    </Tabs.TriggerList>
      WHATEVER CONTENT YOU WANT HERE
  </Tabs.TriggerListFlexContainer>
...
</Tabs>
```

This wrapper is completely optional and the tabs function as normal without it. It just allows flex positioning for complex designs should you need them.

### Screenshots

image
image

### DIFF
--- a/src/components/tabs/Tabs.mdx
+++ b/src/components/tabs/Tabs.mdx
@@ -5,7 +5,7 @@ description: Tabs is a component that provides different sections of content tha
 category: Layout
 ---
 
-The component provides a set of default styles, which can be overwritten by using the `css` prop. It is composed by combining smaller components, such as `Tabs.TriggerList`, `Tabs.Trigger`, and `Tabs.Content`.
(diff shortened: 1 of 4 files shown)


kerryambt:
@thomasdigby if you use Flex the part of the Tabs component that passes down props doesn't see the further nested props and therefore TabsTriggerList and TabsContent don't get passed props like theme. This gives us a little wrapper that passes those props down if needed :)

kerryambt:
have decided to go in a different direction with this, closing for now

PR #284, opened 2022-01-21
fix: slider min max bug

kerryambt:
### Description

The recent change to Slider that added the Slider.Steps component accidentally composed the SliderField in such a way that `min` and `max` values weren't being passed to the Slider. These have now been explicitly passed down.

Also changing steps property to properly be optional in order to remove typescript errors.


kerryambt:
@thomasdigby yeah good point, I'll make one :)

thomasdigby (approved):

Mhoog (approved):

PR #286, opened 2022-01-24
Fix: SearchInput onChange

thomasdigby:


kerryambt (approved):

alexandrunemes (approved):

src/components/search-input/SearchInput.test.tsx
```diff
@@ -17,6 +17,17 @@ describe('SearchInput component', () => {
     expect(container).toMatchSnapshot()
   })
 
+  it('clears text on button click', async () => {
```
andoulla, line 20: I don't think the description matches what the test does. The test asserts that it calls the passed function.

```suggestion
  it('calls the onChange prop on button click', async () => {
```

thomasdigby: Ah yeah, forgot to update that

andoulla (approved):

PR #288, opened 2022-01-25
fix: allow external change handler on date input

alexandrunemes:
Description

The `DateInput` wasn't taking any external change handlers, so if you needed to do something with the set value, there was no reliable way. This adds an optional `handleChange` prop to the component, which gets called inside the `onDateSelected` callback.

### DIFF
--- a/src/components/date-field/DateField.tsx
+++ b/src/components/date-field/DateField.tsx
@@ -13,7 +13,6 @@ type DateFieldProps = DateInputProps & {
   name: string
   prompt?: { link: string; label: string }
   validation?: ValidationOptions
-  handleChange?: (value: Date) => void
 }
 
 export const DateField: React.FC<DateFieldProps> = ({
--- a/src/components/date-field/DateField.tsx
+++ b/src/components/date-field/DateField.tsx
@@ -13,6 +13,7 @@ type DateFieldProps = DateInputProps & {
   name: string
   prompt?: { link: string; label: string }
   validation?: ValidationOptions
+  handleChange?: (value: Date) => void
 }
 
 export const DateField: React.FC<DateFieldProps> = ({
--- a/src/components/date-input/DateInput.tsx
+++ b/src/components/date-input/DateInput.tsx
@@ -18,7 +18,7 @@ export type DateInputProps = Omit<DayzedInterface, 'onDateSelected'> &
     disabled?: boolean
     size?: 'sm' | 'md'
     revalidate?: () => Promise<boolean>
-    handleChange?: (value: Date) => void
+    onChange?: (value?: Date) => void
   }
 
 export const DateInput = React.forwardRef<HTMLInputElement, DateInputProps>(
@@ -37,7 +37,7 @@ export const DateInput = React.forwardRef<HTMLInputElement, DateInputProps>(
         previous: 'Previous month'
       },
       revalidate,
-      handleChange,
+      onChange,
       ...remainingProps
     },
     ref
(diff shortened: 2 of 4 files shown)


src/components/date-field/DateField.tsx
```diff
@@ -13,6 +13,7 @@ type DateFieldProps = DateInputProps & {
   name: string
   prompt?: { link: string; label: string }
   validation?: ValidationOptions
+  handleChange?: (value: Date) => void
```
kerryambt, line 16: do you need to declare this prop here if you're adding it to DateInputProps? DateFieldProps contains all of DateInputProps

src/components/date-input/DateInput.tsx
```diff
@@ -91,6 +93,7 @@ export const DateInput = React.forwardRef<HTMLInputElement, DateInputProps>(
                 setCalendarOpen(false)
                 await setDate(date.date, false)
                 if (revalidate) revalidate()
+                handleChange?.(date.date)
```
thomasdigby, line 96: Could we mirror the name of `onDateSelected` or use `onChange`? We haven't used `handleThing` as a convention for interaction events

src/components/date-field/DateField.tsx
alexandrunemes: No, I don't. I'll remove it

alexandrunemes:
@thomasdigby @kerryambt , I realised the initial approach only handled the scenario where the user actually toggles the calendar and selects a date from there, and completely ignored users trying to manually type the date (I sometimes do that myself with date inputs, if I need to type in my birthdate and I don't want to scroll back for like an hour). Moving the callback into a useEffect hook seems to catch both scenarios.

src/components/date-input/DateInput.tsx
```diff
@@ -47,6 +49,18 @@ export const DateInput = React.forwardRef<HTMLInputElement, DateInputProps>(
     const refDateToday = React.useRef<HTMLButtonElement>(null)
     const refDateSelected = React.useRef<HTMLButtonElement>(null)
 
+    React.useEffect(() => {
+      let isMounted = true
+
+      if (isMounted) {
```
thomasdigby, line 55: Do you see this conditional being necessary even if we're not calling a potentially long-running function?

alexandrunemes: Meh, you're right. We're not doing anything with the result locally, so even if the component unmounts, it shouldn't cause any memory leaks.

src/components/date-input/DateInput.tsx
```diff
@@ -47,6 +49,18 @@ export const DateInput = React.forwardRef<HTMLInputElement, DateInputProps>(
...
+        onChange?.(date)
+      }
+
+      return () => {
+        isMounted = false
+      }
+    }, [date, onChange])
```
kerryambt, line 54: only thing that comes up as a potential issue is do we mind that onChange will be called with every character press? The input box is setting date when it changes. if you type '25/12/2022' or something does it matter that the character presses will all trigger onChange? Or are we deferring responsibility of that to the user in order to not assume too much here?

alexandrunemes: While testing, I noticed that, but managed to bypass it by doing a check on the value inside the callback. While the date isn't "complete" (valid DD-MM-YYYY or some other accepted format), the value that gets passed is `undefined`. 
I don't think it's going to be a problem. @thomasdigby ?

kerryambt: do you think it would be worth while doing a simple regex check on the date in useeffect, and if it's not valid don't call onChange? I can't imagine many use cases where that isn't helpful. again comes to the point of how much we want to defer to the user. I think if we defer it's worth noting in the mdx that this is likely a required thing to add. I think it's an interesting point of where the line is. personally I would check inside useeffect but I'm happy to approve either way as it doesn't seem there's a 'right' answer :)

thomasdigby: I think we should treat this as a regular `onChange` on an `input`, with the added benefit of it also being called when the user selects a date via the pop-up. Like you said @kerryambt, its a balance, but in this particular case I think its too intrusive to not return anything unless the date is valid. We do have the ability to validate the date-field through `react-hook-form` so that's what the user should use to ensure the date is valid.

kerryambt (approved):

thomasdigby (approved):

PR #291, opened 2022-01-27
Fix: Compact version of Calendar popover

thomasdigby:


src/components/calendar/Calendar.tsx
```diff
@@ -18,7 +18,12 @@ import { Day } from './Day'
 const Grid = styled('div', {
   display: 'grid',
   gridTemplateColumns: 'repeat(7, 1fr)',
-  gridGap: '$1 $2'
+  variants: {
+    layout: {
+      compact: { gridGap: '$1 $1' },
```
andoulla, line 23: My question, and it is more of wondering, is why have different variants instead of making the component adjust the gridGap based on the screensize.

But I do see the value in having the variants, so could we combine the variants with the screensize. Something like @sm: {layout:compact}

src/components/calendar/Calendar.tsx
```diff
@@ -90,7 +97,7 @@ export const Calendar: React.FC<CalendarProps> = ({
               {monthNames[month]} {year}
             </Heading>
           </Flex>
-          <Grid css={{ mb: '$3' }}>
+          <Grid layout={layout} css={{ mb: '$3' }}>
```
andoulla, line 100: do we need a snapshot for the different layouts?

thomasdigby: Maybe? I wonder if it might be duplicating already tested behaviour as part of Stitches

andoulla: I think we did this in different places (like the button). it is very minor to be honest

src/components/calendar/Calendar.tsx
thomasdigby: Variants allow parent components to adjust the size independently from viewport size. That gives us more flexibility for when we might want these styles to apply and still gives the parent component the ability to contextually apply that style in a small viewport.

src/components/calendar/Calendar.tsx
thomasdigby: Ill add one 👍

Tangentially I'd be keen to discuss this amongst some other high level changes to make things consistent across the DS

andoulla: Happy too, as I'm having second thoughts about whether snapshots are sufficient. but it is a conversation for face to face and not github

src/components/calendar/Calendar.tsx
andoulla: So we might want to show the comfortable version on a mobile size?

thomasdigby: More that we might want to show the compact version on a tablet/desktop. But really it's just giving us the flexibility to make that decision based on the context, rather than forcing it as part of the component.

thomasdigby:
@D7Torres Hello

thomasdigby:
@D7Torres Test test

D7Torres:
@thomasdigby test

PR #297, opened 2022-02-10
Progress indicator (stepper)

alexandrunemes:
A progress indicator that includes back/next buttons and numbered bullets to indicate steps.

Screenshot 2022-02-10 at 13 14 00

Screenshot 2022-02-10 at 13 14 06

https://user-images.githubusercontent.com/9009155/154486103-313aed07-1c47-42f8-9784-1741d360acfb.mov


src/components/progress-indicator/ProgressIndicator.tsx
```diff
@@ -0,0 +1,118 @@
...
+          )
+        })}
+      </Flex>
+
+      <Button
+        size="sm"
+        onClick={activeStep === steps.length - 1 ? handleFinal : goToNextStep}
+      >
+        {activeStep === steps.length - 1
+          ? labels?.start || 'Start'
```
andoulla, line 113: Would it be simpler to let the parent handle this bit? if they reached the last step they can render a different label if they need to.  This will simplify the component and also make it slightly more flexible (for example if we wanted to have "done" instead of "start" or if we wanted the next label to say "go to next step").

src/components/progress-indicator/ProgressIndicator.test.tsx
```diff
@@ -0,0 +1,54 @@
+import { fireEvent, render, screen } from '@testing-library/react'
+import { axe } from 'jest-axe'
+import * as React from 'react'
+
+import { ProgressIndicatorProvider } from '.'
+import { ProgressIndicator } from './ProgressIndicator'
+
+it('renders a bullet style progress indicator - has no programmatically detectable a11y issues', async () => {
```
andoulla, line 8: Can we add a `describe` that will group all the tests, please?

src/components/progress-indicator/ProgressIndicator.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const { container } = render(
```
andoulla, line 9: Can we also add a test that will generate a snapshot or if you prefer a test that will asses that it we pass an array of 3 steps it will create the corresponding structure with 3 bullets and the buttons

src/components/progress-indicator/ProgressIndicator.tsx
```diff
@@ -0,0 +1,118 @@
...
+        disabled={activeStep === 0}
+        onClick={goToPreviousStep}
+      >
+        {labels?.back || 'Back'}
+      </Button>
+      <Flex>
+        {steps.map((_, index) => {
+          return (
+            <Flex key={`step_${index}`} css={{ alignItems: 'center' }}>
+              <Flex
```
andoulla, line 65: Have you considered making this Flex with styled and then setting the different colours as a state variant?

```
const StyledBullet = styled(Flex, 
{
    cursor: allowSkip ? 'pointer' : 'auto',
     p: '$2',
     justifyContent: 'center',
    alignItems: 'center',
     width: '$2',
     height: '$2',
     borderRadius: '50%',
     border: 'none',
     variants: {
        state: {
              inactive: { bg: tonal50},
              active: {bg:.....}.
             viewed: {bg: ....}
       }
}
```

This will make it a bit more consistent with what we have done in other components but also it will make it easier to extend if we wanted to add more sizes or colours.

src/components/progress-indicator/progress-indicator-context/ProgressIndicatorContext.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+export const ProgressIndicatorProvider: React.FC<ProgressIndicatorProviderProps> = ({
+  children,
+  stepsData,
+  allowSkip,
+  labels
+}) => {
+  const [steps, setSteps] = React.useState<unknown[]>(stepsData)
+  const [activeStep, setActiveStep] = React.useState(0)
+  const [viewedSteps, setviewedSteps] = React.useState<number[]>([0])
```
andoulla, line 26:
```suggestion
  const [viewedSteps, setViewedSteps] = React.useState<number[]>([0])
```

minor typo

src/components/progress-indicator/progress-indicator-context/ProgressIndicatorContext.tsx
```diff
@@ -0,0 +1,74 @@
...
+const ProgressIndicatorContext = React.createContext<ProgressContext>({
+  steps: [],
+  setSteps: () => null,
+  goToNextStep: () => null,
+  goToPreviousStep: () => null,
+  goToStep: () => null,
+  activeStep: 0,
+  setActiveStep: () => null,
+  viewedSteps: [],
+  allowSkip: false,
```
andoulla, line 14: Do we need the allowSkip in the context? It doesn't affect the provider at all

src/components/progress-indicator/ProgressIndicator.tsx
thomasdigby: I wonder if we could go a step further here and export these individual parts of the "stepper"? We could avoid this labelling issue and be a bit more consistent with how we're composing some of our other DS components e.g.

```tsx
<ProgressIndicator>
  <ProgressIndicator.NavigatePrevious>
    {'Go back yo'}
  </ProgressIndicator.NavigatePrevious>
  <ProgressIndicator.Steps />
  <ProgressIndicator.NavigateNext>
    {'Go forward yo'}
  </ProgressIndicator.NavigateNext>
</ProgressIndicator>
```

src/components/progress-indicator/ProgressIndicator.tsx
thomasdigby: Agree with @andoulla, we should use `variants` as much as we can with our `styled()` components 👍

thomasdigby (commented):
I don't think the `ProgressIndicatorProvider` is worth exporting for the user, that's an implementation detail that we can use within `ProgressIndicator`. If we utilise component composition a bit more here, breaking out the steps and navigation buttons, then the `useProgressIndicator()` becomes more useful as well.

src/components/progress-indicator/ProgressIndicator.test.tsx
alexandrunemes: Oh, forgot about that :)

src/components/progress-indicator/ProgressIndicator.test.tsx
alexandrunemes: Fair point. Will add.

src/components/progress-indicator/ProgressIndicator.tsx
alexandrunemes: Thanks! Will update.

src/components/progress-indicator/ProgressIndicator.tsx
alexandrunemes: Oh, this will be much better! Also, love the labels. I should add them as fallback values 😂

src/components/progress-indicator/progress-indicator-context/ProgressIndicatorContext.tsx
alexandrunemes: It's just to keep most of the variations at context level. This way we won't need to pass `allowSkip` to the `Steps` component. It can just get it from the hook.

src/components/progress-indicator/ProgressIndicator.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { ProgressIndicatorSteps } from './ProgressIndicatorSteps'
+import { IProgressIndicatorProps } from './types'
+
+export const ProgressIndicator: React.FC<IProgressIndicatorProps> & {
+  NavigatePrevious: typeof ProgressIndicatorNavigatePrevious
+  NavigateNext: typeof ProgressIndicatorNavigateNext
+  Steps: typeof ProgressIndicatorSteps
+} = ({ children, stepsData, allowSkip }) => {
+  if (!stepsData?.length) {
+    return <Loader />
```
alexandrunemes, line 18: Not sure what to do when valid step data isn't provided. Any thoughts @thomasdigby , @andoulla ?

src/components/progress-indicator/ProgressIndicatorSteps.tsx
```diff
@@ -0,0 +1,94 @@
...
+    allowSkip
+  } = useProgressIndicator()
+
+  const handleStepClick = (index: number) => {
+    if (!allowSkip) return
+    goToStep(index)
+  }
+
+  const getBulletState = (index: number) => {
+    if (activeStep === index) return 'active'
```
andoulla, line 51: I would consider this as more readable, what do you think?

```
if (activeStep === index) return 'active'
if (viewedSteps.includes(index)) return 'viewed'

return "disabled"

```

I would make disabled a state for readability purposes. Also, for consistency as in other components we set all the variants in the variant array and then default one

alexandrunemes: But it's not actually disabled. It's just...`normal` perhaps

src/components/progress-indicator/ProgressIndicatorNavigateNext.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { IProgressIndicatorNavigateNextProps } from './types'
+
+export const ProgressIndicatorNavigateNext: React.FC<IProgressIndicatorNavigateNextProps> &
+  typeof Button = ({ children, finalAction, finalLabel, ...rest }) => {
+  const { goToNextStep, isFinalStep } = useProgressIndicator()
+
+  const handleClick = () => (isFinalStep ? finalAction() : goToNextStep())
+
+  return (
+    <Button size="sm" onClick={handleClick} {...rest}>
```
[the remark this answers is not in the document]
alexandrunemes: it's easily overwritable due to the `...rest` spread. This is just a default value, I'd say. It's the best size for making sure the entire thing (buttons and bullets) fits without squeezing the bullets too much.

src/components/progress-indicator/ProgressIndicatorSteps.tsx
andoulla: but it behaves like disabled. the user can't click the number to move forward, can they?

`default` or `standard` could be alternatives to `normal` if you don't like them

andoulla (approved):

src/components/progress-indicator/ProgressIndicatorSteps.tsx
alexandrunemes: I will go with `normal`, I guess, since the ability to click the number has nothing to do with the colour of the bullet.

src/components/progress-indicator/ProgressIndicator.tsx
alexandrunemes: I think you're right about the parent handling this. Any component that implements can simply return or do whatever it needs to do. Could be showing a loader, or a toast, or redirecting...

src/components/progress-indicator/ProgressIndicator.mdx
```diff
@@ -0,0 +1,43 @@
+---
+title: Progress Indicator
+component: ProgressIndicator
+description: ProgressIndicator provides a bullet list of steps and forward/backwards navigation buttons.
+category: Feedback
+---
+
+`ProgressIndicator` exports multiple components that combine to create a Progress Indicator view.
+
+`ProgressIndicator.NavigatePrevious` represents the backwards navigation element. It must receive text as its child in order to render a label on the underlying Button element. It is automatically disabled while the user is viewing the first step.
```
andoulla, line 10:
```suggestion
`ProgressIndicator.NavigatePrevious` represents the backwards navigation element. It must receive a text prop as its child in order to render a label on the underlying Button element. It is automatically disabled while the user is viewing the first step.
```

src/components/progress-indicator/ProgressIndicator.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+`ProgressIndicator.NavigateNext` represents the forward navigation element. It must also receive text as its child in order to render a label on the underlying Button element. Apart from that, it can also receive `finalLabel` and `finalAction` props. When the user reaches the final step, the `finalLabel` gets rendered, while clicking the element triggers `finalAction`.
```
andoulla, line 12:
```suggestion
`ProgressIndicator.NavigateNext` represents the forward navigation element. It must also receive a text prop as its child in order to render a label on the underlying Button element. Apart from that, it can also receive `finalLabel` and `finalAction` props. When the user reaches the final step, the `finalLabel` gets rendered, while clicking the element triggers `finalAction`.
```

src/components/progress-indicator/ProgressIndicatorSteps.tsx
```diff
@@ -0,0 +1,95 @@
...
+              aria-current={index === activeStep ? 'step' : false}
+              aria-label={`step ${index + 1}`}
+              css={{ cursor: allowSkip ? 'pointer' : 'auto' }}
+            >
+              <StyledText
+                size="xs"
+                color={
+                  viewedSteps.includes(index) || index === activeStep
+                    ? 'light'
+                    : 'dark'
```
thomasdigby, line 75: Can we wrap up all of these text styles in the `StyledBullet` component to avoid rendering an extra `p` tag here?

src/components/progress-indicator/ProgressIndicator.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+`ProgressIndicator.Steps` holds the actual bullet list.
+
+The root `ProgressIndicator` component must be passed `stepsData` in the form of an array.
```
thomasdigby, line 16: What does `stepsData` represent? Other than the length of the array mapping to the number of steps, I'm not totally sure what the values in the array represent?

src/components/progress-indicator/ProgressIndicator.mdx
```diff
@@ -0,0 +1,43 @@
...
+```tsx preview
+<ProgressIndicator stepsData={[1, 2, 3]} allowSkip>
+  <ProgressIndicator.NavigatePrevious>Back</ProgressIndicator.NavigatePrevious>
+  <ProgressIndicator.Steps />
+  <ProgressIndicator.NavigateNext
+    finalLabel="Start"
+    finalAction={() => console.log('Finished')}
+  >
+    Next
+  </ProgressIndicator.NavigateNext>
```
thomasdigby, line 41: Just an idea, and it might be absolutely not the right thing, but keen to hear your feedback. What if we defined the final action/label as another component? e.g.

```tsx
<ProgressIndicator stepsData={[1, 2, 3]}>
  <ProgressIndicator.NavigatePrevious>
    Back
  </ProgressIndicator.NavigatePrevious>
  <ProgressIndicator.Steps />
  <ProgressIndicator.NavigateNext>
    Next
  </ProgressIndicator.NavigateNext>
  <ProgressIndicator.NavigateFinal onClick={() => console.log('Finished')}>
    Start
  </ProgressIndicator.NavigateFinal>
</ProgressIndicator>
```

src/components/progress-indicator/ProgressIndicator.mdx
alexandrunemes: The way I thought of it was that the consumer app will implement a hook or something that would output this array, `stepsData`. For example, in the exam instructions modal, we will have an object that we'll enhance based on the exam's specifics. This `stepsData` array will be fed into the `ProgressIndicator`, which will make it available to its children as `steps`. The stepper will consume this array and basically use the items' indexes, while any other component nested inside the `ProgressIndicator` would also have access to the `steps` also stuff like `activeStep`, which will help it determine what content to render.

src/components/progress-indicator/ProgressIndicatorNavigateNext.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+import { Button } from '../button'
+import { useProgressIndicator } from '.'
+import { IProgressIndicatorNavigateNextProps } from './types'
+
+export const ProgressIndicatorNavigateNext: React.FC<IProgressIndicatorNavigateNextProps> &
+  typeof Button = ({ children, finalAction, finalLabel, ...rest }) => {
+  const { goToNextStep, isFinalStep } = useProgressIndicator()
+
+  const handleClick = () => (isFinalStep ? finalAction() : goToNextStep())
```
thomasdigby, line 11: I wonder even if the `finalAction` function should be declared on the `ProgressIndicator` itself, e.g.

```tsx
<ProgressIndicator onComplete={() => console.log('finished')}>
```

src/components/progress-indicator/ProgressIndicatorNavigatePrevious.tsx
```diff
@@ -0,0 +1,22 @@
...
+  ...rest
+}) => {
+  const { activeStep, goToPreviousStep } = useProgressIndicator()
+  return (
+    <Button
+      size="sm"
+      appearance="outline"
+      disabled={activeStep === 0}
+      onClick={goToPreviousStep}
+      {...rest}
```
thomasdigby, line 17: I would move `{...rest}` before `onClick` and `disabled` as we don't want the user to overwrite that. e.g.

```tsx
<Button
  size="sm"
  appearance="outline"
  {...rest}
  disabled={activeStep === 0}
  onClick={goToPreviousStep}
>
```

src/components/progress-indicator/ProgressIndicatorSteps.tsx
```diff
@@ -0,0 +1,95 @@
...
+    if (activeStep === index) return 'active'
+    if (viewedSteps.includes(index)) return 'viewed'
+    return 'normal'
+  }
+
+  return (
+    <Flex>
+      {steps.map((_, index) => {
+        return (
+          <Flex key={`step_${index}`} css={{ alignItems: 'center' }}>
```
thomasdigby, line 61: I would also remove the `Flex` container here if we can as the styles could apply to the `StyledBullet` as well. That allows us to remove an unnecessary `div` from the output.

src/components/progress-indicator/ProgressIndicatorSteps.tsx
```diff
@@ -0,0 +1,95 @@
...
+            <StyledBullet
+              as={allowSkip ? 'button' : 'div'}
+              onClick={() => handleStepClick(index)}
+              state={getBulletState(index)}
+              aria-current={index === activeStep ? 'step' : false}
```
thomasdigby, line 66: Just a suggestion but I tend to prefer removing any non-matching `aria-current` attributes from the DOM, so setting the falsey return value here to `undefined` instead. `aria-current="false"` is valid but unnecessary so saves us a bit of DOM complexity 🤷

src/components/progress-indicator/ProgressIndicatorSteps.tsx
```diff
@@ -0,0 +1,95 @@
...
+                  viewedSteps.includes(index) || index === activeStep
+                    ? 'light'
+                    : 'dark'
+                }
+              >
+                {index + 1}
+              </StyledText>
+            </StyledBullet>
+            {index < steps.length - 1 && (
+              <Divider
```
thomasdigby, line 82: It might be best to avoid this `Divider` here as the line is more of a visual cue than a necessary semantic/accessible one. We could instead render a line in the background of the steps container and remove this extra `hr` from the DOM.

src/components/progress-indicator/index.ts
```diff
@@ -0,0 +1,5 @@
+export { ProgressIndicator } from './ProgressIndicator'
+export {
+  ProgressIndicatorProvider,
+  useProgressIndicator
+} from './progress-indicator-context/ProgressIndicatorContext'
```
thomasdigby, line 5: Do we need to export these functions?

src/components/progress-indicator/progress-indicator-context/ProgressIndicatorContext.tsx
```diff
@@ -0,0 +1,75 @@
...
+        setSteps,
+        goToNextStep,
+        goToPreviousStep,
+        goToStep,
+        activeStep,
+        setActiveStep,
+        viewedSteps,
+        allowSkip,
+        isFinalStep: activeStep === steps.length - 1
+      }}
```
thomasdigby, line 60: Are we using `setSteps` anywhere in our context consumer? Same with `setActiveStep`? I'd remove anything we don't need here

src/components/progress-indicator/progress-indicator-context/ProgressIndicatorContext.tsx
```diff
@@ -0,0 +1,75 @@
...
+  allowSkip: false,
+  isFinalStep: false
+})
+
+export const ProgressIndicatorProvider: React.FC<ProgressIndicatorProviderProps> = ({
+  children,
+  stepsData,
+  allowSkip
+}) => {
+  const [steps, setSteps] = React.useState<unknown[]>(stepsData)
```
thomasdigby, line 23: Is this necessary? Are we updating `steps` anywhere?

src/components/progress-indicator/types.ts
```diff
@@ -0,0 +1,28 @@
+export type ProgressContext = {
+  steps: unknown[]
+  setSteps: React.Dispatch<React.SetStateAction<unknown>[]>
+  goToNextStep: () => void
+  goToPreviousStep: () => void
+  goToStep: (index: number) => void
+  activeStep: number
+  setActiveStep: React.Dispatch<React.SetStateAction<number>>
+  validateViewedStep?: boolean
```
thomasdigby, line 9: Are these types up to date? I can't find `validateViewedStep` anywhere in this PR

src/components/progress-indicator/ProgressIndicator.mdx
thomasdigby: Could we simplify what's required for this component then? I definitely think it's a bit confusing when all we really need is a "step count" so to speak. That could still be dynamic and handled by the user, but the data passed to progress indicator will be different.

```tsx
<ProgressIndicator stepCount={dynamicDataArray.length}>
```

src/components/progress-indicator/ProgressIndicator.tsx
```diff
@@ -0,0 +1,38 @@
...
+  Steps: typeof ProgressIndicatorSteps
+} = ({ children, stepsData, allowSkip }) => {
+  return (
+    <ProgressIndicatorProvider stepsData={stepsData} allowSkip={allowSkip}>
+      <Flex
+        aria-label="progress"
+        css={{
+          display: 'flex',
+          width: '100%',
+          justifyContent: 'space-around',
```
thomasdigby, line 24: I would think that `space-between` might be preferred in most cases, what are your thoughts on setting this property as an opinionated decision or making it configurable by the user?

src/components/progress-indicator/ProgressIndicator.mdx
alexandrunemes: maybe pass `finalLabel` and `finalAction` as props to the root `ProgressIndicator` and make them available through context? This way the root component becomes the single point that receives input in the entire thing.

src/components/progress-indicator/ProgressIndicatorNavigateNext.tsx
alexandrunemes: Hah. Exactly!

src/components/progress-indicator/ProgressIndicatorNavigatePrevious.tsx
alexandrunemes: Good point

src/components/progress-indicator/ProgressIndicatorSteps.tsx
alexandrunemes: Def

src/components/progress-indicator/ProgressIndicatorSteps.tsx
alexandrunemes: 👍

src/components/progress-indicator/progress-indicator-context/ProgressIndicatorContext.tsx
alexandrunemes: removed

src/components/progress-indicator/progress-indicator-context/ProgressIndicatorContext.tsx
alexandrunemes: No. Can just pass `stepsData` through

src/components/progress-indicator/types.ts
alexandrunemes: Thanks, will remove

src/components/progress-indicator/ProgressIndicatorSteps.tsx
alexandrunemes: 👍

src/components/progress-indicator/ProgressIndicatorSteps.tsx
alexandrunemes: This might be a bit difficult since we need each segment to be able to take a different colour. If it's before the active bullet, it should be `primaryDark`, otherwise `tonal400`. So a single line would be very difficult to pull off. We could, with a linear gradient, but the computations for it would be insane 😂 . I gave ::after pseudo-elements a try, but properly aligning everything is proving a challenge.

alexandrunemes: All sorted out

src/components/progress-indicator/ProgressIndicatorSteps.tsx
```diff
@@ -0,0 +1,90 @@
...
+  ml: '$3',
+  variants: {
+    state: {
+      normal: { bg: '$tonal50', color: '$tonal400' },
+      active: { bg: '$primary', color: 'white' },
+      viewed: { bg: '$primaryDark', color: 'white' }
+    },
+    separator: {
+      normal: {
+        '&:not(:last-child)::after': {
```
thomasdigby, line 27: Could we not apply these styles within the "state" variant? Something like this would work, I had to tweak the sibling logic and where the pseudo-element is positioned, but I think it should render as expected?

```ts
const StyledBullet = styled(Flex, {
  // base styles
  '&:not(:first-child)::after': {
    content: '',
    bg: '$tonal400',
    height: '1px',
    right: '$sizes$2',
    position: 'absolute',
    width: '$space$3'
  },
  variants: {
    state: {
      normal: {
        bg: '$tonal50',
        color: '$tonal400'
      },
      active: {
        bg: '$primary',
        color: 'white',
        '&:not(:first-child)::after': { bg: '$primary' }
      },
      viewed: {
        bg: '$primaryDark',
        color: 'white',
        '&:not(:first-child)::after': { bg: '$primary' }
      }
    }
  }
})
```

src/components/progress-indicator/ProgressIndicatorSteps.tsx
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  return (
+    <Flex css={{ alignItems: 'center' }}>
+      {steps.map((_, index) => {
+        return (
+          <StyledBullet
+            key={`step_${index}`}
+            as={allowSkip ? 'button' : 'div'}
+            onClick={() => handleStepClick(index)}
```
thomasdigby, line 77: I wonder if we could remove the use of an extra function here and inline the `allowSkip` check:

```jsx
onClick={allowSkip ? goToStep(index) : undefined}
```

src/components/progress-indicator/progress-indicator-context/ProgressIndicatorContext.tsx
```diff
@@ -0,0 +1,72 @@
...
+    <ProgressIndicatorContext.Provider
+      value={{
+        steps: Array(stepCount).fill(''),
+        goToNextStep,
+        goToPreviousStep,
+        goToStep,
+        activeStep,
+        viewedSteps,
+        allowSkip,
+        isFinalStep: activeStep === stepCount - 1,
```
thomasdigby, line 55: As I mentioned previously, I think this value may not need to be passed down, and instead, we could use the conditional within `goToNextStep()` to call `onComplete()`. Better to try to simplify the data passed to nested children and locate the logic closer to its actual usage.

```ts
const goToNextStep = () => {
  if (onComplete && activeStep === stepCount - 1) {
    return onComplete()
  }

  setActiveStep((current) => current + 1)
}
```

src/components/progress-indicator/progress-indicator-context/ProgressIndicatorContext.tsx
```diff
@@ -0,0 +1,72 @@
...
+    </ProgressIndicatorContext.Provider>
+  )
+}
+
+export const useProgressIndicator = (): ProgressContext => {
+  const context = React.useContext(ProgressIndicatorContext)
+  if (!context)
+    throw new Error(
+      'useProgressIndicatorContext can only be called within a ProgressIndicatorContext'
+    )
```
thomasdigby, line 69: As we're in library land here I wonder if we should update this message to be a bit clearer to the user, e.g. "Ensure that you wrap any components with the root component `ProgressIndicator`"

thomasdigby (commented):
This is looking good, a few minor changes but it's really optimising what's output and for a library it's important we do that 👍 

However, I _do_ want to talk about the name.
- `ProgressIndicator` is very close to our `ProgressBar` and it wouldn't surprise me if someone assumed it was a progress bar
- Material UI and a bunch of other design system libraries do use the term `Stepper` and it's similar to our usage here, not identical but the intent is the same.
- `Stepper` could also allow us to use actiony names for some components. As an example, I wouldn't be against this style of component names:
```jsx
<Stepper count={3}>
  <Stepper.StepBack>Back</Stepper.StepBack>
  <Stepper.Steps />
  <Stepper.StepForward>Next</Stepper.StepForward>
</Stepper>
```

src/components/progress-indicator/ProgressIndicatorNavigateNext.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { Button } from '../button'
+import { useProgressIndicator } from './progress-indicator-context/ProgressIndicatorContext'
+import { IProgressIndicatorNavigateNextProps } from './types'
+
+export const ProgressIndicatorNavigateNext: React.FC<
+  IProgressIndicatorNavigateNextProps & typeof Button
+> = ({ outputLabel, ...rest }) => {
+  const { goToNextStep, isFinalStep, onComplete } = useProgressIndicator()
+
+  const handleClick = () => (isFinalStep ? onComplete?.() : goToNextStep())
```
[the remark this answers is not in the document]
alexandrunemes: 👍

src/components/progress-indicator/ProgressIndicatorNavigatePrevious.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react'
+
+import { Button } from '../button'
+import { useProgressIndicator } from './progress-indicator-context/ProgressIndicatorContext'
+import { IProgressIndicatorNavigatePreviousProps } from './types'
+
+export const ProgressIndicatorNavigatePrevious: React.FC<
+  IProgressIndicatorNavigatePreviousProps & typeof Button
```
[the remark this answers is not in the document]
alexandrunemes: spot on!

src/components/progress-indicator/ProgressIndicatorNavigateNext.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+  return (
+    <Button size="sm" {...rest} onClick={handleClick}>
+      {outputLabel(isFinalStep)}
```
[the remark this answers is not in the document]
alexandrunemes: What if it gets passed both the label prop and a child node? I'd say `{children || label()}`. What do you think?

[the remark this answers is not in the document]
thomasdigby: Yeah exactly 👍

alexandrunemes:
> ```js
> Stepper
> ```

I'm completely in favour of `Stepper`.

src/components/progress-indicator/progress-indicator-context/ProgressIndicatorContext.tsx
alexandrunemes: 👍

src/components/progress-indicator/ProgressIndicatorSteps.tsx
alexandrunemes: sure.

src/components/progress-indicator/ProgressIndicatorSteps.tsx
alexandrunemes: So the way I built this was that any bullet that's behind the active one would have its corresponding separator (to the right) blue. anything after the active step, grey. As far as I can tell, this would tie the colour of the separator to the viewed/active status of the bullet.

src/components/progress-indicator/ProgressIndicator.mdx
```diff
@@ -0,0 +1,42 @@
...
+
+`ProgressIndicator` exports multiple components that combine to create a Progress Indicator view.
+
+`ProgressIndicator.NavigatePrevious` represents the backwards navigation element. It must receive `outputLabel` as a string in order to render a label on the underlying Button element. It is automatically disabled while the user is viewing the first step.
+
+`ProgressIndicator.NavigateNext` represents the forward navigation element. It must receive `outputLabel` as a function that receives `isFinalStep` as an argument in order to render a normal or a final step label on the underlying Button element.
+
+`ProgressIndicator.Steps` holds the actual bullet list.
+
+The root `ProgressIndicator` component must be passed `stepCount` in the form of a number. It can also take `onComplete` as a function that gets called when the user clicks the `NavigateNext` button while on the last step.
```
shaneakins, line 16: Minor comment:  replace 'in the form of a number' with 'as a number'

src/components/progress-indicator/ProgressIndicator.test.tsx
```diff
@@ -0,0 +1,102 @@
+import { fireEvent, render, screen } from '@testing-library/react'
+import { axe } from 'jest-axe'
+import * as React from 'react'
+
+import { ProgressIndicator } from './ProgressIndicator'
+
+describe('ProgressIndicator', () => {
+  it('renders a bullet style progress indicator - has no programmatically detectable a11y issues', async () => {
+    const { container } = render(
+      <ProgressIndicator stepCount={3} onComplete={jest.fn()}>
```
shaneakins, line 10: You can save yourself some repetition by defining a `props` object with the repeating props right after the describe statement, and then spreading `props` and adding extra props as needed:

_describe('ProgressIndicator', () => {
    const props = {
        stepCount: 3,
        onComplete: jest.fn()
    }
   ...
  }_

_usage with just props object:_

```
const { container } = render(
    <ProgressIndicator {...props}>
    ...
```

_usage with additional props:_

```
const { container } = render(
    <ProgressIndicator {...props} allowSkip >
    ...
```

src/components/progress-indicator/ProgressIndicator.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+    // go to the second item
+    fireEvent.click(screen.getByText('Next'))
+    // go back to the first item
+    fireEvent.click(screen.getByText('Back'))
+    // back button should now be disabled
+    expect(screen.getByText('Back')).toBeDisabled()
+  })
+
+  it('calls final action handler if provided, when on final step', () => {
+    const finalActionSpy = jest.fn()
```
shaneakins, line 86: Any special reason why `finalActionSpy` is defined in just this test?

src/components/progress-indicator/ProgressIndicator.mdx
```diff
@@ -0,0 +1,42 @@
...
+<ProgressIndicator stepCount={3} onComplete={() => console.log('Finished')}>
+  <ProgressIndicator.NavigatePrevious outputLabel="Back" />
+  <ProgressIndicator.Steps />
+  <ProgressIndicator.NavigateNext
+    outputLabel={(isFinalStep) => (isFinalStep ? 'Finish' : 'Next')}
+  />
+</ProgressIndicator>
+```
+
+The root `ProgressIndicator` component can also take an `allowSkip` prop. When this prop is passed, the user is allowed to click the actual bullets in order to navigate.
```
shaneakins, line 28: How about _"The root `ProgressIndicator` component can also take an `allowSkip` prop that allows the user to navigate by clicking the actual bullets."_?

src/components/progress-indicator/ProgressIndicator.test.tsx
alexandrunemes: Yeah, it's not a mandatory prop. I'm testing that if it's passed, it gets called.

src/components/progress-indicator/ProgressIndicator.mdx
alexandrunemes: Thanks!

src/components/progress-indicator/ProgressIndicator.mdx
alexandrunemes: Will do. Thanks!

src/components/progress-indicator/ProgressIndicator.test.tsx
shaneakins: Ahh totally missed that `toHaveBeenCalled`

src/components/progress-indicator/ProgressIndicator.test.tsx
alexandrunemes: 👍

src/components/progress-indicator/progress-indicator-context/ProgressIndicatorContext.tsx
alexandrunemes: Makes sense

shaneakins (approved):

src/components/progress-indicator/ProgressIndicatorSteps.tsx
thomasdigby: Ah ok, that makes sense. Worth tweaking the current styles to remove any duplicated `::after` styles, but otherwise good to stick with your solution

thomasdigby:
> I'm completely in favour of `Stepper`.

Can we change the name in this PR? I'll confirm with design

alexandrunemes:
Already have. Will push in a moment.

src/components/progress-indicator/ProgressIndicatorSteps.tsx
alexandrunemes: 👍

thomasdigby (approved):
🏃

PR #298, opened 2022-02-16
feat: button neutral theme

kerryambt:
### Description

In the parent portal we have a lot of white buttons that are currently not handled by theme. As such, there is inconsistency between their active and hover states because different people have added them at different times. We would like to consolidate this use case into a new button theme, Neutral.

### Screenshots

image

https://user-images.githubusercontent.com/66493855/154307177-23a43508-6a19-45b5-ad55-5e6230f1dab4.mov

Outline variant:
image

### DIFF
--- a/src/components/button/Button.tsx
+++ b/src/components/button/Button.tsx
@@ -1,5 +1,5 @@
 import type * as Stitches from '@stitches/react'
-import { darken } from 'polished'
+import { darken, opacify } from 'polished'
 import * as React from 'react'
 
 import { Box } from '~/components/box'
@@ -74,7 +74,8 @@ export const StyledButton = styled('button', {
       secondary: {},
       success: {},
       warning: {},
-      danger: {}
+      danger: {},
+      neutral: {}
     },
     appearance: {
       solid: {},
@@ -153,6 +154,16 @@ export const StyledButton = styled('button', {
       theme: 'danger',
       appearance: 'solid',
       css: getButtonSolidVariant('$danger', '$dangerMid', '$dangerDark')
+    },
+    {
+      theme: 'neutral',
+      appearance: 'solid',
+      css: getButtonSolidVariant(
+        'white',
+        opacify(-0.1, 'white'),
+        opacify(-0.25, 'white'),
+        '$primary'
+      )
     },
     {
       theme: 'primary',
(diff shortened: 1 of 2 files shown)


thomasdigby:
@kerryambt I like this, my only feedback before merging would be whether we might need other themes on white in future? e.g. `success` with white bg.

Can you check with Javi/Jess/Whoever to get a vibe from them as to where they might go?

Mhoog (approved):

kerryambt:
@thomasdigby just spoke to Javi, they're un sure if we will need those colours in the future, he needs to do research but for now doesn't think we have enough use cases to block this. from a code perspective the thing we want to align on is the hover and click states of our white buttons, which this has done. if we do in the future need other colours, to me it's not a huge break of the DS to just add one line of css to change the text colour, because that wouldn't stop the aligning of the hover and click states. We can always return to this if we find more use cases for other colours, but given they're an unknown for the foreseeable future we should work with what we know we need :)

thomasdigby:
> @thomasdigby just spoke to Javi, they're unsure if we will need those colours in the future, he needs to do research but for now doesn't think we have enough use cases to block this. from a code perspective the thing we want to align on is the hover and click states of our white buttons, which this has done. if we do in the future need other colours, to me, it's not a huge break of the DS to just add one line of CSS to change the text colour because that wouldn't stop the aligning of the hover and click states. We can always return to this if we find more use cases for other colours, but given they're an unknown for the foreseeable future we should work with what we know we need :)

Absolutely not against doing something now which works in the short-term, I was just pondering whether we could use a separate term to allow us to extend the theme in future e.g. `<Button theme="success" applyThemeInReverse={true} />`

For now, let's roll with `theme="neutral"`

kerryambt:
@thomasdigby yeah that's a really interesting thought :o I think if we decide we need other colours that's a very clean way of doing it, will keep that in mind for sure :)

src/components/button/Button.tsx
```diff
@@ -74,7 +74,8 @@ export const StyledButton = styled('button', {
       secondary: {},
       success: {},
       warning: {},
-      danger: {}
+      danger: {},
+      neutral: {}
```
thomasdigby, line 78: Can you add this theme to the docs? Just as another option in the theme example

thomasdigby (approved):

src/components/button/Button.tsx
kerryambt: oops, forgot this! done :)

PR #302, opened 2022-02-19
Chore: add new action to lint and format on merge into main

thomasdigby:
Added a new "Autofix" workflow to run ESLint and Prettier on merge into `main`, the commit will appear as "Apply auto-fix changes" co-authored by a Github actions user and the PR (or last commit) author

Screenshot 2022-02-19 at 10 41 02

### DIFF
--- a/.github/workflows/auto-fix.yml
+++ b/.github/workflows/auto-fix.yml
@@ -1,0 +1,35 @@
+name: auto-fix
+
+on:
+  push:
+    branches:
+      - 'main'
+
+jobs:
+  auto-fix:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout branch
+        uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-fix.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Run ESLint
+        run: yarn lint
+
+      - name: Run Prettier
+        run: yarn format
+
+      - name: Commit changes
+        uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          commit_message: Apply auto-fix changes
```
andoulla, line 35: Very very minor, but if I remember correctly the auto release functionality (when it will be working) will not know how to deal with a commit that doesn't have any prefix. Could we change this to `Chore: Apply auto-fix changes` to tell that plugin to ignore this commit when calculating the semver version?

andoulla (approved):

PR #308, opened 2022-03-07
Add ability to customize the StepperSteps container css

alexandrunemes:
By default, the Stepper.Steps component had all the spacing and dimensions hardcoded, meaning that if you had a small number of bullets on a large screen, these would gather to center of the screen, leaving a lot of empty space between them and the Back/Next buttons. 

We needed a way to make the container's width dynamic and also make sure the separator lines between bullets are spaced and sized accordingly.

Screenshots:
Before:
Screenshot 2022-03-07 at 15 58 03

After:
Screenshot 2022-03-07 at 15 59 27

### DIFF
--- a/src/components/stepper/Stepper.mdx
+++ b/src/components/stepper/Stepper.mdx
@@ -1,6 +1,6 @@
 ---
 title: Stepper
-component: Stepper, Stepper.StepBack, Stepper.StepForward
+component: Stepper, Stepper.StepBack, Stepper.StepForward, Stepper.Steps
 description: Stepper provides a bullet list of steps and forward/backwards navigation buttons.
 category: Navigation
 ---
@@ -11,7 +11,7 @@ category: Navigation
 
 `Stepper.StepForward` represents the forward navigation element. It can receive either a child text node or a `label` prop as a function that receives `activeStep` as an argument in order to possibly render different labels based on the current step the user is on.
 
-`Stepper.Steps` holds the actual bullet list.
+`Stepper.Steps` holds the actual bullet list. It can receive a `stepsWidth` prop as a string in order to customize the width of the container that holds all of the bullets.
 
 The root `Stepper` component must be passed `stepCount` as a number in order to display the desired number of bullet steps. It also takes `onStepChange` as a function that takes `activeStep` as an argument. This can be used to determine what content to display, based on the current step. It can also take `onComplete` as a function that gets called when the user clicks the `StepForward` button while on the last step.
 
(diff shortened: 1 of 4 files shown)


src/components/stepper/Stepper.mdx
```diff
@@ -37,7 +37,7 @@ Below is a more complex example that shows how to dynamically render the button
   }
 >
   <Stepper.StepBack label={(activeStep) => 'Back'} />
-  <Stepper.Steps />
+  <Stepper.Steps width="200px" />
```
andoulla, line 40: width or stepsWidth?

src/components/stepper/StepperSteps.tsx
```diff
@@ -49,7 +49,10 @@ const StyledBullet = styled(Flex, {
   }
 })
 
-export const StepperSteps: React.FC = () => {
+export const StepperSteps: React.FC<IStepperStepsProps> = ({
+  stepsWidth,
+  ...rest
```
andoulla, line 54: Just wondering, why pass a single width value instead of passing a CSS prop to allow customising more things (ie: height too)

it can look something like
```
export const StepperSteps: React.FC<IStepperStepsProps> = ({
 css,
  ...rest

........

 <Flex
      css={{
        alignItems: 'center',
        justifyContent: 'space-between',
...css
      }}
      {...rest}
    >
```
where the bullet is customised we can still get the value as css.width

src/components/stepper/Stepper.mdx
alexandrunemes: ups. nice one. it's supposed to be stepsWidth

src/components/stepper/StepperSteps.tsx
alexandrunemes: That was my first instinct. to allow full customisation. But then I started having second thoughts about opening up the component. IDK, just felt safer to lock it down style-wise. But if you think it's ok to allow any CSS to be passed down, I can update the code.

andoulla: This is a pattern I see in other components in the DS. I didn't really come up with it myself, to be honest.

Also, we have been avoiding single-CSS props when we can. I personally think it will make it more reusable

alexandrunemes: 👍

alexandrunemes:
@thomasdigby @andoulla , what do you think? Can we get these changes merged?

andoulla (approved):

kerryambt (approved):

PR #310, opened 2022-03-24
Fix: various fixed on accordion

andoulla:
Tasks:
- [x] Remove unnecessary box to allow flexible styling on the children of the trigger
- [x] Fix rotating all children instead of just the chevron icon
- [x] Fix ts error when using `type="multiple"` - Can't fix the issue as it needs to be fixed in Radix, however, this change hopefully prevents the error from being thrown in `atom-core`

# screenshots for child alignment
## before from atom-core
image

## after
image

# screenshots for rotation
## before
https://recordit.co/XwRKSnzj4E

## after
https://recordit.co/Cics3jcuzv


kerryambt (approved):

src/components/accordion/AccordionTrigger.tsx
```diff
@@ -29,6 +28,9 @@ const getTriggerVariant = (
   }
 })
 
+const ChevronIcon = () => <Icon className="chevron-icon" is={ChevronDown} />
```
LimeWub, line 31: Huh - I thought you'd be doing something like:

```
const ChevronIcon = styled(Icon, {
   transition: 'transform 300ms',
   '[data-state="open"] > &': {
      transform: 'rotate(180deg)'
    },
    '[data-state="closed"] > &': {
      transform: 'rotate(0deg)'
    }
});
```

Not sure if it would'd worked but that's what I thought you'd try (based on Gary's message) - 
Maybe what I have above doesn't work (?) - but ye that's what I was expecting to see in this PR lol

Ofc what you have is perfectly fine too (just more what I was thinking) :o

Also, If you're just passing in a classname why make a new component for this?

hollg: @andoulla I've got to say...I like @LimeWub 's suggestion more than what we came up with 😄

andoulla: @hollg I don't know why it didn't work when we tried it but now it does 🤦 

@LimeWub I have now pushed a change with your suggestion if you could take a look please

src/components/accordion/AccordionTrigger.tsx
```diff
@@ -29,6 +28,16 @@ const getTriggerVariant = (
   }
 })
 
+const ChevronIcon = styled(Icon, {
```
hollg, line 31: I don't think this name makes sense with the most recent changes. Perhaps this should be `RotatingIcon`?

andoulla: Hm, not sure as then it loses specificity of which icon it is. Would CollapseIcon or StatusIcon be better? Just cause this is what it is it shows whether the accordion item is collapsed or expanded

hollg: This component doesn't specify that the icon is a chevron, though. This component takes an `is` prop and could be passed any other icon. All this component adds on top of `Icon` is the rotation.

alexandrunemes (approved):

src/components/accordion/AccordionTrigger.tsx
hollg: I'm not sure I see why that is necessary instead of `<RotatingIcon is={ChevronDown />` but I definitely don't feel strongly about it; either way everything is named descriptively 👍🏻

andoulla: I was going to say just the same thing, it doesn't matter that match. I will go for the simpler version instead of more descriptive names

andoulla: done

src/components/accordion/AccordionTrigger.tsx
```diff
@@ -41,14 +50,14 @@ const StyledTrigger = styled(Trigger, {
 
   '&[data-state="open"]': {
     borderRadius: '$0 $0 0 0',
-    '& svg': {
+    [`& ${RotatingIcon}`]: {
```
LimeWub, line 53: You don't need these now since you have them defined above. Maybe that's why it's still working - try removing the transform/transition from here

src/components/accordion/AccordionTrigger.tsx
```diff
@@ -41,14 +50,14 @@ const StyledTrigger = styled(Trigger, {
...
       transition: 'transform 300ms',
       transform: 'rotate(180deg)'
     }
   },
   '&[data-state="closed"]': {
     borderRadius: '$0',
-    '& svg': {
+    [`& ${RotatingIcon}`]: {
```
LimeWub, line 60: And here

LimeWub (approved):

PR #311, opened 2022-03-28
fix: add compound variant for outline secondary buttons

avirati:
Problem: 
```tsx
<Button theme="secondary" appearance="outline">Button</Button>
```
The above rendered a button with `color: rgb(0,0,0)` when it should have rendered one with `color: $primaryDark`

Reason: 
Compound Variant for this combination was absent. Hence it was falling back to default color

Fix: 
Adds a compound variant to remedy that

Preview :

https://user-images.githubusercontent.com/1936119/160368094-fe8cedb3-9aa3-4416-a491-93dbd9f09048.mov

### DIFF
--- a/src/components/button/Button.tsx
+++ b/src/components/button/Button.tsx
@@ -178,6 +178,15 @@ export const StyledButton = styled('button', {
         opacify(-0.2, 'white'),
         opacify(-0.35, 'white')
       )
+    },
+    {
+      theme: 'secondary',
+      appearance: 'outline',
+      css: getButtonOutlineVariant(
+        '$primaryDark',
+        darken(0.05, theme.colors.primaryDark.value),
+        darken(0.1, theme.colors.primaryDark.value)
+      )
     }
   ]
 })
--- a/src/components/button/Button.tsx
+++ b/src/components/button/Button.tsx
@@ -184,8 +184,8 @@ export const StyledButton = styled('button', {
       appearance: 'outline',
       css: getButtonOutlineVariant(
         '$primaryDark',
-        darken(0.05, theme.colors.primaryDark.value),
-        darken(0.1, theme.colors.primaryDark.value)
+        darken(0.1, theme.colors.primaryDark.value),
+        darken(0.15, theme.colors.primaryDark.value)
       )
     }
   ]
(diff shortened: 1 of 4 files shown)


src/components/button/Button.tsx
```diff
@@ -178,6 +178,15 @@ export const StyledButton = styled('button', {
...
         opacify(-0.35, 'white')
       )
+    },
+    {
+      theme: 'secondary',
+      appearance: 'outline',
+      css: getButtonOutlineVariant(
+        '$primaryDark',
+        darken(0.05, theme.colors.primaryDark.value),
+        darken(0.1, theme.colors.primaryDark.value)
```
thomasdigby, line 188: Could we match the values from our secondary solid button? https://github.com/Atom-Learning/components/pull/311/files#diff-dbb0a2706cf9edd17c0ff507bbe8bc37e950d6e7cdb77116e7ca9d5d0c29aa72R135

avirati: Makes sense. Done ✔️

kerryambt:
can we update the mdx? :)

avirati:
> can we update the mdx? :)

Done, thanks

kerryambt:
can you update the appearance section of the mdx to state there is an outline variant for secondary? sorry to be a pain :)

avirati:
> can you update the appearance section of the mdx to state there is an outline variant for secondary? sorry to be a pain :)

Thought of doing that but then if we do this, we would need to have all combinations. That might not look good. Buttons with their individual props have clear demonstration. A reader should be able to expect the desired results

kerryambt:
> > can you update the appearance section of the mdx to state there is an outline variant for secondary? sorry to be a pain :)
> 
> Thought of doing that but then if we do this, we would need to have all combinations. That might not look good. Buttons with their individual props have clear demonstration. A reader should be able to expect the desired results

I just mean here where it says `There are the available outline variations for the primary and neutral themes.` we should add `secondary` to that list, because we don't have all combinations available atm. no need to update the preview :)

image

avirati:
Ah okay. Updated 👍

kerryambt (approved):

thomasdigby (approved):

PR #312, opened 2022-03-30
fix: extend type to allow passing CSS

andoulla:
Minor type update to allow the CSS prop

### DIFF
--- a/src/components/accordion/Accordion.tsx
+++ b/src/components/accordion/Accordion.tsx
@@ -1,7 +1,7 @@
 import { Root } from '@radix-ui/react-accordion'
 import React from 'react'
 
-import { CSS, styled } from '~/stitches'
+import { styled } from '~/stitches'
 
 import { AccordionContent } from './AccordionContent'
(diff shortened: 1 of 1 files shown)


src/components/accordion/Accordion.tsx
```diff
@@ -13,6 +13,7 @@ const StyledRoot = styled(Root, {
 
 type AccordionProps = React.ComponentProps<typeof Root> & {
   type: 'single' | 'multiple'
+  css?: CSS
```
alexandrunemes, line 16: Just wondering, but would `React.ComponentProps<typeof StyledRoot>` instead of `React.ComponentProps<typeof Root>` solve the problem? `StyledRoot` is defined with `styled`, so it should take `css` as a prop by default.

andoulla: you are right, will try

alexandrunemes (approved):

thomasdigby (approved):

PR #316, opened 2022-04-07
AP-186: Create a file input

alexandrunemes:
JIRA: https://atomlearningltd.atlassian.net/browse/AP-186

Description: 
Introduces an `input type="file"`.

Note: For this initial version, no drag-and-drop functionality has been implemented. This will be added in a future V2.

Screen recording:

https://user-images.githubusercontent.com/9009155/162151068-e8d74a71-61c7-4d2c-be11-121f3da2c308.mov

### DIFF
--- a/src/components/file-input/FileInput.mdx
+++ b/src/components/file-input/FileInput.mdx
@@ -1,0 +1,51 @@
+---
+title: File Input
+component: FileInput
+description: FileInput is the lowest-level input component for text and number types
+category: Form primitives
+---
+
+`FileInput` is a specialised input that allows file uploads.
+
+The 3 most important props it takes are: `onSelect`, `id` and `label`.
+
+`onSelect` is a callback that accepts the actual file selection (as a FileList), which gets updated after the user interacts with the file browser interface. The file selection can then be used in order to build a payload for something like a `POST` API call.
+
+Note: The actual action of uploading the file to a backend is not this input's responsability.
+
+`id` is really important because under the hood, the actual file input is hidden, while the visible element is a `label` with an `htmlFor` attribute that pairs with the hidden input via the `id` attribute. So make sure the `id` is unique.
+
+`label` is used to add text to the visible button.
+
+```jsx preview
+<FileInput
+  onSelect={(fileSelection) => console.log(fileSelection)}
+  id="unique-identifier"
+  label="Select file"
+/>
+```
+
+The input can also accept multiple files at once, via the `multiple` prop. Another important feature is the `accept` prop. This makes the input filter out unwanted file types and prevent the user from selecting them.
+
+```jsx preview
+<FileInput
+  onSelect={(fileSelection) => console.log(fileSelection)}
+  id="unique-identifier"
+  label="Select file"
+  multiple
+  accept="application/pdf"
+/>
+```
+
+You can also use our icon pack to add an icon next to the button text.
+
+```jsx preview
+<FileInput
+  onSelect={(fileSelection) => console.log(fileSelection)}
+  id="unique-identifier"
+  label="Select file"
+  multiple
+  accept="application/pdf"
+  icon={Upload}
+/>
+```
--- a/src/components/file-input/FileInput.test.tsx
+++ b/src/components/file-input/FileInput.test.tsx
@@ -1,0 +1,42 @@
+import { fireEvent, render, screen, waitFor } from '@testing-library/react'
+import { axe } from 'jest-axe'
+import * as React from 'react'
+
+import { FileInput } from '.'
+import { FileInputProps } from './FileInput'
+
+describe('FileInput component', () => {
+  const props: FileInputProps = {
+    onSelect: jest.fn(),
+    id: 'file-input-id',
+    label: 'Select file'
+  }
+
+  it('renders', async () => {
+    const { container } = render(<FileInput {...props} />)
+
+    expect(container).toMatchSnapshot()
+  })
+
+  it('has no programmatically detectable a11y issues', async () => {
+    const { container } = render(<FileInput {...props} />)
+
+    expect(await axe(container)).toHaveNoViolations()
+  })
+
+  it('uploads a file', async () => {
+    const file = new File(['random content'], 'upload.txt')
+    render(<FileInput {...props} />)
+
+    const input = screen.getByTestId('file-upload') as HTMLInputElement
+
+    await waitFor(() =>
+      fireEvent.change(input, {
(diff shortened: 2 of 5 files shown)


src/components/file-input/FileInput.mdx
```diff
@@ -0,0 +1,51 @@
...
+category: Form primitives
+---
+
+`FileInput` is a specialised input that allows file uploads.
+
+The 3 most important props it takes are: `onSelect`, `id` and `label`.
+
+`onSelect` is a callback that accepts the actual file selection (as a FileList), which gets updated after the user interacts with the file browser interface. The file selection can then be used in order to build a payload for something like a `POST` API call.
+
+Note: The actual action of uploading the file to a backend is not this input's responsability.
```
andoulla, line 14:
```suggestion
Note: The actual action of uploading the file to a backend is the parent's responsibility.
```

src/components/file-input/FileInput.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+`id` is really important because under the hood, the actual file input is hidden, while the visible element is a `label` with an `htmlFor` attribute that pairs with the hidden input via the `id` attribute. So make sure the `id` is unique.
```
andoulla, line 16:
```suggestion
`id` is really important because the actual file input is hidden, while the visible element is a `label` with an `htmlFor` attribute that pairs with the hidden input via the `id` attribute. 

 Note: `id` has to be unique 
```
This way it makes it stand out that the id has to be unique

andoulla (approved):

src/components/file-input/FileInput.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export type FileInputProps = {
+  onSelect: (selection: FileList) => void
+  id: string
+  label: string
+  accept?: string
+  multiple?: boolean
+  appearance?: string
+  css?: CSS
+  icon?: SVGSVGElement
```
LimeWub, line 15: `SVGSVGElement`

Is this a legit type or just typo?

src/components/file-input/FileInput.tsx
```diff
@@ -0,0 +1,57 @@
...
+  return (
+    <>
+      <input
+        type="file"
+        onChange={handleFileSelect}
+        accept={accept}
+        multiple={multiple}
+        id={id}
+        data-testid="file-upload"
+        hidden
```
LimeWub, line 47: Why `hidden`? I'm concerned it will affect screen readers for this element + children (aka the Button)

LimeWub (approved):

src/components/file-input/FileInput.tsx
```diff
@@ -0,0 +1,57 @@
...
+      />
+      <Button as="label" htmlFor={id} appearance={appearance} css={css}>
+        {icon && <Icon is={icon} />}
```
thomasdigby, line 50: This should be an opinionated choice of ours to have an icon, let's remove the conditional. As far as I'm aware, we don't use `icon` in this way in other components.

src/components/file-input/FileInput.tsx
```diff
@@ -0,0 +1,57 @@
...
+        {label}
```
thomasdigby, line 51: Should this be `children` instead? It feels like we're overlapping prop names with a potential `FieldInputField` in the future.

```tsx
<FileInput>Select file</FileInput>
```

src/components/file-input/FileInput.tsx
```diff
@@ -0,0 +1,57 @@
...
+}) => {
+  const [fileSelection, setFileSelection] = React.useState<FileList>()
+
+  React.useEffect(() => {
+    onSelect(fileSelection)
+  }, [fileSelection, onSelect])
+
+  const handleFileSelect = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setFileSelection(event.target.files)
+  }
```
thomasdigby, line 36: Is there any reason to declare two methods here? We're not using `fileSelection` anywhere else, could we just call `onSelect` within `handleFileSelect`?

src/components/file-input/FileInput.tsx
```diff
@@ -0,0 +1,57 @@
...
+      <input
+        type="file"
+        onChange={handleFileSelect}
+        accept={accept}
+        multiple={multiple}
+        id={id}
+        data-testid="file-upload"
+        hidden
+      />
+      <Button as="label" htmlFor={id} appearance={appearance} css={css}>
```
thomasdigby, line 49: I tend to try and avoid passing attributes that don't have a default value, and pass `{...rest}` instead

src/components/file-input/FileInput.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+    const { container } = render(<FileInput {...props} />)
+
+    expect(await axe(container)).toHaveNoViolations()
+  })
+
+  it('uploads a file', async () => {
+    const file = new File(['random content'], 'upload.txt')
+    render(<FileInput {...props} />)
+
+    const input = screen.getByTestId('file-upload') as HTMLInputElement
```
thomasdigby, line 31: Could we target the file input without a `data-testid`? Helps to avoid including that in the output JS

src/components/file-input/FileInput.tsx
andoulla: Ok I will just code the icon in, instead of passing it down

andoulla: done

src/components/file-input/FileInput.tsx
andoulla: done

src/components/file-input/FileInput.tsx
andoulla: done

andoulla: oh, I misunderstood the comment then, I thought it was to remove the useEffect and just call the callback at the same time. Let me look into it and will come back

src/components/button/Button.tsx
```diff
@@ -241,6 +241,7 @@ type ButtonProps = Override<
     as?: React.ComponentType | React.ElementType
     children: React.ReactNode
     isLoading?: boolean
+    htmlFor?: string // styled() omits this prop, however it is needed if need to pair the button with another component (ie: file input)
```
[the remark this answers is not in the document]
andoulla: yes, you are spot on, Alex and I were just debating different solutions for the TS error but didn't think of removing the need for htmlFor. Thanks. I'm pushing this change now

PR #317, opened 2022-04-07
feat: add semantic release

hollg:
Adds automatic versioning and publishing for the NPM package.

I followed this guide, which uses `semantic-release` via `npx` rather than the GitHub Actions package that we tried and failed to use before: https://github.com/zeke/semantic-release-with-github-actions

I've also split the `test` and `release` jobs into separate workflows so that GitHub can display them individually within the "checks" section of the PR view. `test` runs when a PR is made against `main` and will be required in order for a merge (the latter is a change in the repo settings, not in code). `release` runs when something is pushed to `main`, which includes when a PR is merged. But we can also see `release` in the "checks" section, where it will ensure that there is at least one semantic commit message or a semantic title in the PR, so that `semantic-release` can correctly calculate the new version number.

`semantic-release` will get the current version number from our git tags (and create new ones there). It will **not** read from or update the `version` number in `package.json`. There seem to be conflicting conventions about how to highlight in `package.json` that the `version` number there is no longer relevant, but `"version": "0.0.0-semantically-released"` was the most descriptive that I've found, so that's what I picked.


package.json
```diff
@@ -4,7 +4,7 @@
   "main": "dist/index.cjs.js",
   "typings": "dist/index.d.ts",
   "module": "dist/index.js",
-  "version": "1.3.0",
+  "version": "0.0.0-semantically-released",
```
thomasdigby, line 7: This feels like a dealbreaker for me, is there no way that we can reflect the version change back into package.json? Are we blocked by lacking permissions to commit to the repo?

hollg: I didn't look much more into the blockers we faced with the previous approach because I found this one, which side-steps them. The `package.json` of the published version **will** be updated to include the correct version number — does that make it any less of a deal-breaker for you?

thomasdigby: Oh, maybe...

thomasdigby: Our git tags are a bit messed up at the moment, I assume we'd need to tidy up and create historical releases so that this could work as expected?

hollg: From what I've seen, we don't need to create any releases so long as we have git tags. The latest git tag matches what `semantic-release-bot` changed the `package.json` `version` too last, but that doesn't match what we last published, so maybe we just need to push a tag matching the last publish?

LimeWub (approved):
Not my area of expertise but looks okay 👍

.github/workflows/release.yml
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+  release:
+    name: Release
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - uses: actions/setup-node@v1
+        with:
+          node-version: "12.x"
```
andoulla, line 16: just out of curiosity, why run with this version instead of 14.*

hollg: Great spot! No reason at all; this is outdated and I didn't realise when I pasted it in

package.json
```diff
@@ -68,7 +68,6 @@
     "@rollup/plugin-commonjs": "^17.1.0",
     "@rollup/plugin-node-resolve": "^11.2.0",
     "@semantic-release/changelog": "^5.0.1",
-    "@semantic-release/git": "^9.0.0",
```
thomasdigby, line 71: Is this relflected in a `yarn.lock` update?

.github/workflows/release.yml
```diff
@@ -0,0 +1,23 @@
...
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '14'
+      - run: npm ci
+      - run: npm run build --if-present
+      - run: npm test:ci
+      - run: npx semantic-release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
```
thomasdigby, line 23: Do we need to add the slack webhook token here as well?

andoulla (approved):
so happy to see this go in, 15 months after we did the initial investigation!!

.github/workflows/release.yml
hollg: Yep, it's in there now

package.json
hollg: Yep!

andoulla:
@hollg does any of the documentation need updating on this? https://design.atomlearning.technology/components/contributing

hollg:
@andoulla 
>@hollg does any of the documentation need updating on this? https://design.atomlearning.technology/components/contributing

I don't think so; the instructions about commits etc are unchanged.

thomasdigby (approved):

PR #322, opened 2022-04-12
fix: showing toast on top of modal overlay

viniciusvolpe:
### Description
Fixing modal component to not cover the toast component, preventing modal close when clicking on toast.
JIRA: https://atomlearningltd.atlassian.net/browse/PP-161

### Implementation
- Updating Radix modal to fix the overlay issue
- Setting toast porter events to auto so the cursor is correct
- Adding `onPointerDownOutside` props to modal content to prevent the modal being closed when interacting with the Toast

### Screenshots

https://user-images.githubusercontent.com/11462265/162973850-42c98f4f-5b7c-4aa5-af04-5248f33dfe14.mov


package.json
```diff
@@ -131,7 +131,7 @@
     "@radix-ui/react-accordion": "^0.1.1",
     "@radix-ui/react-alert-dialog": "0.1.1",
     "@radix-ui/react-checkbox": "0.1.1",
-    "@radix-ui/react-dialog": "0.1.1",
+    "@radix-ui/react-dialog": "0.1.2",
```
thomasdigby, line 134: Is this upgrade necessary? There may be some overlap with shared utilities across the 0.1.1 packages that we lose when we upgrade. (We should upgrade all of them when we have a chance, just need time to ensure we handle any compatibility changes)

thomasdigby (approved):
Looking good, only one query about the package upgrade. Does closing the toast transfer the focus back to the dialog content? I just want to make sure that we're not losing it if it transfers back to the document

viniciusvolpe:
> Looking good, only one query about the package upgrade. Does closing the toast transfer the focus back to the dialog content? I just want to make sure that we're not losing it if it transfers back to the document

Yes the focus is returned to the modal after closing the toast.

kerryambt (approved):

kerryambt:
all good aside sharing tom's question on the package. I like that this fix works for things other than toasts too :)

thomasdigby:
> Yes the focus is returned to the modal after closing the toast. About the package upgrade it was the only thing that fixed the overlay issue, I guess they changed where the modal div is placed after this version. Do you think we should update any other package to keep them in the same version?

I think we'll be fine for now, but it might be worth considering as part of a guild day task, or something to discuss with the @Atom-Learning/internal-tools-pod

PR #326, opened 2022-04-14
fix: setting modal content and overlay z-index so it appears correctly on top of other components

viniciusvolpe:
### Description
After bumping Radix Dialog version to 0.1.2 the Dialog component was being rendered behind some components
Related PR: https://github.com/Atom-Learning/components/pull/322

### Implementation
- Wrapping dialog into a Portal
- Setting z-index to dialog overlay and content
- Wrapping the dialog content into the overlay to allow scrolling

### Screenshots
image

### DIFF
--- a/src/components/dialog/DialogContent.tsx
+++ b/src/components/dialog/DialogContent.tsx
@@ -1,5 +1,5 @@
 import { Close as CloseIcon } from '@atom-learning/icons'
-import { Close, Content, Overlay } from '@radix-ui/react-dialog'
+import { Close, Content, Overlay, Portal } from '@radix-ui/react-dialog'
 import * as React from 'react'
 
 import { keyframes, styled } from '~/stitches'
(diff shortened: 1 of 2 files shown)


andoulla:
@viniciusvolpe is the screenshot the before right?

alexandrunemes (approved):
The only way I could test this was to actually pull the CSS rules and apply them in my browser on the existing layout. It seems to work, so I'm approving this. What's the benefit from wrapping the layout in Portal? Is it to emulate the previous version's behaviour where it appended the modal to the body instead of nesting it within the layout?

andoulla (approved):

viniciusvolpe:
> @viniciusvolpe is the screenshot the before right?

yes, is the before

viniciusvolpe:
> The only way I could test this was to actually pull the CSS rules and apply them in my browser on the existing layout. It seems to work, so I'm approving this. What's the benefit from wrapping the layout in Portal? Is it to emulate the previous version's behaviour where it appended the modal to the body instead of nesting it within the layout?

Yes, also it was a suggestion from a similar issue in Radix repo: https://github.com/radix-ui/primitives/issues/1317

PR #327, opened 2022-04-20
chore: carousel - useCarousel hook added to expose carousel context o…

michalrozenek:
The hook to get Carousel Context data outside.

Example implementation based on the docs.
The current slide taken from the context.

https://user-images.githubusercontent.com/23363033/164211464-61cf8fd9-14c0-4f67-8c14-29798ee2b56f.mov


avirati (approved):

kerryambt (approved):

alexandrunemes:
So basically, with this hook, the goal is to be able to get the current slide outside of the actual carousel, within the CarouselProvider.

alexandrunemes (approved):

Mhoog (approved):

thomasdigby:
@michalrozenek Can we add some documentation to provide an example of the hook and how it might be used?

michalrozenek:
> @michalrozenek Can we add some documentation to provide an example of the hook and how it might be used?

Great idea @thomasdigby. I added it.

michalrozenek:
> So basically, with this hook, the goal is to be able to get the current slide outside of the actual carousel, within the CarouselProvider.

Yes, this hook let us GET the state of Carousel and react on state changes. Without the hook or HoC we were not able to get Carousel state in Student/Parent/etc apps. Now we will be able to subscribe on state changes and do something if we need.

PR #329, opened 2022-05-06
EmptyStates component

taniya-atomlearning:
## EmptyStates Component

This PR adds the new EmptyStates component to DS. There are 5 variants. `xs, sm, md, lg & xl`

[Ticket](https://atomlearningltd.atlassian.net/browse/INT-50)

[Figma](https://www.figma.com/file/KjgQh7eTwEfJmH6e56k9mX/Component-Documentation?node-id=2974%3A10309)

- xs, sm, md

Screenshot 2022-05-11 at 13 09 14

- lg, xl

Screenshot 2022-05-11 at 13 09 28


package.json
```diff
@@ -110,6 +110,7 @@
     "rollup": "^2.39.0",
     "rollup-plugin-esbuild": "^4.8.2",
     "rollup-plugin-summary": "^1.3.0",
+    "rollup-plugin-svg": "^2.0.0",
```
hollg, line 113: I believe we can get rid of this now, since the semicircle SVG isn't being stored in the library directly?

rollup.config.js
```diff
@@ -4,6 +4,7 @@ import resolve from '@rollup/plugin-node-resolve'
 import hq from 'alias-hq'
 import esbuild from 'rollup-plugin-esbuild'
 import summary from 'rollup-plugin-summary'
+import svg from 'rollup-plugin-svg'
```
hollg, line 7: We should be able to get rid of this (see prev comment)

src/components/empty-states/EmptyStates.mdx
```diff
@@ -0,0 +1,20 @@
+---
+title: Empty States
```
hollg, line 2: This should be singular, i.e. `EmptyState`. Might be simplest to do a find and replace to catch all the references in code etc.

src/components/empty-states/EmptyStates.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+const SimpleExample = () => (
+  <EmptyStates size="lg">
+    <EmptyStates.Title>No users found!</EmptyStates.Title>
+    <EmptyStates.Body>
+      You need to add some users before you can use this feature
+    </EmptyStates.Body>
+  </EmptyStates>
+)
+
+describe('EmptyStates component', () => {
```
hollg, line 16: Could we snapshot each size variant?

src/components/empty-states/EmptyStates.mdx
andoulla, line 2: I know this might sound minor but I think the component should be Empty State (singular) to match the pattern we have used for all the other components. (Button, Label etc).

Also, the plural kind of makes it sound that it is multiple states but it is only showing 1 empty state

src/components/empty-states/EmptyStates.tsx
```diff
@@ -0,0 +1,131 @@
...
+          height: '32px',
+          mb: '$4'
+        },
+        [`& ${EmptyStatesTitle}`]: {
+          fontSize: '$md',
+          fontWeight: '600',
+          mb: '$3'
+        },
+        [`& ${EmptyStatesBody}`]: {
+          color: '$tonal400',
```
taniya-atomlearning, line 28: I had to add color here for each variant of the body as it seems to conflict with the styling of the title when added in the EmptyStatesBody component file.

src/components/empty-states/EmptyStates.tsx
```diff
@@ -0,0 +1,131 @@
...
+      }
+    }
+  }
+})
+
+type IllustrationContainerProps = React.ComponentProps<typeof Image>
+
+const IllustrationContainer: React.FC<IllustrationContainerProps> = (props) => {
+  return (
+    <Box>
```
hollg, line 114: Is this `Box` still required? What does it do?

types/custom.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.svg' {
```
hollg, line 1: This can go now (see above comment)

tsconfig.json
```diff
@@ -17,6 +17,6 @@
     },
     "declaration": true
   },
-  "include": ["./src/*", "types/global.d.ts"],
+  "include": ["./src/*", "types/*"],
```
hollg, line 20: This is another change that was only required when we had svgs inside the library

src/components/empty-states/EmptyStates.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+import { Box } from '~/components/box'
+import { Flex } from '~/components/flex'
+import { Image } from '~/components/image'
+import { styled } from '~/stitches'
+
+import { EmptyStatesBody } from './EmptyStatesBody'
+import { EmptyStatesTitle } from './EmptyStatesTitle'
+
+const StyledBox = styled(Flex, {
```
hollg, line 11: I think it would be nice to give this a descriptive name like you have with the other components. Something like `EmptyStateContainer`.

src/components/empty-states/EmptyStates.tsx
```diff
@@ -0,0 +1,131 @@
...
+  flexDirection: 'column',
+  alignItems: 'center',
+  variants: {
+    size: {
+      xs: {
+        [`& ${Image}`]: {
+          width: '56px',
+          height: '32px',
```
andoulla, line 19: could we use the size tokens for these? `width: '$4', height:'$3'`

src/components/empty-states/EmptyStatesBody.tsx
```diff
@@ -0,0 +1,5 @@
+import { styled } from '~/stitches'
+
+export const EmptyStatesBody = styled('p', {
```
hollg, line 3: Did you ever try using our `Text` component here instead of `'p'`? `Text` already has `fontFamily: '$body'` applied, so unless any other styles on it conflict with the designs for this component, it might save some repetition.

src/components/empty-states/EmptyStates.tsx
```diff
@@ -0,0 +1,131 @@
...
+          mb: '$4'
+        },
+        [`& ${EmptyStatesTitle}`]: {
+          fontSize: '16px',
+          fontWeight: '600',
+          mb: '$3'
+        },
+        [`& ${EmptyStatesBody}`]: {
+          color: '$tonal400',
+          fontSize: '14px',
```
andoulla, line 65: could we do fontSize: "sm" here?

https://design.atomlearning.technology/theme/typography

src/components/empty-states/EmptyStates.tsx
```diff
@@ -0,0 +1,131 @@
...
+        }
+      },
+      md: {
+        [`& ${Image}`]: {
+          width: '126px',
+          height: '72px',
+          mb: '$4'
+        },
+        [`& ${EmptyStatesTitle}`]: {
+          fontSize: '16px',
```
andoulla, line 59: fontSize: "$md"

src/components/empty-states/EmptyStates.tsx
```diff
@@ -0,0 +1,131 @@
...
+        [`& ${EmptyStatesBody}`]: {
+          color: '$tonal400',
+          fontSize: '$sm',
+          fontWeight: '400',
+          mb: '$4'
+        }
+      },
+      sm: {
+        [`& ${Image}`]: {
+          width: '84px',
```
hollg, line 36: `size` sets both `height` and `width` together when they have the same vlaue, e.g. `size: '84px'`

hollg (changes requested):
This is an extremely impressive first outing in the DS! 👏🏻 

I know the nested styling for the variants got a bit complex but the outcome seems like a really nice-to-use API. 

I've requested a few changes but they're mostly just housekeeping to keep this PR consistent with conventions you haven't really been exposed to before / tidying up from when we had those SVGs in the library directly.

src/components/empty-states/EmptyStates.tsx
```diff
@@ -0,0 +1,131 @@
...
+          mb: '$5'
+        }
+      }
+    }
+  }
+})
+
+type IllustrationContainerProps = React.ComponentProps<typeof Image>
+
+const IllustrationContainer: React.FC<IllustrationContainerProps> = (props) => {
```
andoulla, line 112: Why not separate this into its own component file like the title and the body?

src/components/empty-states/EmptyStates.tsx
andoulla, line 11: Could we call this StyledContainer? Very very minor btw

src/components/empty-states/EmptyStates.mdx
andoulla: jinx

src/components/empty-states/EmptyStates.tsx
andoulla: the only concern is that we have a few empty states in assessments where we have a primary coloured body. Will we be able to overwrite this without the !important?

taniya-atomlearning: hmm not sure @hollg ?

package.json
taniya-atomlearning: sure, will remove it wasn't sure whether it'll be useful to leave for future use

src/components/empty-states/EmptyStates.test.tsx
taniya-atomlearning: sure will add

src/components/empty-states/EmptyStates.tsx
taniya-atomlearning: thanks, yea that makes sense

src/components/empty-states/EmptyStates.tsx
taniya-atomlearning: In the design figma I don't have the tokens, is there a place I can see a list of these? @andoulla

andoulla: https://design.atomlearning.technology/overview/introduction layout and space for this and typography for fonts

taniya-atomlearning: Spoke to Miguel, he mentioned the pixels are different to the existing size tokens

taniya-atomlearning: probably something to discuss with Miguel

src/components/empty-states/EmptyStates.tsx
taniya-atomlearning: As discussed @andoulla will leave the customisation for the EmptyStatesBody for v2 of the component

src/components/empty-states/EmptyStates.tsx
taniya-atomlearning: good shout will do

src/components/empty-states/EmptyStates.tsx
taniya-atomlearning: I changed the EmptyStateBody to use Text instead of p tag, I was able to move the color to the element level and we can customise the EmptyStateBody element now :)

src/components/empty-state/EmptyState.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+    const { container } = render(<SimpleExample size="lg" />)
+
+    expect(container).toMatchSnapshot()
+  })
+
+  it('renders a size xl component', async () => {
+    const { container } = render(<SimpleExample size="xl" />)
+
+    expect(container).toMatchSnapshot()
+  })
```
thomasdigby, line 45: I think that we're testing our CSS in JS lib here more than providing a useful test to ensure that the different sizes are rendered correctly. You'll see in the snapshot for this component that all size variant styles are rendered with each test; we could remove them all apart from one and just use that single test to ensure the styles are output correctly.

src/components/empty-state/EmptyState.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { Image } from '~/components/image'
+import { styled } from '~/stitches'
+
+import { EmptyStateBody } from './EmptyStateBody'
+import { EmptyStateImage } from './EmptyStateImage'
+import { EmptyStateTitle } from './EmptyStateTitle'
+
+const EmptyStateContainer = styled(Flex, {
+  flexDirection: 'column',
+  alignItems: 'center',
```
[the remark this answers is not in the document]
taniya-atomlearning: Thanks Tom, not sure about the minimum height, will discuss with the team today. I'll add the padding for each variant

src/components/empty-state/EmptyState.test.tsx
taniya-atomlearning: ok thanks, will try that

taniya-atomlearning: @thomasdigby  I couldn't find any examples in the components repo for this. but I tried something like this. Is this what you mean? Do you know any examples I could look at? Thanks
`  it('renders a component', async () => {
    const { container } = render(<SimpleExample size="lg" />)

    const h2 = document.getElementsByTagName('h2')
    const style = window.getComputedStyle(h2[0])
    expect(style.fontWeight).toBe('')
    expect(style.fontFamily).toBe('')
    expect(style.fontSize).toBe('')
  })`

taniya-atomlearning: sorry I might be misunderstanding your comment, do you mean just have a test with the basic component?
```  it('renders a component', async () => {
    const { container } = render(<SimpleExample />)

  })

Comment (thomasdigby) [reply src/components/empty-state/EmptyState.test.tsx:45]: Yeah exactly, sorry that was my bad explanation. I think we can have a single test to check that it renders as expected

Comment (taniya-atomlearning) [reply src/components/empty-state/EmptyState.test.tsx:45]: ok thanks

File: src/components/empty-state/EmptyState.test.tsx [hunk 18]
```diff
@@ -0,0 +1,38 @@
...
+    <EmptyState.Title>No users found!</EmptyState.Title>
+    <EmptyState.Body>
+      You need to add some users before you can use this feature
+    </EmptyState.Body>
+  </EmptyState>
+)
+
+describe('EmptyState component', () => {
+  it('renders a component', async () => {
+    const { container } = render(<EmptyState />)
```
thomasdigby, line 18: Should this be `SimpleExample`? We'll want to test `EmptyState` with the relevant child components

src/components/empty-state/EmptyState.tsx
```diff
@@ -0,0 +1,117 @@
...
+  variants: {
+    size: {
+      xs: {
```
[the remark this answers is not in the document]
hollg: This is super interesting and I think it's something we should establish a solid convention on. 

My gut feeling is that the reltionship being defined here is fundamentally about styling and that therefore it would be good to keep the logic in CSS (or rather CSS-in-JS) rather than in React code. Having one long file is definitely a cost to that approach, but we do have the option of splitting that big object into chunks and storing each chunk somewhere that feels more intuitive, e.g. nested styled for the title could be co-located with the title component.

That definitely isn't a perfect solution; we'd be introducing a layer of indirection between the definition of the container styles and the nested styles.

Maintaining the very dev-friendly API that Taniya has created is the priority imo, but I'd be keen to hear if you have any thoughts on the distinction between React code vs styling logic, and whether you think it matters etc @thomasdigby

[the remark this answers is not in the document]
LimeWub: For what it's worth - currently on my ToggleGroup component I've introduced a context for the sizing of the children (Which does exactly the same thing as cloneElement, and I could also use cloneElement but I think context is more flexible when it comes to nesting and there was not any loss to using it here afaik (?) Anyway!). So I'm with @thomasdigby here

Compared to having all the size styles on the parent; passing down a size prop was better for me because a) it's easier to read and manage and b) there's a chance the children can be of different visual styles, like a Card vs a Button and it was much cleaner to keep the size related differences in the elements themselves rather than having messy styling on the parent.

So: I would also be for doing things like sizing inside the nested components personally so there's that flexibility there if there ever need to be more variations of the empty state + it's easier to read.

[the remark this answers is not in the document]
taniya-atomlearning: I have updated the component to use the cloneElement now

taniya-atomlearning:
@thomasdigby I have addressed the feedback. Just one thing snapshot test is failing in the build even though its passing locally. Have you seen this before?

thomasdigby:
> @thomasdigby I have addressed the feedback. Just one thing snapshot test is failing in the build even though its passing locally. Have you seen this before?

Thanks! I'll have a quick look at that now

src/components/empty-state/EmptyState.tsx
```diff
@@ -113,8 +38,28 @@ export const EmptyState: React.FC<EmptyStateProps> & {
...
   Title: typeof EmptyStateTitle
   Body: typeof EmptyStateBody
-} = ({ size = 'sm', ...props }) => (
-  <EmptyStateContainer size={size} {...props} />
+} = ({ size = 'sm', children, ...props }) => (
+  <EmptyStateContainer size={size} {...props}>
+    {React.Children.map(children, (child) => {
+      // add relevant check for valid child
+      if (!React.isValidElement(child)) {
+        throw new Error()
```
andoulla, line 46: it is a small change but I think we should be passing an error message here. 
```suggestion
        throw new Error("Invalid child element in EmptyState")
```

or something similar

andoulla (approved):
Just one minor change but I think it is ready to go

src/components/empty-state/EmptyState.mdx
```diff
@@ -0,0 +1,20 @@
+---
+title: Empty State
+component: EmptyState
+description: Empty states are messages that provide an explanation of an interface in absence of content.
+category: Content
+---
+
+`EmptyState` exports a number of components that can be composed together to create an `EmptyState` message.
+The message can include Title and/or Body and sometimes they appear together with an illustration and actionable buttons.
```
thomasdigby, line 9: Do we have an example with an actionable button? I wonder if we should wither remove this copy (to make it easier to get this PR in) or introduce an `EmptyState.Action` component in this PR?

src/components/empty-state/EmptyState.tsx
```diff
@@ -0,0 +1,68 @@
...
+      if (!React.isValidElement(child)) {
+        throw new Error("Invalid child element in EmptyState")
+      }
+
+      if (
+        [
+          (<EmptyStateImage />).type,
+          (<EmptyStateTitle />).type,
+          (<EmptyStateBody />).type
+        ].includes(child.type)
```
thomasdigby, line 54: Could we try this rewrite to avoid invoking each component on every render? This has worked for our `ActionIcon` component (checking `child.type` against the component) and should be a touch more performant and readable.

```jsx
if (
  child.type === EmptyStateImage || 
  child.type === EmptyStateTitle || 
  child.type === EmptyStateBody
)
```

src/components/empty-state/EmptyStateImage.tsx
```diff
@@ -0,0 +1,41 @@
...
+        height: '213px',
+        mb: '$5'
+      }
+    }
+  }
+})
+
+type EmptyStateImage = React.ComponentProps<typeof StyledImage>
+
+export const EmptyStateImage: React.FC<EmptyStateImage> = (props) => (
```
thomasdigby, line 39: I think we can avoid this by exporting `StyledImage` as `EmptyStateImage` directly, e.g. `export const EmptyStateImage = styled(Image, {})`

types/custom.d.ts
```diff
@@ -0,0 +1 @@
+
```
thomasdigby, line 1: Can we remove this file?

thomasdigby (approved):
Just those minor changes then all good from me! 👍

src/components/empty-state/EmptyState.mdx
LimeWub: EmptyState.Action is just any type of button afaik - should there be a component for action which does includes and exports Button?

I think the text here makes sense as it matches the design but the Button is not a new component really.

thomasdigby: We probably ought to make that consistent though right? e.g. `Button size="sm"` where relevant, like how we've handled the other components

LimeWub: Asked Miguel and he says no, they are just flexible. So I think we just leave this to implementation.

Screenshot 2022-05-16 at 12 50 03

src/components/empty-state/EmptyState.tsx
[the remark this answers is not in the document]
LimeWub: Why would we wan't to default to a minimum size? I can ask Miguel about this but not sure I see the benefit. Is this still something we want to check about doing?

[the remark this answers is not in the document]
thomasdigby: Just to provide a bit of breathing room and space around the message, to make it easier to understand in the context of the view, but that might be solved by just increasing the padding a touch instead. It's something we can patch if its a problem

PR #330, opened 2022-05-09
fix: allow on change handler in radio button field

Marcdup1:
### Description

`RadioButtonField` does not expose a way of knowing when the value changes, it thus requires `watching` the form field to know when the value changes and then doing something. This PR adds an optional `onValueChange(value: string)` prop that will get called when a value in the radio group gets changed, allowing the caller to do something with that value.

### DIFF
--- a/src/components/radio-button-field/RadioButtonField.mdx
+++ b/src/components/radio-button-field/RadioButtonField.mdx
@@ -21,5 +21,5 @@ category: Form fields
 If you need to know when the radio group value changes, you can pass `onValueChange`, which takes a `string` type as input.
 
 ```tsx
-<RadioButtonField onValueChange={(value) => doSomethingWith(value)} />
+<RadioButtonField onValueChange={onChangeHandler} />
 ```
--- a/src/components/radio-button-field/RadioButtonField.mdx
+++ b/src/components/radio-button-field/RadioButtonField.mdx
@@ -17,3 +17,9 @@ category: Form fields
   </RadioButtonField>
 </Form>
 ```
+
+If you need to know when the radio group value changes, you can pass `onValueChange`, which takes a `string` type as input.
+
+```tsx
+<RadioButtonField onValueChange={(value) => doSomethingWith(value)} />
+```
--- a/src/components/radio-button-field/RadioButtonField.test.tsx
+++ b/src/components/radio-button-field/RadioButtonField.test.tsx
@@ -1,5 +1,6 @@
 import { IdProvider } from '@radix-ui/react-id'
-import { fireEvent, render, screen, waitFor } from '@testing-library/react'
+import { render, screen, waitFor } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
 import { axe } from 'jest-axe'
 import * as React from 'react'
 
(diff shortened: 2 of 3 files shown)


src/components/radio-button-field/RadioButtonField.mdx
```diff
@@ -17,3 +17,9 @@ category: Form fields
   </RadioButtonField>
 </Form>
 ```
+
+If you need to know when the radio group value changes, you can pass `onValueChange`, which takes a `string` type as input.
+
+```tsx
+<RadioButtonField onValueChange={(value) => doSomethingWith(value)} />
```
hollg, line 24: This can be simplified a little:
```suggestion
<RadioButtonField onValueChange={doSomethingWith} />
```

src/components/radio-button-field/RadioButtonField.test.tsx
```diff
@@ -1,15 +1,25 @@
 import { IdProvider } from '@radix-ui/react-id'
-import { render } from '@testing-library/react'
+import { fireEvent, render, screen, waitFor } from '@testing-library/react'
```
hollg, line 2: I think we should prefer `userEvent` over `fireEvent`, since it's a more realistic simulation of what happens with real user events: https://testing-library.com/docs/user-event/intro/#difference-to-fireevent

src/components/radio-button-field/RadioButtonField.tsx
```diff
@@ -60,7 +62,10 @@ export const RadioButtonField: React.FC<RadioButtonFieldProps> & {
           <RadioButtonGroup
             direction={direction}
             defaultValue={defaultValue}
-            onValueChange={onChange}
+            onValueChange={(value) => {
+              onChange(value)
+              onValueChange?.(value)
```
hollg, line 67: Now that I'm looking at the code, I'm not sure I properly understand the distinction that we're drawing between `onChange` and `onValueChange` 🤔 

They're called in the same set of circumstances, right? So to me it seems more intuitive that the user of this component would just add more stuff to the `onChange` func.

You mentioned a bug were `onChange` gets called multiple times per click. If we solved that bug, would it remove the need for this new prop?

hollg (changes requested):

src/components/radio-button-field/RadioButtonField.tsx
Marcdup1: `onChange` comes from the `render` prop function of the `<Controller />` component - so it doesn't provide a way to expose the value to the user still.

>You mentioned a bug where onChange gets called multiple times per click.

This was pertaining to the `onClick` passed into `<RadioButtonField.Item onClick={onClick} />` however I don't find it intuitive to add an on click to each Item?

src/components/radio-button-field/RadioButtonField.mdx
Marcdup1: Makes sense, updated it now.

src/components/radio-button-field/RadioButtonField.test.tsx
Marcdup1: Oh, interesting! Thank you for that!

src/components/radio-button-field/RadioButtonField.tsx
hollg: Oh, got it! I totally agree that this should be passed to the group rather than the individual item. Thanks for explaining!

hollg (approved):

D7Torres (approved):

PR #331, opened 2022-05-11
fix: update stitches version to fix error

hollg:
We had a weird bug where TS would complain that our styling objects didn't have a `length` property (see image below). Updating to the latest stiches seems to fix it.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -141,7 +141,7 @@
     "@radix-ui/react-tooltip": "0.1.1",
     "@radix-ui/react-visually-hidden": "0.1.1",
     "@reach/combobox": "^0.16.1",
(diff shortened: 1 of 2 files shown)


taniya-atomlearning (approved):

thomasdigby:
Are there any potential compatibility changes from 1.2.5? And is there a need to update the snapshots?

hollg:
> Are there any potential compatibility changes from 1.2.5? And is there a need to update the snapshots?

Apologies, I thought I had included updated snapshots here 😅. They just move some existing styles around in stylesheets. Should be there now! Re compat issues, I had assumed stitches' semver is reliable but you're right that it's worth looking into a bit more. Will do now.

PR #333, opened 2022-05-12
Feat/int 36 toggle group

LimeWub:
## Description
Component that lets users toggle between a group of limited options(Suggested range is 2-6 options).

Functionality based on the [ToggleGroup](https://www.radix-ui.com/docs/primitives/components/toggle-group) radix component, which already allows for: single/multiple select, disabling or partly disabling options, adds keyboard navigation and orientation and more.

Extends visually by allowing for vertical/horizontal display, gaps using stack or joined, rounded corners and so on.

## Screenshots
Screenshot 2022-05-12 at 18 59 22

Screenshot 2022-05-12 at 19 00 15

Screenshot 2022-05-12 at 19 00 32

Screenshot 2022-05-12 at 19 00 52

Screenshot 2022-05-12 at 19 02 09

Screenshot 2022-05-12 at 19 01 22

**Hover:**
Screenshot 2022-05-12 at 19 06 05

**Focus:**
Screenshot 2022-05-12 at 19 05 51

**Selected:**
Screenshot 2022-05-12 at 19 03 51

**Focus + Selected**
Screenshot 2022-05-12 at 19 05 59

**Hover + Selected + Focus**
Screenshot 2022-05-12 at 19 06 25


yarn.lock
```diff
@@ -7943,6 +8042,24 @@ mdast-util-from-markdown@0.8.5, mdast-util-from-markdown@^0.8.0:
     parse-entities "^2.0.0"
     unist-util-stringify-position "^2.0.0"
 
+mdast-util-from-markdown@1.2.0:
```
LimeWub, line 8045: Why is this here! D:
This is so irrelevant to what I've actually added to `package.json`

I've reset and rerun yarn but it's always there - this version was installed as a dependency for the WISIWYG SPIKE. T_T

src/components/toggle-group/ToggleGroupItem.tsx
```diff
@@ -0,0 +1,57 @@
+import * as ToggleGroup from '@radix-ui/react-toggle-group'
+
+import { styled } from '~/stitches'
+
+// TODO: How do I make this reusable from other components?
+const focusIndicatorBlock = ({ position = 'relative', zIndex = 1 } = {}) => {
```
LimeWub, line 6: Help w this pls :)

src/components/stack/Stack.tsx
```diff
@@ -34,28 +34,35 @@ const gap = Object.keys(theme.space).reduce(
...
       },
       'row-reverse': {
         flexDirection: 'row-reverse',
-        alignItems: 'center'
       },
       column: {
         flexDirection: 'column'
       }
     },
+    wrap: {
```
hollg, line 52: Could you explain the need for these changes to `Stack`?

src/components/toggle-group/ToggleGroupButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+}
+
+const px = {
+  sm: '$4',
+  md: '$5',
+  lg: '$5'
+}
+
+const spacingBetweenElements = {
+  sm: '8px',
```
hollg, line 21: Can these values be expressed in terms of our tokens? (I always forget the px values for each token ...)

src/components/toggle-group/ToggleGroupItem.tsx
hollg: Could you expand on what you want to do with this block? E.g. do you want to be able to spread it into _any_ component? Or maybe to create a component that applies the styles to its `children`, which we could wrap around other things?

src/components/toggle-group/ToggleGroupItem.tsx
```diff
@@ -0,0 +1,57 @@
...
+    boxShadow: 'white 0px 0px 0px 2px, $colors$primary 0px 0px 0px 4px'
+  }
+}
+
+export const StyledItem = styled(ToggleGroup.Item, {
+  bg: 'white',
+  color: '$tonal400',
+  border: '1px solid $tonal200',
+  cursor: 'pointer',
+  outline: 'none !important',
```
hollg, line 21: Why do we need `!important` here?

src/components/stack/Stack.tsx
LimeWub: Stack was **way** too specific with alignment and wrapping, that specificity was a hinderance to my use case (and I assume to most use cases). 
I wouldn't be able to use it for doing spacing unless I had more flexibility when it comes to wrapping (I needed no-wrap) and alignment (I needed no alignment so that heights/widths would match depending on orientation).

I put what was used to be the default as the default prop to the component so it should all be safe changes and not break anything not explicitly overriding them.

src/components/toggle-group/ToggleGroupButton.tsx
LimeWub: ye - I'll try to figure out which token they match to :S

src/components/toggle-group/ToggleGroupItem.tsx
LimeWub: I think I would like to be able to store these styles somewhere to import and spread to more focusable block components.

src/components/toggle-group/ToggleGroupItem.tsx
LimeWub: Why do we not need important? :p Is there a chance we would want to override outline for this ever?

I think it was just a quick way of removing it from the focus-visible state tbh, I don't remember why it's separate to the focus-visible style, I assume it's a WIP leftover. 

In fact I'm considering moving this to the focusIndicatorBlock styles....

Will move and test and lyk

src/components/stack/Stack.tsx
LimeWub: I just realised `column` with `align-items: center` may be a problem. Will test/update this.

src/components/toggle-group/ToggleGroupItem.tsx
LimeWub: Moving to focusIndicatorBlock w no `!important` has same effect.

src/components/stack/Stack.tsx
LimeWub: ^ Yup there was a change to behaviour introduced. I've added a line to fix this.

src/components/toggle-group/ToggleGroupItem.tsx
hollg: Are you struggling with where to store them, or how to import them elsewhere? We have a `utils` directory which I think would make sense as a place to store them — but I would only go down that route atm if you've already spoken to Miguel about the idea of using this elsewhere too.

src/components/stack/Stack.tsx
hollg: I feel a little sceptical about these changes. I think `Stack` does something quite specific _by design_. Did you not just want a `Flex`?

LimeWub: > Stack is a layout component that provides an abstraction over the flexbox gap property (~design system)

Why would this component have opinions of alignment, wrap and justify? It doesn't make sense that it would have opinions if it's the 'component to do gaps with'.

Tbf I don't even think we should have defaults for align/gap/etc but these are there (and the new ones have been added) just so the components already using them don't break.

src/components/toggle-group/ToggleGroupItem.tsx
LimeWub: Ye I was struggling about where to store it - I didn't notice the utils directory/wasn't sure if that's where this should go or if there's one just for css utils/stitches - I'm going to move it it in there

LimeWub: moved 👍

src/components/stack/Stack.tsx
hollg: > Why would this component have opinions of alignment, wrap and justify? It doesn't make sense that it would have opinions if it's the 'component to do gaps with'.

This is a good question — I'm not sure! But that's what makes me hesitant. I totally see the logic in what you're saying, it just seems to me that there are two possibilities here: either the code is over-reaching for what it's supposed to do and the docs are an accurate description, or the docs are wrong/outdated and `Stack` is intentionally doing more than just providing gaps.

Gonna ping @thomasdigby and @andoulla since I think one of them wrote the first version of `Stack` and maybe they can share their thoughts.

andoulla: I didn't write it but I remember that intentionality as I struggled to wrap my head around it in the beginning. the intention was to be very specific on what it does as it is just a Flex component with very specific CSS (StackContext follows the same principle). if we make it more flexible then we might as well just use Flex In my opinion.

LimeWub: The difference with Flex is that it Stack does gap spacing to its children. That's why I use Stack.

src/components/stack/Stack.tsx
```diff
@@ -34,28 +34,35 @@ const gap = Object.keys(theme.space).reduce(
...
-        flexDirection: 'row-reverse',
-        alignItems: 'center'
+        flexDirection: 'row-reverse'
       },
       column: {
         flexDirection: 'column'
       }
     },
+    wrap: {
+      true: {
```
andoulla, line 53: Having flexWrap as a truthy variant might be confusing as for all the other variants  justify = "start" which as develop tells me which option it applies

LimeWub: Would you prefer?: `wrap = "wrap" | "no-wrap" | "wrap-reverse"`

I did as true/false as it covers "wrap" | "no-wrap" without repeating the variant name in the value.

src/components/stack/Stack.tsx
thomasdigby: I think the phrase "component to do gaps with" isn't really accurate, perhaps the opening line in the documentation isn't as accurate as it could be, but the stack component kinda just does what it says, it allows you to stack things and provide space between them. Usually, this would just be `Flex` with a `gap` property, but we need a few extra layers to replicate that behaviour when it's unsupported.

I don't think "wrap" aligns with the intent of this component, but I can understand why it might be useful. Is this something required of the toggle group?

src/components/stack/Stack.tsx
andoulla: I think I do because it would be more consistent with the other ones. It will be easier to remember

src/components/stack/Stack.tsx
LimeWub: `flexWrap: 'wrap'` and `alignItems: 'center'` were there before - I just picked them out to props so I'm able to use this component without redoing gap logic with flex.

thomasdigby: Cool, I can get on board with that, although I do think that we ought to consider the overlap between `Flex` and `Stack` components, these changes definitely move them closer together which isn't necessarily a bad thing but we need to document those differences and use-cases for both. (in future)

One thing that we'll need to be careful with is that changing the styles and props for `Stack` will require a major version change, so can we ensure that we set defaults to match how it currently renders.

LimeWub: Just to make sure I'm understanding this:
I've set the defaults to match the old functionality (will do another check to make sure) so I don't think it's really a breaking change BUT do you want me to mark it as a BREAKING CHANGE anyway and get a major version bump?

hollg: @LimeWub apologies for my misunderstanding about what changed here/the reasoning earlier! [this comment](https://github.com/Atom-Learning/components/pull/333#discussion_r875609690) made it all clearer to me :D

I definitely agree about avoiding a breaking change but from what I've seen I think your default props should do that. Maybe we could create some snapshots with the old version and then run them again with the new one to check.

I'm gonna resolve this thread 👍🏻

src/components/icon/Icon.tsx
```diff
@@ -3,7 +3,7 @@ import * as React from 'react'
 import { styled } from '~/stitches'
 import { Override } from '~/utilities/types'
 
-const StyledIcon = styled('svg', {
+export const StyledIcon = styled('svg', {
```
hollg, line 6: How confident are we that we can't target the `Icon` in nested styles without exporting `StyledIcon` directly? If it's the only way then fair enough, but it would be really nice if we could avoid it. I really thought we had managed this before 🤔

LimeWub:
> I missed this in my previous review, but we need some docs for the new component 😄

Yikes! I think I had them in there and I resolved/deleted them. Will re-add!

src/components/icon/Icon.tsx
LimeWub: I'll try it again.

andoulla:
Just a suggestion, but I think we should amend the documentation to reflect some of the things we discussed, on the intentions of the component and the new variants. This way in 6months we can avoid having the same confusion

LimeWub:
> I think we should amend the Stack documentation
I 100% agree on this - But is the Stack documentation in a separate repo rn? I think so and I think @hollg is merging it with this one so I will update it when it's merged (or at least that's the plan)

src/components/icon/Icon.tsx
LimeWub: Ye, It doesn't work with `Icon`

LimeWub:
> Just a suggestion, but I think we should amend the Stack documentation to reflect some of the things we discussed, on the intentions of the component and the new variants

@andoulla Ok I thought I'd just update the table at the bottom of the page here: https://design.atomlearning.technology/components/stack But I think that may auto-update? I don't think the text actually needs any edits.

src/components/stack/Stack.tsx
```diff
@@ -67,17 +80,23 @@ export const Stack: React.FC<StackProps> = ({
   css,
   gap = 2,
   direction = 'row',
+  wrap = 'wrap',
   justify = 'start',
+  align,
```
andoulla, line 85: why not add a default to align?

src/components/toggle-group/ToggleGroup.mdx
```diff
@@ -0,0 +1,86 @@
...
+  <ToggleGroup.Button value="b">B</ToggleGroup.Button>
+  <ToggleGroup.Button value="icon">
+    <Icon is={Upload} />
+  </ToggleGroup.Button>
+</ToggleGroup.Root>
+```
+
+## Full width
+
+`isFullWidth={boolean}`
```
andoulla, line 54:
```suggestion
`isFullWidth={true|false}`
```

just a small suggestion, as the other examples list the values not the type

src/components/toggle-group/ToggleGroup.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+    <ToggleGroup.Button value="button_b">button B</ToggleGroup.Button>
+    <ToggleGroup.Button value="button_icon">
+      <Icon is={Upload} />
+      Button with icon
+    </ToggleGroup.Button>
+  </ToggleGroup.Root>
+)
+
+describe('ToggleGroup', () => {
+  it('renders', async () => {
```
andoulla, line 30: I would suggest testing the other variants too, especially the disabled one.

src/components/toggle-group/ToggleGroupButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+    flexShrink: 0
+  },
+  variants: {
+    size: {
+      sm: getSizeVariant('sm'),
+      md: getSizeVariant('md'),
+      lg: getSizeVariant('lg')
+    },
+    isIconOnly: {
+      true: {
```
andoulla, line 53: not sure if feasible, but it feels that we should be able to pass the isIconOnly into getSizeVariant and just not set margin, instead of setting it and then having to overwrite it

andoulla: Also, if that's not a choice what if we tried having 
```
 isIconOnly: {
      true: {}
}
```

and in the compountVariant part, we tried to set the py and px in the CSS prop there. Will, that help avoid the important?

src/components/toggle-group/ToggleGroupButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+    }
+  ]
+})
+
+export const ToggleGroupButton: React.FC<
+  React.ComponentProps<typeof StyledButton>
+> = ({ size, children, ...rest }) => {
+  const childrenArray = React.Children.toArray(children)
+  const isSingleChild = childrenArray.length <= 1
+  const isIconOnly =
```
andoulla, line 89: I like this, much cleaner than other places

src/components/toggle-group/ToggleGroupItem.tsx
```diff
@@ -0,0 +1,47 @@
...
+      borderColor: 'currentColor !important',
+      color: '$primaryMid'
+    },
+    '&:focus-visible': {
+      ...focusVisibleStyleBlock(),
+      '&[data-state="off"]': {
+        borderColor: '$tonal200 !important'
+      },
+      '&[data-state="on"]': {
+        boxShadow: `inset currentColor 0px 0px 0px 1px, ${
```
andoulla, line 30: should this shadow be added into the theme?

src/components/toggle-group/ToggleGroupRoot.tsx
```diff
@@ -0,0 +1,160 @@
...
+          '&:not(:last-child)': {
+            borderBottomColor: 'transparent'
+          }
+        }
+      }
+    }
+  ]
+})
+
+const orientationToDirection = (o) => {
```
andoulla, line 113: Very minor, I always advocate for using the full word instead of the initial as it increases readability. we don't gain anything with shortening it

src/components/toggle-group/ToggleGroupRoot.tsx
```diff
@@ -0,0 +1,160 @@
...
+  switch (o) {
+    case 'horizontal':
+      return 'row'
+    case 'vertical':
+      return 'column'
+    default:
```
andoulla, line 119: I think we don't need the switch as this is a binary check and always will be, it can either be horizontal or vertical. an if statement is more than enough and we don't need to have this empty default

`return orientation === "horizontal" ? "row": column"`

src/components/toggle-group/ToggleGroup.mdx
```diff
@@ -0,0 +1,86 @@
...
+component: ToggleGroup
+description: Extends functionality the Toggle Group radix component
+category: Content
+---
+
+Lets users toggle between a group of limited options(Suggested range is 2-6 options).
+
+Functionality based on the [`ToggleGroup`](https://www.radix-ui.com/docs/primitives/components/toggle-group) radix component, which already allows for: single/multiple select, disabling or partly disabling options, adds keyboard navigation and orientation and more.
+
+Extends visually by allowing for vertical/horizontal display, gaps using stack or joined, rounded corners and so on.
```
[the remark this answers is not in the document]
LimeWub: Aren't the examples below enough information regarding visual changes? 
I don't see the point of saying I'm adding a white background and primary colour font, I think that's unnecessary info to the user who can just see the examples etc.

src/components/toggle-group/ToggleGroup.mdx
LimeWub: The other values expect an enum type of value though; I'm not necessarily against this - although boolean is always `true|false` - so I don't see a benefit to this change other than opinion

src/components/toggle-group/ToggleGroupItem.tsx
LimeWub: I'm not sure I'm understanding this but I don't thinkkkkkkkk soooooo? It's just a visual for this specific component.

src/components/toggle-group/ToggleGroupRoot.tsx
LimeWub: I agree, I'll change it.

src/components/toggle-group/ToggleGroupRoot.tsx
LimeWub: you're right - I'll change this to either an if or the above ternary

src/utilities/style/focus-visible-style-block.ts
```diff
@@ -0,0 +1,29 @@
...
+  position: string
+  zIndex: number
+  boxShadow: string
+}
+
+export const focusVisibleStyleBlock = ({
+  position = 'relative',
+  zIndex = 1
+}: focusVisibleStyleBlockPropsType = {}): focusVisibleStyleBlockReturnType => {
+  return {
```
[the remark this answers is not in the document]
LimeWub: Ye maybe - not sure how I'd go about it but will have a look.

src/components/toggle-group/ToggleGroupButton.tsx
LimeWub: 1) Not sure how I can pass a prop from a stitches variant - I think it may not be possible but can have a look.

2) If I had compound variant then I'd need to have 3 variants (1 for each size) which I think is messier than the `!important`. 

I do understand no liking the `!important`, although I don't see us ever wanting to override it....

src/components/toggle-group/ToggleGroupItem.tsx
andoulla: oh this is a specific shadow just for this component? 

What I meant is that we do have some default shadows in our theme repo, and I was just wondering if we could either use something from there or add this one there. However, if it is only for this component then we don't need to do it. 

I was just thinking that we were trying to standardise a bit the different shadows we had. (but every rule has exceptions)

src/components/toggle-group/ToggleGroupButton.tsx
andoulla: because I have been caught out saying we will never want to overwrite it and then having to overwrite something, my suggestion is that if we can avoid locking it we should.

I don't see the harm in having 3 compound variants, to be honest. The button for example has quite a lot more

src/components/toggle-group/ToggleGroupItem.tsx
LimeWub: Oh ye it's actually a 'hack' inside shadow to make the border look like 2px wide when selected rather than 1px - very specific to this design.

src/components/stack/Stack.tsx
andoulla: quick q, is Align mandatory for TS or optional?

src/components/toggle-group/ToggleGroup.mdx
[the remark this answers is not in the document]
andoulla: oh then the so on is unnecessary in my opinion. to make it more precise I would just switch to something like. I would just list the things the Dev can control with the variants
```suggestion
Extends the Radix component visually by allowing for vertical/horizontal display and gaps using stack or joined.
```

src/components/toggle-group/ToggleGroup.mdx
andoulla: this is why I said it is a small suggestion.  it is just a consistency thing. not to precious on it though

src/components/toggle-group/ToggleGroup.test.tsx
andoulla: fair enough about the disabled part. what about testing a single option with horizontal orientation.

src/components/toggle-group/ToggleGroupItem.tsx
andoulla: 👍

src/components/toggle-group/ToggleGroup.test.tsx
LimeWub: Orientation is part of ToggleGroup (functional), and direction (visual) is part Stack :o

src/components/toggle-group/ToggleGroup.mdx
LimeWub: imma leave it as it is :)

src/components/toggle-group/ToggleGroup.mdx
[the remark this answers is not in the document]
LimeWub: Okay - I'lll reword 👍

src/components/stack/Stack.tsx
LimeWub: I'm not sure - it uses the `typeof` the stitches setup, I'm not getting good typescript highlighting for it tho...

```
type StackProps = React.ComponentProps<typeof StyledStack> & { css?: CSS }
```

andoulla: because I think it might be mandatory where in that situation users will have to pass it in anyway. We will know as soon as you use it in the feature you need it for right?

LimeWub: It doesn't suggest me props when I use it in ToggleGroup.Root :S

For what it's worth I pass `align={false}` in my use case for ToggleGroup so that it never aligns and just stretches instead.

src/components/toggle-group/ToggleGroupButton.tsx
LimeWub: There were actually already compound variants for isIconOnly+size .-. lol
I stuck a clear padding there - much better than the importants, good catch 👍

src/components/stack/Stack.tsx
andoulla: I wonder why it doesn't suggest Props, is it a stitches bug or something wrong in our setup. but I diverge

src/components/toggle-group/ToggleGroup.mdx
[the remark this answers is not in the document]
LimeWub: done - think it's cleaner now.

hollg (approved):

andoulla (approved):

src/utilities/style/focus-visible-style-block.ts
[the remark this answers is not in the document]
LimeWub: I tried doing this with a Snapshot but couldn't get it to render out the CSS so it wasn't testing right.
I don't know how else to test it.

**code:**
Screenshot 2022-05-20 at 14 05 28

**result:**
Screenshot 2022-05-20 at 14 05 09

If you have any idea what looks wrong with this test lmk :/ planning to merge this today at some point.

PR #335, opened 2022-05-19
Feat: update radix packages

hollg:
Our `radix` package dependencies are a mess and it's beginning to cause some conflicts elsewhere. Notably, attempts to combine the documentation site repo with this one are stalling and these conflicts are part of the reason. (They're being made worse by other conflicts in that branch, but let's not get into that here ...)

This PR upgrades each `radix` package to the latest stable version and makes any required changes to our components to fix tests and/or avoid a breaking change, e.g. adding new props with default values. I'm also pinning each package to a specific version since `radix` does not currently use semver (they're working on it) and therefore specifying a range in `package.json` means we have occasionally got surprise breaking changes from upstream. 

- [ ] `@radix-ui/react-accordion`
- [ ] `@radix-ui/react-alert-dialog`
- [ ] `@radix-ui/react-checkbox`
- [ ] `@radix-ui/react-dropdown-menu`
- [ ] `@radix-ui/react-id`
- [ ] `@radix-ui/react-popover`
- [ ] `@radix-ui/react-progress`
- [ ] `@radix-ui/react-radio-group`
- [ ] `@radix-ui/react-slider`
- [ ] `@radix-ui/react-switch`
- [ ] `@radix-ui/react-tabs`
- [ ] `@radix-ui/react-visually-hidden`


src/components/alert-dialog/AlertDialogContent.tsx
```diff
@@ -63,14 +63,19 @@ const StyledAlertDialogContent = styled(Content, {
 
 type AlertDialogContentProps = React.ComponentProps<
   typeof StyledAlertDialogContent
->
+> & { ariaLabel: string }
 
 export const AlertDialogContent: React.FC<AlertDialogContentProps> = ({
   size = 'sm',
+  ariaLabel = 'Alert',
```
hollg, line 70: This is required to pass our a11y test. I'm open to suggestions for a better default label!

thomasdigby: Is there any reason to not pass `aria-label` directly rather than providing an abstraction?

hollg: I added this to facilitate a default value, thinking that it was a breaking change otherwise. But all that breaks is our internal tests, which can be fixed by directly passing `aria-label` as you suggest. It had been a long day 😅

src/components/alert-dialog/AlertDialog.test.tsx
```diff
@@ -9,7 +9,7 @@ const AlertDialogTest = (props) => (
   <IdProvider>
     <AlertDialog {...props}>
       <AlertDialog.Trigger>TRIGGER</AlertDialog.Trigger>
-      <AlertDialog.Content>CONTENT</AlertDialog.Content>
+      <AlertDialog.Content aria-label="Alert!">CONTENT</AlertDialog.Content>
```
LimeWub, line 12: Interesting change 🤔 
I wonder why it would add this specific word as the label.... Do you think this is fallback logic because we don't have an `AlertDialog.Title` child? .-.

src/components/tooltip/__snapshots__/Tooltip.test.tsx.snap
```diff
@@ -3,10 +3,9 @@
 exports[`Tooltip component renders a tooltip 1`] = `
 <div>
   <button
-    aria-describedby="radix-id-0-1"
+    aria-describedby="radix-0"
     class="c-PJLV"
     data-state="delayed-open"
-    type="button"
```
LimeWub, line 9: Also interesting change 🤔 
I'd think having the type there is actually better....

src/components/dialog/Dialog.tsx
```diff
@@ -13,11 +13,15 @@ type DialogProps = React.ComponentProps<typeof StyledDialog>
...
   Content: typeof DialogContent
+  Description: typeof Description
+  Title: typeof Title
   Trigger: typeof Trigger
 } = (props) => <StyledDialog {...props} />
 
 Dialog.Close = DialogClose
 Dialog.Content = DialogContent
+Dialog.Description = Description
+Dialog.Title = Title
```
LimeWub, line 24: Interesting that we're now forwarding these on when we didn't use to, but we're not forwarding 'Portal' and 'Overlay'. 
Is our Dialog component literally just the radix dialog? (What's the point of it then?)

LimeWub (approved):
I'm approving because it's just bumping dependencies but some of the changes in the deps are.... interesting.

src/components/alert-dialog/AlertDialog.test.tsx
LimeWub: I see a resolved comment about this but I don't see the change to the file in the diff 👀 
wot

hollg: I'm not sure I follow your question, sorry! This is a test file, so I've written this label.

LimeWub: omg I'm blind

src/components/dialog/Dialog.tsx
hollg: `Portal` and `Overlay` are used directly in our `DialogContent`. We've standardised their usage, but `Title` and `Description`'s usage will depend on values that are specific to each usage so I don't think we can standardise that.

I've used the same pattern here that we already have in `AlertDialog`.

src/components/alert-dialog/AlertDialog.test.tsx
LimeWub: But also - does it actually affect the test? I don't see it being there in the snapshot (?) (Am I being blind again?)

hollg: The `a11y` violations test and snapshot are two separate tests

LimeWub: ;-; the test doesn't show content it only shows the trigger duh.

weird and confusing but anyway ignore me.

src/components/dialog/Dialog.tsx
```diff
@@ -1,4 +1,4 @@
-import { Root, Trigger } from '@radix-ui/react-dialog'
+import { Description, Root, Title, Trigger } from '@radix-ui/react-dialog'
```
andoulla, line 1: should the documentation be updated to say that these are now exported?

For the future, we need to find a better way of documenting our DS because it is becoming a bit of a hassle to  have to check our documentation and then Radix to find the combination of information that is split between the two

andoulla (approved):

src/components/dialog/Dialog.tsx
hollg: Ah, I missed these ones, thanks! 

We don't always simply expose what Radix offers directly, e.g. here we're not exporting `Portal` because we use that directly inside our `Dialog.Content`. So we definitely need to keep on top of our own docs so that they are the real source of truth.

We've been _accidentally_ getting new Radix features into the library because they're not using semver, but hopefully now that the versions are all pinned we don't need to worry about discrepancies between the two sets of docs too much. Annoyingly, there is still a manual step in keeping up with their releases so that we can upgrade though, until they start using semver

andoulla: it seems like an odd choice to not have a type of versioning, not necessarily semver if it doesn't work for them.

when we have time maybe something that will check the code for exports coming from radix and add them to the documentation "automatically", like we do with the props

PR #336, opened 2022-05-20
feat: add badge component

andoulla:
### Description:

We need a simple badge component that will enable us to deprecate the old different versions of badges we have (bootstrap etc)

https://www.figma.com/file/KjgQh7eTwEfJmH6e56k9mX/DS-Documentation?node-id=4479%3A31223

**Screenshot:**
image

If the text is too long 
image

### Notes: 
- Future versions will include the indicator version and maybe merge the notification badge but it is TBC


src/components/notification-badge/NotificationBadge.mdx
```diff
@@ -22,12 +22,14 @@ It displays a badge in top right of the content it is wrapping, which displays a
...
 ```tsx preview
-<NotificationBadge value={88}>
-  <ActionIcon appearance="outline" size="lg">
-    <Icon is={Controls} />
-  </ActionIcon>
-</NotificationBadge>
-<NotificationBadge value="hi">
-  <Button theme="warning">I'm a button!</Button>
-</NotificationBadge>
+<>
```
andoulla, line 25: This is currently breaking the documentation
image

src/components/badge/Badge.tsx
```diff
@@ -0,0 +1,71 @@
...
+      info: {
+        bg: '$primaryLight',
+        color: '$primaryMid'
+      }
+    },
+    size: {
+      xs: {
+        fontSize: '$sm',
+        px: '$3',
+        height: '24px'
```
hollg, line 40: Do we need to specify `height` for the `size` variants? If a `Badge` ever gets wrapped onto two lines because of a change of screen size etc then this will break

src/components/notification-badge/NotificationBadge.mdx
hollg: 😅

src/components/badge/Badge.tsx
andoulla: would it be better to ensure the correct spacing using py?

hollg: Hm yeah that makes sense to me!

andoulla: unfortunately, it doesn't work as well as I thought, as I would have to hardcode the spacings and not use the tokens.
If I use our existing tokens it looks quite bigger 
image

However, I know that in case of text not fitting we shouldn't be breaking it into two lines so I can set the no-wrap option on it.

hollg: Won't `no-wrap` just create situations where it the text overflows horizontally, instead? I'd speak to Miguel about the sizes issue

andoulla: ok I will, I'm also researching if there is a way of doing ellipsis without setting any width as that will also solve the problem and is what we need

andoulla: Could you have another look? Now with the ellipsis working setting the height doesn't cause the same problem

LimeWub: The way I did this with toggle group is set a smaller padding than expected and instead of height use 'minHeight' maybe this would work well for this too?

LimeWub: I don't think adding ellipsis is a good idea for usability btw

andoulla: It was requested by miguel. https://www.figma.com/file/KjgQh7eTwEfJmH6e56k9mX/DS-Documentation?node-id=4006%3A15616

LimeWub: I've lit had the same discussion with Miguel for the ToggleGroup buttons so I feel you (although I had successfully pushed back for ToggleGroup) .-.
I stand by the fact ellipsis is rly bad for usability, alas...

andoulla: I will have a look at the solution on the padding and height on Monday. and will also have a chat with Miguel. just so I know what is the issue with the ellipsis? I guess is it the fact that we aren't pairing it with a tooltip so they can see the rest of the content.

LimeWub: okay ye I get the use for it actually (it pains me but it's okay for catching edge cases) ignore moi ;-;

LimeWub: > I guess is it the fact that we aren't pairing it with a tooltip so they can see the rest of the content

Ye the problem I have with it is that they can't see the content.
I understand it's an edge case though (it would just suck if it would cut something that's important).

Btw there's a bug on this rn with scrollbars (Miguel spotted it when he got the code running) so something is wrong with the Box hack below :/

andoulla: ok, so I have switched the height to `minHeight` like it is in the `togglegroup`. however, I'm considering keeping the ellipsis after having a long think over the weekend.  My reasoning is that, if there is a chance that the copy is going to be long, we shouldn't be using this component. We should be using some of the other surface components (like one of the banners we have).  We do have a few choices for showing dynamic copy which we don't have in use cases we need to use a toggle group if that makes sense.

Also, I will talk to Miguel about this issue with the scrollbar as I can't replicate it

src/components/badge/Badge.tsx
```diff
@@ -0,0 +1,82 @@
...
+      info: {
+        bg: '$primaryLight',
+        color: '$primaryMid'
+      }
+    },
+    size: {
+      xs: {
+        fontSize: '$sm',
+        px: '$3',
+        minHeight: 'calc(24px - $space$1)'
```
alexandrunemes, line 41: following the discussion about the layout breaking if the text wraps, and your reply (with which I agree) that for potentially longer strings we shouldn't be using this component, doesn't setting `minHeight` kind of assume that the text will still potentially wrap? Also, is this meant to allow for some vertical _**padding**_ ? _**wink wink**_

src/components/badge/Badge.tsx
LimeWub: > if there is a chance that the copy is going to be long (...)

I'm not disagreeing with this. I was more concerned about the parent width being too narrow, or perhaps the font size too big (A11Y override) with similar effects.

I am not too precious about changing this or keeping the ellipsis rn, because this is just a badge component, but generally speaking I think doing something which cuts out content - making it inaccessible - just to preserve design is bad for the user.



I saw the scrollbars they were overflowY scrollbars. I think the padding + font height was greater than the height(? didn't inspect but looked like it). You shouldn't be getting them with minHeight but the height will be probably b off. If you're going with ellipsis I'd stick with height (minHeight if line breaks) and  give the vertical padding a small value (just in case) and let the centering happen with flex. Then hopefully no scrollbars o>o

andoulla: So I just checked. minHeight doesn't seem to cause any issues other than the fact that it needed some padding to stop cutting the text (the lower part of a g for instance). Now that I have added a tiny bit of padding it works as expected and it doesn't wrap the as the text has a `white-space: nowrap` on it for the ellipsis to work. 

However, since I will keep the ellipsis for now I have switched to height.

andoulla:
@LimeWub @hollg could you please take another look?

src/components/badge/Badge.tsx
andoulla: so the scrolling issue was kind of self-resolved as we can't replicate anymore so I'm closing the thread

src/components/badge/Badge.tsx
andoulla: I have switched to height (https://github.com/Atom-Learning/components/pull/336#discussion_r879426481) and also added some padding to solve a different bug

alexandrunemes (approved):

hollg (approved):

PR #337, opened 2022-05-23
feat(form): added optional onError callback to handleSubmit

Jdrazong:
Added optional onError callback prop that can be passed to Form component. React hook form passes both onError and onSubmit callback to handleSubmit function and calls onSubmit if there are no errors and onError instead if any errors are present.

### DIFF
--- a/src/components/form/Form.mdx
+++ b/src/components/form/Form.mdx
@@ -36,7 +36,7 @@ const InputField = ({ label, name, validation }) => {
 
 ## Form submission
 
-To handle form submission onSubmit and optional onError callback props can be passed to Form. If there are any validation errors present onError will be called, in case of no errors the onSubmit will be called instead.
+To handle form submission, `onSubmit` and optional `onError` callback props can be passed to Form. If there are any validation errors present `onError` will be called, in case of no errors the `onSubmit` will be called instead.
 
 ## ValidationOptions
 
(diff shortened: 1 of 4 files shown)


hollg (approved):

src/components/form/Form.mdx
```diff
@@ -34,6 +34,10 @@ const InputField = ({ label, name, validation }) => {
 }
 ```
 
+## Form submission
+
+To handle form submission onSubmit and optional onError callback props can be passed to Form. If there are any validation errors present onError will be called, in case of no errors the onSubmit will be called instead.
```
thomasdigby, line 39:
```suggestion
To handle form submission, `onSubmit` and optional `onError` callback props can be passed to Form. If there are any validation errors present `onError` will be called, in case of no errors the `onSubmit` will be called instead.
```

src/components/form/Form.test.tsx
```diff
@@ -191,4 +191,41 @@ describe(`Form component`, () => {
       '12345'
     )
   })
+
+  it('should call onError if there are erros in the form and onSubmit when the errors are gone', async () => {
```
thomasdigby, line 195:
```suggestion
  it('should call onError if there are errors in the form and onSubmit when the errors are gone', async () => {
```

PR #340, opened 2022-06-08
fix: the ProgressBar now accepts absolute values and calculates the percentage based on them

Mhoog:
## Issue
- The progress bar only accepted a `value` from 1 to 100, because the `value` was directly used as a percentage for the indicator width.
- The `value` wasn't passed onto the `ProgressBar` component, which meant that the `aria-valuenow` attribute was never set. As a result, `data-state` was always `indeterminate`.

## Changes
- Passed `value` onto the `ProgressBar` component, so that the correct attributes are set.
- The indicator bar's size is now calculated based on `value` and `max`.
- Changed the transition from using `width` to `transition`. This should make it slightly smoother.

### DIFF
--- a/src/components/progress-bar/__snapshots__/ProgressBar.test.tsx.snap
+++ b/src/components/progress-bar/__snapshots__/ProgressBar.test.tsx.snap
@@ -32,16 +32,20 @@ exports[`ProgressBar component renders a progress bar using default props 1`] = 
     aria-label="label"
     aria-valuemax="100"
     aria-valuemin="0"
+    aria-valuenow="50"
+    aria-valuetext="50%"
     class="c-rTKYu c-rTKYu-ixnjYu-theme-primary"
     data-max="100"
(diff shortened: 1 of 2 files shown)


src/components/progress-bar/__snapshots__/ProgressBar.test.tsx.snap
```diff
@@ -32,16 +32,20 @@ exports[`ProgressBar component renders a progress bar using default props 1`] =
     aria-label="label"
     aria-valuemax="100"
     aria-valuemin="0"
+    aria-valuenow="50"
+    aria-valuetext="50%"
     class="c-rTKYu c-rTKYu-ixnjYu-theme-primary"
     data-max="100"
-    data-state="indeterminate"
```
thomasdigby, line 37: 😩

thomasdigby (approved):

PR #342, opened 2022-06-09
feat: table theme light

kerryambt:
### Description

I needed a light theme for the table component, where we have a grey header and no alternating row colours.

https://atomlearningltd.atlassian.net/jira/software/projects/PAR/boards/11?selectedIssue=PAR-126

### Implementation

- Added `light` theme to `Table.Header`
- Extended `Table.Body` to accept a striped prop, which is a boolean.
- Removed the alternating background colour from `Table.Cell` so it can be handled at a `Table.Body` level
- Updated snapshots and mdx

### Screenshots

image

### DIFF
--- a/src/components/table/Table.mdx
+++ b/src/components/table/Table.mdx
@@ -129,10 +129,18 @@ The `Table` component by default renders with a slight border radius. If you wan
 <Table corners="square">...</Table>
 ```
 
-`Table.Header` accepts a theme prop, the current available options are `primary` and `primaryDark`. The default is `primaryDark`.
+`Table.Header` accepts a theme prop, the current available options are `primary`, `primaryDark`, and `neutral`. The default is `primaryDark`.
 
 ```tsx
 <Table>
-  <Table.Header theme="primary">...</Table.Header>
+  <Table.Header theme="neutral">...</Table.Header>
+</Table>
+```
+
+`Table.Body` accepts an appearance prop, which determines whether the rows will be alternate coloured or not. The current options are `alternating` and `simple`, and the default is `alternating`.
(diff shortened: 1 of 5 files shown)


hollg:
Have you spoken with Miguel about these changes? He's got designs ready for a grey version of the Table header so we definitely want to make sure that we're not putting something similar-but-not-the-same in. I don't think changes to whether row colours alternate have come up before, but that's also a conversation he should be involved in

src/components/table/Table.mdx
```diff
@@ -129,10 +129,18 @@ The `Table` component by default renders with a slight border radius. If you wan
...
+`Table.Header` accepts a theme prop, the current available options are `primary`, `primaryDark`, and `neutral`. The default is `primaryDark`.
 
 ```tsx
 <Table>
-  <Table.Header theme="primary">...</Table.Header>
+  <Table.Header theme="neutral">...</Table.Header>
+</Table>
+```
+
+`Table.Body` accepts an appearance prop, which determines whether the rows will be alternate coloured or not. The current options are `alternating` and `simple`, and the default is `alternating`.
```
andoulla: I'm glad this is separated out. not sure about simple as a name though. is there a better description for no highlighting

andoulla (approved):

src/components/table/Table.mdx
kerryambt: @andoulla just chatted to miguel and we've changed it to a striped boolean instead :)

andoulla: thanks

hollg (approved):

PR #344, opened 2022-06-16
fix: tab triggerlist uppercase

kerryambt:
### Description

I noticed that `Tabs.TriggerList` wasn't doing anything when `appearance="uppercase"` was passed in.

### Implementation

In order to stop the textTransform property being overridden, I set the CSS from the uppercase variant to apply to descendant buttons instead of on the object itself.

### DIFF
--- a/src/components/tabs/TabsTriggerList.tsx
+++ b/src/components/tabs/TabsTriggerList.tsx
@@ -11,6 +11,7 @@ import { styled, theme } from '~/stitches'
 
 import { TabTrigger } from './TabTrigger'
 import { passPropsToChildren } from './utils'
+
 interface ListProps extends React.ComponentProps<typeof StyledTriggerList> {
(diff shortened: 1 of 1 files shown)


src/components/tabs/TabsTriggerList.tsx
```diff
@@ -66,7 +67,7 @@ const StyledTriggerList = styled(List, {
     },
     appearance: {
       uppercase: {
-        textTransform: 'uppercase'
+        '& button': { textTransform: 'uppercase' }
```
thomasdigby, line 70: Could we target the `TabTrigger` itself? e.g. `'& ${TabTrigger}': { textTransform: 'uppercase' }`

thomasdigby: Yeah that's all good, an interesting issue but not a blocker 👍

PR #346, opened 2022-06-23
Move documentation site code into main repo

hollg:
This PR moves our docs site code into the DS library repo. This is step 1 of a project to add a git-based CMS that gives non-engineers a nice UI to update the component documentation stored in the library repo and see the changes in the docs site.

The approach is add a layer of nesting to this repo to create a monorepo containing both lib and docs. We use `yarn` workspaces to manage the `site` package's dependency on `lib`. 

Running the docs site locally requires that the `lib`'s `dev` script is running (see [README](https://github.com/Atom-Learning/components/pull/346/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5)). This provides hot-reloading support on the docs site for changes to TS and TSX files (thank you @LimeWub for this suggestion!).

Because so many files have been added and all existing files have been moved, this PR looks _huge_. I would recommend pulling the branch and looking at it in your IDE rather than in the GitHub UI!

## Future work
There are other incremental improvements that could be made here, e.g. moving to `yarn` v3 and using `lerna` to improve some workflows. I have avoided those changes in this PR because this is already so huge in scope. They will go into the IT pod backlog.


.github/workflows/release.yml
```diff
@@ -4,7 +4,8 @@ on:
   push:
     branches:
       - main
-
+    paths:
+    - 'lib/**'
```
hollg, line 8: @thomasdigby My intention here is to prevent `release` from running for any PR that only makes changes to the `site` package. Does this look right to you?

Issue #295, opened 2022-02-09
Feature request: ActionIcon


andoulla:
## Component Name(\*) or Feature Name

`ActionIcon` 
## Description(\*)

Please include the expected behaviour and the acceptance criteria for this component.
The component currently sets the nested icon size to 20px in case the ActionIcon size is set to `lg` or `xl`. However, when using the `appearance=simple` the smaller icon looks very small so doesn't work visually

Proposal:
Amend the code to allow the size to be cascaded down if the appearance is set to simple?

## Designs and UX wireframes(\*)

TBC

PR #350, opened 2022-06-23
fix: action icon sizes

kerryambt:
From https://github.com/Atom-Learning/components/issues/295

#### EDIT 27/6
Following lots of discussion with design, we have decided to make this a PR to align sizes between ActionIcons, Icons, and Buttons. We will be unfortunately introducing breaking changes with this PR, in order to keep the sizing consistent as there is current a difference between what each one considers sm, md, etc. We want a size md ActionIcon to be the same size as a size md ActionIcon, so have implemented that change here.

The main goal behind this was to make icons sized inside action icons properly. This has been achieved with compound variants. 

Every instance of Icon and ActionIcon in the DS has also been checked and updated where necessary in order to keep sizing consistent to what it was.

Current:
Screenshot 2022-06-23 at 16 27 51
Fixed:
image

### Issues

This does pose an issue in that we'll have to audit core in order to not break things, but the more we put this off the worse this will become in terms of we'll just keep making tech debt. We have a ton of ActionIcons that have either overridden CSS or just strange styling, because this sizing has always not matched in the way it should.

I have another guild day to take next week and am happy to leave this unmerged until then, and spend the time auditing core in order to make sure all our ActionIcons retain their desired styles.


src/components/action-icon/ActionIcon.tsx
```diff
@@ -253,10 +263,11 @@ export const ActionIcon = React.forwardRef<HTMLButtonElement, ActionIconProps>(
           )
 
           return React.cloneElement(child, {
-            css: {
-              size: ['lg', 'xl'].includes(size as string) ? 20 : 16,
-              ...(child.props.css ? child.props.css : {})
-            }
+            size:
+              child.props.size && appearance === 'simple'
```
andoulla, line 267: Maybe I'm confused but since we check the size and simple appearance here and return the size icon that is passed, then why do we need the same check within the getIcon on line 201

src/components/action-icon/ActionIcon.tsx
```diff
@@ -82,6 +82,7 @@ const StyledButton = styled('button', {
       solid: {}
     },
     size: {
+      sm: { size: '$2 ' },
```
andoulla, line 85: I get this notification from Github about a TS error  on line 1
image

andoulla (approved):

andoulla:
> This does pose an issue in that we'll have to audit core in order to not break things, but the more we put this off the worse this will become in terms of we'll just keep making tech debt. We have a ton of ActionIcons that have either overridden CSS or just strange styling because this sizing has always not matched in the way it should.

@kerryambt I have a clean up PR with some Icons I replacing in core, happy to mark the ones I had to override the size with the same string comment to help make them more discoverable

hollg:
Have you seen this Jira ticket for a fix to `ActionIcon` styling? Just asking because I don't see a reference to it here and want to avoid any overlapping/conflicting work! https://atomlearningltd.atlassian.net/jira/software/projects/DS/boards/27?selectedIssue=DS-66

src/components/action-icon/ActionIcon.tsx
kerryambt: so we allow size specified on the icon (i.e. if they did something like `<ActionIcon size="md"><Icon size="lg"/></ActionIcon>`) to override the action icon size only when it's appearance simple. the check on line 201 is checking based on the size from the action icon, but in instances where we want to override we allow `child.props.size` to be used.

src/components/action-icon/ActionIcon.tsx
kerryambt: yeah I have fixed this just now, was a type error.

kerryambt:
@andoulla that would be really helpful! 

@hollg I'd never seen that board before, so no :o I will link this PR there now :)

andoulla (approved):

thomasdigby:
I'm not loving the icon to container size ratio for medium, can we reduce the size of the icon just a touch in that context? e.g. 1/2px reduced size

I guess we need to consider how that would work if it was used alongside the standard `Icon`s, would the size difference matter...?

src/components/date-input/DateInput.tsx
```diff
@@ -67,10 +67,15 @@ export const DateInput = React.forwardRef<HTMLInputElement, DateInputProps>(
...
+              css={{
+                position: 'absolute',
+                top: '50%',
+                right: '$1',
+                transform: 'translateY(-50%)'
+              }}
               disabled={disabled}
               label={labels.open}
-              size={size === 'sm' ? 'md' : 'lg'}
+              size={size}
```
thomasdigby, line 78: Shouldn't the size of the `ActionIcon` button container match up with the `Input`? Could we avoid the extra positioning CSS above by retaining some form of conditional logic here like we had before?

MiguelGDLR: Hi, I just found this issue because Kerry commented a similar task I created. Can you tag me when there are discussions about changes in any style? otherwise, I can not keep track of them so design (Figma components) will be inconsistent with development.

kerryambt: @thomasdigby I have updated this to use the conditional logic again :)

kerryambt:
> I'm not loving the icon to container size ratio for medium, can we reduce the size of the icon just a touch in that context? e.g. 1/2px reduced size
> 
> I guess we need to consider how that would work if it was used alongside the standard `Icon`s, would the size difference matter...?

@thomasdigby slightly tweaked the sizing to give a little more space on medium, I agree it looks better. I also realised I could simplify the conditional logic by using compound variants, so have changed up the code a bit.

src/components/action-icon/ActionIcon.tsx
```diff
@@ -215,7 +209,7 @@ export const ActionIcon = React.forwardRef<HTMLButtonElement, ActionIconProps>(
       children,
       theme = 'primary',
       appearance = 'simple',
-      size = 'md',
+      size = 'sm',
```
andoulla, line 212: is it a good idea to change the default value of this? It will cause a bit of a pain in core to update all of the instances where md was assumed the default

src/components/icon/Icon.tsx
```diff
@@ -12,12 +12,32 @@ export const StyledIcon = styled('svg', {
   verticalAlign: 'middle',
   variants: {
     size: {
-      sm: { size: '$1', strokeWidth: '1.75' },
-      md: { size: '$2', strokeWidth: '2' },
-      lg: { size: '$3', strokeWidth: '2' },
-      xl: { size: '$4', strokeWidth: '2' }
+      xs: { size: '$1', strokeWidth: '1.75' },
```
andoulla, line 15: we need to make this update a breaking change when we release it.  This change affects everything

src/components/icon/Icon.tsx
```diff
@@ -29,7 +49,7 @@ type IconProps = Override<
 >
 
 export const Icon: React.FC<IconProps> = React.forwardRef(
-  ({ is: SVG, size = 'md', ...remainingProps }, ref) => (
+  ({ is: SVG, size = 'sm', ...remainingProps }, ref) => (
```
andoulla, line 52: similar to above, I don't think we should change the default value as it will make the migration harder

andoulla (changes requested):

src/components/action-icon/ActionIcon.tsx
kerryambt: the tokens had to change so they matched buttons. so what was md is now sm. changing the default value makes it less likely we need to change as many places.

src/components/icon/Icon.tsx
kerryambt: as above :)

src/components/icon/Icon.tsx
andoulla: to do it you need to add a BREAKING_CHANGE in the commit message, as this is automatically released
image
https://design.atomlearning.technology/components/versioning

kerryambt: thanks! saves me looking for that later :)

thomasdigby:
> > I'm not loving the icon to container size ratio for medium, can we reduce the size of the icon just a touch in that context? e.g. 1/2px reduced size
> > I guess we need to consider how that would work if it was used alongside the standard `Icon`s, would the size difference matter...?
> 
> @thomasdigby slightly tweaked the sizing to give a little more space on medium, I agree it looks better. I also realised I could simplify the conditional logic by using compound variants, so have changed up the code a bit.
> 
> image



> > I'm not loving the icon to container size ratio for medium, can we reduce the size of the icon just a touch in that context? e.g. 1/2px reduced size
> > I guess we need to consider how that would work if it was used alongside the standard `Icon`s, would the size difference matter...?
> 
> @thomasdigby slightly tweaked the sizing to give a little more space on medium, I agree it looks better. I also realised I could simplify the conditional logic by using compound variants, so have changed up the code a bit.
> 
> image

Great, thanks. Looks better now. Will take another look through this update today 👍

src/components/action-icon/ActionIcon.mdx
```diff
@@ -10,16 +10,16 @@ category: Navigation
 ```tsx preview
 <>
   <ActionIcon label="Attach a file" appearance="simple">
-    <Icon is={Clip} size="sm" />
```
LimeWub, line 13: Is there a reason we are removing the `size` prop here?
I kindddddd of don't like removing the explicit size prop. 
Like, I can see the default has changed but if the default ever changes again this will just be more things to add back to the code - plus, I don't really like depending on the defaults of the components for things controlled via props and think all props should be explicit.

andoulla: I on the other hand find the defaults very handy. In my head, the default is the standard of what we should be using (sizes or themes etc) and it points out the intentionality of when we need to change from the default

kerryambt: yeah I'm with @andoulla here - we want to assume certain things with the design system in order to allow fast iteration on ideas. not having defaults goes against that imo.

the size was removed here because it is now properly passed down from ActionIcon, so that Icons within ActionIcons are explicitly and consistently sized to the size of their ActionIcon.

I do see your point, and don't 100% disagree in that there's definitely merit to that line of thought, I just think it's not what we've decided the DS should be. Happy to open up more conversations on the matter but most of the decisions in this PR came from Miguel and I feel like at the end of the day DS owning designer has more weight in the choices the DS makes than developers :)

LimeWub: Okay so, honestly, I (once again) hadn't noticed this was an `.mdx` rather than a `.tsx` file and thought it actually had impact to real components. (I thought it was the icon for the Accordion we were changing or something cause while scanning I combined the filename of this and the file above ;_;) - oops.

Ye, removing the prop doesn't matter on an `.mdx` file and passing the size down from the ActionIcon makes sense. 

My bad for flagging this.

kerryambt: It's better to flag things and it be okay than not flag things! MDX files can be sneaky 😅

kerryambt:
@andoulla @thomasdigby @MiguelGDLR changes have been made that _finally_ have all of us on the same page. I've added breaking change tags as well. should be good to go now finally!

thomasdigby:
Thanks @kerryambt I'll review this evening!

src/components/action-icon/ActionIcon.mdx
```diff
@@ -10,16 +10,16 @@ category: Navigation
...
+    <Icon is={Clip} />
   </ActionIcon>
   <ActionIcon label="Attach a file" appearance="subtle">
-    <Icon is={Clip} size="sm" />
+    <Icon is={Clip} />
```
thomasdigby, line 16: Sorry, this is a "while we're here", but can we remove the `appearance="subtle"` variants in the docs? That variant was removed a while ago so the docs are out of date

src/components/action-icon/ActionIcon.tsx
```diff
@@ -253,10 +254,9 @@ export const ActionIcon = React.forwardRef<HTMLButtonElement, ActionIconProps>(
           )
 
           return React.cloneElement(child, {
-            css: {
-              size: ['lg', 'xl'].includes(size as string) ? 20 : 16,
-              ...(child.props.css ? child.props.css : {})
-            }
+            size: child.props.size || size,
+            isInActionIcon: !child.props.size,
```
thomasdigby, line 258: Is there any way we can avoid this prop? I get the benefit of being explicit about the context and making styling decisions in `Icon` easier, but it feels like we're coupling the components a bit too much. I wonder if just a different name might help? Worth bearing in mind that this prop will be exposed via the props table in the docs

I also wonder if we can more explicit about converting the `child.props.size` here to a boolean, I didn't initially understand the logic here.

src/components/notification-badge/__snapshots__/NotificationBadge.test.tsx.snap
```diff
@@ -54,18 +54,18 @@ exports[`NotificationBadge component renders a NotificationBadge 1`] = `
...
   .c-fDzwZw-rloUt-isRounded-true {
     border-radius: var(--radii-round);
   }
 
-  .c-dbrbZt-hemXWm-size-md {
-    height: var(--sizes-2);
-    width: var(--sizes-2);
+  .c-dbrbZt-fLGGKn-size-lg {
+    height: var(--sizes-3);
+    width: var(--sizes-3);
```
thomasdigby, line 68: Did we need to change the size prop in `NotificationBadge`? Seems like both `height`/`width` values are changing in this snapshot

kerryambt: it's just the icon used in the test. because it's size lg, which previously was using incorrect sizing, it's had all its value changed.

src/components/action-icon/ActionIcon.tsx
kerryambt: I think a different name might suffice, I think it's better than having a bunch of complicated logic to determine size solely on the action icon. maybe like `isWrapped` or `isInContainer`?

thomasdigby:
@andoulla @kerryambt I'm not against skipping the breaking change and just bumping it as a minor update. I want to get to a place where we're actually sticking to Semver soon, but we've done plenty of breaking changes previous to this without a major bump.

src/components/action-icon/ActionIcon.tsx
thomasdigby: Actually, I think it's fine as is, I don't think there's a need to decouple from `ActionIcon` when there's a clear dependency on it. 👍

lib/src/components/action-icon/ActionIcon.tsx
```diff
@@ -253,10 +254,9 @@ export const ActionIcon = React.forwardRef<HTMLButtonElement, ActionIconProps>(
           )
 
           return React.cloneElement(child, {
-            css: {
-              size: ['lg', 'xl'].includes(size as string) ? 20 : 16,
-              ...(child.props.css ? child.props.css : {})
-            }
+            size: child.props.size || size,
+            isInActionIcon: !child.props.size,
```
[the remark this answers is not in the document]
andoulla: I guess we could just say these are the sizes and don't override them (just like we do with button). Less flexibility sometimes means that it is easier to use the component

andoulla (approved):

lib/src/components/action-icon/ActionIcon.mdx
```diff
@@ -66,7 +60,7 @@ It is recommended to pair `ActionIcon` with `Tooltip` to provide a clear visual
 <Tooltip>
   <Tooltip.Trigger>
```
andoulla, line 61: I think this section is misleading. We have a TooltipActionIcon in core where we do just that. Maybe we either do @kerryambt idea where we just give the option to show a tooltip as a thing this component does or we move the TooltipAction icon here next to it, so people know it exists. 

@thomasdigby @hollg 


Obviously, this needs to be part of a follow-up and not here

thomasdigby (approved):

lib/src/components/icon/Icon.tsx
```diff
@@ -15,8 +15,33 @@ export const StyledIcon = styled('svg', {
...
+      css: { size: '$1', strokeWidth: '1.75' }
+    },
+    {
+      size: 'sm',
+      isInActionIcon: true,
+      css: { size: '$1', strokeWidth: '2' }
+    },
+    {
+      size: 'md',
+      isInActionIcon: true,
```
LimeWub, line 36: Is there a reason why we are depending on `isInActionIcon` prop in `Icon` to do the sizing of this?

I don't think the `Icon` should even know if it's in ActionIcon - can we not target this - or just pass the right size from ActionIcon itself?

kerryambt:
@hollg @LimeWub following a conversation with Myrto I realised I could avoid compound variants _and_ conditional logic and use a simple map to size the Icon based on the ActionIcon size. This has the benefit of working with the new Icon stroke widths and removing the unnecessary compound variant CSS. Could you take a look when you get time? :)

PR #354, opened 2022-06-29
Update components to be used in TextEditor

LimeWub:
This PR solves some actual issues I had with using design system compinents in the new SlateJS TextEditor we're developing (+ some small things I noticed along the way).

Required for: https://github.com/Atom-Learning/atom-core/pull/1799

### Main things this solves:
* I need forwardRefs to be added to our components that can go in a TextEditor in atom-core; SlateJS needs them to do contentEditable magic/manipulations on the components without throwing errors in the console every time something is typed.
No more of these:
Screenshot 2022-06-29 at 10 45 33

* StackContent needs to forwardRef, once again due to me wanting to use it as the Slate `Editable` element on the TextEditor so we get spacing. Without the forwardRef the page explodes so I'm hoping the forwardRef will resolve this; and also give us the spacing between the elements as needed.
No more of this either:
Screenshot 2022-06-29 at 10 50 06

* I also refactored StackContent to no longer do cloneElements to apply its styles - Tbf this started with me trying to make it more generic so I could just nest the `Editable` inside of it rather than making the `Editable` **be it** (`<Editable as={StackContent} />`) BUT it's a bad idea to make it more generic and targeting direct children is probably the way to go anyway. 
Regardless, I can imagine it not doing cloneElements and element checking helps my case with using it with Slate; and I think it makes much more sense to apply styles in... well... styles - rather than cloningElements and passing down CSS (overcomplicating it). This also simplifies us needing to check and clone specific elements which extend the base text elements. AFAIK we no longer need to specifically check if the element (to be cloned) is eg. a MarkdownImage rather than an Image, cause anything that uses Image just gets the style we need. (win!) 

### Smaller fixes:
* I had forgotten to export ToggleGroup way back when I developed it, this actually exports it.
* I was working on the Markdown elements while trying to get StackContent to work and did some minor renaming. I think the renaming makes more sense in terms of keeping the pattern of `Component` / `StyledComponent` so I'm also adding to this PR rather than reverting it.
* I added forwardRefs to Accordion since I was working on it to debug some typescript issues with list and it's a good change to it so worth adding in rather than reverting.

**Update**: I have tested whether this PR solves the expected problems, as mentioned above - it does! I have also done a pass on the website to check for StackContent regressions and nothing jumps out to me as having regressed so I think this is a safe change; but feel free to have a look for yourselves via this build, which uses a pre-release tagged version of this PR: https://int-162-fe-update-text-editor-and-preview-to-render-des-pre.atomlearning.review/


src/components/list/List.tsx
```diff
@@ -6,7 +6,7 @@ import { textVariantSize } from '../text'
```
LimeWub, line 6: This file is giving me TS errors locally - can you help me resolve?
Screenshot 2022-06-29 at 11 05 05

hollg: Do we need a different type instead of `React.FC` when we're forwarding refs around?

LimeWub: Perhaps? The other components that I changed to forwardRef did not complain though; but it does seem there's something wrong with the type...

src/components/accordion/Accordion.tsx
```diff
@@ -11,21 +11,23 @@ const StyledRoot = styled(Root, {
   width: '100%'
 })
 
-type AccordionProps = React.ComponentProps<typeof StyledRoot> & {
+type AccordionPropsType = React.ComponentProps<typeof StyledRoot> & {
```
hollg, line 14: What's the reason for appending `Type` here?

LimeWub: OCD / Trying to have things use the same pattern

LimeWub: ....and I'm just realising it's not even that consistent ;-;

It's just `AccordionType` has to have `Type` in the name to not clash with `Accordion`; then the other type in the file should also have `Type` in my head for consistency although I am not noticing this is super inconsistent across the entire codebase really (including files I've edited) so... :( 
But ye I just renamed it so it matches the other 'Type'

LimeWub: I'll switch it back to what it was for now and if we ever decide on a convention for this we just change things then.

hollg: Cool, thanks. It does seem like a convention for defining a _thing_ and that thing's _type_ separately would be a useful one to have!

LimeWub: Although, I don't think this is an unusual case to need to specify the Type of the component in a separate definition. I think we do this a lot in core - but ye, previous comment still stands. Just going to revert for now.

hollg (approved):

thomasdigby (approved):

PR #355, opened 2022-06-29
Move docs site into lib repo

hollg:
This PR moves our docs site code into the DS library repo. This is step 1 of a project to add a git-based CMS that gives non-engineers a nice UI to update the component documentation stored in the library repo and see the changes in the docs site.

The approach is to use `yarn workspaces` to turn this repo into a monorepo with two packages: `lib` and `site`. `yarn` handles `site`'s dependency on `lib` for us. 

Running the docs site locally requires that `lib`'s `dev` script is running (see [README](https://github.com/Atom-Learning/components/pull/346/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5)). This provides hot-reloading support on the docs site for changes to TS and TSX files (thank you @LimeWub for this suggestion!).

Because so many files have been added and all existing files have been moved, this PR looks huge. I would recommend pulling the branch and looking at it in your IDE rather than in the GitHub UI!

## Future work
There are other incremental improvements that could be made here, e.g. moving to yarn v3 and using lerna to improve some workflows. I have avoided those changes in this PR because this is already so huge in scope. They will go into the IT pod backlog.


.github/workflows/release.yml
```diff
@@ -4,7 +4,8 @@ on:
   push:
     branches:
       - main
-
+    paths:
```
hollg, line 7: @thomasdigby My intention here is to prevent `release` from ever running on a merge that doesn't make changes in the `lib` package. Does this look right to you?

I previously had `paths` indented at the same level as `push`, but I started getting emails from GitHub about attempted `release` runs for my PR even though it hadn't merged, so I thought it seemed like the two filters were being treated as an `OR` when they were on the same level

thomasdigby: I'll take a look, definitely need to cover that use-case 👍

lib/.gitignore
```diff
@@ -0,0 +1,8 @@
+node_modules
+dist
+.cache
+.eslintcache
+yarn-error.log
+coverage
+stats.html
+.DS_Store
```
thomasdigby, line 8: Shouldn't we have a single `.gitignore` at the root?

lib/src/components/markdown-content/components/MarkdownImage.tsx
```diff
@@ -11,5 +11,5 @@ type MarkdownImageProps = {
 }
 
 export const MarkdownImage: React.FC<MarkdownImageProps> = ({ node, css }) => (
-  <Image src={node.url} alt={node.alt} css={css} />
+  <Image src={node.url} alt={node.alt ?? undefined} css={css} />
```
thomasdigby, line 14: This isn't quite well enough supported yet for us to use, unless you've seen ESBuild transpile it away? https://caniuse.com/mdn-javascript_operators_nullish_coalescing

package.json
```diff
@@ -1,209 +1,41 @@
...
-              "pattern": "main",
-              "notifyOnFail": true,
-              "notifyOnSuccess": true
-            }
-          ]
-        }
-      ]
-    ]
-  }
+	"private": true,
```
thomasdigby, line 2: Shouldn't this be `private: false`?

site/.editorconfig
```diff
@@ -0,0 +1,14 @@
...
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false
```
thomasdigby, line 14: Are we able to pull most of these dot files to the root?

package.json
hollg: `"private": true` is required in order for `workspaces` to work, to prevent you from publishing the workspace root. It doesn't affect the nested packages, though.

lib/rollup.config.js
```diff
@@ -28,6 +29,7 @@ export default {
     { dir: 'dist', format: 'esm', preserveModules: true }
   ],
   plugins: [
+    // peerDepsExternal(),
```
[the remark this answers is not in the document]
hollg: Ah, this is a hangover from a previous attempt to fix the issue with multiple versions of React running at once. I can just get rid of the commented code. Thanks for catching!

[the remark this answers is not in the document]
thomasdigby: Ah I also didn't see that it was additional, not commented out from the original config 👍

lib/src/components/markdown-content/components/MarkdownImage.tsx
hollg: What's our support threshold? I thought it was supported in the latest version of all our supported browsers and that that's good enough 😅

thomasdigby: Our baseline still includes Safari 12 on `core`, but I just checked the output from the build and it transforms it to a compatible spec so we're all good 👍

site/.editorconfig
hollg: You're right that there are some we can move. I did try moving them all, but when I got into things like `lintstagedrc` and overlaps with `eslintrc` etc, they felt a bit more complex. I looked at the `atom-core` setup for reference, and each package there has its own config for most of those files. 

I'll move the stuff that seems uncontroversial now, and then I think the other stuff might be simpler to consolidate if we setup `lerna` in this repo and can use it to config scripts to target certain packages more easily etc.

thomasdigby (approved):

LimeWub:
Just to confirm, is this still supposed to be in there?

Screenshot 2022-07-01 at 11 24 57

There's another one in here
Screenshot 2022-07-01 at 11 25 51

README.md
```diff
@@ -1,26 +1,16 @@
...
-const Component = () => {
-  return (
-    <Box as="article">
-      <Heading css={{ mb: '$3' }}>This is a heading</Heading>
-      <Text>This is a paragraph of text</Text>
-    </Box>
-  )
-}
-```
+In another terminal, run `yarn dev:site` to run the documentation site at `http://localhost:3000`, taking the output as the previous commands as its input.
```
LimeWub, line 12: Do we want to be losing all the previous documentation from this file?

ie. This url could be useful:
```
You can read more about the components included in `@atom-learning/components` at [https://design.atomlearning.technology](https://design.atomlearning.technology)
```

We could rephrase to:
```
When a PR is merged into `main`, this repo is automatically released under `@atom-learning/components` and published at [https://design.atomlearning.technology](https://design.atomlearning.technology)
```

README.md
```diff
@@ -1,26 +1,16 @@
...
 
-![badge](https://img.shields.io/npm/v/@atom-learning/components)
-![badge](https://img.shields.io/github/workflow/status/Atom-Learning/components/Test%20&%20validate)
-![badge](https://img.shields.io/bundlephobia/minzip/@atom-learning/components)
+This repo contains both the Design System (`./lib`) and its documentation site (`./site`).
 
-`@atom-learning/components` is a collection of generic React components built using [stitches](https://stitches.dev), it includes components that cover layout, content, data collection, media and a host of other UI concerns. It uses `@atom-learning/theme` to provide the design tokens for color, typography, spacing and layout.
+## Working with the repo locally ft. useful scripts
 
-```
```
LimeWub, line 9: This is also useful?

README.md
```diff
@@ -1,26 +1,16 @@
-# Atom Learning Components
+# Atom Learning Design System
 
-![badge](https://img.shields.io/npm/v/@atom-learning/components)
```
LimeWub, line 3: Also this intro is *potentially* useful? Or it like, explains a bit what this repo is.

LimeWub (approved):

PR #356, opened 2022-06-30
Fix: DialogContent - handle case where child can be null

avirati:
## Description
We have started seeing some failing tests which have been traced to the changes introduced in `v1.17.0`

```
TypeError: Cannot read property 'type' of null
        at /Users/avinah/workspace/atom-core/.yarn/__virtual__/@atom-learning-components-virtual-a3a313397e/0/cache/@atom-learning-components-npm-1.17.0-c6922d29f2-c0ffecf250.zip/node_modules/@atom-learning/components/dist/index.cjs.js:1:30825
```

Adding a null check in attempt to fix it

### DIFF
--- a/src/components/dialog/DialogContent.tsx
+++ b/src/components/dialog/DialogContent.tsx
@@ -91,7 +91,8 @@ export const DialogContent: React.FC<DialogContentProps> = ({
     <StyledDialogOverlay id={modalOverlayId}>
       {React.Children.map(
         children,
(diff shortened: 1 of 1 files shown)


hollg:
@LimeWub and @bstyles3 have also been seeing failing tests which seem to be tracked back to `1.17.0` — tagging here so they can see whether this looks like a fix to their issues too.

viniciusvolpe (approved):

thomasdigby (approved):

PR #361, opened 2022-07-11
feat: Stepper New States

kriszzy01:
### Description

This PR introduces the following updates found in the design documentation [here](https://www.figma.com/file/KjgQh7eTwEfJmH6e56k9mX/DS-Documentation?node-id=5466%3A42338)

- Switch orientation  horizontally and vertically
- Utilize component as a controlled component whose state can be controlled externally from a parent component
- Update step states with corresponding icons, background colours and colours.
- Add step labels

Screenshot 2022-07-11 at 13 45 01
Screenshot 2022-07-11 at 13 45 32


andoulla:
@kriszzy01 could we please add some screenshots of the other colours introduced and also could we add the link of the figma file in the description? It will make it easier to review

lib/src/components/stepper/Stepper.mdx
```diff
@@ -45,13 +45,55 @@ Below is a more complex example that shows how to dynamically render the button
...
+      // do something before
+      goToNextStep()
+    }}
+  >
+    Next
+  </Stepper.StepForward>
+</Stepper>
+```
+
+The root `Stepper` component also allows the user to pass a `step` prop in order to render step labels. When the `step` prop is passed, the `Stepper` component will be used as a controlled component thereby taking control of it's state. When used as a controlled component, navigation across steps, and identifying the state of a step would be controlled by the array of steps passed to the component.
```
andoulla, line 64:
```suggestion
The root `Stepper` component also allows the user to pass a `step` prop in order to render step labels. When the `step` prop is passed, the `Stepper` component will be used as a controlled component thereby taking control of its state. When used as a controlled component, navigation across steps, and identifying the state of a step would be controlled by the array of steps passed to the component.
```

lib/src/components/stepper/Stepper.mdx
```diff
@@ -45,13 +45,55 @@ Below is a more complex example that shows how to dynamically render the button
...
+
+Below is an example that shows how to use the controlled component Each object in the `step` array must include a `label` and a `status` property to determine what content to render based on the current position.
```
andoulla, line 66:
```suggestion
Below is an example that shows how to use the controlled component. Each object in the `steps` array must include a `label` and a `status` property to determine what content to render based on the current position.
```

lib/src/components/stepper/Stepper.test.tsx
```diff
@@ -179,4 +179,119 @@ describe('Stepper', () => {
...
+    render(
+      <Stepper allowSkip {...props} orientation="horizontal">
+        <Stepper.StepBack label={() => 'Back'} />
+        <Stepper.Steps />
+        <Stepper.StepForward
+          label={(activeStep) => (activeStep === 2 ? 'Start' : 'Next')}
+        />
+      </Stepper>
+    )
+    expect(screen.getByTestId('horizontal')).toBeVisible
```
andoulla, line 193: This needs to be `.toBeVisible()` I don't think it works without the `()` could we please update all

lib/src/components/stepper/Stepper.test.tsx
```diff
@@ -179,4 +179,119 @@ describe('Stepper', () => {
...
+        <Stepper.StepForward
+          label={(activeStep) => (activeStep === 2 ? 'Start' : 'Next')}
+        />
+      </Stepper>
+    )
+
+    expect(screen.getByTestId('vertical')).toBeVisible
+  })
+
+  it('renders in horizontal orientation', () => {
```
andoulla, line 210: This test seems to be identical to the one on line 183.  I can't find any differences

lib/src/components/stepper/Stepper.test.tsx
```diff
@@ -179,4 +179,119 @@ describe('Stepper', () => {
...
+        <Stepper.StepForward
+          label={(activeStep) => (activeStep === 2 ? 'Start' : 'Next')}
+        />
+      </Stepper>
+    )
+
+    expect(screen.getByTestId('horizontal')).toBeVisible
+  })
+
+  it('renders the correct number of bullets when a controlled component', () => {
```
andoulla, line 224:
```suggestion
  it('renders the correct number of bullets when used as a controlled component', () => {
```

Just a suggestion as it makes more sense for me?

lib/src/components/stepper/types.ts
```diff
@@ -1,27 +1,38 @@
...
 export type Context = {
-  steps: unknown[]
+  steps: Step[]
   goToPreviousStep: () => void
   goToNextStep: () => void
   goToStep: (index: number) => void
   activeStep: number
   viewedSteps: number[]
   allowSkip?: boolean
+  orientation?: 'vertical' | 'horizontal'
```
elliotlaws, line 16: Would it be worth creating a separate type named 'Orientation' that can be reused in these Types?

lib/src/components/stepper/Stepper.tsx
```diff
@@ -11,20 +11,36 @@ export const Stepper: React.FC<IStepperProps> & {
...
       <Box
         aria-label="progress"
+        aria-orientation={orientation}
+        data-testid={orientation}
         css={{
           width: '100%',
           display: 'grid',
-          gridTemplateColumns: '1fr auto 1fr'
+          gridTemplateColumns: '1fr auto 1fr',
+          ...css
```
andoulla, line 43: very good, we needed this

lib/src/components/stepper/StepperSteps.tsx
```diff
@@ -1,108 +1,215 @@
...
+  const getSeparatorState = (index: number) => {
+    const bulletStatus = steps[index]?.status
+
+    if (bulletStatus === 'completed') return 'success'
+
+    if (
+      bulletStatus === 'active' ||
+      bulletStatus === 'viewed' ||
+      index <= Math.max(...viewedSteps)
+    )
```
elliotlaws, line 148: Might be worth enclosing this return in curly braces as it's gone onto a new line

lib/src/components/stepper/Stepper.test.tsx
kriszzy01: Yes it was a duplication. Will be removed

lib/src/components/stepper/types.ts
kriszzy01: Yeah, makes sense to.

lib/src/components/stepper/Stepper.test.tsx
kriszzy01: Actually makes more sense with your suggestion 👍

lib/src/components/stepper/StepperSteps.tsx
```diff
@@ -1,108 +1,215 @@
...
-        ...css
-      }}
-      {...rest}
-    >
-      {steps.map((_, index) => {
+    <StepsContainer css={css} {...rest} orientation={orientation}>
+      {steps.map((step, index) => {
         return (
-          <StyledBullet
+          <StepContainer
```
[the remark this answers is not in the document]
kriszzy01: We actually StepsContainer within a StepContainer, not StepContainer within a StepContainer. I think renaming StepsContainer to StepperStepsContainer would create a less confusing component name?

lib/src/components/stepper/StepperSteps.tsx
```diff
@@ -1,108 +1,215 @@
...
 import { Flex } from '../flex'
+import { Icon } from '../icon'
+import { Text } from '../text'
+import { Box } from '../box'
+
 import { useStepper } from './stepper-context/StepperContext'
+
 import { IStepperStepsProps } from './types'
 
 const StyledBullet = styled(Flex, {
```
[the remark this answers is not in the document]
kriszzy01: Do you mean like having a separate component for `StepContainer`, `BulletContainer` etc?

lib/src/components/stepper/StepperSteps.tsx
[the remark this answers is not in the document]
kriszzy01: We're nesting `StepContainer` in `StepsContainer`. I think we can make the component names less similar to prevent this confusion.

lib/src/components/stepper/StepperSteps.tsx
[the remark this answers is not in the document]
thomasdigby: Yeah I'm with @andoulla here, let's reorganise and rename if it helps improve readability.

lib/src/components/stepper/StepperSteps.tsx
```diff
@@ -1,108 +1,232 @@
...
-  position: 'relative',
   p: '$2',
   justifyContent: 'center',
   alignItems: 'center',
-  width: '$2',
-  height: '$2',
+  width: '$3',
+  height: '$3',
```
thomasdigby, line 20: Just an FYI, we can use the `size` key if you want `height` and `width` to be the same. `size: '$3'`

lib/src/components/stepper/types.ts
```diff
@@ -1,27 +1,48 @@
...
+export enum StatusEnum {
+  Active = 'ACTIVE',
+  Viewed = 'VIEWED',
+  Success = 'SUCCESS',
+  Normal = 'NORMAL'
+}
+
+export type Status = 'active' | 'viewed' | 'success' | 'normal'
+export interface Step {
+  label: string
```
thomasdigby, line 12: Is `label` required? I assume it would be optional so you can render a controlled `Stepper` without the need for a label

lib/src/components/stepper/stepper-context/StepperContext.tsx
```diff
@@ -40,23 +48,26 @@ export const StepperProvider: React.FC<StepperProviderProps> = ({
   }
 
   const goToPreviousStep = () => {
+    if (steps[activeStep]?.status) return
     setActiveStep((current) => current - 1)
   }
 
   const goToStep = (index: number) => {
+    if (steps[index]?.status) return
```
thomasdigby, line 56: What are we testing for here? That any status is set on the current step? If we're looking to remove the ability to update the stepper state if the component is controlled then it might be worth just checking for `steps`, that should be a simpler way to determine if we should modify the internal state or not.

I'd even maybe consider only adding `goToX` functions to the `value` of the provider if `steps` isn't passed into `Stepper`. Then we can treat them as optional in child components and avoid the duplicated status check within each function

lib/src/components/stepper/types.ts
kriszzy01: It is for a controlled component. If the component is to be used as uncontrolled, we omit the `steps` prop and steps are calculated using the `stepCount` prop.

lib/src/components/stepper/stepper-context/StepperContext.tsx
thomasdigby: I've just realised that `steps` can be an array of empty strings. In that case it might be worth just having a single const of `isControlled` available throughout the logic instead of repeating it.

lib/src/components/stepper/StepperSteps.tsx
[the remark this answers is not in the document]
kriszzy01: Is there an approach you recommend for this? Would a components folder that houses them make sense?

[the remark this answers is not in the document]
andoulla: I think something similar to Carousel will work nicely https://github.com/Atom-Learning/components/tree/main/lib/src/components/carousel

lib/src/components/stepper/types.ts
```diff
@@ -1,37 +1,47 @@
 import { CSS } from '~/stitches'
 
+export enum StatusEnum {
+  Active = 'ACTIVE',
```
andoulla, line 4: This is a great step in the right direction. But may I offer an extra suggestion? in core we have our statuses like 
image

Also, if we all caps the first part of the assignment and lowercase the second part, it will mean that we can simplify the`getSeparatorState` function in stepperSteps as it will just check if it is viewed.

andoulla: Also do we need a separate type for status?

andoulla: yes, and this is a pattern you can find in other places too so it is very consistent

lib/src/components/stepper/types.ts
thomasdigby: Having the `Stepper` controlled or uncontrolled shouldn't change whether a label is rendered or not right? We should be able to render the `Stepper` as it is without labels, but still have it controlled.

lib/src/components/stepper/stepper-context/StepperContext.tsx
```diff
@@ -9,28 +9,35 @@ const StepperContext = React.createContext<Context>({
...
   React.useEffect(() => {
     setviewedSteps((prev) =>
       prev.includes(activeStep) ? prev : [...prev, activeStep]
     )
+
     onStepChange?.(activeStep)
   }, [activeStep, onStepChange])
 
   const goToNextStep = () => {
+    if (isControlled) return
```
thomasdigby, line 40: Do we need this check here if we're returning `undefined` in the provider value below?

lib/src/components/stepper/stepper-context/StepperContext.tsx
```diff
@@ -9,28 +9,35 @@ const StepperContext = React.createContext<Context>({
...
-  onStepChange
+  onStepChange,
+  orientation,
+  steps
 }) => {
   const [activeStep, setActiveStep] = React.useState(0)
 
   const [viewedSteps, setviewedSteps] = React.useState<number[]>([0])
 
+  const isControlled = !stepCount
```
thomasdigby, line 29: To make it easier for users to understand how `stepCount` and `steps` works, could we add an `invariant` here to throw an error if both are used? https://github.com/Atom-Learning/components/blob/71bbcdde366d39930e3e60c7a5723cb10bd3cfe8/lib/src/components/form/Form.tsx#L90-L93

lib/src/components/stepper/Stepper.mdx
```diff
@@ -45,13 +45,49 @@ Below is a more complex example that shows how to dynamically render the button
...
+    }}
+  >
+    Next
+  </Stepper.StepForward>
 </Stepper>
-```
\ No newline at end of file
+```
+
+The root `Stepper` component also allows the user to pass a `step` prop in order to render step labels. When the `step` prop is passed, the `Stepper` component will be used as a controlled component thereby taking control of its state. When used as a controlled component, navigation across steps, and identifying the state of a step would be controlled by the array of steps passed to the component.
```
thomasdigby, line 64:
```suggestion
The root `Stepper` component also allows the user to pass a `steps` prop in order to render step labels. When the `steps` prop is passed, the `Stepper` component will be used as a controlled component thereby taking control of its state. When used as a controlled component, navigation across steps, and identifying the state of a step would be controlled by the array of steps passed to the component.
```

lib/src/components/stepper/Stepper.test.tsx
```diff
@@ -179,4 +179,104 @@ describe('Stepper', () => {
...
+      <Stepper allowSkip {...props} orientation="horizontal">
+        <Stepper.StepBack label={() => 'Back'} />
+        <Stepper.Steps />
+        <Stepper.StepForward
+          label={(activeStep) => (activeStep === 2 ? 'Start' : 'Next')}
+        />
+      </Stepper>
+    )
+
+    expect(screen.getByTestId('horizontal')).toBeVisible()
```
thomasdigby, line 208: Can we test for `aria-orientation` instead of a custom `test-id`? That way we can remove the `data-test-id` from the output HTML

thomasdigby:
Seeing a few issues with how the stepper renders with various props. Without an `orientation` prop it's missing some styles/logic to render correctly, I assume we need to set a default here. Also, I know we spoke before about not tackling the horizontal labels yet, but it seems odd that we're able to pass them into the `steps` array and then it doesn't render? Could we at least render them without the line inbetween? Or just some simple logic so the text wraps properly, I think that would help in the short-term.

```
<Stepper
  steps={[
    { label: 'Step 1', status: 'success' },
    { label: 'Step 2', status: 'active' },
    { label: 'Step 2', status: 'normal' }
  ]}
>
  <Stepper.Steps />
</Stepper>
```
Screenshot 2022-07-12 at 11 00 51

```
<Stepper
  orientation="horizontal"
  steps={[
    { label: 'Step 1', status: 'success' },
    { label: 'Step 2', status: 'active' },
    { label: 'Step 2', status: 'normal' }
  ]}
>
  <Stepper.Steps />
</Stepper>
```
Screenshot 2022-07-12 at 11 00 35

```
<Stepper
  orientation="vertical"
  steps={[
    { label: 'Step 1', status: 'success' },
    { label: 'Step 2', status: 'active' },
    { label: 'Step 2', status: 'normal' }
  ]}
>
  <Stepper.Steps />
</Stepper>
```
Screenshot 2022-07-12 at 11 00 17

lib/src/components/stepper/Stepper.test.tsx
kriszzy01: this was my approach yesterday but the test failed.

`expect(container).toHaveAttribute('aria-orientation', 'horizontal')`

lib/src/components/stepper/stepper-context/StepperContext.tsx
kriszzy01: The `stepCount` value used within this function will flag a typescript error. `stepCount` is optional and we're using it for conditional logic within the function

One way to get rid of this is to pass the `goToX` functions to the provider without doing the check, the do something like

`const goToX = isControlled
    ? undefined
    : () => {
        //goToX logic
      }  
      
    Provider
      value={{
        goToX, //no checked for controlled component
      }}
    >  
  `
      
  I'm not sure what your thoughts are on both approaches

kriszzy01:
@thomasdigby Would it make sense for the `Stepper.Steps` component to have default width / height? Due to the current implementation, we need to pass width / height in the css prop for the separators to be properly calculated. Since this is not required, omitting this will cause the component to render correctly. I think it may be worth having default width / height which can be overridden.

lib/src/components/stepper/Stepper.test.tsx
kriszzy01: Just found out that the we shouldn't be using aria-orientation attribute here

lib/sandbox/index.tsx
```diff
@@ -15,7 +15,50 @@ const App = () => (
...
+    </Stepper>
+    <Stepper
+      css={{ my: '$5', display: 'grid', gridTemplateColumns: '1fr auto 1fr' }}
+      allowSkip
+      stepCount={3}
+    >
+      <Stepper.StepBack>Back</Stepper.StepBack>
+      <Stepper.Steps />
+      <Stepper.StepForward>Next</Stepper.StepForward>
+    </Stepper>
```
thomasdigby, line 61: We shouldn't commit any changes to the sandbox, can you revert this back to its original state?

lib/src/components/stepper/types.ts
thomasdigby: I still think `label` shouldn't be required here, what if we wanted a controlled `Stepper` without labels?

lib/src/components/stepper/types.ts
```diff
@@ -1,27 +1,48 @@
 import { CSS } from '~/stitches'
 
+export enum StatusEnum {
+  ACTIVE = 'active',
+  VIEWED = 'viewed',
+  SUCCESS = 'success',
+  NORMAL = 'normal'
+}
+
+export type Status = 'active' | 'viewed' | 'success' | 'normal'
```
thomasdigby, line 10: I don't think we need `Status` and `StatusEnum` here, we should use `StatusEnum` where we're currently using `Status` and remove the "Enum" postfix

lib/src/components/stepper/StepperSteps.tsx
```diff
@@ -53,56 +27,81 @@ export const StepperSteps: React.FC<IStepperStepsProps> = ({
...
+            >
+              {step.status === 'success' ? (
+                <Icon is={Ok} data-testid="success-icon" />
+              ) : (
+                index + 1
+              )}
+            </StepperStepBullet>
+
+            {step.label && (
+              <StepperStepLabel orientation={orientation} state={bulletState}>
```
thomasdigby, line 98: Can we update the label to be a `span`? I wonder also if a label is rendered we could use `aria-labelledby` on the bullet so that the button is labelled as the actual label text? e.g.

```tsx
<StepperStepBullet
  aria-label={!step.label && `Step ${index + 1}`}
  aria-labelledby={step.label && 'step-${index}'}
/>
<StepperStepLabel id={`step-${index}`}>
```

lib/src/components/stepper/StepperSteps.tsx
```diff
@@ -53,56 +27,81 @@ export const StepperSteps: React.FC<IStepperStepsProps> = ({
...
+              state={bulletState}
+              aria-current={index === activeStep ? 'step' : undefined}
+              aria-label={`step ${index + 1}`}
+              css={{
+                cursor:
+                  allowSkip && viewedSteps.includes(index) ? 'pointer' : 'auto'
+              }}
+            >
+              {step.status === 'success' ? (
+                <Icon is={Ok} data-testid="success-icon" />
```
thomasdigby, line 91: Can we remove this `data-testid` and replace the test to check if an SVG is rendered instead? Ideally, we don't want to export any `data-testid`s from our library code.

lib/src/components/stepper/StepperSteps.tsx
```diff
@@ -53,56 +27,81 @@ export const StepperSteps: React.FC<IStepperStepsProps> = ({
...
-            {index + 1}
-          </StyledBullet>
+            <StepperStepBullet
+              as={allowSkip ? 'button' : 'div'}
+              onClick={() =>
+                allowSkip && viewedSteps.includes(index)
+                  ? goToStep?.(index)
+                  : undefined
+              }
+              state={bulletState}
```
thomasdigby, line 82: Could we update this prop/variant name to `status`? That will better reflect the actual data we're passing through here and won't overlap with "react state"

lib/src/components/stepper/StepperSteps.tsx
```diff
@@ -53,56 +27,81 @@ export const StepperSteps: React.FC<IStepperStepsProps> = ({
...
-        justifyContent: 'space-between',
-        ...css
-      }}
-      {...rest}
-    >
-      {steps.map((_, index) => {
+    <StepperStepsContainer css={css} {...rest} orientation={orientation}>
+      {steps.map((step, index) => {
+        const bulletState = getBulletState(index)
+        const seperatorState = getSeparatorState(index)
```
thomasdigby, line 62: Same with the naming convention for these, e.g.  `getSeparatorStatus()`

lib/src/components/stepper/Stepper.tsx
```diff
@@ -11,20 +12,38 @@ export const Stepper: React.FC<IStepperProps> & {
...
       allowSkip={allowSkip}
       onComplete={onComplete}
-      onStepChange={onStepChange}
+      onStepChange={stepCount ? onStepChange : undefined}
+      orientation={orientation}
+      steps={steps || Array(count).fill('')}
     >
       <Box
         aria-label="progress"
+        data-testid={orientation}
```
thomasdigby, line 43: Same issue with the `data-testid` here

lib/src/components/stepper/Stepper.test.tsx
```diff
@@ -179,4 +181,129 @@ describe('Stepper', () => {
...
+        ]}
+      >
+        <Stepper.StepBack label={() => 'Back'} />
+        <Stepper.Steps />
+        <Stepper.StepForward
+          label={(activeStep) => (activeStep === 2 ? 'Start' : 'Next')}
+        />
+      </Stepper>
+    )
+    expect(container).toMatchSnapshot()
```
thomasdigby, line 228: Could we replace this snapshot test with a test asserting that the number of bullets matches the length of the array?

lib/src/components/stepper/Stepper.test.tsx
thomasdigby: What do you mean? A testing-library recommendation to not use it?

lib/src/components/stepper/Stepper.mdx
```diff
@@ -45,13 +45,49 @@ Below is a more complex example that shows how to dynamically render the button
...
+</Stepper>
+```
+
+The root `Stepper` component also allows the user to pass a `orientation` prop in order to the alternate the component direction .
+
+Below is an example that shows how to render steps in vertical direction.
+
+```tsx preview
+<Stepper stepCount={2} orientation="vertical">
+  <Stepper.Steps css={{ width: '200px' }} />
```
thomasdigby, line 91: This `width` shouldn't be necessary in vertical orientation

lib/src/components/stepper/Stepper.mdx
```diff
@@ -45,13 +45,49 @@ Below is a more complex example that shows how to dynamically render the button
 ```
 
 Below is how to override the forward default behavior if some logic or validation need to be done before moving to the next step
+
 ```tsx preview
 <Stepper stepCount={3}>
```
thomasdigby, line 50: We need to add the grid CSS to these components in the docs so the render correctly

lib/src/components/stepper/Stepper.mdx
```diff
@@ -45,13 +45,49 @@ Below is a more complex example that shows how to dynamically render the button
...
+      label: 'Step 1',
+      status: 'viewed'
+    },
+    {
+      label: 'Step 2',
+      status: 'active'
+    }
+  ]}
+>
+  <Stepper.Steps css={{ width: '200px' }} />
```
thomasdigby, line 81: Same here, we can remove the width, I would probably add an extra step or two here to show off the success state as well, maybe:

```tsx
<Stepper
  steps={[
    { label: 'Step 1', status: 'success' },
    { label: 'Step 2', status: 'active' },
    { label: 'Step 3', status: 'normal' }
  ]}
>
  <Stepper.Steps />
</Stepper>
```

Screenshot 2022-07-13 at 11 24 18

lib/src/components/stepper/StepperSteps.tsx
kriszzy01: Could we pass a `title` prop of "check" to the icon and check for that? Would this be a better approach?

thomasdigby: I wouldn't want to add a title just for querying the element in a test, it could change how the UI is interpreted by screenreaders (need to test that really). Any other suggestions? I just think we should avoid adding something purely for the test

lib/src/components/stepper/Stepper.mdx
```diff
@@ -35,23 +38,71 @@ Below is a more complex example that shows how to dynamically render the button
...
+      label: 'Step 1',
+      status: 'viewed'
+    },
+    {
+      label: 'Step 2',
+      status: 'active'
+    },
+    {
+      label: 'Step 3',
+      status: 'success'
```
thomasdigby, line 88: Isn't this a bit weird rendering success after the `active` step? 😅 I always thought it was "success" || "viewed" - "active" - "normal"

kriszzy01: Just wanted to show off all the states we have😅. Will update

lib/src/components/stepper/types.ts
kriszzy01: Good point!

thomasdigby (approved):
Nice! 🙌

lib/src/components/stepper/types.ts
```diff
@@ -1,27 +1,46 @@
...
-  goToNextStep: () => void
-  goToStep: (index: number) => void
+  steps: Step[]
+  goToPreviousStep?: () => void
+  goToNextStep?: () => void
+  goToStep?: (index: number) => void
   activeStep: number
   viewedSteps: number[]
   allowSkip?: boolean
+  orientation?: Orientation
```
hollg, line 24: This is a minor nitpick, but could we name this `direction`, to keep it consistent with the equivalent prop on our `Stack` component?

lib/src/components/stepper/Stepper.tsx
```diff
@@ -11,20 +12,37 @@ export const Stepper: React.FC<IStepperProps> & {
...
+  children,
+  stepCount,
+  allowSkip,
+  onComplete,
+  onStepChange,
+  orientation = 'horizontal',
+  steps,
+  css
+}) => {
+  invariant(
```
hollg, line 25: Awesome, super clear error message 👍🏻

lib/src/components/stepper/Stepper.mdx
```diff
@@ -35,23 +38,67 @@ Below is a more complex example that shows how to dynamically render the button
   onStepChange={(activeStep) =>
     console.log('Do something with the active step')
   }
+  css={{ display: 'grid', gridTemplateColumns: '1fr auto 1fr' }}
```
LimeWub, line 41: Out of curiosity - how important is this CSS here? Do we need them for this example to look right?

hollg (approved):

PR #366, opened 2022-07-14
fix: stepper documentation error

kriszzy01:


thomasdigby (approved):
Just FYI, in future using `chore` or `docs` instead of `fix` in your commit will prevent a new release for this change 👍

hollg (approved):

PR #367, opened 2022-07-15
fix: SearchInput throw onChange actual event on clear

LimeWub:
Continues from https://github.com/Atom-Learning/components/pull/358

@kerryambt 's solution wasn't actually throwing onChange events as expected (rather it was calling the `onChange` function). 
This approach throws the clear event as a new onChange event with empty value which means it can be caught appropriately on the components itself and any parent components (forms).

https://user-images.githubusercontent.com/6905473/179212318-c25921d5-727c-422e-a957-9e4a62f87142.mov


lib/src/components/search-input/SearchInput.tsx
```diff
@@ -63,12 +65,20 @@ export const SearchInput: React.FC<SearchInputProps> = ({
...
-    setActiveIcon(INPUT_ICON.SEARCH)
-    onChange?.({
-      currentTarget: { value: '' },
-      target: { value: '' }
-    } as React.ChangeEvent<HTMLInputElement>)
+    const inputEl = inputElRef.current
+    if (!inputEl) return
+
+    // https://stackoverflow.com/a/46012210
+    const nativeInputValueSetter = Object.getOwnPropertyDescriptor(
```
kerryambt, line 72: I initially went down this route with my fix and it was suggested I could simplify to what I did. have you looked at my initial commit to see how it worked with getting around typescript errors? https://github.com/Atom-Learning/components/pull/358/commits/4967839131eeab46882df10f58345264ffd9034a :)

LimeWub: Oh nice! Sorry I was spammed in gmail and missed this!

I'll check out the commit u linked and make the changes! I think it's a better solution than the final one cause well - it works better lol ^^

LimeWub: Thanks for the pointer! I actually fixed it with some `?.`s but it was useful to understand why it was complaining :)

kerryambt (approved):

hollg (approved):

PR #368, opened 2022-07-18
Feat: Hide Label Support in Stepper

kriszzy01:


lib/src/components/stepper/types.ts
```diff
@@ -31,6 +32,7 @@ export type StepperProviderProps = {
   onStepChange?: (activeStep: number) => void
   direction?: Direction
   steps: Step[]
+  hideLabels: boolean
 }
 
 export interface IStepperProps {
```
kerryambt, line 38: I know it wasn't your change that made this, but it seems really redundant to me to have an interface and a type that are basically identical. where we would use the interface can we instead go `<StepperProviderProps & { css?: CSS }>` or something to that effect?

kerryambt: oh yeah :o maybe we can reuse something? idk. out of scope for this though :)

elliotlaws (approved):

PR #370, opened 2022-07-25
Upgrade to Radix v1.0.0

hollg:
Radix has finally released version `1.0.0` and aligned with semver! https://www.radix-ui.com/docs/primitives/overview/releases#july-21-2022

This PR upgrades all `@radix-ui` packages to version `^1.0.0`. Most changes from the libraries don't really affect us because we either weren't using the relevant props anyway or we weren't even exposing the relevant components. Some new components have `.Portal` subcomponents though, which I have exposed here because I know they've been useful before for components that already had them, e.g. the `Dialog`.

We've previously had some bugs caused by breaking changes sneaking into what looked like patch/minor releases in these libraries so this should be a stability improvement. The packages all share internal util dependencies so anything that makes it easy to keep all the packages up to date also avoids the need for multiple versions of those shared dependencies 👍🏻


site/package.json
```diff
@@ -26,7 +26,6 @@
     "@atom-learning/icons": "1.9.0",
     "@atom-learning/theme": "1.0.1",
     "@next/mdx": "^10.0.6",
-    "@radix-ui/react-id": "0.1.5",
```
hollg, line 29: The up-to-date primitives don't need this any more 👋🏻

thomasdigby: I assumed this was because they moved to the React 18 `useId`, just want to check that these Radix updates are compatible with React 16/17?

thomasdigby: @hollg Can you double check that this still isn't required? Just want to be sure before we run it across our core test-suite and see ID mismatch errors

lib/package.json
```diff
@@ -126,23 +126,23 @@
...
   },
   "dependencies": {
-    "@atom-learning/theme": "1.1.0",
-    "@radix-ui/react-accordion": "0.1.6",
-    "@radix-ui/react-alert-dialog": "0.1.7",
-    "@radix-ui/react-checkbox": "0.1.5",
-    "@radix-ui/react-dialog": "0.1.7",
-    "@radix-ui/react-dropdown-menu": "0.1.6",
+"@atom-learning/theme": "1.1.0",    
+"@radix-ui/react-accordion": "^1.0.0",
```
thomasdigby, line 130: The indentation on this is weird, did you run a `yarn install` after updating it? I assume it'll pop back into place when you do so

site/package.json
hollg: The package is deprecated, and the deprecation notice links to this page about how to handle SSR: https://www.radix-ui.com/docs/primitives/overview/server-side-rendering

There isn't anything specifically about FE unit tests mentioned. Do you have anything in mind for what would constitute a reasonable check about this? I've rerun all the tests here several times to make sure it's consistent and it is, but am happy to do further testing if there's something you think would be helpful.

lib/src/components/action-icon/ActionIcon.mdx
```diff
@@ -57,14 +57,16 @@ Passing `disabled` makes the Action Icon take a disabled appearance, where any c
...
 ```tsx preview
-<Tooltip>
-  <Tooltip.Trigger>
-    <ActionIcon size="lg" label="Add to basket">
-      <Icon is={Basket} />
-    </ActionIcon>
-  </Tooltip.Trigger>
-  <Tooltip.Content>Add to basket</Tooltip.Content>
-</Tooltip>
+<ToolTip.Provider>
```
thomasdigby, line 60: As I assume we'll be using the provider at the root level of the app, and as `Tooltip.Provider` is not technically required for the `Tooltip` to work, do we want/need to include it in the docs here?

hollg: Isn't it? A bunch of tests started failing without it 🤔

thomasdigby: Maybe I'm not up to date with how the provider works for this... Are we expecting to use a single instance of `Tooltip.Provider` at the root of the app? Or are we going to wrap each tooltip in components/core directly?

hollg: Sorry, yes, single instance at the app root. But as far as I can tell it is required. It's only here so that this example can render correctly. Maybe the simplest way forward is just to add a comment explaining that the provider would normally be in the root.

lib/src/components/data-table/DataTable.test.tsx
```diff
@@ -44,13 +45,21 @@ const data = [
...
+)
+
 describe('DataTable component', () => {
   it('renders', () => {
     const { container } = render(
       <DataTable columns={columns} data={data}>
         <DataTable.GlobalFilter label="User search" css={{ mb: '$4' }} />
         <DataTable.Table sortable css={{ mb: '$4' }} />
-        <DataTable.Pagination pageSize={5} />
+        <PaginationWithToolTipProvider pageSize={5} />
```
thomasdigby, line 62: Minor comment, but as we want to use a single `Tooltip.Provider` for the app, maybe it makes more sense here to wrap the entire table in the provider, rather than just the pagination. Hopefully it's a clear indicator that it's a "global" requirement, as opposed to a specific change to pagination e.g.

```tsx
const Wrapper: React.FC = () => (
  <Tooltip.Provider>{children}</Tooltip.Provider>
)
```

lib/src/components/dropdown-menu/__snapshots__/DropdownMenu.test.tsx.snap
```diff
@@ -103,64 +103,6 @@ exports[`DropdownMenu component renders 1`] = `
     style="outline: none; opacity: 0; position: fixed; pointer-events: none;"
     tabindex="0"
   />
-  <div
```
thomasdigby, line 106: Do you think we'll run into any issues with the update to remove portalling by default in the app?

site/package.json
thomasdigby: It seems like we're still using it within the tests though as `IdProvider` unless I've misunderstood what this package is used for 🤔

lib/src/components/tooltip/Tooltip.tsx
```diff
@@ -9,6 +9,7 @@ type TooltipProps = React.ComponentProps<typeof Root>
 
 export const Tooltip: React.FC<TooltipProps> & {
   Content: typeof TooltipContent
+  Portal: typeof Portal
```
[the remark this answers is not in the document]
hollg: What's the relevant difference between `Dialog` and `Tooltip` in this context? I had just assumed they'd be the same in this regard but am happy to be corrected!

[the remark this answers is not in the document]
thomasdigby: Nothing in terms of implementation from Radix's side, it's just whether it's relevant to expose it to our users. I may be misunderstanding its use, but I could see an argument for potential Dialog positioning or layering problems that could be solved with `Portal` that I can't for Tooltip. The content is always, necessarily so, going to render next to the trigger, that's not the case with Dialog.

site/package.json
hollg: Ahh sorry, you're right that it's still in those tests. I have some changes to that stuff in a separate branch from last time I was working on this. WIll merge them in now. Totally my fault!

hollg: I've pulled those other changes over now, so there's no `IdProvider` in any test and snapshots are stable.

lib/src/components/action-icon/ActionIcon.mdx
LimeWub: On a slightly off topic note - this example is wrong - as in: it generates bad/non-accessible code of two nested buttons:

Screenshot 2022-10-03 at 12 09 33

The correct way of adding tooltips to Button like components is by using 'asChild' for active buttons so there's no button nesting and doing [nesting with a span in the middle instead for inactive buttons ](https://www.radix-ui.com/docs/primitives/components/tooltip#displaying-a-tooltip-from-a-disabled-button).

As for having a Tooltip.Provider in the demo: I don't hate but if we want to we can probably add the `Tooltip.Provider` in the App file in `next` instead to avoid. (I think it's probably cleaner here as `Tooltip` has Provider in its example). Although there should be text if there isn't already in Tooltip to say it's one per App (cause I didn't know and was using multiple Providers?)

lib/src/components/tooltip/Tooltip.tsx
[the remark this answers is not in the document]
LimeWub: Not sure I'm 100% understanding this convo but rendering in a portal is also useful if we want an item that's nested inside an `overflow: hidden` container to not be chopped off by said overflow hidden. So if the parent of the trigger had an overflow which would chop off the tooltip we could use Portal to solve that ^^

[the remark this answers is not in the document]
thomasdigby: I like it, that sounds good enough to me. Let's keep it as you have it here 👍

lib/src/components/action-icon/ActionIcon.mdx
hollg: @LimeWub Thanks for this! Did this come up through the issue you and Kat were looking at for our app's dashboard?

hollg: Have updated this example. The `Tooltip` docs did already mention this but I've also changed the wording slightly to hopefully make it clearer.

lib/src/components/data-table/DataTable.test.tsx
hollg: My original thought here was to couple it specifically with `DataTable.Pagination` to make it clear that it's only relevant because of that component. I like your point about communicating that the provider is a global req though. Have updated the wrapper and included a comment to communicate the link with `DataTable.Pagination`.

thomasdigby (approved):

lib/src/components/dropdown-menu/__snapshots__/DropdownMenu.test.tsx.snap
hollg: Eurghhh, looking into this, it does seem like we might open the door to some `z-index` shenanigans. I'll look into adding a `z-index` to the dropdown vs including the portal in this component to avoid a regression

lib/src/components/action-icon/ActionIcon.mdx
LimeWub: @hollg Yes that's why I noticed the nested buttons and how I found out about the special nesting needed for inactive buttons :)

PR #380, opened 2022-08-05
feat: calendar year picker

kerryambt:
### Description

As we have added the date input for DOB on the new signup flow in core, we wanted to add a year selector to the calendar. You can now click on the year when viewing the calendar, and you'll be shown a grid of years you can click on and scroll through with the arrows. Clicking a year sets the year for the current set date but keeps the day and month, and takes you back to the calendar screen.

This year grid is constrained within the minDate and maxDate values (if set), so you can't use it to set a date you otherwise wouldn't be able to set.

### Screenshots

image
image
image
image
image
image

### DIFF
--- a/lib/src/components/calendar/Calendar.tsx
+++ b/lib/src/components/calendar/Calendar.tsx
@@ -160,6 +160,9 @@ export const Calendar: React.FC<CalendarProps> = ({
         >
           {yearList.map((year, i) => {
             const isCurrentYear = year === date.getFullYear()
+            if (!year)
+              return <Box key={i} css={{ width: '$6', height: '$4' }} />
+
             return (
               <StyledButton
                 key={`${year}${i}`}
@@ -167,9 +170,8 @@ export const Calendar: React.FC<CalendarProps> = ({
                 isRounded
                 onClick={() => handleSetYear(year)}
                 selected={isCurrentYear}
-                disabled={year === 0}
               >
-                {year !== 0 && year}
+                {year}
               </StyledButton>
             )
           })}
--- a/lib/src/components/calendar/Calendar.tsx
+++ b/lib/src/components/calendar/Calendar.tsx
@@ -6,13 +6,13 @@ import * as React from 'react'
 import { ActionIcon } from '~/components/action-icon'
 import { Box } from '~/components/box'
 import { Flex } from '~/components/flex'
-import { Heading } from '~/components/heading'
 import { Icon } from '~/components/icon'
 import { Text } from '~/components/text'
+import { Button } from '~/components/button'
 import type { CSS } from '~/stitches'
 import { styled } from '~/stitches'
 
-import { monthNamesShort, weekdayNamesShort } from './constants'
+import { monthNamesShort, weekdayNamesShort, DEFAULT_LABELS } from './constants'
 import { Day } from './Day'
 
 const Grid = styled('div', {
@@ -21,10 +21,35 @@ const Grid = styled('div', {
   gridGap: '$1 $2'
 })
 
+const StyledButton = styled(Button, {
+  color: '$tonal600',
+  p: '$3',
+  width: '$6',
+  variants: {
+    selected: {
+      false: {
+        color: '$tonal600 !important',
+        fontWeight: '400',
+        '&:hover': {
+          bg: '$tonal100 !important',
+          color: '$tonal600 !important'
+        },
+        '&[disabled]': { bg: 'white !important' }
+      }
+    }
+  }
+})
+
 export type CalendarTranslationProps = {
   monthNames?: string[]
   weekdayNames?: string[]
-  labels?: { open: string; next: string; previous: string }
+  labels?: {
+    open: string
+    next: string
+    previous: string
+    nextYear: string
+    previousYear: string
+  }
 }
 
 type CalendarProps = DayzedInterface &
@@ -32,6 +57,7 @@ type CalendarProps = DayzedInterface &
     css?: CSS
     refDateToday?: React.RefObject<HTMLButtonElement>
     refDateSelected?: React.RefObject<HTMLButtonElement>
+    setYear: (date: Date) => Promise<void>
   }
 
 const offsetWeekdayNames = (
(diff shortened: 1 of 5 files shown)


lib/src/components/calendar/Calendar.tsx
```diff
@@ -21,17 +21,43 @@ const Grid = styled('div', {
...
 })
 
+const StyledButton = styled(Button, {
+  color: '$tonal600',
+  p: '$3',
+  width: '$6',
+  variants: {
+    selected: {
+      false: {
+        color: '$tonal600 !important',
```
valllentinnaa, line 31: If the styles are conditionally rendered, do we need the overriding with` important`

lib/src/components/calendar/Calendar.tsx
```diff
@@ -67,70 +125,112 @@ export const Calendar: React.FC<CalendarProps> = ({
...
-        </Box>
-      ))}
+              )}
+            </Grid>
+            <Grid>
+              {weeks.map((week, weekIndex) =>
+                week.map((dateObj, index) => {
+                  const key = `${month}${year}${weekIndex}${index}`
+
+                  if (!dateObj) return <div key={key} />
```
valllentinnaa, line 209: If we want to check for falsy value shouldn't we use
`Object.entries(dateObj).length` instead?

lib/src/components/calendar/Calendar.tsx
```diff
@@ -67,70 +125,112 @@ export const Calendar: React.FC<CalendarProps> = ({
...
-                    {...getDateProps({ dateObj })}
-                    type="button"
+                {year !== 0 && year}
+              </StyledButton>
+            )
+          })}
+        </Grid>
+      )}
+      {!showYears &&
+        calendars.map(({ month, year, weeks }) => (
```
valllentinnaa, line 179: It's a bit hard to determine what these variables hold when there are no types. However, I know we are in a hurry, just decided to share my thoughts on it.

lib/src/components/calendar/Calendar.tsx
```diff
@@ -67,70 +125,112 @@ export const Calendar: React.FC<CalendarProps> = ({
...
-                    isOutsideMonth={prevMonth || nextMonth}
-                    isSelected={selected}
-                    isToday={today}
-                    key={key}
-                    ref={
-                      selected ? refDateSelected : today ? refDateToday : null
-                    }
-                    {...getDateProps({ dateObj })}
-                    type="button"
+                {year !== 0 && year}
```
valllentinnaa, line 172: if year is a  type Number, we shouldn't need to check `year === 0 ` or `year !== 0` because 0 will evaluate to false anyways

lib/src/components/calendar/Calendar.tsx
hollg, line 31: What conflict are these `!important`s solving?

kerryambt: the styles from button themes seem to take precedence over these styled variants here, so important is needed. not sure if that's intended behaviour from stitches but that's how it works atm

kerryambt: this is overriding the styling from themes applied to the buttons, for some reason that's taking precedence over styled variants.

lib/src/components/calendar/Calendar.tsx
kerryambt: I guess we could write `!!year && year` but that just reads really weird. if you just go `year` it'll output 0 as text which you don't want. I think this is the most readable.

lib/src/components/calendar/Calendar.tsx
kerryambt: this isn't changed code, but if I hover on calendars it tells me it's an array of Calendar types, so there is typing it's just not explicit per destructured item. I think this is fine tbh.

lib/src/components/calendar/Calendar.tsx
kerryambt: I think it can also be a string, which is why this is like this

lib/src/components/calendar/Calendar.tsx
hollg: Ah, that makes sense! I didn't catch that you were targetting `Button`, rather than `button`. Thanks for explaining!

lib/src/components/calendar/Calendar.tsx
hollg, line 172: If `year` is falsey, do we still want to display the button? I.e. is this condition definitely in the right place, or could it be moved before the `return` statement? (If that does work, it could solve the issue of how to write the condition too, since you could just do `if (year) return ...`)

kerryambt: the reasoning behind this was to preserve placement in the grid - I guess we could could go like `{year ? <Button>{year}</Button> : <emptydiv>}` but I don't think that's necessarily better? thoughts? :)

elliotlaws:
When I tab through the years it selects years from right to left, instead of left to right. Is that the intended behaviour?

kerryambt:
> When I tab through the years it selects years from right to left, instead of left to right. Is that the intended behaviour?

@elliotlaws this is probably an unintended side effect of using rtl grid placing to place things correctly in the grid. not sure how big of an issue it is?

lib/src/components/calendar/Calendar.tsx
valllentinnaa: got it

lib/src/components/calendar/Calendar.tsx
valllentinnaa: maybe Boolean(year) if we want it to output the boolean value

kerryambt: no because `Boolean(2020)` is  `true` when we want `2020`

elliotlaws (approved):

lib/src/components/calendar/Calendar.tsx
hollg: 🤔 I think an empty div actually seems better than a button that doesn't do anything _semantically_ speaking

kerryambt: @hollg yeah that's def true semantically. I've added this and it's a lot cleaner now with the early return for a div :)

hollg (approved):

PR #381, opened 2022-08-09
feat: chip component

kerryambt:
### Description

https://atomlearningltd.atlassian.net/jira/software/projects/DS/boards/27?selectedIssue=DS-54

Tackling some of the DS backlog as part of my guild day, I've created the Chip component.

Chip has two modes, filter - which can be selected or unselected, and input.

### Implementation

- Created component
- Added tests
- Added documentation

### Screenshots

image


lib/src/components/chip/Chip.tsx
```diff
@@ -0,0 +1,167 @@
...
+    >
+      {((icon && !isFilterMode) || internalSelected) && (
+        <Icon
+          aria-label={isFilterMode ? 'SelectedIcon' : 'ChipIcon'}
+          is={isFilterMode ? Ok : icon}
+          size={size === 'lg' ? 'md' : 'sm'}
+          css={{ color: '$primary', mr: '$1' }}
+        />
+      )}
+      <Text size={size === 'lg' ? 'md' : 'sm'}>{label}</Text>
```
hollg, line 149: Is a `p` tag semantically sound here? To me it feels almost equivalent to putting a `p` inside a `button`. I wonder whether there is some ARIA role that we can give the container to communicate the significance of the text within

lib/src/components/chip/Chip.test.tsx
```diff
@@ -0,0 +1,81 @@
+import { fireEvent, render, screen } from '@testing-library/react'
+import { Hamburger } from '@atom-learning/icons'
+import { axe } from 'jest-axe'
+import * as React from 'react'
+
+import { Chip } from './Chip'
+
+describe('Badge component', () => {
```
hollg, line 8:
```suggestion
describe('Chip component', () => {
```

lib/src/components/chip/Chip.tsx
```diff
@@ -0,0 +1,167 @@
...
+      size: 'sm',
+      css: { pr: '$2' }
+    }
+  ]
+})
+
+interface ChipProps {
+  label: string
+  selected?: boolean
+  icon?: SVGSVGElement
```
hollg, line 103: Did you consider an API for this component where the user would add the text/icon as `children`, rather than passing them as props? I am always keen to use that pattern where practical because it makes the component more flexible in case at some point down the line we want a `Chip` with an icon in a different position, text with one word in bold, etc.

lib/src/components/chip/Chip.tsx
```diff
@@ -0,0 +1,167 @@
...
+      disabled={disabled}
+      {...(isFilterMode && {
+        onClick: () => handleSelectedChange(!internalSelected)
+      })}
+      tabIndex={0}
+      {...remainingProps}
+    >
+      {((icon && !isFilterMode) || internalSelected) && (
+        <Icon
+          aria-label={isFilterMode ? 'SelectedIcon' : 'ChipIcon'}
```
hollg, line 143: I might be misunderstanding something here(!) but I don't think `'SelectedIcon'` or `'ChipIcon'` will mean much to the end user.

lib/src/components/chip/Chip.tsx
```diff
@@ -0,0 +1,167 @@
...
+  return (
+    <ChipContainer
+      size={size}
+      mode={mode}
+      selected={isFilterMode && internalSelected}
+      disabled={disabled}
+      {...(isFilterMode && {
+        onClick: () => handleSelectedChange(!internalSelected)
+      })}
+      tabIndex={0}
```
hollg, line 138: Could you ELI5 what `tabIndex` is doing here?

lib/src/components/chip/Chip.tsx
kerryambt: yeah maybe a span would be better :o not sure on the role though

lib/src/components/chip/Chip.test.tsx
kerryambt: oops ._. this is why we need snippets!

lib/src/components/chip/Chip.tsx
kerryambt: wasn't sure how else to check these specific icons in testing, might have to resort to data-test-id instead

LimeWub:
For visibility (I've already DM'ed): I have been working on an alternative approach for this which I think is a bit simpler depending on the use case, let's catch up.

LimeWub:
Link to refactor branch for this (WIP): https://github.com/Atom-Learning/components/tree/feat/chip-component-refactor

(Splitting into 2 components, 1 ToggleGroup, 1 Dismissible) - I'll chat to Miguel regarding naming.
<Icon /> to be nested (and cloneElement to apply sizing)

I'll try to pick this up and wrap up within this/next week. ^^

LimeWub:
Refactor PR: https://github.com/Atom-Learning/components/pull/384 (yes I know it really spiralled out)

LimeWub:
Closing this as refactor exists so we don't confuse people ^^

PR #384, opened 2022-08-17
Feat/chip component refactor

LimeWub:
## Description

Development work for the `Chip` component. 

Jira: https://atomlearningltd.atlassian.net/jira/software/projects/DS/boards/27?selectedIssue=DS-54

## Implementation
Refactors over @kerryambt 's work to split the `Dismissible` and `Toggle` variations of the component. 

During this process it was discussed w design and agreed that both of these `Chip` types can only exist in groups. 
Thus, all the style relevant logic has remained in the primitive `Chip`; while I have separated toggle relevant logic to `ChipToggleGroup` which combines `Chip` it with the radix `ToggleGroup` and all the dismiss relevant logic to`ChipDismissibleGroup` which combines `Chip` with our own dismissible group logic.

Rather than keep the dismiss logic in `ChipDismissibleGroup` I abstracted it out to pair of primitive, logic heavy components so it's more reusable: `Dismissible` and `DismissibleGroup` are now also being exported and can be used in the future for things like dismissible errors, banners or any other kind of content that can be removed from the view.

_**(!) I am having issues with a typescript error when forwarding refs (which you can see in the failing tests) so any help with that would be amazing!**_

### Chip (Primitive/Styles)
Screenshot 2022-08-18 at 13 10 53

Screenshot 2022-08-18 at 13 10 39

### Chip Toggle Group
ezgif-4-d6b44ccf34

### Chip Dismissible Group
ezgif-4-0538779931

### Dismissible Item (Primitive - Singular)
ezgif-4-8d2c579ee9

### Dismissible Group (Primitive - Multiple)
ezgif-4-0b376ad613


lib/package.json
```diff
@@ -135,6 +135,7 @@
     "@radix-ui/react-progress": "0.1.4",
     "@radix-ui/react-radio-group": "0.1.5",
     "@radix-ui/react-slider": "0.1.4",
+    "@radix-ui/react-slot": "^1.0.0",
```
hollg, line 138:
```suggestion
    "@radix-ui/react-slot": "1.0.0",
```

lib/src/components/chip-dismissible-group/ChipDismissibleGroup.mdx
```diff
@@ -0,0 +1,29 @@
+---
+title: Chip Dismissible Group
+component: ChipDismissibleGroup
+description: Combines the DismissibleGroup logic together with the Chip primitive styling
```
hollg, line 4: This feels like a description of the _implementation_, rather than a description of the functionality provided. I think it would be useful to have the latter here

LimeWub: The explanation functionality provided is in the DismissibleGroup (also I copied the structure of this description from ToggleGroup) :o 
What kind of description would u be expecting here?

thomasdigby (commented):
I have a slight concern that we're introducing a "dismissable" component without the relevant accessible roles or attributes. A toast, notification, banner, inline message, or anything that can be dismissed will also probably need to be announced, or at least have a special role relevant to the context within which it's used. We ought to be careful in suggesting these components should be used and I'd probably recommend keeping that logic internal to `ChipDismissableGroup` for now or at least not exposing it as a publically consumable component. What are your thoughts?

(I'd also probably say the same for the `Chip` and `ChipGroup` components, anything that doesn't have a use outside of the specific components created within the design system should be kept internal)

lib/src/components/chip-dismissible-group/ChipDismissibleGroup.mdx
```diff
@@ -0,0 +1,29 @@
...
+category: Feedback
+---
+
+Used to visualise groups of dismissible information, in fields such as an entity or different attributes. Can also be used to represent removable filtering criteria.
+
+```tsx preview
+<ChipDismissibleGroup.Root
```
thomasdigby, line 11: We've avoided using the `Component.Root` convention across the DS, are we able to use `ChipDismissibleGroup` instead? Seems that only `ToggleGroup` uses that convention

LimeWub: Why have we chosen to to avoid the `.Root` convention when the library which we are basing our components on is using said convention? Doesn't that lead to code inconsistently using `.Root` vs no root if we mix use of our own components and radix components?

LimeWub:
> I have a slight concern that we're introducing a "dismissable" component without the relevant accessible roles or attributes. A toast, notification, banner, inline message, or anything that can be dismissed will also probably need to be announced, or at least have a special role relevant to the context within which it's used. 
I can spend some time adding some aria-roles to this. I'll probably just inspect the roles of the Dialog and Collapsible radix component + do a google search in case there's something specific to this case and just add what I think is required.

> We ought to be careful in suggesting these components should be used and ...
Fair dos - is this in regards to the documentation?

> I'd probably recommend keeping that logic internal to ChipDismissableGroup for now or at least not exposing it as a publically consumable component. What are your thoughts?

I don't want to move the logic back into ChipDismissibleGroup as that's a step back. I prefer to just make this component accessible and have it as a potentially useful utility.
As for exposing the components - the main reason they are exposes are so I can use them in the documentation site and write docs about them. Why are we against exposing them?

> (I'd also probably say the same for the Chip and ChipGroup components, anything that doesn't have a use outside of the specific components created within the design system should be kept internal)
(as above - they are exposed so I can write the docs for them)

lib/src/components/chip-dismissible-group/ChipDismissibleGroup.mdx
thomasdigby: I don't think it does, we can consume packages within the DS that have an entirely different API. I can't recall the reason why we chose this convention over Root but it's important to keep it consistent across our components. I'd create a separate ticket so we can revisit the convention of a named root component.

thomasdigby:
> > I have a slight concern that we're introducing a "dismissable" component without the relevant accessible roles or attributes. A toast, notification, banner, inline message, or anything that can be dismissed will also probably need to be announced, or at least have a special role relevant to the context within which it's used. 
> 
> 
> 
> I can spend some time adding some aria-roles to this. I'll probably just inspect the roles of the Dialog and Collapsible radix component + do a google search in case there's something specific to this case and just add what I think is required.
> 
> 
> 
> > We ought to be careful in suggesting these components should be used and ...
> 
> 
> 
> Fair dos - is this in regards to the documentation?
> 
> 
> 
> > I'd probably recommend keeping that logic internal to ChipDismissableGroup for now or at least not exposing it as a publically consumable component. What are your thoughts?
> 
> 
> 
> 
> 
> I don't want to move the logic back into ChipDismissibleGroup as that's a step back. I prefer to just make this component accessible and have it as a potentially useful utility.
> 
> As for exposing the components - the main reason they are exposed is so I can use them in the documentation site and write docs about them. Why are we against exposing them?
> 
> 
> 
> > (I'd also probably say the same for the Chip and ChipGroup components, anything that doesn't have a use outside of the specific components created within the design system should be kept internal)
> 
> 
> 
> (as above - they are exposed so I can write the docs for them)

I don't think adding a11y attributes to the Dismissible component solves the problem, it requires an understanding of the context and specific use case for those attributes to be correct.

The reason why I don't think they should be made public is that a dev can't just drop them in and use them, they would need to add extra logic and a11y specific roles/attrs to make sure they're accessible, something we don't do for any other components.

To be clear, I don't need you to move any files or logic around, the structure you have is fine, I just want to avoid making these components available public. I'm happy for us to revisit in future if there's a use case that requires external access to these primitive components, but I'd bet that if that does happen we would be better off building a specific component for that case anyway, with the correct attributes and semantics. e.g. `Notification` or `Banner`

lib/src/components/chip-dismissible-group/ChipDismissibleGroup.mdx
LimeWub: https://github.com/Atom-Learning/components/pull/384/commits/e0bc794c3f20fef256337cc1620310e44c1794e6

I've done this here. I think it looks slightly more complicated than just exporting Root but it's good to be consistent for now.

lib/src/components/chip-dismissible-group/ChipDismissibleGroup.mdx
LimeWub: Lmk about this and I'll change it ^^

hollg: Sorry, I totally missed your reply! I just opened the docs site to look at other examples and there are a whole bunch which do have this style of description, so I'm happy to mark this as resolved and maybe we can think about rationlising our approach to them in general at some point in the future 👍🏻

lib/src/components/chip-dismissible-group/ChipDismissibleGroupItem.tsx
```diff
@@ -0,0 +1,39 @@
...
+  ...rest
+}) => {
+  return (
+    <DismissibleGroup.Item asChild {...rest}>
+      <StyledChipDismissibleGroupItem size={size}>
+        <Chip.Content>{children}</Chip.Content>
+        <DismissibleGroup.Trigger asChild>
+          <ActionIcon
+            label="Dismiss" // size css will be replaced with size prop when action icon size fixes are done
+            css={{
```
LimeWub, line 28: I just noticed this. I'll need to remove the size css since ActionIcon changes got merged.

lib/src/components/index.ts
```diff
@@ -10,10 +10,15 @@ export { Button, StyledButton } from './button'
...
 export { Checkbox } from './checkbox'
 export { CheckboxField } from './checkbox-field'
+export { Chip, ChipGroup } from './chip'
+export { ChipDismissibleGroup } from './chip-dismissible-group'
+export { ChipToggleGroup } from './chip-toggle-group'
 export { Combobox } from './combobox'
 export { Dialog } from './dialog'
 export { DateInput } from './date-input'
 export { DateField } from './date-field'
+export { Dismissible } from './dismissible'
```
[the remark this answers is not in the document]
LimeWub: No I have to be exporting it because otherwise I cannot do documentation for it. What I did is I moved this is a 'Primitives' folder in the navigation.

lib/src/components/chip/Chip.mdx
```diff
@@ -0,0 +1,66 @@
+---
+title: Chip
+component: Chip, ChipGroup
+description: A component in the shape of a pill providing visual cues to prompt users to enter information or filter content.
+category: Feedback
+---
+
+`Chip` itself is a primitive. It has no functional logic in itself, however, it is used to provide common styles for all the `Chip`-based components.
```
thomasdigby, line 8: Can we add a similar description and notice for `Dismissible` and `DismissibleGroup` to make sure devs are aware that they will need to add the relevant logic and test for accessibility if they use them?

lib/src/components/chip-dismissible-group/ChipDismissibleGroupRoot.tsx
```diff
@@ -0,0 +1,14 @@
...
+import { DismissibleGroup } from '~/components/dismissible-group'
+
+type TChipDismissibleGroupRootProps = React.ComponentProps<typeof ChipGroup> &
+  React.ComponentProps<typeof DismissibleGroup>
+
+export const ChipDismissibleGroupRoot = React.forwardRef<
+  HTMLDivElement,
+  TChipDismissibleGroupRootProps
+>((props, ref) => {
+  return <ChipGroup as={DismissibleGroup} ref={ref} {...props} />
```
thomasdigby, line 13: Really cool that we can wrap a component around already defined functionality. It behaves as a dismissible group and has all the correct roles and semantics for that, but visually looks like a chip group.

`ChipGroup as={DismissibleGroup}` and `ChipGroup as={ToggleGroup.Root}` is a great internal API 👍

lib/src/components/chip/Chip.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+export type TChipRootProps = React.ComponentProps<typeof StyledRoot> & {
+  size?: 'sm' | 'md' | 'lg'
+  asWorkaround?: React.ElementType // (!?) `asWorkaround` rather than `as` because, it seems, when we extend this via `styled()` stitches overrides this component from the first argument for the value in `as`
+}
+
+const ChipRoot: React.ForwardRefExoticComponent<TChipRootProps> =
+  React.forwardRef(({ asWorkaround, size = 'md', ...rest }, ref) => {
+    return (
+      <ChipRootProvider size={size}>
```
thomasdigby, line 114: We might have discussed this before, so apologies if I've forgotten, but what's the purpose of the context and `Chip.Content` component here? What's preventing us from returning the styled component and passing the relevant props down to the children in `ChipRoot`?

LimeWub: The `Chip.Content` is here as an extra div with extra margin on each side to achieve the design requirements.

The context is here so that nested children can read the size they are supposed to be without worrying about passing props to them.

I'm seeing I'm using it in Content to pass a size down to the Icon. I think I will make a change so there's a ChipIcon which reads directly from the context - so we avoid any cloneElements.

But ye - both are here basically for styling.

> What's preventing us from returning the styled component and passing the relevant props down to the children in ChipRoot?
I'm not sure I understand this alternative :o How would we pass the roots down? cloneElement?

lib/src/components/chip/Chip.mdx
LimeWub: Sure, how does this sound?

> `Dismissible` itself is a primitive. It has not been designed so it should not be used out-of-the-box as a user facing component. Test the component is accessible depending on each implementation use case.

thomasdigby: Yeah that's great, thanks!

lib/src/components/dismissible-group/DismissibleGroupItem.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+export const DismissibleGroupItem: React.FC<TDismissibleGroupItemProps> = ({
+  children,
+  disabled: itemDisabled = false,
+  ...rest
+}) => {
+  const groupContext = React.useContext(DismissibleGroupContext)
+  if (groupContext === undefined) {
+    throw new Error(
+      'DismissibleGroup.Item should be use withing a DismissibleGroup'
```
thomasdigby, line 20:
```suggestion
      'DismissibleGroup.Item can only be used within a DismissibleGroup'
```

lib/src/components/dismissible/Dismissible.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+const DismissibleImplementation = () => (
+  <Dismissible value="a" onDismiss={mockOnDismiss} data-cy="root">
+    A
+    <Dismissible.Trigger data-cy="trigger" />
+  </Dismissible>
+)
+
+const DismissibleImplementationCustomOverrides = () => (
+  <Dismissible value="a" onDismiss={mockOnDismiss} asChild>
+    <div data-cy="custom-root">
```
thomasdigby, line 18: Super pedantic, but can we replace all of our `data-cy` attributes in tests with `data-testid`? `data-cy` is used primarily (exclusively?) for Cypress and we're not using Cypress in our tests here.

LimeWub: Not pedantic at all! It will make the tests cleaner as well since I wont need to be doing querySelects (I think)

LimeWub:
@thomasdigby I've addressed all the comments ([and added a utility you may be interested to check out to support overriding + passing responsive variants through to children](https://github.com/Atom-Learning/components/pull/384/files#diff-e7580072058b9562234697cfee60e57ff652f78b1ae99c2fec8273f8c2ffe95d) ). Lmk if all is good with this now or if there's any more tweaking to do ^^

thomasdigby (approved):

PR #385, opened 2022-08-18
feat: creating bulk actions bar component

viniciusvolpe:
## Description
Creating a bulk actions bar component to be used with lists or tables and allowing executing actions in all selected items.
The `BulkActionsBar` can have only children of type `BulkActionsBar.Action` and the actions can have two modes: main actions that are displayed as buttons and other actions that are displayed as dropdown items. The mode is defined by the `isMain` property and the main actions are moved to the dropdown on small screens

Jira: https://atomlearningltd.atlassian.net/browse/PP-585

## Implementation
- Created bulk actions component
- added tests and documentation

## Screenshots

Large screens:
image

Small screens:
image


hollg:
These components look great and I am extremely appreciative of how thorough you've been with the docs and tests! I do have a couple of concerns about the API though unfortunately.

We've run into trouble in the past when we've used a parent component to loop through child components to make decisions about what to render. It makes the API of these components difficult to extend in the future, because we can't add new types of valid child components without changing the implementation of the parent component (i.e. it breaks [the open/closed principle](https://stackify.com/solid-design-open-closed-principle/)). We also see this problem when we use boolean props to communicate a choice between two mutually exclusive types, like `isMain` in this API.

For example, if the design team realise down the line that they need a third type of action, we would need to replace `isMain` with some enum prop that accepts three values, otherwise we'd be stuck adding new props, e.g. `isSecondary`, and updating the code in the parent with some logic about the different combinations of those props which gets ... messy 😅. And replacing `isMain` still means updating existing uses of the components and making changes to the implementation of the parent to support the new type.

The way that we usually try to avoid these issues in this library is to use a more compositional API, i.e. take some of these decisions out of the parent component and give them to the consumer. In this case, I think that could be achieved with an API somewhere along these lines:

```tsx
<BulkActionsBar
  label="1 item selected - Sorted by session name"
  cancelLabel="Cancel"
  onCancel={() => {}}
>
  <BulkActionsBar.Action
    icon={Edit}
    text="Edit"
    onClick={() => {}}
  />
  <BulkActionsBar.Action
    icon={Bolt}
    text="Other action"
    onClick={() => {}}
  />
  <BulkActionsBar.Dropdown>
    <BulkActionsBar.DropdownAction
      icon={Trash}
      text="Delete"
      onClick={() => {}}
    />
  </BulkActionsBar.Dropdown>
</BulkActionsBar>
```

If we know from Miguel that the dropdown is always going to use the same icon as a trigger, then that can be built into `BulkActionsBar.Dropdown` so that we're still providing a useful convenience to the consumer (i.e. preventing them from using the same `Dropdown.Trigger` code in every use of these components). But it also leaves open the possibility of introducing a new type of dropdown further down the line if the scope of what the design team would like to use these components for changes. And introducing a new type would have no impact on the original type.

I would go as far as to suggest _not_ enforcing child types at all. The design team should be able to extend these components over time and as edge cases are found for them. If another pod's designer needs an extra vertical divider between two groups of "main" actions, or maybe a tooltip, I don't think an engineer should have to come into this lib, edit the parent component, make a new release and update `atom-core` to the new version, for example.

Does that make sense? Let me know if not and I'm very happy to chat further about it on Slack or on a call etc.

hollg:
@viniciusvolpe I've just realised that the design documentation used for this component has come from Miguel's table work, which isn't marked as being ready for development yet. Did you speak to him about building this before it reached that milestone?

thomasdigby:
@viniciusvolpe Is this PR still relevant/useful?

PR #386, opened 2022-08-26
Feat: DataTable components

hollg:
MVP for the new`DataTable` component, which provides advanced table features (i.e. the stuff we currently have in `EnhancedTable`), but with a more composition-friendly and extensible API.

This initial version:
- [X] Establishes provider / subcomponent API
- [x] Adds sortable columns feature
- [X] Adds pagination feature
- [X] Adds global search feature

## Implementation
We're using the new verison of `react-table` (now known as `@tanstack/react-table`) for all the table state and complex logic. The basic idea is that `DataTable` configures and manages the instance of `useReactTable()` from that library and exposes its state and methods via a new custom hook called `useDataTable`.

`useDataTable` also exposes some additonal methods that allow child components to interact with the table state in useful ways. For example, `setIsPaginated` allows the new `DataTable.Pagination` component to configure pagination at the `DataTable` level as soon as it's rendered. This is a pattern I'd like to lean heavily into with future extensions of this component and API. It should allow us to create a whole range of just-drop-them-in-and-they-work components that abstract the complexities of any common use cases away from the consumer **without** building it all into one component that becomes more and more difficult to add new features to (_cough `EnhancedTable` cough_).

Because we can call `useDataTable` anywhere within a `DataTable`, we always have the option of manually passing its state into our more fundamental `Table` components, e.g. `Table.Row` etc. This is important because it is also how we'll build new `DataTable` components. For example, we're going to need selectable and expandable rows in the near-ish future.

I've also added components like `DataTable.Row` and `DataTable.Body` to provide useful default behaviours that we would otherwise be writing quite a lot of boilerplate code for every time we used any of these components. `DataTable.Row` takes a `Row` type from `@tanstack/react-table` as a prop and renders each of its cells as `Table.Cell`s. `DataTable.Body` calls `useDataTable` to get all the rows, and passes each one into a `DataTable.Row`. The idea here is that most of our tables are pretty similar, so we should be able to bundle that common behaviour up into easy-to-access defaults while still giving access to the low-level stuff for those who need it.

### Primary theme, not striped, with search and pagination:
Screenshot 2022-09-28 at 08 49 24
Code:
```tsx
<DataTable
  columns={columns}
  data={data}
  defaultSort={{ column: 'name', direction: 'asc' }}
>
  <DataTable.Search
    css={{ mb: '$4', maxWidth: '300px' }}
    label="User search"
    placeholder="Search for a user"
  />
    <DataTable.Table css={{ mb: '$4' }} theme="primary" striped={false} />
    <DataTable.Pagination pageSize={5} />
</DataTable>
```

### Light theme, striped, with search and pagination, search label hidden:
Screenshot 2022-09-28 at 08 53 17

Code:
```tsx
<DataTable
  columns={columns}
  data={data}
  defaultSort={{ column: 'name', direction: 'asc' }}
>
  <DataTable.Search
    css={{ mb: '$4', maxWidth: '300px' }}
    label="User search"
    placeholder="Search for a user"
    hideLabel
  />
    <DataTable.Table css={{ mb: '$4' }} theme="light" striped />
    <DataTable.Pagination pageSize={5} />
</DataTable>
```

## Search demo

https://user-images.githubusercontent.com/21319237/192722539-ad42d11f-9615-4c9c-9ed0-2cb696a39f1b.mov

## Sort / pagination demo

https://user-images.githubusercontent.com/21319237/192760428-cb814fb3-bfec-48df-9552-492aadbaa928.mov

## Unfinished business
1. Annoyingly, I haven't been able to create any previewable or editable code examples for the docs site. The component we use to display them only supports React markup, but the table requires some imports and `const` declarations. (Actually, the `const` declarations can be avoided by pasing a _gross_ IIFE as a prop, but that doesn't solve the import issue).
2. The `DataTable.Pagination` component includes some copy for things like "1 of 4 pages". I have [added a ticket to the DS backlog](https://atomlearningltd.atlassian.net/browse/DS-130?atlOrigin=eyJpIjoiYTg4YWYxMTlkMjA3NDkzYmJmYTEwYjE4ODYwNjNiMzUiLCJwIjoiamlyYS1zbGFjay1pbnQifQ) to set something up to manage this generally (especially re different locales), e.g. by allowing the consumer to override default copy.


lib/src/components/data-table/pagination/Pagination.tsx
```diff
@@ -0,0 +1,84 @@
...
+    setPageSize,
+    getTotalRows
+  } = useDataTable()
+
+  React.useEffect(() => {
+    applyPagination()
+  }, [applyPagination])
+
+  React.useEffect(() => {
+    setPageSize(defaultPageSize)
```
thomasdigby, line 44: I like how we can co-locate pagination configuration with the component 👍 And as we're always going to require a delay to fetch data in this context, the `useEffect` won't delay applying the correct layout

lib/src/components/data-table/pagination/Buttons.tsx
```diff
@@ -0,0 +1,80 @@
...
+    iconSize?: 'sm' | 'md' | 'lg'
+  }
+export const TooltipActionIcon: React.FC<
+  Omit<TooltipActionIconProps, 'children' | 'is'>
+> = ({ icon, iconSize, label, ...rest }) => {
+  return (
+    <Tooltip>
+      <Tooltip.Trigger asChild>
+        <ActionIcon label={label} {...rest}>
+          <Icon is={icon} size={iconSize} />
```
thomasdigby, line 74: If there's not a specific reason to retain it, we should remove `iconSize` here as `ActionIcon` should now control `Icon` sizing.

lib/src/components/data-table/DataTable.mdx
```diff
@@ -0,0 +1,113 @@
...
+  { name: 'betty', hobby: 'acting' }
+]
+```
+
+#### With Defaults
+
+```tsx
+<DataTable.Provider columns={columns} data={data}>
+  <DataTable>
+    <DataTable.Header />
```
thomasdigby, line 56: Should this be `DataTable.Head` to align with standard table naming conventions? 🤷

lib/src/components/data-table/pagination/Buttons.tsx
andoulla: Just an FYI, I had talked to Miguel about removing the tooltip action icon and just making the ActionIcon to have the option to show a tooltip. As part of that, it was to remove the option of changing the icon size and let the action icon just control the size of it. It would be the thing I would do on Wednesday during my guild day now that the big ActionIcon PR is merged.

lib/src/components/data-table/DataTable.mdx
```diff
@@ -0,0 +1,113 @@
...
+  {() => {
+    const { getHeaderGroups, getRowModel } = useDataTable()
+
+    return (
+      <Table>
+        <Table.Header>
+          {getHeaderGroups().map((headerGroup) => (
+            <Table.Row key={headerGroup.id}>
+              {headerGroup.headers.map((header) => (
+                <DataTable.HeaderCell
```
thomasdigby: It feels a little incongruous to mix `Table` and `DataTable` components here... I don't know what I'd suggest otherwise though. Adding conditional `DataTable` logic to `Table.HeaderCell` isn't ideal, removing `sortable` support from "roll your own" to keep the component simple is not helpful to the user... 🤔 

Could we substitute all components in "roll your own" for `DataTable` aliased versions of `Table` components? e.g. `DataTable.Cell`. Or would that clash with `DataTable.Header` and `DataTable.Body` that have already been created? 🤔

lib/src/components/data-table/DataTableHeaderCell.tsx
```diff
@@ -0,0 +1,33 @@
...
+  header,
+  sortable,
+  children,
+  ...props
+}) => {
+  const sort = header.column.getIsSorted()
+
+  return (
+    <Table.HeaderCell
+      onClick={sortable ? header.column.getToggleSortingHandler() : undefined}
```
thomasdigby, line 26: Can we add `cursor: 'pointer'` to headers that are sortable? Or is there then no way to differentiate between sortable columns and non-sortable columns?

lib/src/components/data-table/DataTableHeaderCell.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { Icon } from '../icon'
+import { ChevronDown, ChevronUp } from '@atom-learning/icons'
+type DataTableHeaderCellProps = React.ComponentProps<typeof Table.Header> & {
+  header: Header<Record<string, unknown>, unknown>
+  sortable?: boolean
+}
+
+const sortIcons = {
+  asc: <Icon is={ChevronUp} size="sm" css={{ ml: '$1' }} />,
+  desc: <Icon is={ChevronDown} size="sm" css={{ ml: '$1' }} />
```
thomasdigby, line 13: Would like your opinion on `SortUp` and `SortDown` as icons instead of the chevrons. I'm... on the fence, but there's something about the chevrons that don't _quite_ feel right.

lib/src/components/data-table/pagination/Buttons.tsx
```diff
@@ -0,0 +1,80 @@
...
+  pageIndex: number
+  pageCount: number
+  gotoPage: (pageNumber: number) => void
+}> = ({ gotoPage, pageCount, pageIndex }) => {
+  return (
+    <Flex css={{ alignItems: 'center' }}>
+      <Select
+        defaultValue={pageIndex}
+        size="sm"
+        css={{ flex: 'none', mr: '$3' }}
```
thomasdigby, line 44: Are we able to reduce the width of this `Select` so that it's closer to the width of the numbers? Or maxes out at 3 character width? I rendered these at 74px max-width, feels a bit nicer?

Screenshot 2022-09-15 at 16 48 12
Screenshot 2022-09-15 at 16 48 07

lib/src/components/data-table/DataTable.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+Other `DataTable` components use `useDataTable` and provide useful default behaviours for different functionality. If they don't meet your requirements, you can build your own implementation by calling `useDataTable` in a new component yourself. If your use case is a common one, please make a PR to add the component to the library. The root `DataTable` component is a re-export of our `Table` component for convenience, since it doesn't need to contain any logic itself.
+
+### Using defaults vs using rolling your own
+
+Here are two ways of building the same `DataTable` from this set of columns and data. The first uses the highest-level components that are bundled into `DataTable` for useful default behaviours with minimal code. The second uses a combination of lower-level `DataTable` components and the underlying `Table` components to achieve the same thing. This demonstrates how you could create more custom table UIs without the need to extend the high-level components.
+
+```tsx
+const columnHelper = createColumnHelper<{
```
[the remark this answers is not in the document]
thomasdigby: 🤔

I _think_ we probably ought to **avoid** exporting it from the DS as it will pollute our exports, and there's the potential for collisions that we can't control. It feels like a very data-specific fn, which AFAIC is not a DS problem, you get your data in order and you give it to us to render. Are the other frequently used utils similar in use?

lib/src/components/data-table/DataTable.mdx
thomasdigby: We'd probably need to update the `Table` component as well, but I wouldn't necessarily be against that...

lib/src/components/data-table/DataTable.mdx
thomasdigby: That looks much better, nice separation between the data table provider/hook and render components 👍 I don't think we need aliases if we keep this like this.

lib/src/components/data-table/DataTable.mdx
```diff
@@ -0,0 +1,121 @@
...
+```
+
+#### With Defaults
+
+```tsx
+<DataTable.Provider columns={columns} data={data}>
+  <DataTable>
+    <DataTable.Header sortable />
+    <DataTable.Body />
+  </DataTable>
```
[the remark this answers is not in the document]
hollg: Definitely achievable, but I haven't been able to make a firm decision on desirable 😅 

I think this gets at the same issue as a couple of your other comments, and the first question in the PR description. I think that API would be _great_ in all the cases where it works. But as we add more features to `DataTable` (selectable rows, bulk actions, expandable rows, re-orderable rows, whatever), I don't want to reach a situation where we have `DataTable.Table`, which sounds like it should sort of just be able to do all the stuff, but we can't use it if we want certain sets or combinations of those features. 

However, the same thing can probably be said about `DataTable.Body` and `DataTable.Row`! Maybe it becomes more of an issue with `DataTable.Table` though because it's one more level of misdirection and the styling on the base `Table` components is already a _little_ obscured, e.g. variants on `Table.Body` set the styles for the `Table.Rows` within.

Do you have any thoughts on how we might define the desirable scope of these "bundle of defaults" type components?

lib/src/components/data-table/DataTableHeaderCell.tsx
```diff
@@ -0,0 +1,33 @@
...
+}) => {
+  const sort = header.column.getIsSorted()
+
+  return (
+    <Table.HeaderCell
+      onClick={sortable ? header.column.getToggleSortingHandler() : undefined}
+      {...props}
+    >
+      {flexRender(header.column.columnDef.header, header.getContext())}
+      {sort && sortIcons[sort as string]}
```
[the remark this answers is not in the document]
hollg: Oh yeah nice, great idea! Honestly I haven't actually touched the styling or logic of the sorting/pagination other than to move them over from `atom-core` and get them to work with the new API so I'm sure there's a fair amount of low-hanging fruit like this. I'm going to try to prioritise the more complex work on the API and tests, so these comments might stay unresolved for a little while but I will get to them!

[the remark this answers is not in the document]
thomasdigby: Ah yeah cool, that's all good, I'm pretty happy with 90% of this so just finding little bits to refine while we get the chance 👍

lib/src/components/data-table/DataTable.mdx
```diff
@@ -0,0 +1,121 @@
+---
+title: DataTable
+component: DataTable, DataTable.Provider, DataTable.Body, DataTable.Header, DataTable.Row, DataTable.HeaderCell
```
LimeWub, line 3: (Unless there's a use case for Provider to be used separately from a DataTable:)

Does the DataTable and DataTable.Provider need to be 2 separate components? Can't the DataTable export itself nested in a DataTable.Provider directly.

ie.
```
const DataTableInternal = () => {} // The current data table
const DataTable = ({providerRelatedProps, ...rest}) => <DataTableProvider><DataTableInternal {...rest} /></DataTableProvider>
```

I've done similar to this patter here: Screenshot 2022-09-16 at 12 00 53

hollg: This is a super interesting thought! The one thing we definitely need to avoid is putting the `DataTable.Provider` inside a self-closing component that doesn't give the consumer the opportunity to put other components inside it. But maybe one of the components that bundle up default behaviours could do this 🤔

lib/src/components/data-table/context.tsx
```diff
@@ -0,0 +1,71 @@
...
+  children,
+  sortable
+}: TableProviderProps): JSX.Element => {
+  const [isPaginated, setIsPaginated] = React.useState<boolean>(false)
+
+  const applyPagination = React.useCallback(() => {
+    setIsPaginated(true)
+  }, [])
+
+  const getTotalRows = () => data.length
```
LimeWub, line 37: Minor: You know my opinion about this being a function :p 
I think `const totalRows = data.length` is enough u_u

lib/src/components/data-table/pagination/Buttons.tsx
LimeWub: @andoulla 

FYI (irrelevant to this PR) there's a design I've recently reviewed where the ActionIcon shows a tooltip when **disabled**. So an ActionIcon which shows a tooltip onClick may not be enough and perhaps we should just wrap ActionIcon's that need to show a Tooltip (at any state) with a `Tooltip.Trigger` (so it works for both enabled/disabled ActionIcons - and any other Buttons/Links tbf). Basically I think the way it's used here is better than extending it like so: "making the ActionIcon to have the option to show a tooltip"?

https://www.figma.com/file/KjgQh7eTwEfJmH6e56k9mX?node-id=7252:52083#266376228

lib/src/components/data-table/pagination/Pagination.tsx
```diff
@@ -0,0 +1,84 @@
...
+})
+
+type PaginationProps = React.ComponentProps<typeof StyledNav> & {
+  pageSize?: number
+}
+
+/** Applies pagination to parent DataTableProvider and renders UI to switch pages etc */
+export const Pagination: React.FC<PaginationProps> = ({
+  css,
+  pageSize: defaultPageSize = 10,
```
LimeWub, line 24: Can this change in some other way other than the prop right here? (Is the naming of `default` actually correct?)

lib/src/components/data-table/pagination/Pagination.tsx
```diff
@@ -0,0 +1,84 @@
...
+  React.useEffect(() => {
+    setPageSize(defaultPageSize)
+  }, [setPageSize, defaultPageSize])
+
+  const { pageIndex, pageSize } = getState().pagination
+  const recordsCountFrom = pageIndex * pageSize + 1
+  const recordsCountTo = recordsCountFrom + getRowModel().rows.length - 1
+
+  return (
+    <StyledNav {...props} css={{ mt: '$4', ...css } as CSS}>
```
LimeWub, line 52: margin-top?

Can't the element over this have margin-bottom (or wrap them both in a Stack?)

lib/src/components/data-table/pagination/Pagination.tsx
```diff
@@ -0,0 +1,84 @@
...
+      <Text size="sm" css={{ flexBasis: '25%' }}>
+        {`${recordsCountFrom.toString()} - ${recordsCountTo.toString()} of ${getTotalRows()} items`}
+      </Text>
+      <Flex
```
LimeWub, line 56: i don't think u need this `Flex`wrapper here.
I also don't think u need any of the flexBasis set on any of these components 🤔

lib/src/components/data-table/pagination/Buttons.tsx
LimeWub, line 44: Why do u need flex: none? 👀

lib/src/components/data-table/DataTable.mdx
hollg: Yeah exactly, the root `DataTable` component is now the context provider, à la Radix. I think it tidies things up pretty significantly!

lib/src/components/data-table/DataTable.mdx
```diff
@@ -0,0 +1,121 @@
...
+Note that `useDataTable` can only be called by a child component of `DataTable.Provider`. In a real example, you'll probably have a separate named component which makes the `useDataTable` call, because if you're not using the defaults as above then you probably have some complex logic involved. In this example we've got an inline child component for simplicity.
+
+```tsx
+<DataTable.Provider columns={columns} data={data}>
+  {() => {
+    const { getHeaderGroups, getRowModel } = useDataTable()
+
+    return (
+      <Table>
+        <Table.Header>
```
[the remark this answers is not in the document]
hollg: My feeling at the moment is that we shouldn't alias the `Table` subcomponents as `DataTable` subcomponents, because I think doing so makes it less clear what `DataTable` is actually for (which is almost certainly an issue these docs have at the moment too — lots of room for improvement here). Specifically, `DataTable` does two things:

1. (`DataTable` root component) Provide an interface to `@tanstack/react-table` via a Context provider/hook as a simple and extensible API for building interactive tables
2. (`DataTable.Whatever` subcomponents) Provide useful default implementations for common features of interactive tables

We can think of each `DataTable` subcomponent as an implementation of this idea: `DataTable state management + UI components = useful table feature, compatible with any other DataTable instance`. To me, this makes it very intuitive that I can build _my own_ implementation of that idea when needed. The `Table` components are the UI components for tables in our Design System, so obviously I'll lean heavily on those for my implementation. But I can also include other things, like `Button`s and `Icon`s etc.

This changes if we alias the `Table` components on `DataTable`. For example if we have `DataTable.Row = Table.Row` then there are two categories of `DataTable` subcomponents: one that includes its own logic and state management via `useDataTable()`, e.g. `DataTable.Pagination`, and one that is _just_ a UI component, with no logic at all (e.g. `DataTable.Row`). But I don't think we'd be providing any clear way of distinguishing between them without going into the docs or source code. Whereas if we don't alias anything, then the pre-existing namespaces of `DataTable` and `Table` provide that distinction for us already.

Aliasing those components would also make it more complex to name the `DataTable` subcomponents that are _not_ `Table` component aliases. E.g. we'd need a new name for the current `DataTable.Body`, which provides a default implementation using `useDataTable`, to free that name up for the `DataTable.Body = Table.Body` setup.

> My reasoning for this is if in the future we want to do a specific override for a table component relevant to DataTable we can just extend the exported component.

If I'm understanding this point correctly, then I think the API that I've described solves (or perhaps side-steps) this problem because we can make whatever changes we want to our default implementations, and any code in `atom-core` that isn't using those implementation _shouldn't_ be affected by changes to them anyway.

This has  been a bit of a stream-of-consciousness for me but I hope it makes sense and am keen to hear your thoughts!

lib/src/components/data-table/DataTableHeaderCell.tsx
hollg: TBH I'll ask Miguel about this as I've just copied what already exists in `EnhancedTable` atm. They definitely don't look great!

lib/src/components/data-table/pagination/Buttons.tsx
LimeWub: I think I don't necessarily see why 'rebuilding the pattern' is a big problem. I'm assuming it's just a tiny bit of additional nesting?

Is the usecase for tooltip + action icon increased visual a11y/potentially less cognitive load? If so I can kinda see why u may want to combine it tbh...

lib/src/components/data-table/DataTableHead.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { TABLE_HEADER_THEMES } from '../table/TableHeader'
+import { useDataTable } from './DataTableContext'
+import { DataTable } from './index'
+
+type DataTableProps = {
+  sortable?: boolean
+  theme?: keyof typeof TABLE_HEADER_THEMES
+}
+
+export const DataTableHead: React.FC<DataTableProps> = ({
```
LimeWub, line 13: Shouldn't this be called `DataTable.Header` if it's building a `Table.Header` structure?

lib/src/components/data-table/DataTableHeader.tsx
```diff
@@ -0,0 +1,33 @@
...
+  header: Header<Record<string, unknown>, unknown>
+  sortable?: boolean
+}
+
+const sortIcons = {
+  asc: <Icon is={ChevronUp} size="sm" css={{ ml: '$1' }} />,
+  desc: <Icon is={ChevronDown} size="sm" css={{ ml: '$1' }} />
+}
+
+export const DataTableHeader: React.FC<DataTableHeaderCellProps> = ({
```
LimeWub, line 16: DataTableHeaderCell?

lib/src/components/data-table/DataTableTable.tsx
```diff
@@ -0,0 +1,21 @@
...
+> & {
+  sortable?: boolean
+}
+
+export const DataTableTable: React.FC<DataTableTableProps> = ({
+  sortable = true,
+  ...props
+}) => (
+  <Table {...props}>
+    <DataTable.Head sortable={sortable} />
```
LimeWub, line 18: Rather than passing down sortable to the `Header` can the sortable setting live in the context? Then u don't need to prop-drill and the people doing custom implementations of HeaderCell don't need to care about passing sortable in 2 places (Table/HeaderCell) (?)

lib/src/components/data-table/DataTableHeader.tsx
```diff
@@ -0,0 +1,33 @@
...
+  header,
+  sortable,
+  children,
+  ...props
+}) => {
+  const sort = header.column.getIsSorted()
+
+  return (
+    <Table.HeaderCell
+      onClick={sortable ? header.column.getToggleSortingHandler() : undefined}
```
LimeWub, line 26: What does this do? Does it actually do the sorting?
Shouldn't the sorting happen when the sorting icon is clicked (ActionIcon so button so accessible) rather than a HeaderCell - which is not accessible?

lib/src/components/data-table/DataTable.mdx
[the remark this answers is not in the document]
LimeWub: Thanks for explaining. I think I get it and agree with you. 

Off the top of my head I think the Data version (implementation w data related logic) of these components will work fine for extending and there should be no need to have both the Data version and the building block which accepts children (Table) version exported both.

However, can we still have DataTable.Cell exported? (U can use that in the row as well)
```
const DataTableCell = (data, ....rest) => (
<Table.Cell {...rest}>
     {flexRender(data.column.columnDef.cell, data.getContext())}
</Table.Cell>
)
```

We already have `HeaderCell` exported (with its own logic) and I think it we have ALL of the Table components exported in their DataTable version.


Speaking of all components. Are we not extending the TableFooter/TableFooterCell in this PR? Or has it just been forgotten about :o ?

lib/src/components/data-table/pagination/Buttons.tsx
```diff
@@ -0,0 +1,80 @@
...
+  disabled: boolean
+  onClick: () => void
+}> = ({ direction, ...remainingProps }) => {
+  const isNext = direction === 'next'
+
+  return (
+    <TooltipActionIcon
+      appearance="outline"
+      icon={isNext ? ChevronRight : ChevronLeft}
+      iconSize="sm"
```
LimeWub, line 24: iconSize? Why do we need this in addition to size? (I believe ActionIcon does the correct sizing for its children now btw)

lib/src/components/data-table/DataTableTable.tsx
LimeWub: Plus. If you want to bind the sort to the ActionIcon in the header (my preference) u can lit make this its own component and get the 'sortable' from the context and avoid even more prop drilling - then we can use the sort icon in other placements (ie if we want sorting to also live in other place - say in the footer or so.) 
Anyway having the sortable in the context means other components (inc custom can know about it so it's a win imo :o)

lib/src/components/data-table/DataTable.mdx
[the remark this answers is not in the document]
hollg: > However, can we still have DataTable.Cell exported? (U can use that in the row as well)

Oh good call, thanks! I'll add that today.

> Speaking of all components. Are we not extending the TableFooter/TableFooterCell in this PR? Or has it just been forgotten about :o ?

To be honest I hadn't they've been at the bottom of my list of priorities because I didn't think we were really using them, but I should do some actual discovery on that and confirm whether they are needed in order to replace existing tables!

lib/src/components/data-table/DataTableHeader.tsx
hollg: I don't have answers to all these questions atm as the sorting and pagination UI is just copied over from what existed in `atom-core`, but there is no icon when the column hasn't been sorted yet, so without a change to that logic we can't implement that change. Will look into it!

lib/src/components/data-table/DataTableTable.tsx
hollg: Do you mean can the prop itself be passed to the provider? I've tried to avoid that because I've had some difficult-to-put-into-words gut feeling that it makes more sense as config for the UI than for the data, but I do think what you're saying makes a lot of sense and I think it's also the only way to avoid the need to make `sortable` configurable in a bunch of different places. 

I reckon moving it to `DataTable` is the right thing to do. But there is a question remaining about the API for specifying the default column to sort by, so you might not see me push this change until I figure that out too!

lib/src/components/data-table/DataTableHead.tsx
hollg: The existing `Table` components aren't named according to the actual names of the HTML elements they render. @thomasdigby mentioned it in another comment somewhere. I think I've got all the `DataTable` components named according to the name of the underlying HTML now, but it does introduce inconsistency with `Table`. Not sure whether we should stick with the not-technically-correct `Table` names or bite the bullet and get this set right. 

Keen to hear both your thoughts!

LimeWub: Ah I seeeeeeeeeeeee... I hadn't noticed!

I'd be tempted to stick with the naming we have for table now for the following reasons:

- HeaderCell as child of Head makes my naming OCD sad (yes I know this is lit what HTML is doing)
(- More sadness because in other html choices 'head' is document head w all the hidden/meta stuff and 'header' is the actual top of the body with the content)
- Table.Footer (tfoot) has a Table.FooterCell in our implementation - which doesn't map to any special html component (td)
- We still call td as 'Cell' and not `Data` (td = table data) or 'DataCell' so we are not matching html there
- People who have already seen/used the table namings or are replacing bits with them to extend functionality might be confused that one thing has 2 different potential names. (Head.... Header? HeaderCell -> Header?????)

So ye - I'm personally of the opinion we just keep the names we have for Table stuff as they are now

lib/src/components/data-table/DataTableTable.tsx
LimeWub: `Do you mean can the prop itself be passed to the provider?`

I don't _think_ I'm saying that. I meant the same way pagination being on is 'enabled' in the provider.
So if DataTable is sortable it sets 'sortingEnabled' (or something like that) on in the provider - rather that prop drilling 'sortable' to the header cell. 
HeaderCell (or whatever wants to check if it does sorting) just reads from the context.

lib/src/components/data-table/DataTableHeader.tsx
LimeWub: ohhhh so the entire cell is clickable - it has no icon if it's not sorted by but has an icon if it is.
The Cell itself is not a button - but has an onClick.

Honestly I think we should ask design about whether the contents of a sortable header should just be a button (for a11y) - because if there's no chance of other buttons being in there it makes semantic sense. But if other buttons may be in the header we need to rethink the 'cell as button' idea .-. 
(can leave as is but I don't think it's accessible rn)

lib/src/components/data-table/pagination/Pagination.tsx
LimeWub: You could add a Stack around them - but that will add elements to the DOM. 
Tbf, it's nbd and now that u explained it it makes sense - so idm if there's changes made to this or not :p

lib/src/components/data-table/DataTableTable.tsx
hollg: If it's _not_ a prop on `DataTable`, then I think we end up having to do the prop drilling (or worse). 

It needs to be configurable on `DataTable.Table`, because when a consumer uses that component, they don't have direct access to anything at a lower level. But it also needs to be configurable on lower level components, because we want to support a consumer using `DataTable.Head` and their own table body implementation.

LimeWub: That's not what I'm saying. Lets have a quick call to try to clarify this. @hollg

lib/src/components/data-table/DataTable.tsx
```diff
@@ -1,12 +1,12 @@
 import * as React from 'react'
 
-import { DataTableBody } from './DataTableBody'
-import { DataTableDataCell } from './DataTableDataCell'
+import { DataTableBody } from './DataTable.Body'
```
LimeWub, line 3: 👀 New naming pattern why? Has this been discussed somewhere?

hollg: Tom suggested yesterday that it might make the file names easier to read so I thought I'd try it out. What do you think? Breaking up the names does make things a bit easier to read imo but also makes things a bit confusing because I'm used to anything after a `.` to tell me about the _type_ of file (e.g. `.test`). So it feels like we'd now have `.` meaning two possible things in file names, if that makes sense.

lib/src/components/data-table/pagination/Pagination.tsx
hollg: Good catch! That name was to differentiate it from the `pageSize` that comes from the context provider, but you're absolutely right: "default" doesn't actually describe the relevant concept here. I'll try renaming the one that comes from context instead.

lib/src/components/data-table/DataTable.Search.tsx
```diff
@@ -0,0 +1,47 @@
+import * as React from 'react'
+import { debounce } from 'throttle-debounce'
+
+import { SearchInput } from '../search-input'
+import { useDataTable } from './DataTableContext'
+import { Label } from '../label'
+
+type DataTableSearchProps = React.ComponentProps<typeof SearchInput> & {
+  label: string
+  hideLabel?: boolean
```
hollg, line 10: I **hate** this prop but I couldn't think of a better way to both force the consumer to give the searchbox a label _and_ support it being hidden. If anyone sees one, please share 🙏🏻

lib/src/components/data-table/DataTable.Header.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { Table } from '../table'
+import { Icon } from '../icon'
+import { SortDown, SortUp } from '@atom-learning/icons'
+
+import { useDataTable } from './DataTableContext'
+type DataTableHeaderProps = React.ComponentProps<typeof Table.HeaderCell> & {
+  header: Header<Record<string, unknown>, unknown>
+}
+
+const sortIcons = {
```
LimeWub, line 14: superMinor but this could also just be done as a  function 

```
const renderSortIcon = (sortDirection<'asc' | 'desc'>) => { 
  switch (sortDirection) {
   case 'asc':
   return //ASC ICON CODE//
   case 'desc'
   return //DESC ICON CODE//
  }
}
```

Might be slliiiiiightly easier to follow although what u have now is fine too - I just haven't seen us using objects to render optionally like this b4.

lib/src/components/data-table/DataTable.mdx
```diff
@@ -0,0 +1,160 @@
...
+}>()
+
+const columns = [
+  columnHelper.accessor('name', {
+    cell: (info) => info.getValue()
+  }),
+  columnHelper.accessor('hobby', {
+    cell: (info) => info.getValue()
+  }),
+  // Columns created with columnHelper.display won't get be sortable.
```
LimeWub, line 41: typo :o 
> won't get be sortable

lib/src/components/data-table/pagination/Pagination.tsx
```diff
@@ -0,0 +1,78 @@
...
+  alignItems: 'center'
+})
+
+type PaginationProps = React.ComponentProps<typeof StyledNav> & {
+  pageSize?: number
+}
+
+/** Applies pagination to parent DataTableProvider and renders UI to switch pages etc */
+export const Pagination: React.FC<PaginationProps> = ({
+  css,
```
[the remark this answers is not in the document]
hollg: Thanks!

lib/src/components/data-table/DataTable.tsx
```diff
@@ -0,0 +1,89 @@
...
+ */
+export const DataTable: TDataTable = (props) => <DataTableProvider {...props} />
+
+DataTable.Body = DataTableBody
+DataTable.DataCell = DataTableDataCell
+DataTable.Head = DataTableHead
+DataTable.HeaderCell = DataTableHeaderCell
+DataTable.Pagination = Pagination
+DataTable.Row = DataTableRow
+DataTable.GloabalFilter = DataTableGlobalFilter
```
LimeWub, line 88: Typo `GloabalFilter`

lib/src/components/data-table/DataTableGlobalFilter.tsx
```diff
@@ -0,0 +1,47 @@
+import * as React from 'react'
```
LimeWub, line 1: Can we approve/merge SearchInputField/Field VisibilityHidden PR and use that, for vis hidden? ;)

hollg: Happy to change this to use visually hidden, but the search isn't a `Form` and I don't think it ought to be, which counts us out of using anything with a `FieldWrapper`.

hollg: I've made a little util to conditionally wrap something in `VisuallyHidden` and used it here now. Pls may I merge? 🙏🏻

PR #388, opened 2022-08-30
feat/added new appbar component 

elliotlaws:
Ticket https://atomlearningltd.atlassian.net/browse/DS-50
Design https://www.figma.com/file/KjgQh7eTwEfJmH6e56k9mX/DS-Documentation?node-id=6892%3A45237

### Implementation

The component has been created to allow consumers to decide what elements are contained within it. Components  include `AppBar.Brand` and `AppBar.ActionIcon`, and provide some standard UI that may be required across all apps. However, `AppBar` is by no way limited to this group of components.  

- Added `AppBar` compound component
- Replaced hardcoded zIndices with constants 

### Testing locally
To test locally, copy and paste the contents of this [text file](https://github.com/Atom-Learning/components/files/9464775/sandbox-appbar.txt) into lib/sandbox/index.tsx to generate the same UI as shown in the screenshots below.

### Screenshots
image

#### On scroll - box shadow applied
image

#### Adding additional components 
image


lib/src/components/app-bar/Actions.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { EllypsisVertical } from '@atom-learning/icons'
+
+import { styled, CSS } from '~/stitches'
+
+type ActionsOverflowMenuProps = Omit<
+  React.ComponentProps<typeof BaseActionIcon>,
+  'children'
+>
+
+const StyledActionIcon = styled(BaseActionIcon, {
```
LimeWub, line 15: Is this a different theme for the ActionIcon from the ones already supported?

lib/src/components/app-bar/Actions.tsx
```diff
@@ -0,0 +1,62 @@
...
+  children,
+  label = 'Click to expand'
+}) => {
+  return (
+    <Flex css={{ alignItems: 'center', mr: '-$2' }}>
+      <DropdownMenu>
+        <DropdownMenu.Trigger asChild>
+          <ActionIcon icon={EllypsisVertical} label={label} />
+        </DropdownMenu.Trigger>
+        <DropdownMenu.Content sideOffset={16}>{children}</DropdownMenu.Content>
```
LimeWub, line 36: sideOffset? What's this?

lib/src/components/app-bar/Actions.tsx
```diff
@@ -0,0 +1,62 @@
...
+  },
+  '& > svg': { color: '$tonal400' }
+})
+
+const ActionsOverflowMenu: React.FC<ActionsOverflowMenuProps> = ({
+  children,
+  label = 'Click to expand'
+}) => {
+  return (
+    <Flex css={{ alignItems: 'center', mr: '-$2' }}>
```
LimeWub, line 31: Why negative margin here?

lib/src/components/app-bar/Actions.tsx
```diff
@@ -0,0 +1,62 @@
...
+  )
+}
+
+type ActionIconProps = Omit<
+  React.ComponentProps<typeof StyledActionIcon>,
+  'size' | 'children'
+> & { icon: React.FC<React.SVGProps<SVGSVGElement>>; label: string }
+
+const ActionIcon: React.ForwardRefExoticComponent<ActionIconProps> =
+  React.forwardRef(({ icon, label, ...rest }, forwardedRef) => {
```
LimeWub, line 54: We tend to just use nesting to add Icons to ActionIcons (ie. lit how you've done it in the implementation here haha). For consistency - I think this should just be rendering 'children' rather than the specific Icon component.

LimeWub: You also don't need to explicitly destructure `label` since this is in the `rest` for the ActionIcon.

Basically, what this is doing imo, is just explicitly asking the size of the icon to be 'large' - am I correct?

lib/src/components/app-bar/AppBar.mdx
```diff
@@ -0,0 +1,103 @@
...
+</AppBar>
+```
+
+By default, `AppBar` uses a flex layout with flex items placed immediatley after each other. To align flex items, an optional css prop with `justifyContent` can be passed.
+
+Alternatively, if you want to avoid wrapping groups of items with a `div` purely for alignment purposes, you could use a spacer.
+With the following code, our collapsible trigger (hamburger) and brand will be aligned to the left and our actions aligned to the right.
+
+```tsx preview
+const Spacer = styled('div', {
```
LimeWub, line 46: Perhaps Spacer should be a component of the AppBar ie. AppBar.Spacer - if you think this is going to be useful/required to space things out ^^

lib/src/components/app-bar/Brand.tsx
```diff
@@ -0,0 +1,57 @@
...
+        color: '$tonal400',
+        '& > p': {
+          color: '$tonal400'
+        },
+        '&:hover, &:focus': { textDecoration: 'none' }
+      }}
+    >
+      <Image
+        src={Sources[theme]}
+        alt="An Atom Learning user riding a pencil rocket"
```
LimeWub, line 43: o_o what's this Alt?
Is this what the image actually shows?

I'd be tempted to have an empty `alt` for the logo or just the text of where it actually takes you (dashboard?) we can perhaps have this as a prop so it's also translatable. 

Currently a user would tab to the link and a text reader would say "An Atom Learning user riding a pencil rocket" - which is probably very confusing haha

lib/src/components/app-bar/Actions.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+import { styled, CSS } from '~/stitches'
+
+type ActionsOverflowMenuProps = Omit<
+  React.ComponentProps<typeof BaseActionIcon>,
+  'children'
+>
+
+const StyledActionIcon = styled(BaseActionIcon, {
+  borderRadius: '4px',
```
kerryambt, line 16: can we use DS token here? i.e. `$0`

lib/src/components/app-bar/Actions.tsx
```diff
@@ -0,0 +1,62 @@
...
+    <Flex css={{ gap: '$1', alignItems: 'center', ...css }}>{children}</Flex>
+  )
+}
+
+type ActionIconProps = Omit<
+  React.ComponentProps<typeof StyledActionIcon>,
+  'size' | 'children'
+> & { icon: React.FC<React.SVGProps<SVGSVGElement>>; label: string }
+
+const ActionIcon: React.ForwardRefExoticComponent<ActionIconProps> =
```
kerryambt, line 53: I think it's unclear to name this the same as another component we have in the DS, even if you've overridden that import I think we should avoid this. Can we call this `AppBarActionIcon`?

kerryambt: also am I reading right that this just removes the composition from ActionIcon to make it by one component? If so, is that something we should do?

lib/src/components/app-bar/Actions.tsx
```diff
@@ -0,0 +1,62 @@
...
+  React.forwardRef(({ icon, label, ...rest }, forwardedRef) => {
+    return (
+      <StyledActionIcon label={label} size="lg" ref={forwardedRef} {...rest}>
+        <Icon is={icon} />
+      </StyledActionIcon>
+    )
+  })
+
+export { Actions, ActionIcon, ActionsOverflowMenu }
```
kerryambt, line 62: is there a reason to not prefix each of these const declarations with `export` and remove the need for this line?

lib/src/components/app-bar/AppBar.mdx
```diff
@@ -0,0 +1,103 @@
+---
+title: AppBar
+component: AppBar,AppBar.Collapsible,AppBar.CollapsibleTrigger,AppBar.CollapsibleContent,AppBar.Brand,AppBar.Actions,AppBar.ActionIcon,AppBar.ActionsOverflowMenu
+description: The top App bar provides content and actions related to the current screen. It's used for branding, screen titles, navigation, and actions.
+category: Surfaces
+---
+
+`AppBar` exports many components that combine to form the elements that you would expect to commonly see in a AppBar. Although the App bar is not limited to these components.
```
kerryambt, line 8: minor but I think you should start the second sentence with `However,` rather than `Although`, and potentially expand on the point to give examples of what else could be used.

lib/src/components/app-bar/AppBar.mdx
kerryambt, line 46: these tsx blocks don't let you define consts like this, so you'll have to hard code this onto the div

LimeWub (commented):
Reviewed halfway through but need to refocus on something for release - will continue reviewing later but submitting for now.

lib/src/components/app-bar/AppBar.tsx
```diff
@@ -0,0 +1,93 @@
...
+  CollapsibleContent: typeof CollapsibleContent
+  Brand: typeof Brand
+  Actions: typeof Actions
+  ActionIcon: typeof ActionIcon
+  ActionsOverflowMenu: typeof ActionsOverflowMenu
+  Divider: typeof Divider
+}
+
+const Divider = () => (
+  <BaseDivider
```
kerryambt, line 28: again not a fan of overriding the component names here. `AppBarDivider`?

lib/src/components/app-bar/AppBar.tsx
```diff
@@ -0,0 +1,93 @@
...
+    orientation="vertical"
+    css={{ minHeight: 28, background: '$tonal100' }}
```
kerryambt, line 30: background -> bg

lib/src/components/app-bar/AppBar.tsx
```diff
@@ -0,0 +1,93 @@
...
+  display: 'flex',
+  alignItems: 'center',
+  width: '100vw',
+  top: '0',
+  height: '$6',
+  zIndex: 1,
+  borderBottom: '1px solid $tonal100',
+  transition: 'box-shadow .2s ease-out',
+  variants: {
+    scrolled: {
```
kerryambt, line 46: should we name this `hasScrolled` for consistency?

lib/src/components/app-bar/AppBar.tsx
```diff
@@ -0,0 +1,93 @@
...
+}
+
+AppBar.Collapsible = Collapsible
+AppBar.CollapsibleTrigger = CollapsibleTrigger
+AppBar.CollapsibleContent = CollapsibleContent
+AppBar.Brand = Brand
+AppBar.Actions = Actions
+AppBar.ActionIcon = ActionIcon
+AppBar.ActionsOverflowMenu = ActionsOverflowMenu
+AppBar.Divider = Divider
```
kerryambt, line 93: can we also add `AppBar.displayName = 'AppBar'` to the end of this file?

lib/src/components/app-bar/Brand.tsx
```diff
@@ -0,0 +1,57 @@
...
+  multi: logo
+} as const
+
+type BrandProps = {
+  href: string
+  theme?: keyof typeof Sources
+  css?: CSS
+}
+
+export const Brand: React.FC<BrandProps> = ({
```
kerryambt, line 23: I think we should maybe have more specific names for these types of components, in case we ever want a brand component that isn't tied to app bar. AppBarBrand makes more sense I think

lib/src/components/app-bar/Brand.tsx
```diff
@@ -0,0 +1,57 @@
...
+        '& > p': {
+          color: '$tonal400'
+        },
+        '&:hover, &:focus': { textDecoration: 'none' }
+      }}
+    >
+      <Image
+        src={Sources[theme]}
+        alt="An Atom Learning user riding a pencil rocket"
+        height={24}
```
kerryambt, line 44: why are we setting this prop as well as using css? do we need this?

lib/src/components/app-bar/Collapsible.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+const StyledTrigger = styled(CollapsiblePrimitive.Trigger, {
+  mr: '$2',
+  ml: '-$2',
+  p: 0,
+  color: '$tonal400',
+  display: 'flex',
+  alignItems: 'center',
+  justifyContent: 'center',
+  background: 'none',
```
kerryambt, line 45: bg :)

lib/src/components/app-bar/Collapsible.tsx
```diff
@@ -0,0 +1,136 @@
...
+  border: 'none',
+  cursor: 'pointer',
+  height: '$4',
+  width: '$4'
```
kerryambt, line 49: you can shorten this to `size: '$4'`

lib/src/components/app-bar/utils.ts
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+import { throttle } from 'throttle-debounce'
+
+export const useScrollPosition = (): { x: number; y: number } => {
```
kerryambt, line 4: this should probably live in the higher level utils folder rather than inside the app bar component

lib/src/components/app-bar/Actions.tsx
elliotlaws: Happy to rename the component to AppBarActionIcon.

The idea behind creating a separate component, in this case, is so we can set the size prop to "lg"

I'm also a fan of just passing in an icon as prop as opposed to relying upon the consumer to always pass in an icon as children. It constrains the component to what we need it to be. Open to other suggestions though

lib/src/components/app-bar/Actions.tsx
elliotlaws: I think it's easier to see at a glance which components are exported, instead of having to go to the individual declarations.

lib/src/components/app-bar/AppBar.mdx
elliotlaws: Good point, I'll update that

lib/src/components/app-bar/AppBar.mdx
elliotlaws: That's a shame 😭

lib/src/components/app-bar/Collapsible.tsx
elliotlaws: Nice!

lib/src/components/app-bar/Actions.tsx
elliotlaws: I've added some additional styles as per Miguels designs

lib/src/components/app-bar/Actions.tsx
elliotlaws: The distance in pixels from the dropdown trigger

lib/src/components/app-bar/Actions.tsx
elliotlaws: See Miguel's designs, it's to align the icon with the edge of the main content.

lib/src/components/app-bar/Actions.tsx
elliotlaws: Regarding the first comment - I actually think it makes the API simpler - the consumer doesn't have to worry about importing an icon and using that as children because, in this case, we're constraining the component to not accept children and only an icon (which the base ActionIcon only allows anyway). Any thoughts?

I'll remove the explicit assignment of `label` and just spread the props :) 

Yeah, pretty much, all this is doing is assigning a set size and making the component API simpler so the consumer doesn't have to import and use an icon component every time like this.

``` tsx
<AppBar.ActionIcon icon={Settings} label="Settings" />
```

instead of having to do this 
``` tsx
<AppBar.ActionIcon label="Settings">
  <Icon as={Settings}/>
</AppBar.ActionIcon>
```

I'm not sure what the benefits of doing the latter are, in this particular case?

lib/src/components/app-bar/AppBar.mdx
elliotlaws: Yeah I agree, I've actually created a ticket in the DS backlog to create this as a separate component because I think it will be useful when using flex layouts

https://atomlearningltd.atlassian.net/browse/DS-120

lib/src/components/app-bar/Brand.tsx
elliotlaws: Not sure haha, I'll add it as a prop so the consumer can change this

lib/src/components/app-bar/Brand.tsx
elliotlaws: Nope, I'll remove it!

lib/src/components/app-bar/Brand.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export const AppBarBrand: React.FC<AppBarBrandProps> = ({
+  children,
+  theme = 'multi',
+  href,
+  css,
+  logoAlt = 'Atom Learning logo'
+}) => {
+  return (
+    <Link
```
hollg, line 32: I'm not 100% sure about using the `Link` component here, just to override its styling. Perhaps this should be a `styled('a', { ... })` with the required styling provided directly? I don't feel super strongly about it though so could definitely be convinced otherwise.

lib/src/components/app-bar/Brand.tsx
```diff
@@ -0,0 +1,60 @@
...
+        ...css
+      }}
+    >
+      <Image
+        src={Sources[theme]}
+        alt={logoAlt}
+        css={{
+          mr: '$3',
+          mb: 5,
+          '&[src$=".svg"]': {
```
hollg, line 51: What does the `$` in this syntax do? 👀

lib/src/components/app-bar/Collapsible.tsx
```diff
@@ -0,0 +1,135 @@
...
+  left: 0,
+  bottom: 0,
+  right: 0,
+  zIndex: MAX_Z_INDEX,
+  position: 'fixed',
+  '&[data-state="open"]': { animation: `${open} 0.15s linear` },
+  '&[data-state="closed"]': { animation: `${close} 0.15s linear` }
+})
+
+const StyledRoot = styled(CollapsiblePrimitive.Root, {})
```
hollg, line 35: Can this be skipped, if it doesn't add any styling?

hollg: Oh, maybe it's required to allow the consumer to pass CSS in

lib/src/utilities/hooks/useScrollPosition.ts
```diff
@@ -0,0 +1,21 @@
+import { useEffect, useState } from 'react'
+import { throttle } from 'throttle-debounce'
+
+export const useScrollPosition = (): { x: number; y: number } => {
```
hollg, line 4: I have a vague feeling that we already had a util for this, or at least had this logic written inside another component. Or maybe that was in `atom-core` 🤔

lib/src/components/app-bar/Actions.tsx
hollg: On one hand, I'm glad to see people are sceptical about having an icon as a prop because it's definitely the sort of thing we've run into trouble with in the past, but on the other hand, I think I agree with @elliotlaws that in this case I can't really see any specific problems with it.

Specifically it has caused issues in the past when we come back later and decide we want to _do something different_ with the icon. Now you need a second prop to describe where the icon should be positioned, or to change its line weight etc. Or two, to do both! But in this case I think we can basically be 100% confident that we're never going to want to support `children` _other than_ an `Icon`, which is perhaps the relevant distinction?

lib/src/components/app-bar/Actions.tsx
hollg: I'd be inclined to say we should stick with `export const ...` for consistency throughout the codebase at least for now, and then maybe take a decision about whether to switch _everything_ over to this style if there's a lot of interest in the team.

elliotlaws: Makes sense, happy to change for consistency

lib/src/components/app-bar/Actions.tsx
elliotlaws: Yeah definitely, I understand the need for avoiding this pattern in most cases, but in this case, I think I will leave it as is for now. 

I don't see any benefit to having the consumer specifically render an icon as children - the extra flexibility is redundant when we have a set way we want the `ActionIcon` to look

lib/src/components/app-bar/Brand.tsx
elliotlaws: Happy to change this

lib/src/components/app-bar/Brand.tsx
elliotlaws: I stole this from the atom-core I believe, but Google led me to this...

Targets the SVG element where its source value ends with `.svg`

lib/src/utilities/hooks/useScrollPosition.ts
elliotlaws: I stoke it from here

https://github.com/Atom-Learning/marketing-site/blob/develop/frontend/utils/hooks/use-scroll-position.ts

lib/src/components/app-bar/Actions.tsx
```diff
@@ -0,0 +1,62 @@
...
+const AppBarActionIcon: React.ForwardRefExoticComponent<AppBarActionIconProps> =
+  React.forwardRef(({ icon, ...rest }, forwardedRef) => {
+    return (
+      <StyledActionIcon size="lg" ref={forwardedRef} {...rest}>
+        <Icon is={icon} />
+      </StyledActionIcon>
+    )
+  })
+
+export { Actions, AppBarActionIcon, ActionsOverflowMenu }
```
kerryambt, line 62: can we update this like the others please?

lib/src/components/app-bar/Actions.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+type AppBarActionIconProps = Omit<
+  React.ComponentProps<typeof StyledActionIcon>,
+  'size' | 'children'
+> & { icon: React.FC<React.SVGProps<SVGSVGElement>>; label: string }
+
+const AppBarActionIcon: React.ForwardRefExoticComponent<AppBarActionIconProps> =
+  React.forwardRef(({ icon, ...rest }, forwardedRef) => {
+    return (
+      <StyledActionIcon size="lg" ref={forwardedRef} {...rest}>
```
kerryambt, line 56: FYI action icon sizes are going to change tomorrow, will be communicating in DS/FEG channels later but FYI this needs to be updated after that

lib/src/components/app-bar/AppBar.tsx
```diff
@@ -0,0 +1,95 @@
+import React from 'react'
+
+import { CSS, styled } from '~/stitches'
+import { useScrollPosition } from '~/utilities/hooks/useScrollPosition'
+
+import { Divider as BaseDivider } from '../divider'
```
kerryambt, line 6: Do we need this `x as y` given we're not overriding the name below now?

lib/src/components/app-bar/Actions.tsx
```diff
@@ -0,0 +1,62 @@
+import React from 'react'
+import { ActionIcon as BaseActionIcon } from '../action-icon'
```
kerryambt, line 2: again I think we can remove this `x as y`

lib/src/components/app-bar/Actions.tsx
elliotlaws: Updated this to "md" and added a slighter larger gap between icons as they looked slightly cramped after size change

lib/src/components/app-bar/Brand.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+const Sources = {
+  primary: logoPrimary,
+  white: logoLight,
+  multi: logo
+} as const
+
+type AppBarBrandProps = {
+  href: string
+  theme?: keyof typeof Sources
```
hollg, line 18: nice!

hollg (approved):

kerryambt (approved):

lib/src/components/app-bar/Actions.tsx
LimeWub: Doesn't that imply that the `Actions` are always going to be at the end of the toolbar? I thought this was supposed to be super-flexible so wasn't sure we wanted to be assuming that. Can u confirm?

elliotlaws: This did cross my mind too and I raised it with @hollg, but he seemed to think it was reasonable to expect the Actions to always be at the end.

I'm on the fence about this, any thoughts from you guys?

hollg: Yeah, Miguel said that if there are actions then they always go on the end. If we're really concerned about at least giving the consumer the ability to override the negative `margin-right` we could always pass a `css` prop down to this `Flex`?

lib/src/components/app-bar/Actions.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { DropdownMenu } from '../dropdown-menu'
+import { Flex } from '../flex'
+import { Icon } from '../icon'
+
+type ActionsOverflowMenuProps = Omit<
+  React.ComponentProps<typeof ActionIcon>,
+  'children'
+>
+
+const StyledActionIcon = styled(ActionIcon, {
```
LimeWub, line 16: Did you try to add this ActionIcon style in as a variant of the ActionIcon rather than the overrides here? Only `???` would be how to do the `&[data-state="open"]` style - but I'd debate you could just add that bit to the ActionIcon styles if we plan on using them for any collapsible triggers.

borderRadius: 0 should be covered via the 'isRounded' prop btw.

lib/src/components/app-bar/Actions.tsx
elliotlaws: That sounds good to me, any further thoughts from you @LimeWub before I make this change?

LimeWub: I think it would be good to confirm with Miguel tbf because that's where I got the idea that they were flexible placement.

If they are not flexible then this is fine 👍 
If they are - we probs need some type of 'smart' negative margin for non-background ActionIcons or something.

lib/src/components/app-bar/Collapsible.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+const StyledTrigger = styled(CollapsiblePrimitive.Trigger, {
+  mr: '$2',
+  ml: '-$2',
```
LimeWub, line 39: It's the same w the negative margin here (assumes left-most placement for Collapsible)

lib/src/components/app-bar/Collapsible.tsx
```diff
@@ -0,0 +1,133 @@
...
+const CollaspibleContext = React.createContext<
+  CollapsibleContextValue | undefined
+>(undefined)
+
+export const useCollapsibleContext = () => {
+  const context = React.useContext(CollaspibleContext)
+
+  if (context === undefined) {
+    throw new Error(
+      'useMobileMenuContext must be used within a MobileMenuContextProvider'
```
LimeWub, line 65: `useMobileMenuContext` leftover naming?

lib/src/components/app-bar/Collapsible.tsx
```diff
@@ -0,0 +1,133 @@
...
+  return context
+}
+
+export const Collapsible: React.FC<React.ComponentProps<typeof StyledRoot>> = ({
+  children,
+  css,
+  ...props
+}) => {
+  const [isOpen, setIsOpen] = React.useState(false)
+  const onOpenToggle = () => setIsOpen(!isOpen)
```
LimeWub, line 78: can this be called `toggleIsOpen` - reading it I thought it was an `onOpen` callback function

lib/src/components/app-bar/Collapsible.tsx
```diff
@@ -0,0 +1,133 @@
...
+> = ({ children, css }) => {
+  const { isOpen, onOpenToggle } = useCollapsibleContext()
+
+  return (
+    <RemoveScroll enabled={isOpen} forwardProps>
+      <StyledContent>
+        <Flex css={{ flexDirection: 'column', px: '$4', py: '$5', ...css }}>
+          {children}
+        </Flex>
+        <Box
```
LimeWub, line 123: Is this a 'close' button? Or just an overlay?
Are we supporting other escape mechanisms? (ie. Esc key?)

lib/src/components/app-bar/Collapsible.tsx
```diff
@@ -0,0 +1,133 @@
...
+  from: { height: 0, opacity: 0 },
+  to: { height: 'var(--radix-collapsible-content-height)', opacity: 1 }
+})
+
+const close = keyframes({
+  from: { height: 'var(--radix-collapsible-content-height)', opacity: 1 },
+  to: { height: 0, opacity: 0 }
+})
+
+const StyledContent = styled(CollapsiblePrimitive.Content, {
```
LimeWub, line 20: Are the `Collapsible.Content` styles final? Could you please show me a screenshot of the design for them? 
They aren't what I would expect but perhaps I'm thinking of the wrong component (?)

lib/src/components/app-bar/Collapsible.tsx
```diff
@@ -0,0 +1,133 @@
...
+  )
+}
+
+export const CollapsibleTrigger: React.FC<
+  React.ComponentProps<typeof StyledTrigger>
+> = ({ children, css, ...props }) => {
+  const { isOpen } = useCollapsibleContext()
+
+  return (
+    <StyledTrigger css={css} aria-label="toggle mobile menu" {...props}>
```
LimeWub, line 105: `aria-label="toggle mobile menu"` Specific label for something that's kinda generic.

I'm supposing this should fall to implementation to add an appropriate/translatable label. Perhaps label should be a prop?

LimeWub: U also don't need to expand css if you have {...rest} expanded (v.minor - but can we use {...rest} for things that aren't the full props ? :o )

lib/src/components/app-bar/Collapsible.tsx
```diff
@@ -0,0 +1,133 @@
...
+interface CollapsibleContextValue {
+  isOpen: boolean
+  onOpenToggle: () => void
+}
+
+const CollaspibleContext = React.createContext<
+  CollapsibleContextValue | undefined
+>(undefined)
+
+export const useCollapsibleContext = () => {
```
LimeWub, line 60: Is this a context which is not relevant to the radix components? (just an empty context for `isOpen` to sit in?)

lib/src/components/app-bar/Collapsible.tsx
```diff
@@ -0,0 +1,133 @@
+import React from 'react'
+import * as CollapsiblePrimitive from '@radix-ui/react-collapsible'
```
LimeWub, line 2: Also curious: Did you consider using Dialog or NavigationMenu for this rather than Collapsible? Just saying because it seems you want this to be doing Dialog like things (trapping focus -which i don't think is happening rn-, not scrolling -which u've used a 3rd party for-, having a clickable to esc overlay -I think you've done this w a box?-) and you are kinda coding some of these for Collapsible - but Dialog or NavigationMenu may already had been doing these? :o 

NavigationMenu would probs work best for the full 'AppBar'/'TopBar' context but there are triggers/Links/Subcontent.

Just wondering if this was considered :p

lib/src/components/app-bar/Actions.tsx
```diff
@@ -0,0 +1,61 @@
+import { EllypsisVertical } from '@atom-learning/icons'
```
LimeWub, line 1: General Naming/Minor: Could these components include 'AppBar' in their file name (as they already do in the actual components)?

AppBarBrand, AppBarActionIcon, generally `AppBar${ComponentName}`

LimeWub, line 1: Why is this component named Actions? is Actions just the spacing part? Shouldn't the spacing be separate anyway to include all other types of children components of the AppBar; then this component can just be any necessary ActionIcon overrides?


I kind of don't even think it's necessary to do anything for the Dropdown Menu as it seems u are just using that out of the box (but if we do need specific things for it then that could also be a separate thing?)

LimeWub:
Hey - just did a proper pass through this in terms of logically what this is trying to achieve. 

I started by reviewing the actual code so some of the comments are code specific - and in terms of code I think it's fine but in terms of logically what we're doing and how we are going to be using this - I'm not sure about how some things have been separated and why some radix tools have been used over others so I've left some comments (which will also help me understand this better ^^).

Sorry for going off and probs spamming you. Happy to discuss any of these on a call :)

lib/src/components/top-bar/TopBarActionIcon.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react'
+
+import { styled } from '~/stitches'
+
+import { ActionIcon } from '../action-icon'
+import { Icon } from '../icon'
+
+const StyledActionIcon = styled(ActionIcon, {
+  borderRadius: '$0',
+  '&[data-state="open"]': {
```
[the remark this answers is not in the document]
elliotlaws: I'll remove both of them

lib/src/components/top-bar/TopBar.tsx
```diff
@@ -0,0 +1,77 @@
...
+      }
+    }
+  }
+})
+
+export const TopBar: React.FC<{ css?: CSS }> & TopBarSubComponents = ({
+  children,
+  css
+}) => {
+  const { y: hasScrolled } = useScrollPosition()
```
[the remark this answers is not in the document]
elliotlaws: I like this, have changed

lib/src/components/top-bar/TopBarBrand.tsx
```diff
@@ -0,0 +1,59 @@
+import logo from '@atom-learning/theme/lib/assets/logo.svg'
+import logoLight from '@atom-learning/theme/lib/assets/logo-light.svg'
+import logoPrimary from '@atom-learning/theme/lib/assets/logo-primary.svg'
+import React from 'react'
+
+import { CSS, styled } from '~/stitches'
+
+import { Image } from '../image'
+
+const Sources = {
```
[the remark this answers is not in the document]
elliotlaws: It's capitalised because it's being used as a type in AppBarBrandProps. Very similar to how you would use an enum 

_Taken from TypeScript docs_ 
```ts
export const Colors = {
  red: "RED",
  blue: "BLUE",
  green: "GREEN",
} as const;
// or use an 'export default'
export default {
  red: "RED",
  blue: "BLUE",
  green: "GREEN",
} as const;
```

lib/src/components/top-bar/TopBar.test.tsx
```diff
@@ -0,0 +1,34 @@
+import { Search, SwitchOff } from '@atom-learning/icons'
+import { IdProvider } from '@radix-ui/react-id'
+import { render } from '@testing-library/react'
+import { axe } from 'jest-axe'
+import * as React from 'react'
+
+import { TopBar } from '.'
+
+const ExampleTopBar = () => (
+  <IdProvider>
```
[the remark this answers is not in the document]
elliotlaws: If I'm being honest I'm not entirely sure, it's in a lot of other component tests. 
it's a Radix thing https://www.radix-ui.com/docs/primitives/utilities/id-provider

Although it seems redundant to use it in a testing environment because there's no concept of client/server and hydration. 

I've removed it

LimeWub (approved):

lib/src/components/top-bar/TopBarActionIcon.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+
+import { styled } from '~/stitches'
+
+import { ActionIcon } from '../action-icon'
+import { Icon } from '../icon'
+
+const StyledActionIcon = styled(ActionIcon, {
+  '&:hover': {
```
thomasdigby, line 9: What's the reason for overriding our `ActionIcon` hover styles? Do we not have a matching variant/theme that we can use? @MiguelGDLR

lib/src/components/top-bar/TopBarActionIcon.tsx
```diff
@@ -0,0 +1,27 @@
...
+type TopBarActionIconProps = Omit<
+  React.ComponentProps<typeof StyledActionIcon>,
+  'size' | 'children'
+> & { icon: React.FC<React.SVGProps<SVGSVGElement>>; label: string }
+
+export const TopBarActionIcon: React.ForwardRefExoticComponent<TopBarActionIconProps> =
+  React.forwardRef(({ icon, ...rest }, forwardedRef) => {
+    return (
+      <StyledActionIcon size="md" ref={forwardedRef} {...rest}>
+        <Icon is={icon} />
```
thomasdigby, line 24: I'm torn between this being an inconsistent API with our other components and the alternative being overly verbose. We've avoided this convention in the past `<TopBar.ActionIcon icon={Search} />` and used component composition instead `<TopBar.ActionIcon><Icon is={Search} /></TopBar.ActionIcon>`.

It makes this component slightly redundant though if all we're changing are the hover styles, and I like that we're namespacing it against the top bar. Not sure what to suggest, sorry.

lib/src/components/top-bar/TopBarBrand.tsx
```diff
@@ -0,0 +1,59 @@
...
+  css?: CSS
+  logoAlt?: string
+}
+
+const StyledLink = styled('a', {
+  display: 'flex',
+  alignItems: 'center',
+  textDecoration: 'none',
+  color: '$tonal400',
+  '& > p': {
```
thomasdigby, line 28: This seems to require a specific element passed in as a child, could we avoid this by being explicit about the component used to render the brand name? e.g.
```jsx
<TopBar.Brand>
  <TopBar.BrandName>Atom Learning</TopBar.BrandName>
</TopBar.Brand>
```

Then we can be consistent with the style that the name is rendered with and avoid side-effect styles like this.

lib/src/components/top-bar/TopBar.tsx
```diff
@@ -0,0 +1,71 @@
...
+        css={{
+          alignItems: 'center',
+          height: '$4',
+          mx: '$4',
+          width: '100%',
+          '@md': {
+            mx: '$5'
+          },
+          ...css
+        }}
```
thomasdigby, line 59: Is there any reason we didn't use `styled` for this component instead of inline `css`?

lib/src/components/top-bar/TopBar.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+Alternatively, if you want to avoid wrapping groups of items with a `div` purely for alignment purposes, you could use a spacer. In this example, the styling of the `Flex` component has been extended so that it stretches to fill the available space.
+
+```tsx preview
+<TopBar>
+  <TopBar.Brand href="atomlearning.co.uk">
+    <Text>Admin Panel</Text>
+  </TopBar.Brand>
+  <Flex css={{ flex: 1, justifySelf: 'stretch', alignSelf: 'stretch' }} />
+  <Stack gap="2" direction="row">
```
thomasdigby, line 20: Could we create a specific `TopBar.Actions` component to encapsulate some style decisions around alignment? Apologies if you've already discussed this with others, I'm just imagining sticking the `NavigationMenu` in there and trying to understand how the other components would be positioned relative to it

lib/src/components/top-bar/TopBarActionIcon.tsx
MiguelGDLR: We already have discussed about it, it would be great if someone can share some light on [ this ticket ](https://atomlearningltd.atlassian.net/browse/DS-128?atlOrigin=eyJpIjoiODdjYmI4ZmViODIwNDMwZmI3ODY4MDA3NzI5ZDE3MjMiLCJwIjoiaiJ9) so that we can make a decision for this. I don't care so much about the hover effect (I suggested we do that to the Action icons) but I do care about the icon colour, we can't have blue icons in the topbar.

thomasdigby: That feels like a pre-requisite to this work, we probably should have decided what to do with the Action Icon variants before starting this. For now I'd probably suggest using `appearance="simple"` and `theme="neutral"`. Would that be ok?

MiguelGDLR: Yes that's what I suggested, but according to devs that appearance doesn't exist in code. Only appears as an example in the docs but not even in the props table.

thomasdigby: Happy to chat in person @MiguelGDLR if it's easier, I can dedicate some time to it as well so we unblock everything

thomasdigby: @MiguelGDLR `appearance="simple"` is the default, I wasn't suggesting to use `subtle`

thomasdigby: For what it's worth, I think we can probably update `appearance="simple"` to use a light block colour on hover just like @elliotlaws has added here. We already have a `neutral` theme to render dark grey instead of blue, so it's just updating the hover/focus styles.

elliotlaws: Am I right in thinking that we agreed to leave the hover styles in `TopBarActionIcon` until `ActionIcon` could be updated, is that correct @MiguelGDLR ?

thomasdigby: Sorry, I forgot to message yesterday. @MiguelGDLR and I quickly discussed it and decided that `appearance="simple"` and `theme="neutral"` with no custom hover styles will be fine for the nav 👍

lib/src/components/top-bar/TopBarActionIcon.tsx
elliotlaws: I was torn too 🤔 but ended up going with passing in the icon after seeing Chakra UI do it https://chakra-ui.com/docs/components/icon-button 

I like how the consumer doesn't have to worry about specifically passing an `Icon` as children (if that's the only thing that can be passed in as children anyway).

lib/src/components/top-bar/TopBar.mdx
elliotlaws: I originally had `TopBar.Actions` but removed it from the API after speaking with the internal tools folks. 

We agreed that `TopBar.Actions` didn't add anything different to what could already be achieved using our existing `Stack` component.

lib/src/components/top-bar/TopBar.tsx
elliotlaws: No particular reason, other than being lazy to try and name it something meaningful 😂

lib/src/components/top-bar/TopBarBrand.tsx
elliotlaws: I agree, I'll see what I can do to make this more explicit

lib/src/components/top-bar/TopBar.mdx
```diff
@@ -0,0 +1,59 @@
...
+    <TopBar.ActionIcon icon={SwitchOff} label="Light/Dark mode" />
+  </Stack>
+</TopBar>
+```
+
+## TopBar.Brand
+
+`TopBar.Brand` renders a styled link.
+
+```tsx preview
```
thomasdigby, line 32: Can we remove `preview` here to just render a code example?

lib/src/components/top-bar/TopBar.mdx
```diff
@@ -0,0 +1,59 @@
...
+}
+```
+
+## TopBar.BrandLogo
+
+`TopBar.BrandLogo` renders an image with set styles and is primarily intended for the Atom logo.
+
+## TopBar.BrandText
+
+`TopBar.BrandText` renders a `Text` component with set styles and is primarily intended to display the name of the current app.
```
thomasdigby, line 53: Did you consider `BrandName` here? I wonder if it might be a bit more meaningful than `BrandText`

lib/src/components/top-bar/TopBarActionIcon.tsx
thomasdigby: Yeah, I get it, probably one for an internal DS discussion around API consistency against ease of use across all of our components

lib/src/components/top-bar/TopBarBrand.tsx
```diff
@@ -0,0 +1,46 @@
...
+  alt = 'Atom Learning logo',
+  css
+}: TopBarBrandLogoProps): JSX.Element => {
+  return (
+    <Image
+      src={src}
+      alt={alt}
+      css={{
+        mr: '$3',
+        mb: 5,
```
thomasdigby, line 25: Do we need this margin-bottom if it's centred vertically in the nav?

lib/package.json
```diff
@@ -114,6 +114,7 @@
     "semantic-release-slack-bot": "^2.1.1",
     "size-limit": "^7.0.5",
     "stitches-reset": "^1.0.0",
+    "svg-jest": "^1.0.1",
```
thomasdigby, line 117: Can we remove any packages you initially added for the image parsing?

lib/sandbox/index.tsx
```diff
@@ -1,22 +1,103 @@
+import {
```
thomasdigby, line 1: Need to revert these changes 👍

lib/src/components/top-bar/TopBar.tsx
thomasdigby: Did you want to update this before merging or keep it as is?

thomasdigby (commented):
Just a couple of minor comments, but looking good otherwise!

lib/src/components/top-bar/TopBarBrand.tsx
MiguelGDLR: this is a little hack because the logo has too much weight on the bottom
[examples here](https://www.figma.com/file/KjgQh7eTwEfJmH6e56k9mX/DS-Documentation?node-id=7592%3A53271)

lib/src/components/top-bar/TopBar.tsx
elliotlaws: I'll update now to keep things consistent

lib/src/components/top-bar/TopBar.mdx
elliotlaws: Happy to change

lib/jest.config.js
```diff
@@ -6,7 +6,8 @@ module.exports = {
   testPathIgnorePatterns: ['/node_modules/'],
   setupFilesAfterEnv: ['./jest-setup.ts'],
   moduleNameMapper: {
-    '~/(.*)': '<rootDir>/src/$1' // matches the path alias in tsconfig.json
+    '~/(.*)': '<rootDir>/src/$1', // matches the path alias in tsconfig.json
+    '^.+\\.svg$': 'svg-jest'
```
thomasdigby, line 10: Still need to remove this and the reference in `yarn.lock`. Just run a `yarn` in the root of `lib` and it should remove it automatically.

thomasdigby (approved):
Just that final fix then all good to merge 👍

PR #389, opened 2022-08-30
feat:added navigation menu component

elliotlaws:
Ticket: https://atomlearningltd.atlassian.net/browse/DS-118
Designs: https://www.figma.com/file/KjgQh7eTwEfJmH6e56k9mX/DS-Documentation?node-id=2

### Implementation:
- Used Radix UI Navigation Menu component - see https://www.radix-ui.com/docs/primitives/components/navigation-menu
- To align dropdown content directly under its corresponding trigger, an offset has been applied to the dropdown content; see https://github.com/radix-ui/primitives/issues/1462 for more info
- `Nav.DropdownItem` is unopinionated about what its contents are. You'll notice that I haven't created a DropdownItem component which accepts a `title`, `subtitle` or `icon` prop for example, as this limits the flexibility of this component. 

### Testing locally
To test locally, copy and paste the contents of this [text file](https://github.com/AtomLearning/components/files/9493961/sandbox.navigation.menu.txt) into `lib/sandbox/index.tsx` to generate the same UI as shown in the screenshots below.

You will need to run `yarn` to install `@radix-ui/react-navigation-menu`

#### Hover
image

 #### Focus
image

#### Active
image

#### Custom grid layout
image


lib/src/components/navigation/Nav.mdx
```diff
@@ -0,0 +1,150 @@
...
+## Changing the layout of the dropdown content
+
+By default, dropdown items inside of `Nav.DropdownContent` will stack, if you want to change the way items are displayed, you can add custom styling to `Nav.DropdownContent`.
+
+`Nav.DropdownItem` can render whatever you choose, in the below example `ContentGridItem` is used to render an icon, title and subtitle.
+
+`Nav.DropdownContent` styling has also been changed to allow a grid layout.
+
+```tsx preview
+const LinkTitle = styled('div', {
```
kerryambt, line 66: preview can't handle content like this, needs to be either one react component or a few in a fragment afaik

lib/src/components/navigation/Nav.tsx
```diff
@@ -0,0 +1,182 @@
...
+        </ViewportPosition>
+      </StyledMenu>
+    </NavContext.Provider>
+  )
+}
+
+Nav.Link = NavLink
+Nav.Dropdown = NavDropdown
+Nav.DropdownContent = NavDropdownContent
+Nav.DropdownItem = NavDropdownItem
```
kerryambt, line 182: + Nav.displayName = 'Nav'

lib/src/components/navigation/Nav.tsx
```diff
@@ -0,0 +1,182 @@
+import React from 'react'
```
kerryambt, line 1: I think in general for readability this file could be split into more than one file. it's a bit hard to follow atm.

lib/src/components/navigation/Nav.tsx
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    if (trigger && listWidth && activeItem === itemValue) {
+      const listCenter = listWidth / 2
+
+      const triggerOffsetRight =
+        listWidth -
+        trigger.offsetLeft -
+        trigger.offsetWidth +
+        trigger.offsetWidth / 2
```
kerryambt, line 142: does this need any brackets? and/or could it have some for clarity? I think I remember ordering from school but still this is a bit of a headache to read

lib/src/components/index.ts
```diff
@@ -58,3 +58,4 @@ export { ToggleGroup } from './toggle-group'
 export { Tooltip } from './tooltip'
 export { ValidationError } from './validation-error'
 export { Video } from './video'
+export { Nav } from './navigation'
```
thomasdigby, line 61: Have you thought about potentially using `Navigation` or `NavigationMenu` instead of `Nav`? Appreciate that `NavigationMenu.DropdownItem` might be a bit verbose, but I think I prefer that to shortening component names.

lib/package.json
```diff
@@ -131,6 +131,7 @@
     "@radix-ui/react-checkbox": "0.1.5",
     "@radix-ui/react-dialog": "0.1.7",
     "@radix-ui/react-dropdown-menu": "0.1.6",
+    "@radix-ui/react-navigation-menu": "^1.0.0",
```
thomasdigby, line 134: Probably one to add to the backlog, upgraded all other Radix packages to v1. We're potentially duplicating shared logic across components if one is limited to v0 and the other to v1, e.g. `@radix-ui/react-use-layout-effect`

lib/src/components/index.ts
elliotlaws: This did cross my mind too, I'm happy to change to `NavigationMenu`

lib/src/components/navigation/Nav.tsx
elliotlaws: I'll see what I can do to split it out further 🙂

lib/src/components/navigation/Nav.tsx
elliotlaws: I'm not sure where the brackets would go in this case, also if I try to add brackets, VS code automatically removes them on save. 

Open to other suggestions though

lib/package.json
elliotlaws: Good point, didn't think about this!

lib/src/components/navigation/NavigationMenu.tsx
```diff
@@ -0,0 +1,131 @@
...
+import { NavigationMenuContext } from './NavigationMenuContext'
+
+type NavigationMenuSubComponents = {
+  Link: typeof NavigationMenuLink
+  Dropdown: typeof NavigationMenuDropdown
+  DropdownContent: typeof NavigationMenuDropdownContent
+  DropdownItem: typeof NavigationMenuDropdownItem
+}
+
+const fadeIn = keyframes({
```
thomasdigby, line 22: You could reference animations from `utilities/style/keyframe-animations` here, I forgot they existed! 😅

lib/src/components/navigation/NavigationMenu.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const onNodeUpdate = (trigger: HTMLButtonElement, itemValue: string) => {
+    if (!activeItem) {
+      setOffset(null)
+      return trigger
+    }
+
+    if (trigger && listWidth && activeItem === itemValue) {
+      const listCenter = listWidth / 2
+
+      const triggerOffsetRight =
```
thomasdigby, line 83: Curious if this offset calculation works responsively? I haven't spent too much time digging into how this works, but just wanted to check

lib/src/components/navigation/__snapshots__/NavigationMenu.test.tsx.snap
```diff
@@ -0,0 +1,218 @@
...
+    animation: k-feVUdh 0.15s ease;
+  }
+
+  .c-PJLV-ijNcvSj-css[data-state="closed"] {
+    animation: k-kJITOZ 0.15s ease;
+  }
+
+  .c-PJLV-ifwUrEa-css {
+    display: none;
+    transform: translateX(nullpx);
```
thomasdigby, line 150: Is this expected due to `offset` not being calculated yet? Can we ensure that it falls back to `0` if falsey

lib/src/components/navigation/NavigationMenu.tsx
```diff
@@ -0,0 +1,131 @@
...
+    >
+      <StyledMenu onValueChange={setActiveItem} css={css}>
+        <StyledList ref={listRef}>{children}</StyledList>
+        <ViewportPosition>
+          <StyledViewport
+            css={{
+              // Avoid transitioning from initial position when first opening
+              display: !offset ? 'none' : undefined,
+              transform: `translateX(${offset}px)`,
+              transition: 'all 0.3s ease',
```
thomasdigby, line 111: I'm actually not a huge fan of the x-axis animation here, I wonder if removing it completely would just make it a little easier to use.

The Radix animation example works well because the entire dropdown seems to move across to each trigger and the contents within smoothly animate. In our animation, as we're instantly changing the content it feels a bit more jarring. Would be keen to hear your thoughts.

Also, have you considered adding a slight delay on the dropdown to allow the user to move the mouse over and past the trigger without triggering it?

lib/package.json
hollg: I actually have a branch where I've started this process from a guild day

lib/src/components/navigation/NavigationMenu.mdx
```diff
@@ -0,0 +1,169 @@
...
+    <NavigationMenu.Link active={isActive} {...props}>
+      Introduction
+    </NavigationMenu.Link>
+  )
+}
+```
+
+## Changing the layout of the dropdown content
+
+By default, dropdown items inside of `NavigationMenu.DropdownContent` will stack, if you want to change the way items are displayed, you can add custom styling to `NavigationMenu.DropdownContent`.
```
hollg, line 65: Outrageously picky of me, but:
```suggestion
By default, dropdown items inside of `NavigationMenu.DropdownContent` will stack. If you want to change the way items are displayed, you can add custom styling to `NavigationMenu.DropdownContent`.
```

lib/src/components/navigation/NavigationMenu.tsx
elliotlaws: Ah nice!

lib/src/components/navigation/NavigationMenu.tsx
elliotlaws: Unfortunately, it doesn't work responsively. The offset is fixed and isn't changed depending on screen size, so on smaller screen sizes, there is a possibility that the dropdown content will be clipped. 

The reason I've not bothered with this is that I'm assuming that at smaller screen sizes, the `NavigationMenu` component will not be used. I would expect a menu specifically designed for mobile to be rendered instead. 

Any thoughts on this?

elliotlaws:
Updated sandbox code has been updated in the description. You can also find it here 
[sandbox navigation menu.txt](https://github.com/Atom-Learning/components/files/9493974/sandbox.navigation.menu.txt)

lib/src/components/navigation/NavigationMenu.tsx
elliotlaws: I was unsure about the animation initially too. I've checked again rendering more dropdown items and it does seem slightly jarring, so I've removed x-axis animation. 

In regards to adding a delay on trigger - I've tried to do this by adding an `animationDelay` to the `fadeIn` animation, however, it doesn't work as intended (it initially shows and hides, then fades in). I've tried multiple different solutions to get this work but the only way I've managed to achieve some sort of delay is by altering the keyframes for `fadeIn` animation so that at 75% the opacity is still set to 0. Would be keen to know if you have any other ideas

lib/src/components/navigation/NavigationMenu.tsx
thomasdigby: I wonder if we could calculate if the offset value sits outside of the viewport and apply some CSS e.g. `max-width: 100vw; left: 0;`? I think we ought to do something about it, just to ensure that we cover ourselves if an overzealous marketing team member adds a whole bunch of content that makes it wider than the viewport even on desktop.

lib/src/components/navigation/NavigationMenu.tsx
thomasdigby: On the delay, it looks like it would need an update to the Radix component itself https://github.com/radix-ui/primitives/pull/1478 so we can leave it as is for now!

lib/src/components/navigation/NavigationMenuItem.tsx
```diff
@@ -0,0 +1,155 @@
+import React from 'react'
+import { ChevronDown } from '@atom-learning/icons'
+import { styled } from '~/stitches'
+import { Icon } from '../icon'
+import * as PrimitiveNavigationMenu from '@radix-ui/react-navigation-menu'
```
thomasdigby, line 5: Can we keep these imports consistent? We're using `import * as NavigationMenuPrimitive` in `NavigationMenuDropdown`

lib/src/components/navigation/NavigationMenu.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+import { NavigationMenu } from '.'
+
+const ExampleNav = () => (
+  <IdProvider>
+    <NavigationMenu>
+      <NavigationMenu.Link href="/introduction">
+        Introduction
+      </NavigationMenu.Link>
+      <NavigationMenu.Dropdown title="Theme">
```
thomasdigby, line 15: Just a quick one on this API, do you think we're restricting ourselves with the trigger as a string? It feels like this could be a component within `Dropdown`. Or is this set-up like this to avoid overriding/clashing with the icon behaviour?

```jsx
<NavigationMenu>
  <NavigationMenu.Link href="/introduction">
    Introduction
  </NavigationMenu.Link>
  <NavigationMenu.Dropdown>
    <NavigationMenu.DropdownTrigger>
      Theme
    </NavigationMenu.DropdownTrigger>
    <NavigationMenu.DropdownContent>
      <NavigationMenu.DropdownItem href="https://app.atomlearning.co.uk/colours">
        Colours
    </NavigationMenu.DropdownContent>
  </NavigationMenu.Dropdown>
</NavigationMenu>
```

lib/src/components/navigation/NavItem.tsx
```diff
@@ -0,0 +1,155 @@
...
+)
+
+const NavDropdownItem = React.forwardRef<
+  HTMLAnchorElement,
+  React.PropsWithChildren<NavLinkProps>
+>((props, forwardedRef) => {
+  return <NavLink ref={forwardedRef} variant="dropdownItem" {...props} />
+})
+
+export { NavDropdownTrigger, NavLink, NavDropdownContent, NavDropdownItem }
```
kerryambt: can we change these exports as per your other PR? want to keep these consistent

lib/src/components/navigation/Nav.tsx
kerryambt: maybe some of the named components could be in separate files? just seems a little big atm

lib/src/components/navigation/NavItem.tsx
elliotlaws: This should be updated now

lib/src/components/navigation/NavigationMenu.test.tsx
elliotlaws: To avoid overriding/clashing with the icon was one of the reasons, but another reason was that NavigationMenuPrimitive.Item requires a value prop for the viewport offset to apply (so we can render the dropdown menu directly under trigger). 

It is possible but we end up with a lot more complexity (see below). Any thoughts? 

```tsx
export const NavigationMenuDropdown: React.FC<NavigationMenuDropdownProps> = ({
  children,
  title,
  ...props
}) => {
  const [itemValue, setItemValue] = React.useState<string | undefined>('')

  return (
    <NavigationMenuPrimitive.Item value={itemValue}>
      {React.Children.map(children, (child) => {
        if (!React.isValidElement(child)) {
          throw new Error('something')
        }

        const isValidChildType =
          child.type === NavigationMenuDropdownTrigger ||
          child.type === NavigationMenuDropdownContent

        invariant(
          isValidChildType,
          `Children of type ${child?.type} aren't permitted. Only an ${NavigationMenuDropdownTrigger.displayName} or ${NavigationMenuDropdownContent.displayName} component is allowed in ${NavigationMenuDropdownTrigger.displayName}`
        )

        if (child.type === NavigationMenuDropdownTrigger) {
          return React.cloneElement(child as React.ReactElement, {
            itemValue,
            setItemValue
          })
        } else if (child.type === NavigationMenuDropdownContent) {
          return child
        }
      })}
    </NavigationMenuPrimitive.Item>
  )
}
```

Trigger 
```tsx
type NavigationMenuTriggerCommonProps = {
  active?: boolean
  setItemValue?: React.Dispatch<React.SetStateAction<string>>
  itemValue?: string
}

type NavigationMenuDropdownTriggerPropVariants =
  | {
      value?: string
      children: string
    }
  | {
      value: string
      children: React.ReactNode
    }
  | {
      value: string
    }

type NavigationMenuDropdownTriggerProps =
  NavigationMenuDropdownTriggerPropVariants & NavigationMenuTriggerCommonProps

export const NavigationMenuDropdownTrigger = React.forwardRef<
  HTMLButtonElement,
  NavigationMenuDropdownTriggerProps
>(
  (
    { children, active, value, itemValue, setItemValue, ...props },
    forwardedRef
  ) => {
    const { onNodeUpdate } = useNavigationMenuContext()

    React.useEffect(() => {
      if (setItemValue) {
        const val = typeof children === 'string' ? children : value
        setItemValue(val || '')
      }
    }, [children, setItemValue, value])

    return (
      <StyledTrigger
        active={active}
        {...props}
        ref={(node: HTMLButtonElement) =>
          onNodeUpdate(node, itemValue as string)
        }
      >
        {children}
        <Icon
          is={ChevronDown}
          css={{
            '[data-state=open] &': { transform: 'rotate(-180deg)' },
            '@media (prefers-reduced-motion: no-preference)': {
              transition: 'transform .2s ease'
            }
          }}
          size="sm"
        />
      </StyledTrigger>
    )
  }
)

```

elliotlaws: With the above code, we allow the consumer to pass in whatever they want as a trigger. However, if the children is not of type string, then the consumer needs to pass in a `value` prop to `NavigationMenu.Trigger`, so that we can assign `NavigationMenu.Item` with a valid string value.

thomasdigby: Eesh yeah, probably best to keep it as it is for now. We can always revise the API in future if we find an elegant way to handle nav positioning.

lib/src/components/navigation/NavigationMenu.tsx
elliotlaws: Myself, Gary, Myrto and Miguel have spoken about this and we've agreed that this can be implemented as part of a separate ticket see https://atomlearningltd.atlassian.net/browse/DS-127

I've created a copy of this branch and linked it in the ticket so that whoever picks it up can pick up where we left off.

thomasdigby (approved):

lib/sandbox/index.tsx
```diff
@@ -1,22 +1,193 @@
+import {
```
thomasdigby, line 1: Can we remove all the changes from this file before merging?

elliotlaws: Oops, yeah sure

lib/src/components/navigation/NavigationMenu.mdx
```diff
@@ -0,0 +1,170 @@
...
+## Changing the layout of the dropdown content
+
+By default, dropdown items inside of `NavigationMenu.DropdownContent` will stack.
+If you want to change the way items are displayed, you can add custom styling to `NavigationMenu.DropdownContent`.
+
+`NavigationMenu.DropdownItem` can render whatever you choose, in the below example `ContentGridItem` is used to render an icon, title and subtitle.
+
+`NavigationMenu.DropdownContent` styling has also been changed to allow a grid layout.
+
+```tsx preview
```
thomasdigby, line 72: Have you run the site locally to test if this works?

elliotlaws: I did at one point but not since my updates, will check now

elliotlaws: Updated this now

PR #390, opened 2022-09-01
feat: inline message

kerryambt:
### Description

https://atomlearningltd.atlassian.net/jira/software/projects/DS/boards/27?selectedIssue=DS-65

**This is a breaking change**

We are removing the ValidationError component and replacing it with InlineMessage, which has the same functionality plus more themes, sizes, and optional icons.

Using it is as simple as `<InlineMessage>Message here</InlineMessage>`.

This not only creates the new component but removes the old one, and updates all use of that component in things like field components.

### Screenshots

image
image


lib/src/components/inline-message/InlineMessage.tsx
```diff
@@ -0,0 +1,52 @@
...
+  mr: '$2'
+})
+
+interface IInlineMessageProps {
+  showIcon?: boolean
+  css?: CSS
+  size?: 'xs' | 'sm' | 'md'
+}
+
+type TInlineMessage = React.ComponentProps<typeof InlineMessageContainer> &
```
hollg, line 31: Doesn't `React.ComponentProps<typeof InlineMessageContainer>` already include the `css` field? I.e. can we not omit it from `IInlineMessageProps`?

I think the naming here is also a little confusing. I'm used to `SomeComponentProps` or `IComponentProps` being the type/interface that defines the props for `SomeComponent`. Having something explicitly called `IInlineMessageProps` which does _not_ define the type of `InlineMessage`'s props feels quite misleading. 

Maybe this could be simplified with something like the following?

```ts
type InlineMessageProps = React.ComponentProps<typeof InlineMessageContainer> & {
  showIcon?: boolean
  size?: 'xs' | 'sm' | 'md'
}
```

lib/src/components/inline-message/InlineMessage.tsx
```diff
@@ -0,0 +1,52 @@
...
+  children
+}) => (
+  <InlineMessageContainer theme={theme} css={css}>
+    {showIcon && (
+      <InlineMessageIcon
+        size="sm"
+        is={INLINE_MESSAGE_ICONS[theme as InlineMessageTheme]}
+      />
+    )}
+    <Text as="span" size={size} css={{ color: 'inherit', mt: '$0' }}>
```
hollg, line 48: Does `Text` have some `margin-top` by default? 😳

kerryambt: no? did you read this as `0` instead of `$0`?

this is done to make the alignment match the designs, because capsize makes them a little too high up compared to the Icon. On the previous component a transform was used, but margin seemed much easier to understand :)

lib/src/components/inline-message/InlineMessage.tsx
kerryambt: I've used this pattern a bunch of times and seen it used in other places, I think the key for me is if the props are numerous to put it in a separate interface. given your point about css being part of the styled output (my bad, always forget that!) then I think we can condense here as you suggest. will do now :)

elliotlaws: Could I maybe suggest an alternative to avoid type casting `theme` and reduce the amount of code required.

``` tsx
type InlineMessageProps = {
  showIcon?: boolean
  css?: CSS
  size?: 'xs' | 'sm' | 'md'
  theme: InlineMessageTheme
}

export const InlineMessage: React.FC<InlineMessageProps> = ({
  css,
  showIcon = true,
  theme = 'error',
  size = 'sm',
  children
}) => (
  <InlineMessageContainer theme={theme} css={css}>
    {showIcon && (
      <InlineMessageIcon size="sm" is={INLINE_MESSAGE_ICONS[theme]} />
    )}
    <Text as="span" size={size} css={{ color: 'inherit', mt: '$0' }}>
      {children}
    </Text>
  </InlineMessageContainer>
)
```

lib/src/components/inline-message/InlineMessage.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { Flex } from '~/components/flex'
+import { Text } from '~/components/text'
+import { Icon } from '~/components/icon'
+
+import { CSS, styled } from '~/stitches'
+import { InlineMessageTheme } from './InlineMessage.types'
+
+const InlineMessageContainer = styled(Flex, {
+  variants: {
+    theme: INLINE_MESSAGE_THEMES
```
hollg, line 16: Has this pattern of moving a theme definition out of the component file been used elsewhere in the lib? My instinct is that it makes this component slightly more difficult to parse without much obvious gain. The file isn't particularly long and the theme definition is only an extra 5 lines so I wouldn't be worried about it taking up space. But maybe there is some other consideration that I'm not seeing?

lib/src/components/inline-message/InlineMessage.tsx
hollg: Are there not fields on the Stitches type that we'll lose that way? E.g. `as`, possibly various HTML attributes? IMO we should be passing all valid `Flex` props from `InlineMessage` to the `InlineMessageContainer` and I'm not sure TypeScript will let us if we're not explicit about it

kerryambt: yeah I've changed this, fair suggestion Elliot but we should definitely use `React.ComponentProps<typeof ThingGeneratedFromStyled>` for the reason Gary says :)

lib/src/components/inline-message/InlineMessage.tsx
hollg: > no? did you read this as 0 instead of $0?

_yup_!

Will just ping @thomasdigby re margin vs transform because I suspect that came from him originally and I don't know anything about capsize 🤷🏻

lib/src/components/inline-message/InlineMessage.tsx
hollg: Ha, good catch! Yes, I think we should also spread them into the container component

lib/src/components/inline-message/InlineMessage.tsx
kerryambt: I've used this in core, not in this lib. my reasoning behind it is that if you create the variants in another const, you can compose that with `Record<ThemeType, whatever>` and you get the benefit of errors whenever you haven't made a theme for all your variants. It's super useful when extending components, because adding another theme will error if you haven't updated config.

lib/src/components/inline-message/InlineMessage.tsx
kerryambt: yeah nice catch! I'll add that spread now

MiguelGDLR (commented):
Everything looks good on my end. Great work

lib/src/components/inline-message/InlineMessage.config.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import { InlineMessageTheme } from "./InlineMessage.types"
+
+import type { CSS } from '~/stitches'
+
+export const INLINE_MESSAGE_THEMES: Record<InlineMessageTheme, CSS> = {
+    success: { color: "$success" },
+    warning: { color: "$warningText", "& svg": { color: "$warningDark" }}, 
+    info: { color: "$primary" },
+    system: { color: "$tonal400" },
```
andoulla, line 13: In `Badge` this grey variant is called neutral, why not use the same term here? the benefit is that as devs we will associate gray with neutral and it will be easier to remember.

lib/src/components/inline-message/InlineMessage.mdx
```diff
@@ -0,0 +1,50 @@
+---
+title: Inline Message
+component: InlineMessage
+description: Inline feedback message with an optional icon
+category: Feedback
+---
+
+Inline notices consist of an inline feedback message in the form of simple text that inform the user of relevant information, revealed in context.
```
andoulla, line 8: Extremely silly, but I would suggest the word message instead of notices because, as a non-native speaker, it took a second pass to realise that it was notices the noun instead of the verb. Again,  it might be just me.

lib/src/components/inline-message/InlineMessage.mdx
```diff
@@ -0,0 +1,50 @@
...
+
+```tsx preview
+<InlineMessage>This is an Inline Message</InlineMessage>
+```
+
+## Theme
+
+These are the available `theme`s for this component: `success`, `warning`, `error`, `system` and `info`. The default is `error` (due to most frequent context).
```
andoulla, line 16: I understand why you chose the default as error, but since we are going to have to replace all the instances in the code base anyway, we might as well set the preferred default (I assume is either system or info, right?). Since the code base has only the error variant, it isn't hard to just pass the error theme down.

lib/src/components/inline-message/InlineMessage.tsx
andoulla: +1 on Gary's comment.

lib/src/components/inline-message/InlineMessage.mdx
```diff
@@ -0,0 +1,50 @@
...
+  <InlineMessage size="md">This is a md Inline Message</InlineMessage>
+</>
+```
+
+## Icon
+
+Each theme has a specific icon it should export, so showing or hiding an icon is simply done using the `showIcon` prop. Default is `true`.
+
+```tsx preview
+<InlineMessage showIcon={false}>This is an Inline Message</InlineMessage>
```
andoulla, line 49: Does this break the composition pattern we are using? Why not export the Icon component (InlineMessage.Icon) and let the consumer decide if they want the icon or not? 
We could even try to pass the theme down to the icon component if it is in the Children array.

Unless this is intentional to never allow the icon to be on the right-hand side

site/src/sandbox/components.tsx
```diff
@@ -1104,9 +1104,9 @@ post.
               <Button size="sm" type="submit">
                 Click to show validation errors
               </Button>
-              <ValidationError>
```
andoulla, line 1107: Didn't know this file existed, wondering what it does

lib/src/components/inline-message/InlineMessage.mdx
kerryambt: oh nice catch! I just copypasted this from DS docs but that works better, I'll change now

lib/src/components/inline-message/InlineMessage.mdx
kerryambt: @MiguelGDLR which do you think should be default? I'm not sure tbh

site/src/sandbox/components.tsx
kerryambt: this sandbox component? I think it's like a storybook type thing we used to use? I figured it should be updated as the component no longer exists but I don't actually know where this page is viewable

andoulla: of course, it should be updated. I was just so oblivious of its existence, I will later investigate what it does

lib/src/components/inline-message/InlineMessage.mdx
kerryambt: knew someone was going to say this :P my reasoning that this is slightly against compositional pattern is:

- we always want the icon to be on the left
- the icon matches the theme and should not be changed
- even if I made `<InlineMessage.Icon>` and made that inherit the theme so that it always knew the right Icon to use, what would the API look like? You couldn't go `<InlineMessage><InlineMessage.Icon/>text here</InlineMessage>` because how would you differentiate the text from the icon when using `children` to compose? Write some overly complicated function to pick out the Icon? The other option would be `<InlineMessage><InlineMessage.Icon/><InlineMessage.Text>text here</InlineMessage.Text></InlineMessage>` but imo that is waaaaaaaay too complicated for the use case. I don't think we gain enough from 100% composition pattern here. We are assuming a lot with the Icon on this component and I think not using a prop would be over engineering. If the icon could move, or be different, then I'd be all for it. Given that is not the case, I think a prop is fine. 

definitely a good question to ask, hopefully it's okay how it is though :)

lib/src/components/inline-message/InlineMessage.config.ts
MiguelGDLR: the concept is slightly different between Badges and Inline messages (strictly semantic) but I don't love _System_ either and _Neutral_ seems ok to me. I will change it in Figma too, thanks for the suggestion

lib/src/components/inline-message/InlineMessage.mdx
andoulla: This is why I added the last sentence, because if we are going to "enforce" the icon position then how you have done it is the way.

andoulla (approved):

lib/src/components/inline-message/InlineMessage.tsx
thomasdigby: `transform` moves the element without affecting other elements around it. It's a purely visual change. Whereas margin could require layout calculations for surrounding elements. It doesn't really matter but it makes sense to use transform if possible.

thomasdigby (approved):

hollg (approved):

PR #394, opened 2022-09-15
fix/ds-91 removing marging bottom from label wrapper when there is no label

viniciusvolpe:
### Description
Fixing field wrapper component to not add any margin when there is no label to be displayed. This extra margin was causing issues when using fields that don't need labels.

Jira: https://atomlearningltd.atlassian.net/browse/DS-91

### Implementation
- removing margin when there is no label
- fixing tests

### Screenshot
image


lib/src/components/field-wrapper/FieldWrapper.tsx
```diff
@@ -31,7 +31,11 @@ export const FieldWrapper: React.FC<FieldWrapperProps> = ({
 }) => (
   <Box css={css}>
     <Flex
-      css={{ justifyContent: 'space-between', alignItems: 'center', mb: '$3' }}
+      css={{
+        justifyContent: 'space-between',
+        alignItems: 'center',
+        mb: label ? '$3' : '0'
```
hollg, line 37: I don't think I understand the use case for a `FieldWrapper` without a `label`. `label` is even a required field on the prop type at the moment. 

@LimeWub the Jira ticket claims to point back to a comment from you, but the link doesn't work for me. Do you know what the deal is here?

LimeWub: I think it's from this: https://github.com/Atom-Learning/atom-core/pull/1779#discussion_r912927800

It's not about not having a label. It's about having a visually hidden label if the design shows no labels.

LimeWub: fwiw: In the usecase from that PR we just added regular labels over the field

hollg: Ah, thanks @LimeWub! It looks like this made it into Jira without the second half of your comment, which included the proper solution. I don't remember seeing this ticket, so if it was in `Ready for dev` then we perhaps need to examine what process (or lack of) led to that. Let's have that conversation within the pod.

@viniciusvolpe inputs in forms should always have labels for accessibility reasons, so we don't actually want to support _not_ providing one. As Myrto mentions above, what we really want to support is a _visually hidden_ label. I'm really sorry that a ticket got into `Ready for dev` in the DS backlog with bad (or a lack of) information in it. Starting next week, Myrto and I have regular meetings to further refine and add info for engineers to those tickets so hopefully this can be avoided in the future!

viniciusvolpe: Thanks both, added the change as requested

LimeWub (approved):

hollg (approved):

PR #395, opened 2022-09-15
fix/ds-75 adding image to empty state doc

viniciusvolpe:
### Description
Adding an image to empty state doc so it's possible to see how the component looks.

Jira: https://atomlearningltd.atlassian.net/browse/DS-75

image


hollg:
@viniciusvolpe Thanks for picking this up Vini! Could you add a screenshot of what the page looks like with the image to the PR description?

andoulla (approved):

hollg (approved):

PR #396, opened 2022-09-20
fix: empty state image size

kerryambt:
### Description

Small bug fix for the Empty State image, if you used an SVG as the image default SVG sizes were overriding the sizes passed in, meaning the image needed to be sized using a box.

This fix makes the passed in sizes take precedence rather than using an outer box which would add bloat to the dom.

https://atomlearningltd.atlassian.net/jira/software/projects/DS/boards/27?selectedIssue=DS-122

### Screenshots

Fixed:
image
Broken:
image

### DIFF
--- a/lib/src/components/empty-state/EmptyStateImage.tsx
+++ b/lib/src/components/empty-state/EmptyStateImage.tsx
@@ -5,27 +5,27 @@ export const EmptyStateImage = styled(Image, {
   variants: {
     size: {
       xs: {
-        width: '56px',
-        height: '32px',
+        width: '56px !important',
+        height: '32px !important',
         mb: '$4'
       },
       sm: {
-        size: '84px',
+        size: '84px !important',
         mb: '$4'
       },
       md: {
-        width: '126px',
-        height: '72px',
+        width: '126px !important',
+        height: '72px !important',
         mb: '$4'
       },
       lg: {
-        width: '190px',
-        height: '142px',
+        width: '190px !important',
+        height: '142px !important',
         mb: '$5'
       },
       xl: {
-        width: '285px',
-        height: '213px',
+        width: '285px !important',
+        height: '213px !important',
         mb: '$5'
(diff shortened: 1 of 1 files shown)


LimeWub:
Can't we fix this properly (ie. fix the actual image component to be ok w width/height) ? ._.

kerryambt:
> Can't

My reason for not doing this was can we be sure that somewhere a component isn't depending on the current setup of width/height for svgs? I don't want this to be a breaking change. afaik because of how stitches works and that the svg style is set in `'&[src$=".svg"]': { ... }` we can't just spread css in to an image component to override that because `&` levels always take precedence over base levels. i.e. if you composed the component to hard override its own css with passed in css (i.e. `css={{ stuff, ...css }}` it'd still not care about width and height passed as css because of the `&` block. So the only solution would be change what happens for svgs, but then I come back to my point of 'can we be sure that wont break something?'

unless I totally misunderstand in which case, my bad :)

LimeWub:
@kerryambt 
Currently the image component has one assumption: that we want the image to be fluid.

It shouldn't have this assumption and the fluid functionality should instead be a variant `isFluid`(?). For a start: We can introduce an anti-pattern to default this variant to `true` (tmp until someone does the work to replace all uses of images which might be a significant task). This means that all current uses of `Image` will be getting the same css.
Updating snapshots will be somewhat painful but otherwise nothing else will change AND it will allow for new uses where we pass `isFluid` as `false` and can actually override width/height.

That's how I was thinking of it.

If you are not comfy with making this change (after all updating the snapshots would be PITA + I think there's one more ticket for fixing the Image component?) - then I'd prefer the Box version of this as it is clearer/consistent with other use cases and will be more obvious to be replaced when Image is fixed to not assume we want it to be fluid.

thomasdigby (approved):

LimeWub (approved):

PR #397, opened 2022-09-20
feat: search field

kerryambt:
### Description

https://atomlearningltd.atlassian.net/jira/software/projects/DS/boards/27?selectedIssue=DS-100

We did not have a form field component for `SearchInput`, this adds that component.

The only real complexity here, and difference from `InputField` in terms of differences of Input -> Field, is that `SearchInput` already used a ref. React-hook-form's `register` uses refs to register the form. The ref passed in from that register is a completely different type to the type of ref we'd normally use on a component, so attempting to access `ref.current` wasn't working. Eventually after a lot of digging I found an example on the react-hook-form docs for sharing refs, where you can construct the ref on the input element from two types of refs. This way we have all the information registered properly but can also access `ref.current` in order to clear the input and properly bubble up onChange events (a problem we have solved in the past by adding the ref).

Otherwise this is pretty standard, just adds a SearchField component that wraps SearchInput in a FieldWrapper, as well as things like tests and mdx.

### Screenshots

image

https://user-images.githubusercontent.com/66493855/191293332-3a6d2c6e-50dc-4701-8aca-f6bfd57708b2.mov


lib/src/components/search-field/SearchField.tsx
```diff
@@ -0,0 +1,52 @@
+import * as React from 'react'
+import { useFormContext } from 'react-hook-form'
+
+import { FieldWrapper } from '~/components/field-wrapper'
+import { useFieldError, ValidationOptions } from '~/components/form'
+import { SearchInput, SearchInputProps } from '~/components/search-input'
+import type { CSS } from '~/stitches'
+
+type SearchFieldProps = SearchInputProps & {
+  css?: CSS
```
hollg, line 10: `css` is already on `SearchInputProps — can we skip repeating it here?

kerryambt: oops! nice catch, fixed :)

hollg:
Could you also add an example of the `SearchField` to the `Forms` section of `site/src/sandbox/components.tsx`? That's what renders the page of _every component_ at https://design.atomlearning.technology/components

lib/src/components/search-input/SearchInput.tsx
```diff
@@ -48,82 +48,90 @@ const StyledSearchInput = styled(Input, {
...
+      size = 'md',
+      css,
+      value,
+      clearText = 'Clear',
+      onChange,
+      ...remainingProps
+    },
+    ref: (instance: HTMLInputElement | null) => void
+  ) => {
+    const [inputElRef] = useCallbackRef()
```
LimeWub, line 63: If possible - could you pls try adding this to the component

```
React.useImperativeHandle(ref, () => inputElRef);
```

instead of the other changes made in the `ref={()=>}` below and below?

If I remember well and this one line works - I think it'll be much easier to understand what we are doing (and easier to comment if needed as it's 1 line) and we wont be messing with internals of the `useCallbackRef` hook

kerryambt: the actual code was very close, `React.useImperativeHandle(ref, () => inputElRef.current as HTMLInputElement)` (type casting to avoid errors), needed the .current to work basically. very interesting though! thanks for sharing :)

LimeWub: Ah! Glad u got this version working! (and glad I _almost_ remembered it)

LimeWub (approved):

PR #398, opened 2022-09-23
CMS

LimeWub:
## Description

Replaces `/site` directory with a new `Next.js` + `NetlifyCMS` implementation to allow non-technical users to edit and manage design system documentation.

**Jira**: https://atomlearningltd.atlassian.net/browse/GT-57

During this new build some additional changes have been required to be made in the components lib. **These have now all been merged in as separate PRs but for posterity sake I'm leaving them as references here:**

* ColorScheme component and some initial color schemes
* Two components have been modified to use colorscheme: `Tabs` and `Accordion`. 
* StackContent spacing uses `& > *`
I use StackContent to space the contents of the documentation site and noticed it was not adding spacings to code. So I made it more generic to add spacing to all direct children - this should not affect anything other than snapshots.
* Heading and Text no longer to have a hardcoded color
* Made the Image component actually respect size without a Box required around it here

The content has been added in a separate PR here: https://github.com/Atom-Learning/components/pull/416

## Screenshots
Screenshot 2022-10-20 at 10 39 44
Screenshot 2022-10-20 at 10 43 06
Screenshot 2022-10-20 at 10 40 17
Screenshot 2022-10-20 at 10 41 36
Screenshot 2022-10-20 at 10 42 20


color-scheme/README.md
```diff
@@ -0,0 +1,20 @@
+## Color Schemes (v.ALPHA)
```
hollg, line 1: I wonder whether it's worth having a separate PR just for this, too, so that we can get all the inevitable discussion handled as smoothly as possible.

LimeWub: https://github.com/Atom-Learning/components/pull/415

lib/src/components/text/Text.tsx
```diff
@@ -38,7 +38,6 @@ export const textVariantSize = ({ applyCapsize = true } = {}): Record<
 })
 
 export const StyledText = styled('p', {
-  color: '$tonal600',
```
thomasdigby, line 41: Are we combining the global typography theming update into this PR? I think we ought to split it out if so, there are probably a few conversations to be had specifically on that implementation that I assume aren't related to the CMS work here

lib/src/components/stack/Stack.tsx
```diff
@@ -46,12 +45,7 @@ const StyledStack = styled('div', {
     },
     gap: {
       ...createThemeVariants('space', {
-        mt: '-$key',
-        ml: '-$key',
-        '& > *': {
-          mt: '$key',
-          ml: '$key'
-        }
+        gap: '$key',
```
thomasdigby, line 48: I wish we could do this, but we can't yet. We're still missing support from Safari v12/13 https://caniuse.com/flexbox-gap

If there's a way that we can combine both, or provide a fallback for older browsers, then I'm all for it 👍 Unfortunately as we can't test for support in `@supports()`, I don't think there's a CSS only way to detect support and fallback to margins.

LimeWub: I thought we'd be safe because of the caniuse global use of v12/13 Safari totalling less than 1% global (and I assume UK mid/hi-income families wouldn't be in that 1% of global).
https://caniuse.com/usage-table

Screenshot 2022-10-31 at 12 31 46
Screenshot 2022-10-31 at 12 33 26


Do we have proof in our analytics that our users still have Safari 12/13? We might be able to drop support if that's no longer the case

LimeWub:
I've rebased and cleaned this up so afaik this one should be good for a proper re-review now. @hollg @thomasdigby 
This PR is JUST replacing site with the cms version without any content changes.
The PR with the content (separate PR linked in description) needs one more manual check of the content for anything that may have gone out of sync in terms of documentation and deleting the `.mdx` files but will ping on that one separately once it's done.

documentation/pages/_app.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { Layout as AppLayout } from "~/components/app";
+
+
+globalCss({
+  '*': { boxSizing: 'border-box' },
+  body: { margin: 0 }
+})()
+
+
+// TODO: Test in build.
```
hollg, line 17: @LimeWub one of the comments you mentioned?

LimeWub: yes :p will do another pass for little leftover bits like this.

lib/scripts/generate-component-props.mjs
```diff
@@ -29,6 +29,9 @@ const run = async () => {
       tsConfigParser.parse(path)
     )
 
+    await fs.mkdir('./dist', { recursive: true }, (err) => {
```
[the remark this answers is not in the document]
LimeWub: So. I just tested this by running it when dist is already there (and when it's not).
It seems it doesn't matter and it runs fine in both cases and doesn't override other things in the folder when they are there (ie after lib has built) - BUT I can add the `.exists` check if you think it's worth.


Update: I think this just works because of the `{recursive: true}` option as per this SO answer: https://stackoverflow.com/a/24311711

documentation/pages/_app.tsx
LimeWub: All cleaned up now 👍

documentation/public/favicon/browserconfig.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
```
LimeWub, line 1: FYI: I added a favicon

documentation/components/admin/editor-component/SizeTokenListEditorComponent.tsx
```diff
@@ -0,0 +1,28 @@
+export const SizeTokenListEditorComponent = {
```
hollg, line 1: What does the `EditorComponent` part of these names signify? I don't think I totally get where these fit in — is this the config for the "custom Netlify CMS component"? Maybe that's what `EditorComponent` means?

documentation/components/app/Navigation.tsx
```diff
@@ -0,0 +1,86 @@
...
+import * as React from 'react'
+import { Link } from '.'
+
+const navigationStructure = buildConstants['navigation-structure']
+
+type NavigationProps = {
+  onNavigate?: (pathname) => void
+}
+
+const NavigationTitle = ({ level, css, ...rest }) => {
```
hollg, line 15: Could we get a type for these props? Might need to get a bit fancy to capture all possibilities, e.g.

```ts
type NavigationTitleProps = (React.ComponentProps<typeof Heading> & {level: 1, css: CSS}) | (React.ComponentProps<typeof Text> & {level: 2, css: CSS}) | (React.ComponentProps<typeof Text> & {level: 3, css: CSS}) 
```

documentation/components/app/Navigation.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const renderRecursive = React.useMemo(() => {
+    const render = (navigationStructureLevel, level) => {
+      if (!navigationStructureLevel?.length) return null;
+      return (
+        <Box as="ul" css={{ listStyleType: 'none', m: 0, mb: '$4', p: 0, pl: [1, 2].includes(level) ? 0 : '$4' }}>
+          {navigationStructureLevel.map(({ title, href, children }) => {
+            if (!title) return;
+            return (
+              <Box as="li" css={{ mb: [1].includes(level) ? '$6' : 0 }} key={href}>
```
hollg, line 63: Is there any functional difference between this and `level === 1 ? ...` ? If not, I think that might be a bit easier to parse

lib/src/components/form/__snapshots__/Form.test.tsx.snap
```diff
@@ -51,7 +51,7 @@ exports[`Form component renders a form 1`] = `
 
 .sxfigyv[disabled] {
   background-color: var(--colors-tonal300);
-  color: var(--colors-tonal700);
+  color: var(--colors-tonal600);
```
hollg, line 54: Why is this changing?

LimeWub: tonal700 doesn't exist as a colour. Not sure why the snapshot by itself updated here - I thought I had updated this one already in the colours PR but maybe it got missed (?)

documentation/components/app/Navigation.tsx
LimeWub: Ye you're right. Pretty sure this is the same as `level === 1`.  I'll change it.

lib/src/components/form/__snapshots__/Form.test.tsx.snap
hollg: Hm, weird. My worry was just that whatever's causing this could also cause unexpected snapshot failures in `atom-core`, but I don't _think_ that seems likely, right?

documentation/components/app/Navigation.tsx
LimeWub: 👀 do we have to? this looks really complex.

but, if i remember well, level is literally just a number in this one

documentation/components/admin/editor-component/SizeTokenListEditorComponent.tsx
LimeWub: ye Editor Component is the admin setup for the custom components

documentation/components/app/Navigation.tsx
hollg: IMO the complexity is already in your code — defining the type just makes it explicit (and therefore easier to spot, reason about etc) rather than letting it be hidden. The existing code already has specific valid combinations of props, but they are left implicit.

E.g. if we add a new prop, `foo` to `Header`, and have a `<NavigationTitle level={2} foo={bar} {...} />`, that `foo` will be passed to `Text` even though it isn't a valid prop for that component. We have to help TS help us to prevent that sort of nonsense.

lib/src/components/form/__snapshots__/Form.test.tsx.snap
LimeWub: I'll have a look to try and figure out what happened here. My guess is we just missed the snapshot in another PR but it is indeed suspicious.

LimeWub: Well. I have looked into it - I have no clue what happened but there's no real changes to this component in this PR. All changes in components in lib in this PR are formatting.
Rerunning the snapshot doesn't change it either.

I went back to main to check for uses of `tonal700` and apart from this one all the other files using it have been deleted (old site files).
Screenshot 2023-02-03 at 18 49 01
but running the snapshot with tonal700 in main doesn't error. 

Also deleting all the other mentions of tonal700 in main and rerunning tests STILL leaves the snapshot to tonal700.

I'm literally so confused.

LimeWub: All the formatting changes were the pre-commit formatter but I removed this bit.

documentation/components/app/Navigation.tsx
LimeWub: This is too complex for me to do at this time of the evening - every other piece of feedback has been addressed tho.
If you'd like to make a commit on this on Monday that'd be awesome, otherwise I can try it out on Tuesday.

(I still think it'd be quite complex for something fairly simple - perhaps we could just simplify the props and just allow `css` if that would be ezier 🤷 )

documentation/components/app/Navigation.tsx
LimeWub: done

lib/src/components/form/__snapshots__/Form.test.tsx.snap
hollg: I don't think I follow exactly what has happened — let's catch up about it this morning. Sounds like the worst case scenario is that we just leave it as it was (skipped) and come back to figure out what I was doing at a later point

documentation/components/app/Navigation.tsx
hollg: > I still think it'd be quite complex for something fairly simple

I'm not sure I follow what you mean here — the type is just a description of the logic that you already have in place. If you're happy with that logic, why would you not want to put the type in that describes it?

LimeWub: Simplified it: https://github.com/Atom-Learning/components/pull/398/commits/97f2a4e6756f84edb1cbf7cc12b23238c81ff836

hollg (approved):

documentation/components/admin/editor-component/CodeBlockEditorComponent.tsx
```diff
@@ -2,45 +2,52 @@ import * as React from "react";
...
+  pattern:
+    /^<CodeBlock live={(.*?)} preview={(.*?)} code={`(.*?)`} language={"(.*?)"} \/>$/ms,
+  fromBlock: function (match) {
+    return {
+      live: match[1],
+      preview: match[2],
+      code: { code: match[3], lang: match[4] },
+    };
+  },
+  /**
```
LimeWub, line 34: I'm realising today I could do this significantly simpler.
By doing `match[1] === 'true'` in the `fromBlock` 👀 
Might change and test it real quick as it avoids the below complication if it works.

LimeWub: I've replaced this commit with the simpler version.

package.json
```diff
@@ -3,19 +3,20 @@
   "name": "design-system",
   "workspaces": [
     "lib",
-    "site"
+    "documentation"
   ],
   "scripts": {
-    "dev:lib": "cd lib && NODE_ENV=development rollup -c -w",
-    "build:lib": "cd lib && NODE_ENV=production rollup -c",
+    "dev:lib": "cd lib && yarn build:proptypes && yarn dev",
```
thomasdigby, line 9: Is it necessary/useful to build proptypes in dev? Could we run it within `dev:docs`/`build:docs` instead?

LimeWub: In this iteration `docs` is the documentation website.
There is no longer a `build:docs` which generates proptypes from lib and a dist from the mdx files.

Proptypes generation itself is part of the lib package.json so I moved it to this command 👀 . There could be a `proptypes:lib` separate command to `cd lib` and run its own command tho.

I don't know if it is useful building the proptypes in dev...
Maybe (???) it is so that the latest (pre-changes) proptypes are reflected in the component props table (?) unsure. 
Tbf, I did the build version first then copied it and changed build to 'dev' for the dev version here.

Is there a benefit to splitting this command out?

documentation/components/markdown/editor-component/ComponentProps.tsx
```diff
@@ -0,0 +1,137 @@
...
+import {
+  Text,
+  Flex,
+  Icon,
+  Link,
+  Tooltip,
+  Table,
+  Box
+} from '@atom-learning/components'
+import docgen from '@atom-learning/components/dist/docgen.json'
```
thomasdigby, line 11: I wonder if it makes sense to move this docgen script/output to the documentation site as we're making this structural change. Its only used for the documentation site, so actually we should have made this change when we moved the docs site into this repo.

Not a biggie but it would simplify the build step for the library and reduce the number of files bundled in the published .tar for `@atom-learning/components`

documentation/components/app/Header.tsx
```diff
@@ -82,18 +82,19 @@ const StyledNavigation = styled('nav', {
   }
 })
 
-export const Navigation: React.FC<NavigationProps> = ({ items, version }) => {
+const version = buildConstants['version']
```
thomasdigby, line 85: Is this version going to be automated with each release?

documentation/components/admin/widget/UuidWidget.jsx
```diff
@@ -0,0 +1,48 @@
...
+    );
+}
+
+// (!) We can't use functional components to define a NetlifyCMS ControlComponent!
+// If functional it throws `Uncaught TypeError: this.wrappedControlValid is not a function.`!
+class UuidControlComponent extends React.Component {
+  componentDidMount() {
+    const { value, onChange } = this.props;
+    if (!value) {
+      onChange(nanoid());
```
thomasdigby, line 27: Could we drop in a small `uid` package for this instead of using the function above? We use this in `core` and maybe even here in other parts of the lib 🤔 https://github.com/lukeed/uuid

thomasdigby (approved):

documentation/components/admin/widget/UuidWidget.jsx
LimeWub: Yes we could :) This was just me avoiding installing a dependency. :p I can replace with uuid.

package.json
thomasdigby: What would be the process for an engineer to run the sandbox locally, they'd run `yarn dev:lib` right? Ah no, they'd just go `yarn dev:sandbox` which would skip the need to run this build step at all. In which case, I think this is fine 👍

documentation/components/app/Header.tsx
LimeWub: This version will get fetched every time the pre-build node script runs which generates the `buildConstants`. This is every time the `build:docs` or `dev:docs` command runs, which will be on every release (when the site is rebuilt/updated) or any time a developer runs any of the above commands (quite often).

documentation/components/markdown/editor-component/ComponentProps.tsx
LimeWub: Imma have a quick look at that now.

documentation/components/admin/widget/UuidWidget.jsx
LimeWub: done

documentation/components/markdown/editor-component/ComponentProps.tsx
LimeWub: > Not a biggie but it would simplify the build step for the library and reduce the number of files bundled in the published .tar for @atom-learning/components

So umm, I don't think we want to do this. Do we not need the docgen for the types to be exported from this?

Either way - I'm a bit scared to touch this rn but maybe we can look at it separately?
ComponentProps and the docsgen it uses are a bit borked and show the wrong things/don't always show all the props anyway and we could use some time to look into and fix it up. Maybe if this could be separated we could do in that bit of work?

thomasdigby: Yeah, we can definitely look at it separately. The types are built separately to this docgen thing, so we have the ability to change this 👍

documentation/pages/api/auth.js
```diff
@@ -0,0 +1,17 @@
+// Props: https://github.com/vencax/netlify-cms-github-oauth-provider
```
LimeWub, line 1: Hey @electrocally and @timatomlearning 

The files that do the GitHub OAuth handshake are this one and `callback.js` below as well as [oauth.config.js](https://github.com/Atom-Learning/components/pull/398/files#diff-d488474cdc00118b771665486dda31a97287c02e5903d01061ac8eeb8c6a40b4) . Could you have a look and confirm if they look okay to you?

Wherever we deploy we need these three .env variables:
`OAUTH_CLIENT_ID`, `OAUTH_CLIENT_SECRET`, `ORIGIN` - the first two can be found in the OAuth app for this which I believe has already been created in the Atom Learning Organisation and Origin is the url we'll want to use without `https://`

Everything else mentioned as `process.env.WHATEVER` is optional and has not been tested but I'm leaving in case we want to use it in the future.

We will also need to update the `branch` and `api_root` to match the new url in [documentation/public/admin/config.yml](https://github.com/Atom-Learning/components/pull/398/files#diff-f06744a33030643dd5d32dfc5adc60773a7fc0095cc8acd2894f8c5a6db52bd6).

documentation/lib/build/constants.json
```diff
@@ -0,0 +1 @@
+{"version":"v2.31.1"}
```
LimeWub, line 1: gitignore me

documentation/pages/api/auth.js
LimeWub: Hey did you mean `res.render(script)` instead of `res.send`? Not sure what `resp` is 👀

electrocally: I did indeed! Sorry, that must have been a typo 😅

LimeWub: Hey, I tried this and it throws a 500 in the callback instead so I'll just leave it as the `.send` it seems to be important here :o

PR #399, opened 2022-09-28
fix(TopBar): add size variant

mzeenkala23:
Ticket: https://atomlearningltd.atlassian.net/browse/DS-129

**This PR is blocking this task** https://atomlearningltd.atlassian.net/jira/software/projects/STU/boards/1?selectedIssue=STU-729 as we need to make the changes to the DS Component before we can use the Component locally in the Student App.

From our discussion with @avirati  @hollg  @andoulla and @MiguelGDLR  we needed to allow the option to increase the height of the `TopBar` Component so it would be easier for the Student Pod to use and customise the `Topbar` to their specific needs. This was done by implementing a size variant for the `TopBar`, following these designs https://www.figma.com/file/KjgQh7eTwEfJmH6e56k9mX/DS-Documentation?node-id=2973%3A5433

During my discussion with @hollg he suggested the ideal solution would be if each size of the TopBar would automatically set the size of the `TopBarBrandLogo`  Component.  So to accomplish this I followed the stitches  [documentation](https://stitches.dev/docs/styling#target-a-react-component) as this is an implementation which has never been done before in the DS Repo. It also involves adding classes  and the `.toString`  method which we have been reluctant to use. But @hollg was fine with this as the Components using this implementation are specifically scoped to the `TopBar` and using this weird implementation would allow us to create a nice API for the consumer.

## TopBar and TopBarBrandLogo `md` size variant
Screenshot 2022-09-28 at 12 15 37

## TopBar and TopBarBrandLogo `lg` size variant
Screenshot 2022-09-28 at 12 16 24

### DIFF
--- a/lib/src/components/top-bar/TopBar.mdx
+++ b/lib/src/components/top-bar/TopBar.mdx
@@ -25,6 +25,40 @@ Alternatively, if you want to avoid wrapping groups of items with a `div` purely
 </TopBar>
 ```
 
+## TopBar Size Variant
+
+`TopBar` also takes in one of these size props `md | lg`, this size is used to dictate the height of the `TopBar` and the `TopBar.BrandLogo`.
+
+If size `md` is passed to the `TopBar` it will be 64px in height and the `TopBar.BrandLogo` will be 24px in height.
+
+```tsx preview
+<TopBar size="md">
+  <TopBar.Brand href="atomlearning.co.uk">
+    <TopBar.BrandLogo src={logo} />
+  </TopBar.Brand>
+</TopBar>
+```
+
+If size `lg` is passed to the `TopBar` it will be 96px in height and the `TopBar.BrandLogo` will be 32px in height.
+
+```tsx preview
+<TopBar size="lg">
+  <TopBar.Brand href="atomlearning.co.uk">
+    <TopBar.BrandLogo src={logo} />
+  </TopBar.Brand>
+</TopBar>
+```
+
+If no size prop is added to the `TopBar` by default it will have the size `md`.
+
+```tsx preview
+<TopBar>
+  <TopBar.Brand href="atomlearning.co.uk">
+    <TopBar.BrandLogo src={logo} />
+  </TopBar.Brand>
+</TopBar>
+```
+
 ## TopBar.Brand
 
 `TopBar.Brand` renders a styled link.
--- a/lib/src/components/top-bar/TopBar.mdx
+++ b/lib/src/components/top-bar/TopBar.mdx
@@ -27,12 +27,10 @@ Alternatively, if you want to avoid wrapping groups of items with a `div` purely
 
 ## TopBar Size Variant
 
-`TopBar` also takes in one of these size props `md | lg`, this size is used to dictate the height of the `TopBar` and the `TopBar.BrandLogo`.
-
-If size `md` is passed to the `TopBar` it will be 64px in height and the `TopBar.BrandLogo` will be 24px in height.
+TopBar has a default size of `md` which means that the default height of the component is 64px and the `Topbar.BrandLogo` is 24px.
 
 ```tsx preview
-<TopBar size="md">
+<TopBar>
   <TopBar.Brand href="atomlearning.co.uk">
     <TopBar.BrandLogo src={logo} />
   </TopBar.Brand>
@@ -43,16 +41,6 @@ If size `lg` is passed to the `TopBar` it will be 96px in height and the `TopBar
 
 ```tsx preview
 <TopBar size="lg">
-  <TopBar.Brand href="atomlearning.co.uk">
-    <TopBar.BrandLogo src={logo} />
-  </TopBar.Brand>
-</TopBar>
-```
-
-If no size prop is added to the `TopBar` by default it will have the size `md`.
-
-```tsx preview
-<TopBar>
   <TopBar.Brand href="atomlearning.co.uk">
     <TopBar.BrandLogo src={logo} />
   </TopBar.Brand>
--- a/lib/src/components/top-bar/TopBar.test.tsx
+++ b/lib/src/components/top-bar/TopBar.test.tsx
@@ -17,7 +17,7 @@ const ExampleTopBar = (size) => (
 )
 
 describe('TopBar component', () => {
-  it('renders without size prop', async () => {
+  it('renders', async () => {
     const { container } = render(<ExampleTopBar />, {
       container: document.body
     })
@@ -25,27 +25,13 @@ describe('TopBar component', () => {
     expect(container).toMatchSnapshot()
   })
 
-  it('has no programmatically detectable a11y issues, when no size prop is passed', async () => {
+  it('has no programmatically detectable a11y issues', async () => {
     const { container } = render(<ExampleTopBar />)
 
     expect(await axe(container)).toHaveNoViolations()
   })
 
-  it('renders with the size prop md', async () => {
-    const { container } = render(<ExampleTopBar size="md" />, {
-      container: document.body
-    })
-
-    expect(container).toMatchSnapshot()
-  })
-
-  it('has no programmatically detectable a11y issues, when the size prop md is passed', async () => {
-    const { container } = render(<ExampleTopBar size="md" />)
-
-    expect(await axe(container)).toHaveNoViolations()
-  })
-
-  it('renders with the size prop lg', async () => {
+  it('renders the lg variant', async () => {
     const { container } = render(<ExampleTopBar size="lg" />, {
       container: document.body
     })
@@ -53,7 +39,7 @@ describe('TopBar component', () => {
     expect(container).toMatchSnapshot()
   })
 
-  it('has no programmatically detectable a11y issues, when the size prop lg is passed', async () => {
+  it('has no programmatically detectable a11y issues, in the lg variant', async () => {
     const { container } = render(<ExampleTopBar size="lg" />)
 
     expect(await axe(container)).toHaveNoViolations()
--- a/lib/src/components/top-bar/TopBar.test.tsx
+++ b/lib/src/components/top-bar/TopBar.test.tsx
@@ -5,7 +5,7 @@ import * as React from 'react'
 
 import { TopBar } from '.'
 
-const ExampleTopBar = (size) => (
+const ExampleTopBar = ({size}) => (
   <TopBar size={size}>
     <TopBar.Brand href="atomlearning.co.uk">
       <TopBar.BrandName>Admin Panel</TopBar.BrandName>
--- a/lib/src/components/top-bar/TopBar.test.tsx
+++ b/lib/src/components/top-bar/TopBar.test.tsx
@@ -5,8 +5,8 @@ import * as React from 'react'
 
 import { TopBar } from '.'
 
-const ExampleTopBar = () => (
-  <TopBar>
+const ExampleTopBar = (size) => (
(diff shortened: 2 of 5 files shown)


MiguelGDLR (commented):
All looks good from my side, thanks

avirati (approved):

lib/src/components/top-bar/TopBar.mdx
```diff
@@ -25,6 +25,40 @@ Alternatively, if you want to avoid wrapping groups of items with a `div` purely
 </TopBar>
 ```
 
+## TopBar Size Variant
+
+`TopBar` also takes in one of these size props `md | lg`, this size is used to dictate the height of the `TopBar` and the `TopBar.BrandLogo`.
+
+If size `md` is passed to the `TopBar` it will be 64px in height and the `TopBar.BrandLogo` will be 24px in height.
```
andoulla, line 32: Can I suggest a rewording here?

```
TopBar has a default size of `md` which means that the default height of the component is 64px and the `Topbar.BrandLogo` is 24px.

<TopBar>
  <TopBar.Brand href="atomlearning.co.uk">
    <TopBar.BrandLogo src={logo} />
  </TopBar.Brand>
</TopBar>

 
If size `lg` is passed to the `TopBar` it will be 96px in height and the `TopBar.BrandLogo` will be 32px in height.


<TopBar size="lg">
  <TopBar.Brand href="atomlearning.co.uk">
    <TopBar.BrandLogo src={logo} />
  </TopBar.Brand>
</TopBar>
```
and remove the lines 52- 61

lib/src/components/top-bar/TopBar.test.tsx
```diff
@@ -17,17 +17,45 @@ const ExampleTopBar = () => (
 )
 
 describe('TopBar component', () => {
-  it('renders', async () => {
+  it('renders without size prop', async () => {
```
andoulla, line 20: No need for the `without size prop` bit here. We do at least one snapshot with the default values
```suggestion
  it('renders', async () => {
```

lib/src/components/top-bar/TopBar.test.tsx
```diff
@@ -17,17 +17,45 @@ const ExampleTopBar = () => (
...
     const { container } = render(<ExampleTopBar />, {
       container: document.body
     })
 
     expect(container).toMatchSnapshot()
   })
 
-  it('has no programmatically detectable a11y issues', async () => {
+  it('has no programmatically detectable a11y issues, when no size prop is passed', async () => {
```
andoulla, line 28:
```suggestion
  it('has no programmatically detectable a11y issues, async () => {
```

lib/src/components/top-bar/TopBar.test.tsx
```diff
@@ -17,17 +17,45 @@ const ExampleTopBar = () => (
...
     const { container } = render(<ExampleTopBar />)
 
     expect(await axe(container)).toHaveNoViolations()
   })
+
+  it('renders with the size prop md', async () => {
```
andoulla, line 34: these 2 tests are redundant as they assert the same thing as the first 2 tests.

lib/src/components/top-bar/TopBar.test.tsx
```diff
@@ -17,17 +17,45 @@ const ExampleTopBar = () => (
...
+    expect(container).toMatchSnapshot()
+  })
+
+  it('has no programmatically detectable a11y issues, when the size prop md is passed', async () => {
+    const { container } = render(<ExampleTopBar size="md" />)
+
+    expect(await axe(container)).toHaveNoViolations()
+  })
+
+  it('renders with the size prop lg', async () => {
```
andoulla, line 48:
```suggestion
  it('renders the lg variant', async () => {
```

lib/src/components/top-bar/TopBar.test.tsx
```diff
@@ -17,17 +17,45 @@ const ExampleTopBar = () => (
...
+    const { container } = render(<ExampleTopBar size="lg" />, {
+      container: document.body
+    })
+
+    expect(container).toMatchSnapshot()
+  })
+
+  it('has no programmatically detectable a11y issues, when the size prop lg is passed', async () => {
```
andoulla, line 56:
```suggestion
  it('has no programmatically detectable a11y issues, in the lg variant', async () => {
```

lib/src/components/top-bar/TopBar.tsx
```diff
@@ -36,30 +35,64 @@ const StyledRoot = styled('div', {
...
   mx: '$4',
   width: '100%',
   '@md': {
     mx: '$5'
+  },
+  variants: {
+    size: {
+      md: {
+        height: '$4',
+        [`& ${TopBarBrandLogo}`]: {
```
LimeWub, line 61: Would it not be more 'correct' to also give a size variant to the TopBarBrandLogo to do this within itself? Rather than targeting it from here?

elliotlaws: If there is ever going to be a case where the sizes of the `<TopBar />` and `<TopBarBrandLogo />` need to be changed independently then I think it would be good to be able to pass in the size into `<TopBarBrandLogo />` too. 

If not, then this implementation is probably the only way to go about the automatic sizing.

lib/src/components/top-bar/TopBar.mdx
mzeenkala23: yeah I think that looks better

lib/src/components/top-bar/TopBar.test.tsx
mzeenkala23: thanks will make this change

lib/src/components/top-bar/TopBar.test.tsx
mzeenkala23: will update

lib/src/components/top-bar/TopBar.test.tsx
mzeenkala23: will change this also

lib/src/components/top-bar/TopBar.test.tsx
mzeenkala23: and this

lib/src/components/top-bar/TopBar.test.tsx
mzeenkala23: and this too

lib/src/components/top-bar/TopBar.tsx
hollg: The requirement to have the logo's size always match the size of the top bar came from Miguel. We have other ways of rendering the Atom logo if we need it in other contexts so I feel pretty relaxed about locking this in to the size of `TopBar`.

I suggested this approach to @mzeenkala23 because I am keen for us to lean on the stuff that's already happening (stitches) where possible, rather than adding additional overhead (context) just for styling. The only other option I know of is for `TopBar` to check the name/type of all its children and manually add/override a `size` prop for the logo, but I _really_ dislike that pattern; it couples the implementation of the parent together with any additional children in a really obscure way.

We did have a discussion about the need to add `.toString()` etc and the fact that it isn't done much (if at all?) elsehwere in the codebase but imo it's lower risk to introduce it in a component that is _very_ narrowly scoped to being used within `TopBar` rather than in some common top-level component wich will quickly appear all across `atom-core`.

I'm happy to say keep this implementation and then I will be the one to take the heat from any parties who feel very strongly about it after the fact. As @LimeWub suggested, speaking to the whole guild and formalising some patterns here is probably really useful, but at least we'll have a data point to use in that discussion 🤷🏻

lib/src/components/top-bar/TopBar.tsx
```diff
@@ -36,30 +35,64 @@ const StyledRoot = styled('div', {
       true: {
         boxShadow: '0px 4px 4px -2px rgba(31, 31, 31, 0.1);'
       }
+    },
+    size: {
```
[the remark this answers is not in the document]
mzeenkala23: ahh yes good spot i missed that thanks

lib/src/components/top-bar/TopBar.test.tsx
andoulla: don't know if you know you can just commit suggestions and then pull the changes down and run the tests once more

lib/src/components/top-bar/TopBar.tsx
andoulla: Sorry to pipe in and confuse everyone, but a discussion on one use case might be hard to get anywhere with it. This is a very specific use case where the topbar should never have any other combination of parent and children heights other than these 2. 

The ActionIcon also determines the child size based on the parent size. Everywhere else, we allow for flexibility (if memory serves me right) thinking of buttons etc. 

Where I'm getting at is to have a good productive discussion during the guild, we need more use cases so people can compare and be able to conclude.

LimeWub: @andoulla There's use cases multiple components usually the ones that have nested icons or do specific spacing based on the size of the parent. In terms of sizing SVGs/Icons we do this in ActionIcon, Icons in Chips and this logo and all 3 places do it differently. 
* ActionIcon passes the size prop directly to its children
* Chips uses context to store the size, ChipIcon pulls from context to size itself
* This PR uses CSS nesting to target the child component

So I think we can at least /mention/ it in the guild considering there's multiple solutions being used right now.


@mzeenkala23 I'm happy with this to stay as is for this PR; it's a legit way of doing it and there doesn't need to be a change as there's no specific pattern for it agreed yet ^^

lib/src/components/top-bar/TopBar.tsx
```diff
@@ -36,30 +35,64 @@ const StyledRoot = styled('div', {
...
+        height: '$4',
+        [`& ${TopBarBrandLogo}`]: {
+          '&[src$=".svg"]': {
+            height: 24,
+            width: 'auto'
+          }
+        }
+      },
+      lg: {
+        height: 80,
```
LimeWub, line 69: Just spoke to Miguel - you only need to set the height on one of the 2 components  and then just let the items inside it center (whichever one makes sense - I think the nested one may be the one making more sense but u can test when u write the CSS).

The '80px' height for the chunky/image links will be handled in that component when it's built ^^

mzeenkala23: lovely thanks for the update

lib/src/components/top-bar/TopBar.test.tsx
```diff
@@ -5,8 +5,8 @@ import * as React from 'react'
 
 import { TopBar } from '.'
 
-const ExampleTopBar = () => (
-  <TopBar>
+const ExampleTopBar = (size) => (
```
LimeWub, line 8: Does this definitely work? 
Does it not need to be `({size})` rather than `(size)` to get size from the destructured props?

mzeenkala23: ahh yes you are right will update

LimeWub: (I'm confused because the snapshot below does have the `lg` version so it seems it has worked BUT I don't get how :p This might just be me misunderstanding this 👀 )

mzeenkala23: yeah pretty weird

andoulla (approved):

LimeWub (approved):

hollg (approved):

andoulla (approved):

PR #400, opened 2022-09-30
Fix: Render TopBar logo in docs

hollg:
The Atom logo wasn't displaying in the `TopBar` docs page because it was relying on an import for the image itself, and the preview code blocks don't support imports.

Jira ticket: https://atomlearningltd.atlassian.net/browse/DS-131?atlOrigin=eyJpIjoiYzRlNGFiZGVkZWVkNDdjOWE1MWE4NThmZWEzYzNlNDUiLCJwIjoiamlyYS1zbGFjay1pbnQifQ

Fixed by referencing a publicly accessibly version of the image from our website CMS.

Before:
image

After:
Screenshot 2022-09-30 at 09 22 43


package.json
```diff
@@ -42,6 +42,6 @@
     }
   },
   "engines": {
-    "node": "^14.20.1"
+    "node": "^14.20"
```
hollg, line 45: This was added yesterday to force Digital Ocean to use a version of node that was able to compile the docs site without whining too much. I had to nudge it down here to a version that our GitHub Actions setup can use for the Test flow.

elliotlaws (approved):

thomasdigby (approved):

PR #401, opened 2022-09-30
Collapsible component

Mhoog:
https://user-images.githubusercontent.com/2501587/193253222-1f9ffbc7-2503-4a2b-b369-b2ea67777da9.mp4


lib/package.json
```diff
@@ -130,6 +130,7 @@
     "@radix-ui/react-accordion": "0.1.6",
     "@radix-ui/react-alert-dialog": "0.1.7",
     "@radix-ui/react-checkbox": "0.1.5",
+    "@radix-ui/react-collapsible": "^1.0.0",
```
thomasdigby, line 133: @hollg I think we're going to need that 1.0.0 update across all packages sooner rather than later. Do you want to chat about it next week and we can get it over the line?

lib/src/components/collapsible/Collapsible.mdx
```diff
@@ -0,0 +1,31 @@
...
+- `Collapsible`
+- `Collapsible.Trigger`
+- `Collapsible.Content`
+
+These components have some minimal styling applied to align the items within and provide space between them. The `Collapsible.Trigger` component also renders a Chevron icon to indicate the current state of the collapsible.
+
+This component is based on the `Collapsible` component from Radix UI, you can find more information in [their docs](https://www.radix-ui.com/docs/primitives/components/collapsible).
+
+```tsx preview
+<Collapsible space={3}>
```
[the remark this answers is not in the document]
Mhoog: Yes it adds spacing between `Trigger` and `Content`, and spacing around the two:
https://github.com/Atom-Learning/components/pull/401/files#diff-ef4d0d2c20d772d2d7c173f38b2b372b0402470ec82066410faf12ab829fce6fR9

MiguelGDLR (commented):
Hey, I didn't have a design ready for this component yet. I think it's fine to release v.1 now and have it improved later on when I have time to do research. I already have a ticket on the backlog for this. Thanks for tagging @thomasdigby

lib/src/components/collapsible/Collapsible.Content.tsx
```diff
@@ -0,0 +1,34 @@
...
+    '&[data-state="closed"]': { animation: `${close} 200ms ease-out` }
+  }
+})
+
+export const CollapsibleContent = React.forwardRef<
+  React.ElementRef<typeof StyledContent>,
+  React.ComponentPropsWithRef<typeof StyledContent> & { showDivider?: boolean }
+>(({ children, showDivider, ...props }, ref) => (
+  <StyledContent ref={ref} {...props}>
+    {showDivider && <Divider />}
```
LimeWub, line 31: Why not let the user pass in `Divider` as a child if they want to - rather than use the `showDivider` prop?

thomasdigby: Agree with @LimeWub , I had a similar thought. Maybe worth discussing again if need be when @MiguelGDLR finalises a consistent design for it

Mhoog: The main reason was to keep the space consistent.

Since you control the spacing with the `space` attribute, it would mean that you'd have to set the margin on the `Divider` manually. Something that could easily be overlooked or done wrong when setting/updating the size on the `Collapsible` component.

Happy to remove it from this component though, if you think that's not an issue (or an issue that we should solve this way).

lib/src/components/collapsible/Collapsible.tsx
```diff
@@ -0,0 +1,53 @@
+import { Root } from '@radix-ui/react-collapsible'
+import * as React from 'react'
+
+import { styled, theme } from '~/stitches'
+
+import { CollapsibleContent } from './Collapsible.Content'
+import { CollapsibleTrigger } from './Collapsible.Trigger'
+
+const space = Object.keys(theme.space).reduce(
```
LimeWub, line 9: Rather than doing this space calculations and `&` nesting - could you perhaps do spacing using `Stack` ? Perhaps by doing `<StyledRoot as={Stack} />` further down and letting the space/ gap work with Stack's gap?


So it's kinda like 
```
<StyledCollapsibleRoot as={Stack} {...rest} direction="column" /> // {...rest} includes things like `gap` from Stack and also whatever props Collapsible's root exposes
     {children}
</StyledCollapsibleRoot>
```

and used like:
```
<Collapsible gap={2} /> 
    <Collapsible.Trigger>Trigger stuff</Collapsible.Trigger>
    <Collapsible.Content><Divider />Content stuff</Collapsible.Content>
</Collapsible>
```

lib/src/components/collapsible/Collapsible.tsx
```diff
@@ -0,0 +1,53 @@
...
+type TCollapsible = React.ForwardRefExoticComponent<
+  React.ComponentProps<typeof StyledRoot>
+> & {
+  Content: typeof CollapsibleContent
+  Trigger: typeof CollapsibleTrigger
+}
+
+export const Collapsible = React.forwardRef(
+  ({ children, css, ...props }, ref) => (
+    <StyledRoot ref={ref} css={{ width: '100%', ...css }} {...props}>
```
LimeWub, line 46: this width 100% could be in the `styled` for `StyledRoot` instead of using css here?

lib/src/components/collapsible/Collapsible.Content.tsx
Mhoog: image

The space is inside the `Trigger` (to make the clickable area as big as possible) and inside the `Content` (except the top, to prevent double space there).

The divider has to be rendered inside the content, otherwise, it would be visible at all times, instead of only when the content is expanded.

You then want to have space below the `Divider` to match the space above it created by the padding in `Trigger`.

lib/src/components/collapsible/Collapsible.tsx
Mhoog: I wanted to use `padding` instead of `gap` / `margin` to make sure the clickable area of the `Trigger` would be as big as possible.

My initial thought was to pass the `space` prop down to the `children` so that I didn't need the messy CSS, but then I'd have to define the space variants for both the `Trigger` and `Content` components (+ code to manually add props in the root component). Might make it messier again 😕 

What do you think?

lib/src/components/collapsible/Collapsible.Content.tsx
LimeWub: Ah I see. I hadn;t noticed this was doing padding for the spacing.

I think the main problem here is that this is not designed. If it were designed we'd probably have 1-3 sizes for the Trigger that are basically the button sizes (so padding and border bottom if we want) that would be t-shirt size like our other components and you would just have an API like:
```
<Collapsible /> 
    <Collapsible.Trigger size="sm">Trigger stuff</Collapsible.Trigger>
    <Collapsible.Content>Content stuff</Collapsible.Content>
</Collapsible>
```

or 
```
<Collapsible size="sm" /> // and you pass down the size to its children either as props or via context 
    <Collapsible.Trigger>Trigger stuff</Collapsible.Trigger>
    <Collapsible.Content>Content stuff</Collapsible.Content>
</Collapsible>
```

This way it's clear that you are not setting spacing - you are setting the size of the Trigger OR the size of the Collapsible (as perhaps the Content also needs appropriate padding depending on the size).

I just think these nested component should be dealing with their own size base based on the context they are in/props they have rather than naming their sizing "space"; as space implies a gap. And also as that `& > div` style targeting shouldn't be necessary if we were just styling the nested components themselves.

May I ask: Where's this component coming from without the design? Does it live in a design of a page you are working on? Does it need to support all our 'sizing'/'spacing' options?

Mhoog: The new question container has a collapsible element:

image

I took the name `space` because we use those values to describe `padding`, `margin`, etc.: https://design.atomlearning.technology/theme/layout-and-space

I think having `space` instead of mapping t-shirt sizes to specific space-variables makes this more flexible. If we went for t-shirt sizes it could always happen that the spacing needed isn't available and we'd have to adjust the t-shirt sizes again later.

I agree that adding styles to the individual components would be nicer. I'll implement the second solution, and allow overriding the passed down prop.

lib/src/components/collapsible/Collapsible.tsx
LimeWub: Ye I'd pass down a size to its children (or do it via a context) and let them handle it but I also wouldn't provide 10 options for the size

lib/src/components/collapsible/Collapsible.Content.tsx
LimeWub: You can also just do the 1 size you need for it for now and when this gets actually designed we adjust it ^^

LimeWub: I also kinda wonder since this hasn't been properly designed if it should go in the DS as is rn or if it should stay somewhere in core (?) @MiguelGDLR what do u think?

MiguelGDLR: The reason I didn't mind in this case if we released the first version on the DS is because this way it will be more noticeable that we need to improve it in the future and it will be visible to designers. What's in Core is not visible to us.  If you see a problem then it can go to Core.  I already have the ticket in the backlog to design it in the future.

hollg: @Mhoog 
> The main reason was to keep the space consistent.

>Since you control the spacing with the space attribute, it would mean that you'd have to set the margin on the Divider manually. Something that could easily be overlooked or done wrong when setting/updating the size on the Collapsible component.

Could you show an example of what the code would look like if we took the "bring your own children" approach with the `Divider`? I'm not sure I follow what margin we're talking about.

I'm generally in favour of that approach anyway, but especially here since we know @MiguelGDLR will come back to this and give it a full design in the future. Since we know it's going to change, I'd be keen to start with the least-opinionated version that we can since I suspect that will make it as easy as possible to adapt later.

lib/src/components/collapsible/Collapsible.tsx
Mhoog: Updated the components so that you can set `space` on all three. If you only set it on the root component, it will pass to the children.

I've kept it flexible with the different `space` options now, but we can review that and maybe choose a few (or just one) once there is a design in place.

lib/src/components/collapsible/Collapsible.Content.tsx
LimeWub: @Mhoog  Do you need more than 1 space option for your use case?

lib/src/components/collapsible/Collapsible.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+    render(<Component />)
+
+    const trigger = screen.getByRole('button', { name: triggerText })
+    const content = screen.getByTestId('content')
+
+    expect(trigger).toHaveAttribute('data-state', 'closed')
+    expect(trigger).toHaveAttribute('aria-expanded', 'false')
+
+    expect(content).toHaveAttribute('data-state', 'closed')
+    expect(screen.queryByText(contentText)).toBeNull()
```
kerryambt, line 50: is `.not.toBeInTheDocument()` better here? (and same below)

lib/src/components/collapsible/Collapsible.tsx
```diff
@@ -0,0 +1,51 @@
...
+        if (
+          !React.isValidElement(child) ||
+          (child.type !== Collapsible.Trigger &&
+            child.type !== Collapsible.Content)
+        )
+          throw new Error(
+            `Only ${Collapsible.Trigger.displayName} and ${Collapsible.Content.displayName} are valid children for ${Collapsible.displayName}.`
+          )
+
+        return React.cloneElement(
```
kerryambt, line 35: we have a `passPropsToChildren` util, I think with one small change to add an option to not render children if they aren't a specific type (rather than just not passing props) we could avoid duplicating a lot of this logic. thoughts?

lib/src/components/collapsible/Collapsible.tsx
```diff
@@ -0,0 +1,51 @@
...
+            React.ComponentProps<
+              typeof Collapsible.Trigger | typeof Collapsible.Content
+            >
+          >,
+          { space: child.props.space || space }
+        )
+      })}
+    </StyledRoot>
+  )
+) as TCollapsible
```
kerryambt, line 46: is there a reason we type cast here rather than typing the component itself in `forwardRef`?

Mhoog: Adding compound components while using `forwardRef` returns Typescript errors:
https://stackoverflow.com/questions/70202711/how-to-attach-a-compound-component-when-using-react-forward-ref-property-does-n

As far as I could find, there's not really a clean way around it 🙁

lib/src/components/collapsible/Collapsible.Content.tsx
Mhoog: Not for my use case. But I don't think I should set `$5` (which is needed for my use case) as the default space option, as that's quite big.

lib/src/components/collapsible/Collapsible.tsx
kerryambt: ah that's annoying. this is probably best then :)

kerryambt (approved):

LimeWub:
Hey, for what it's worth I don't think this component should be going in the DS.

We have had a chat about what 'Collapsible' does (vs Accordion), and I'm of the opinion that collapsible functionality is really too generic to pinpoint one use/design of and stick it in the DS. IMO, there's a naming and a use-case issue, one component with the name of `collapsible` would just not be enough to cover all `collapsible` use cases while it occupies that very generic name. Collapsible is very basic, low level functionality. It is used for a 'collapsible field' for the promo code, it is a singular accordion, it can be used for sections of information in various places that don't look like this even for things that don't do an accordion like animation - it's just a show/hide. There are multiple components that can have collapsible functionality that do not look like this use case.

Even if it did go in as some super generic collapsible, as is that is not the case because it has styles to suit the student app use-case and an API which is not inspired from a specific design; which would have all the options accounted for and consideration for how this can be most reusable in terms of fitting into the designs. Currently this API is made up and doesn't match other APIs we have worked on recently; that is because it has not been thought off by designers. From that perspective alone I wouldn't want this to get merged have us start using something that does not match how we usually structure APIs only to potentially have to replace its usecases whenever it is designed - if it even fits with this specific design in the end.

I understand the argument about putting things in the DS to have visibility for them needing to be designed - but I disagree with making up APIs for components which may become a pattern or might not. (So I suppose I disagree with the argument?)

I would not be against this component being developed, according to specs produced by a designer, with options to match our current/expected use cases - if there are enough usecases for this specific design, with a name less vague than collapsible (unless a general decision had been made about this being the only or the by-far most used "collapsible")

Personally I think this should be simplified for your use case and added in core, for now.

Mhoog:
@LimeWub, 

I agree that we have loads of different use cases for this `Collapsible` that currently wouldn't fit the styling of this component. I do think it would provide value as a Primitive to make sure we create collapsible elements in our app that are accessible. Currently we usually just hide/show a `Box` without any additional context.

Once we have a `Collapsible` primitive in the design system, we can then create different components that make use of this primitive for different use cases.

LimeWub:
@Mhoog We already have an accessible collapsible primitive which is available, and that's the one from [radix-ui/collapsible](https://www.radix-ui.com/docs/primitives/components/collapsible). People should just use this directly when they need collapsing rather than a show/hide `Box`. I don't think this component in the DS is the solution but I do think people should be familiar with what radix itself provides. 🤷 

Perhaps there would be an argument for a demo/documentation and re-export of `radix-ui/collapsible` directly (with no additions, literally a re-export) so that people can see it in a design system as something they can use and the import from radix happens there. I'm not for adding any additional API currently and imo as long as we don't have a specific narrowed down design (if we ever get one) people should be aware and use the unstyled/unopinionated primitive directly.

PR #402, opened 2022-10-04
feat: added util to create variants based on theme tokens

Mhoog:
Added a small utility function to create a set of Stitches variants based on tokens in our theme. We were doing this in a few different components, and as I was adding this for our Collapsible component (#401) I thought it would be good to have a util function instead.

### Example
```ts
createThemeVariants('space', { p: '$key' })

// { 
//   0: { p: '$space$0' },
//   1: { p: '$space$1' }, 
//   ...etc
// }


thomasdigby (approved):

hollg (approved):
Neat idea!

PR #403, opened 2022-10-07
Feat/DS-56 number input

elliotlaws:
Ticket - https://atomlearningltd.atlassian.net/browse/DS-56
Designs - https://www.figma.com/file/KjgQh7eTwEfJmH6e56k9mX/DS-Documentation?node-id=7252%3A48640

Implementation
- Created NumberInput component form primitive 
- Created NumberInputField component form field 
- Followed accessibility guidelines for `spinbutton` see [MDN docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/spinbutton_role#keyboard_interactions)

Screenshots - NumberInput
image

image

Screenshots - NumberInputField
image

image


lib/src/components/number-input/NumberInput.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+```tsx preview
+<NumberInput
+  name="age"
+  max={11}
+  disabledTooltipContent={{ increment: 'Ages over 11 are not allowed' }}
+/>
+```
+
+The input box will allow out of range numeric values, when an `onBlur` event is triggered the value is validated against the min and max values.
```
LimeWub, line 26: I do wonder how this will work with generic form validation. I don't thiiiiink we have other components that validate themselves when onBlur (lmk if I'm wrong). I think it makes sense that this one does but curious if this difference in behaviour may cause some issues.

What does it do if it validates wrong? Does it set the value back to a valid one or just error?

lib/src/components/number-input/NumberInput.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+## Keyboard interactions
+
+A negative `tabindex` has been added to the increment and decrement buttons to remove them from the default tabbing order. Instead, users can use the below keys to interact with the component.
+
+- When you hit the ⬆ or ⬇ key, the input value will be increased or decreased by step.
```
LimeWub, line 32: Can we also do this for left and right key as per the docs linked below?

lib/src/components/number-input/NumberInput.mdx
```diff
@@ -0,0 +1,36 @@
...
+- When you hit the Home button, the value will jump to the min value.
+- When you hit the End button, the value will jump to the max value.
+
+See [MDN docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/spinbutton_role#keyboard_interactions) for more information on accessibility.
```
LimeWub, line 36: Noice!

lib/src/components/number-input/NumberInput.tsx
```diff
@@ -0,0 +1,200 @@
...
+import { Input } from '../input'
+import { NumberInputStepper } from './NumberInputStepper'
+
+export interface NumberInputProps {
+  name: string
+  min?: number
+  max?: number
+  step?: number
+  initialValue?: number
+  isDisabled?: boolean
```
LimeWub, line 17: Minor but `disabled` rather than `isDisabled` here is preferred. Just a code style thing where me and Gary have decided to prefer props with no `isX` even if boolean so the components look more like regular html elements would look like. Internally booleans with `is` is great and this preferance is perhaps up for debate if it doesn't work out but we're trying it for now :) 
same for `isReadOnly` => `readonly` (same as the html attribute). 

Feel free to rename them internally to have the `isReadonly` or `isWhatever` on them but just prefer the regular style for the props directly ^^

lib/src/components/number-input/NumberInput.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+    const isAtMax = value === max
+    const isAtMin = value === min
+
+    const clamp = React.useCallback(
+      (value: number) => Math.min(Math.max(value, min), max),
+      [max, min]
+    )
+
+    const validateOnBlur = React.useCallback(() => {
```
LimeWub, line 60: Rather than validate on blur here have you considered just forcing the value between min and max? Just so we don't get other conflicts with validation (I dunno if we would but I have seen this logic before....) Perhaps do this with another prop to `contain` them between the range or so. And let validation be handled by implementation?

lib/src/components/number-input/NumberInput.tsx
```diff
@@ -0,0 +1,200 @@
...
+         * We want to increase or decrease the input's value
+         * based on if the user the arrow keys.
+         *
+         * @see https://www.w3.org/TR/wai-aria-practices-1.1/#keyboard-interaction-17
+         */
+        const eventKey = event.key
+
+        const keyMap: Record<string, React.KeyboardEventHandler> = {
+          ArrowUp: increment,
+          ArrowDown: decrement,
```
LimeWub, line 128: ArrowRight and ArrowLeft too :3

lib/src/components/number-input/NumberInput.tsx
```diff
@@ -0,0 +1,200 @@
...
+          icon={Minus}
+          css={{
+            borderRight: 'none',
+            borderRadius: '$0 0px 0px $0'
+          }}
+          size={size}
+          isRootDisabled={isDisabled}
+          disabled={isAtMin}
+          disabledTooltipContent={disabledTooltipContent.decrement}
+          data-testid="decrement-btn"
```
LimeWub, line 182: Could this `testid` perhaps be added directly in the test rather than from here so we're not always adding it to the document?

lib/src/components/number-input/NumberInput.tsx
```diff
@@ -0,0 +1,200 @@
...
+        />
+        <Input {...inputProps} />
+        <NumberInputStepper
+          onClick={increment}
+          icon={Plus}
+          css={{ borderLeft: 'none', borderRadius: '0px $0 $0 0px' }}
+          size={size}
+          isRootDisabled={isDisabled}
+          disabled={isAtMax}
```
LimeWub, line 191: `disabled={isDisabled || isAtMax}` (assuming isDisabled doesn't already know it's disabled because isAtMax.

perhaps you can just pass `tooltipContent` in conditionally and that's what you use to show/not show the tooltip. Then it's easier to read and understand rather than `isRootDisabled`/`disabled` and all the extra logic in the button

lib/src/components/number-input/NumberInput.tsx
```diff
@@ -0,0 +1,200 @@
...
+      'aria-valuemin': min,
+      'aria-valuemax': max,
+      'aria-invalid': isInvalid,
+      'aria-valuenow': Number(value),
+      role: 'spinbutton',
+      ...(isInvalid && { state: 'error' })
+    }
+
+    return (
+      <Flex
```
LimeWub, line 165: I know it's just one line but could you use styled and moved this to the top? 
Unless you can do the disable fade per disabled element: Button or Input; in which case can this opacity be moved to the child elements (in their styled instead).

LimeWub (commented):
Nice stuff and good consideration of a11y!
Done a 1st pass and left some comments + questions ^^

lib/src/components/number-input/NumberInputStepper.tsx
```diff
@@ -0,0 +1,77 @@
...
+  '&:hover': {
+    bg: '$tonal50',
+    svg: {
+      color: '$tonal400'
+    }
+  },
+  '&:active': {
+    bg: '$tonal100',
+    svg: {
+      color: '$tonal500'
```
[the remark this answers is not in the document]
elliotlaws: This changes the colour to blue ($primary) on click, so going to leave this as is I think

lib/src/components/number-input/NumberInput.tsx
elliotlaws: I'm not really sure how to do this without creating extra complexity around adding additional props to the individual buttons.

lib/src/utilities/mergeRefs.ts
```diff
@@ -0,0 +1,38 @@
+import { useMemo } from 'react'
```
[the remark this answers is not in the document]
elliotlaws: `mergeRefs` is required here so that a forwarded ref and an internal ref (using `useRef`) can be used and passed to the component.

lib/src/components/number-input/NumberInput.tsx
```diff
@@ -0,0 +1,200 @@
...
+      name,
+      type: 'number',
+      value,
+      ...rest,
+      onChange: onInputChange,
+      onBlur: validateOnBlur,
+      onKeyDown,
+      size,
+      css: { borderRadius: '0px', width: '$6' },
+      ref: mergeRefs(inputRef, forwardedRef),
```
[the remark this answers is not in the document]
elliotlaws: See comment above

lib/src/components/number-input/NumberInput.tsx
elliotlaws: What's the benefit of using styled in this case? 

If it's only a one-liner and the component already accepts a `css` prop, I'm not sure I see the advantage of declaring it separately.

elliotlaws: Updated this so that all disabled styles live on the buttons/input

lib/src/components/number-input/NumberInput.tsx
elliotlaws: Added `showTooltip` to `NumberInputStepper` which now renders tooltip content conditionally based on this

lib/src/components/number-input/NumberInputStepper.tsx
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    return (
+      <Tooltip open={tooltipOpen} onOpenChange={onOpenChange}>
+        <Tooltip.Trigger asChild>
+          <StyledSpan aria-hidden>
+            <ActionIcon
+              tabIndex={-1}
+              appearance="outline"
+              css={{ ...getButtonStyles({ isRootDisabled, disabled }), ...css }}
```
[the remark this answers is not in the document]
elliotlaws: Updated to use `styled` and removed need for extra data to be passed

lib/src/components/number-input/NumberInputStepper.tsx
```diff
@@ -0,0 +1,77 @@
...
+interface NumberInputStepperProps
+  extends Omit<React.ComponentProps<typeof ActionIcon>, 'children' | 'label'> {
+  icon: React.FC<React.SVGProps<SVGSVGElement>>
+  isRootDisabled: boolean
+  disabledTooltipContent: string
+}
+
+const StyledSpan = styled('span', { zIndex: 1 })
+
+const getButtonStyles = ({ isRootDisabled, disabled }): CSS => ({
```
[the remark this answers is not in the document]
elliotlaws: Updated to use `styled` and removed need for extra data to be passed

lib/src/components/number-input/NumberInputStepper.tsx
```diff
@@ -0,0 +1,77 @@
...
+      disabledTooltipContent,
+      isRootDisabled,
+      css,
+      disabled,
+      ...rest
+    } = props
+    const [tooltipOpen, setTooltipOpen] = React.useState(false)
+
+    const onOpenChange = (open: boolean) => {
+      if (isRootDisabled || !disabled) return
```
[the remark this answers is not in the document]
elliotlaws: Updated this implementation so that `Tooltip.Content` only renders is `showTooltip` is true

lib/src/utilities/mergeRefs.ts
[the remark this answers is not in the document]
LimeWub: Is this something that could be solved by passing the ref up with `useImperativeHandle`?
It should be a 1 line added and had worked for Kerry when she needed to do something similar I believe:
See: https://github.com/Atom-Learning/components/pull/397/commits/25a4ac4516f3e498e1bbb773d6b2a2f3283ff490

She had also done a custom ref related solution but I think the one liner is simpler, less code and makes more sense (at least to me) :p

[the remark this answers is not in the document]
elliotlaws: Yep you're right, I've used `useImperativeHandle` before in a similar scenario, so not sure why I reached for the `mergeRefs` function 🤷‍♂️ 

I've deleted `mergeRefs` and now using `useImperativeHandle`. Tested and it still works

lib/src/components/number-input/NumberInput.tsx
elliotlaws: I did have this implemented originally, but @MiguelGDLR  didn't seem to like it - he said that forcing a value change automatically without user interaction is bad for UX

lib/src/components/number-input/NumberInput.mdx
elliotlaws: I've removed the validation logic inside of `NumberInput` now. 

I've updated `NumberInputField` to pass in `'aria-invalid': true` if `react-hook-form` returns an error

lib/src/components/number-input/NumberInputStepper.tsx
```diff
@@ -0,0 +1,77 @@
...
+      <Tooltip open={tooltipOpen} onOpenChange={onOpenChange}>
+        <Tooltip.Trigger asChild>
+          <StyledSpan aria-hidden>
+            <ActionIcon
+              tabIndex={-1}
+              appearance="outline"
+              css={{ ...getButtonStyles({ isRootDisabled, disabled }), ...css }}
+              ref={forwardedRef}
+              disabled={disabled}
+              label=""
```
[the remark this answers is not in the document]
elliotlaws: Yeah, you're right; I'll add the labels and also add props for these

Regarding the removal of focus on the buttons, happy to add this in comments in code

[See the MDN docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/spinbutton_role#:~:text=The%20spinbutton%20role%20is%20user,and%20decrement%20the%20spinbutton%20value.)

> Spinbuttons often have three components, including a text field that displays the current value, an increment button, and a decrement button. The text field is usually the only focusable component because the increment and decrement functions are keyboard accessible via arrow keys. Typically, the text field also allows users to directly edit the value.

lib/src/components/number-input/NumberInputStepper.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+    const onOpenChange = (open: boolean) => {
+      if (isRootDisabled || !disabled) return
+      setTooltipOpen(open)
+    }
+
+    return (
+      <Tooltip open={tooltipOpen} onOpenChange={onOpenChange}>
+        <Tooltip.Trigger asChild>
+          <StyledSpan aria-hidden>
```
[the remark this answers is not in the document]
elliotlaws: I'll remove `aria-hidden` as tabIndex should be enough here

I'll also add a zIndex of -1 to the span too to make things explicit

lib/src/components/number-input/NumberInput.tsx
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    const disabledTooltipContent = {
+      decrement: `Cannot enter values below ${min}`,
+      increment: `Cannot enter values above ${max}`,
+      ...disabledTooltipContentProp
+    }
+
+    const isAtMax = value === max
+    const isAtMin = value === min
```
LimeWub, line 64: If value has been typed in and is over `max` or under `min` this will be false.

Should this be using greater equal `>=` and less equal than `<=`?

lib/src/components/number-input/NumberInputStepper.tsx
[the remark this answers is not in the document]
LimeWub: That's great thanks!

lib/src/components/number-input/NumberInputStepper.tsx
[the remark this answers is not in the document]
elliotlaws: I believe these are new colours so will have to leave as is

lib/src/components/number-input/NumberInput.tsx
elliotlaws: Oops missed that, cheers!

lib/src/components/number-input/NumberInput.tsx
```diff
@@ -0,0 +1,206 @@
...
+      () => inputRef.current as HTMLInputElement
+    )
+
+    const stepperButtonLabels = {
+      increment: 'increment',
+      decrement: 'decrement',
+      ...stepperButtonLabelsProp
+    }
+
+    const disabledTooltipContent = {
```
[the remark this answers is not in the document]
kerryambt: would be inclined to set defaults, we do it some other components and I think there's a point for wanting things to be as simple to plug and play as possible? I definitely see either side of the argument though, it's hard to know what's 100% best but personally I'd have a default :)

LimeWub (approved):

lib/src/components/number-input/NumberInput.tsx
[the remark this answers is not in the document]
LimeWub: Imma also tag @thomasdigby for an opinion on this when he's back ^^

@kerryambt do we add defaults for translations on other components 👀 ? haven't noticed that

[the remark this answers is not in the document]
kerryambt: only example that comes to mind is in `Slider.Value` but we probably do it elsewhere :o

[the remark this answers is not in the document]
hollg: We have a ticket in the DS backlog to come up with a proper strategy for handling these pieces of copy in a scalable way. I'm happy with either of the approaches here in the meantime 🤷🏻

hollg (approved):

LimeWub:
FYI - I've left comments for the last commit of this: https://github.com/Atom-Learning/components/commit/0ae97f5325db3c19a23df12d46904396703c76c2#comments

Github doesn't resolve them to show in the full changes tabs - so here's the ez access

LimeWub:
LGTM :)

PR #404, opened 2022-10-13
Image `fluid` variant

LimeWub:
Currently the code assumes the images are always fluid. This leads us to always have to do code like:
```
<Box css={{width: 100, height: 100}}><Image src="" width="is ignored" css={{width: "is ignored"}} /></Box>
```

With this it no longer assumes that - so we can do:

```
<Image css={{width: 100, height: 100}} src="" />
```

or if we want the original "fluid image" behaviour:

```
<Image src="" fluid />
```

Basically: I've pulled the fluidity out in a `fluid` variant. It defaults to `false`, so I will need to update all uses of images in core depending on if they need fluidity or not.


hollg:
The `fluid` variant makes sense to me, but I've been fully default-values-should-be-false-pilled (by you!) now and therefore feel very icky about `fluid={false}`. Do we have a clear idea of how much extra work we're talking about for getting this into `atom-core` if we did default `fluid` to false?

LimeWub:
this many in core + probs a couple more from components: 
Screenshot 2022-10-13 at 11 11 59

I think I can probably just do a find and replace from `<Image` to `<Image fluid` tbh and have these "resolved"; then do a pass through and check I didn't break anything - perhaps even remove some `<Box />` wrappers where we've needed the hack. All in all - I don't think it will actually take this long tbf. I can probs do a version of it in half a day or so?

hollg:
@LimeWub cool, I'd be keen to go for that then. We can do a `canary` release or something with your work here too, so that you can make all the required changes in `atom-core` before we release the new "real" version of the lib. Keeps the time pressure of the work in `atom-core`! 

I don't think you'll have access to make that release, so feel free to ping me when you'd like it to happen.

hollg (approved):

LimeWub:
@thomasdigby  Just pinging you cause this PR has one approval in case it can go in tomorrow so I can update core with it as well :)

andoulla (approved):

lib/src/components/image/Image.tsx
```diff
@@ -7,16 +7,22 @@ import { Override } from '~/utilities/types'
...
+          width: 'auto'
+        },
+        '&[width][height]': {
+          height: 'auto'
+        },
+        '&[src$=".svg"]': {
+          width: '100%',
+          height: 'auto',
+          maxWidth: 'none'
+        }
```
LimeWub, line 23: ._. this is actually wrong lol.
```
 maxWidth: '100%',
  variants: {
    fluid: {
      true: {
        '&[width]': {
          width: 'auto' // useless ?
        },
        '&[width][height]': {
          height: 'auto' // useless?
        },
        '&[src$=".svg"]': { // This is what's important
          width: '100%',
          height: 'auto',
          maxWidth: 'none'
        }
```
So should be something like: 
```
 maxWidth: '100%',
  variants: {
    fluid: {
      true: { // This is what's important
          width: '100%',
          height: 'auto',
          maxWidth: 'none'
        }
```

LimeWub: Actually this is.... supposed - to work. But I don't get why this is supposed to be a 'fluid' image D:

PR #406, opened 2022-10-17
Fix: field element prop warnings for `fieldId`

Marcdup1:
### Description
When using `<TextareaField />` or `<SelectField />` components TS complains saying that the `fieldId` prop is required because their component props extend from `FieldWrapperProps`, but when you add the prop we get a warning in the console: `React does not recognize the 'fieldId' prop on a DOM element.`. In https://github.com/Atom-Learning/components/commit/ee8f9ad66f74e26c047b37c5436e27a329e2ff47 I omit `fieldId` from the props, however in e0b149a29507720f31f69930b384c014c3af3cc9 I created a type that we can use for field elements that are wrapped with `FieldWrapper` in order to not repeat the type and keep it consistent.

image

### DIFF
--- a/lib/src/components/checkbox-field/CheckboxField.tsx
+++ b/lib/src/components/checkbox-field/CheckboxField.tsx
@@ -2,18 +2,14 @@ import * as React from 'react'
 import { Controller, useFormContext } from 'react-hook-form'
 
 import { Checkbox } from '~/components/checkbox'
-import { InlineFieldWrapper } from '~/components/field-wrapper'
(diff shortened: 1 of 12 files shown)


elliotlaws (approved):

Marcdup1:
> Hey @Marcdup1 I've just merged #409 which achieves the same thing, but your solution is much more extensive. Feel free to override the work I've done

Oh I see, okay thanks for flagging @elliotlaws, I will resolve the conflicts quickly

kerryambt (approved):

PR #408, opened 2022-10-18
Adding New Combobox with input supports validations and errors

valllentinnaa:


lib/src/components/combobox-field/ComboboxField.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+type ComboboxFieldProps = React.ComponentProps<typeof StyledComboboxField>
+
+export const ComboboxField: React.FC<ComboboxFieldProps> & {
+  Input: typeof ComboboxFieldInput
+  Popover: typeof ComboboxFieldPopover
+  List: typeof ComboboxFieldList
+  Option: typeof ComboboxFieldOption
+  OptionText: typeof ComboboxOptionText
+} = (props) => <StyledComboboxField {...props} />
```
kerryambt, line 23: should this not expose children as part of props and then compose like `<StyledComboboxField {...props}>{children</StyledComboboxField>`

lib/src/components/combobox-field/ComboboxFieldInput.tsx
```diff
@@ -0,0 +1,101 @@
...
+type ComboboxFieldInputProps = InputProps & {
+  css?: CSS
+  description?: string
+  label: string
+  name: string
+  prompt?: { link: string; label: string }
+  validation?: ValidationOptions
+}
+
+const Input = styled(BaseComboboxInput, {
```
kerryambt, line 22: Is there no way we can use our `Input` component here to keep DS styles consistent?

LimeWub:
We have a combobox component as is - can we not wrap that in a field to show validations etc rather than redoing it?

hollg:
I gather this PR is for [this ticket](https://atomlearningltd.atlassian.net/jira/software/projects/PAG/boards/12?selectedIssue=PAG-291)?

I think we need to take some time to get aligned on what is needed here and how it should be implemented. The fact that this work builds a whole new `Combobox` API from scratch and hasn't had any input from @MiguelGDLR is ringing alarm bells for me.

As @LimeWub mentioned, we already have a `Combobox` component in the library and it's not immediately clear to me why that `Combobox` isn't being composed into a field with `FieldWrapper` here, which is how the other field components work. If `Combobox` itself lacks functionality, then I think we should be looking at extending/improving it rather than introducing a separate `Combobox` API altogether.

We have a [ticket in the DS backlog](https://atomlearningltd.atlassian.net/jira/software/projects/DS/boards/27?selectedIssue=DS-77&text=select) for building new versions of the `Select` component with single/multi-select variants, typeahead, rich content in the dropdown, etc which I suspect will cover much of the behaviour needed for this PR. But that is a complex piece of work so @LimeWub and I have been trying to plan it out ahead of time. That planning hasn't been completed yet, so it's a little alarming to see what seems like the same work going on already!

In general I really don't think we should be starting work to add such big chunks of new UI code to the DS library without design input from @MiguelGDLR and/or technical input from GT pod engs. We're trying very hard to keep the library consistent and high-quality, and when we don't get included in these dicsussions until the work is already done, that often involves asking for big rewrites because established conventions haven't been followed etc. Similarly, we all need to be aware of the context we're working in, i.e. what other work is planned or already happening in the library (see above). Miguel manages to the DS backlog Jira board and GT pod are heavily involved with it so we can help with that context too.

I'll bring this up at the FE roundtable today to make sure everyone is on the same page and provide an opportunity to hash out any details about the general process but maybe we can get some time this week to dig into the right approach for this PR in particular? @kerryambt I'd be keen for you to be involved in that too if possible?

PR #410, opened 2022-10-20
RFC: DataTable Adapter pattern

hollg:
# The problem
In the GT pod we've come up against the need to get data _out_ of a `DataTable`. We're going to introduce table rows that can be re-ordered via drag and drop. This will change the internal state of the `DataTable`, but we also need to be able to synchronise external changes. In this particular case, we eventually need to pass the updated order into an API request to update the database after the user has confirmed the changes in the UI.

This RFC is an attempt to codify a generic pattern for solving this problem. **It is not a suggestion for anything new to _build_ in the library**.

# Suggest solution: adapter pattern
The temptation to build a new abstraction into the `DataTable` API is _huge_, but while we're relatively early in its lifetime and we're yet to see a range of different use cases for this functionality play out, a well-documented and simple-to-use solution that lives wholly in the implementation seems safest to me.

My suggestion for that pattern is the creation of specific "adapter" components which subscribe to `useDataTable()` and then use its output to fire events, sync with a database over the network, update other context providers, etc.

```tsx
const ExampleDataTableAdapter = () => {
  const { getState } = useDataTable()

  const {
    pagination: { pageIndex }
  } = getState()

  console.log('Page index is:', pageIndex)
}

const App = () => {
  return (
      <ExampleDataTableAdapter>
        <DataTable.Table sortable css={{ mb: '$4' }} />
        <DataTable.Pagination pageSize={5} />
      </ExampleDataTableAdapter>
    </DataTable>
  )
}
```

# Alternatives considered

One possible solution for the use case outlined above would be to include a separate `DataTable.DnDContextProvider` component in the `DataTable` API which takes an `onSortChange` prop. We could include all the logic we need in that function without problem. 

But we will have a nicer overall API if we abstract that context provider away so that the consumer doesn't need to use it directly. Ideally we want to replicate the pattern of the `DataTable.Pagination` pattern wherever possible; in this case I think that means passing `sortable` (or some other prop) to the table rows and having the `DataTable` components figure the rest out for you automatically. 

Additionally, that `DataTable.DndContextProvider` would only allow the consumer to sync changes based on the drag-and-drop sort order changing. What if they want to sync changes based on something else? Do we need to add a new such component for every use case, each taking a different type of event handler? This would suck.

An obvious alternative idea might be to build a universal `onChange` prop into the `DataTable` itself. It could take the output of the `useDataTable` hook as its input. But we wouldn't be able to call _other_ hooks in that function (because it wouldn't be either a React component or a hook itself). I suspect it will be common to want to synchronise changes in a `DataTable` with state in another context provider (and this will be what we hope to do in GT Pod's initial use case).

The workaround for this problem would be to call the second hook in the same component that renders the `DataTable` and use its output in the function passed to `onChange`.

### DIFF
--- a/lib/sandbox/index.tsx
+++ b/lib/sandbox/index.tsx
@@ -1,22 +1,58 @@
 import * as React from 'react'
 import * as ReactDOM from 'react-dom'
 import { reset } from 'stitches-reset'
-
-import { Box, Flex, globalCss } from '../src'
+import { createColumnHelper } from '@tanstack/react-table'
+import { Box, Flex, globalCss, DataTable, useDataTable, Tooltip } from '../src'
 
 globalCss({ ...reset, '*': { boxSizing: 'border-box' } })()
-
+const columnHelper = createColumnHelper<{
+  firstName: string
+  lastName: string
+}>()
 const App = () => (
-  <Flex
-    css={{
-      minHeight: '100vh',
-      justifyContent: 'center',
-      alignItems: 'center',
-      flexDirection: 'column'
-    }}
-  >
-    <Box />
-  </Flex>
+  <Tooltip.Provider>
+    <Flex
+      css={{
+        minHeight: '100vh',
+        justifyContent: 'center',
+        alignItems: 'center',
+        flexDirection: 'column'
+      }}
+    >
+      <Box css={{ width: '700px' }}>
+        <DataTable
+          data={[
+            { firstName: 'william', lastName: 'johnson' },
+            { firstName: 'patty', lastName: 'fontaine' }
+          ]}
+          columns={[
+            columnHelper.accessor('firstName', {
+              cell: (info) => info.getValue()
+            }),
+            columnHelper.accessor('lastName', {
+              cell: (info) => info.getValue()
+            })
+          ]}
+        >
+          <ExampleAdapter>
+            <DataTable.Table />
+            <DataTable.Pagination pageSize={1} />
+          </ExampleAdapter>
+        </DataTable>
+      </Box>
+    </Flex>
+  </Tooltip.Provider>
 )
+
+const ExampleAdapter = ({ children }) => {
+  const { getState } = useDataTable()
+  const {
+    pagination: { pageIndex }
+  } = getState()
+
+  console.log('pageIndex:', pageIndex)
+
+  return children
+}
 
 ReactDOM.render(<App />, document.getElementById('root'))
--- a/lib/src/components/data-table/DataTable.mdx
+++ b/lib/src/components/data-table/DataTable.mdx
@@ -154,3 +154,35 @@ If `DataTable`'s `isSortable` state is `true`, then `DataTable.Header` will be c
 ### Pagination
 
 `DataTable.Pagination` can be passed as a child to `DataTable` to render the pagination UI and configure the parent `DataTable` to paginate its data.
+
+## Adapter Pattern
+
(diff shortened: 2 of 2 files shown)


LimeWub:
Hey! In the example I think you mean this:

```
const App = () => {
  return (
        <ExampleDataTableAdapter />
        <DataTable.Table sortable css={{ mb: '$4' }} />
        <DataTable.Pagination pageSize={5} />
    </DataTable>
  )
}
```
as `ExamDataTableAdapter` doesn't return children - but also there's no need for it wrap the children anyway - in this example


> One possible solution for the use case outlined above would be to include a separate DataTable.DnDContextProvider component in the DataTable API which takes an onSortChange prop. We could include all the logic we need in that function without problem.

I see where you're going with this as an example. Logic like the `onSortChange` would be nicely encapsulated in a 'logic only' component nested in the context it needs to read from. However, The Dnd context is needed to wrap around children so it's not a 'logic only' component as it has a non-null return value even though it adds no extra DOM elements (but adds context). `DataTable.DndSortable` is the same style of component as `DataTable.Pagination` with the only change being that it wraps around other elements (the rows).

Another "example" I could think of is **_if_** we wanted to display content from the table outside of the TableContext entirely for some reason. ie. We have a header that's like **Table with context. Viewing X of Y results** which is not a table component but needs to read from table context. An adapter to just 'read the needed things and pass them up and out of the context' could be used then.

Although! Even on that case, what is stopping us from pulling the `Table` context further out - ie the same way I've wrapped a `<Tabs />` context much further out from the 2 places I needed to use them to feed the context into a greater area of the site.
Screenshot 2022-10-21 at 11 39 20

So:
```
<DataTable data={[]} columns={[]}>
  <header>
       <HeadingDisplayingDataTableRelevantInfo />
  </header>
  <section>
     <DataTable.Table />
  </section>
</DataTable>
```

> An obvious alternative idea might be to build a universal onChange prop into the DataTable itself. It could take the output of the useDataTable hook as its input.

`onValueChange` is how radix does it so I don't mind this idea. However, I would expect `onValueChange` to have output in the same format as the input from the DataTable. So basically if you wanted you could have 2 syncronised tables just by passing the `onValueChange` data from one *changed* table to the other.

so data in:
```
const data = [
  { name: 'chrissy', hobby: 'bare-knuckle boxing' },
  { name: 'agatha', hobby: 'crossfit' },
  { name: 'betty', hobby: 'acting' }
]
```

and data out:
```
const data = [
  { name: 'betty', hobby: 'acting' },
  { name: 'chrissy', hobby: 'muai thai' }
]
```

That's what I would expect and `onValueChange`/`onDataChange` to be doing. Not giving the entire context back.

Perhaps even have an `onColumnChange` for column modifications (if that's ever useful).

> I suspect it will be common to want to synchronise changes in a DataTable with state in another context provider (and this will be what we hope to do in GT Pod's initial use case).

Why? 

I don't think I'm fully getting what you're saying here:
> The workaround for this problem would be to call the second hook in the same component that renders the DataTable and use its output in the function passed to onChange.

Anyway, while I don't hate the idea for an Adaptor and it could make sense perhaps in some niche situation; the more I think about it the more I feel like just moving the `<DataTable />`  context higher up the hierarchy would solve the problems.
I don't think DndContext needs an adaptor or anything like that since it's already in the context.
I get a vibe that using an adaptor could be an anti-pattern of sorts (something in it feels like a bit of a hack) and I don't want us to overcomplicate things. If it's really truly needed we can deal with that then and there - but moving the context further up is probably simpler (?)

PR #411, opened 2022-10-24
Feat: Section, SectionItem components

avirati:
Ticket : https://atomlearningltd.atlassian.net/browse/STU-736
Designs - https://www.figma.com/file/KmSAa9g0YJc1lfctTVyl2W/Navigation?node-id=1225%3A16722

Implementation - 

1. Created `Section` component
2. Created `SectionItem` component

Screenshots - 

Screenshot 2022-10-24 at 12 24 03

Screenshot 2022-10-24 at 12 24 09


hollg:
> Thanks @avirati, has this been run past @MiguelGDLR or @hollg at all? I have a feeling that this may need to change to be useful within the context of the component library. I might even argue that this shouldn't go in the design system... keen to hear your thoughts @MiguelGDLR @hollg

I'm inclined to agree with Tom about this needing changes to make sense in this library. As an example, the `Section` component currently takes an optional `heading` prop which is conditonally rendered as an `h3`. So there's no way to use that component in a context that already has an `h3` heading without breaking the rules of semantic HTML. The way we normally approach these problems in the library is to create a composable API, i.e. one where the `Section` would render its `children` and the consumer could render the heading they want within if necessary. Perhaps you know that it should only ever be an `h3` because of your pod's specific requirements for it, but a component that can only be used in one app, in such a specific use case, isn't appropriate for this library really.

On the more general points, neither Miguel nor I knew anything about this and components should not be going into the library that haven't been through Miguel. He puts a _huge_ amount of work into ensuring that the library is consistent, serves all of our apps' varied needs, and is extremely thoroughly documented with regards to what contexts each component should be used in, things product designers need to be aware of, possible variants, responsive design, etc. We can't ask Miguel to keep going through that process on components that we've built without him and have already started using in the apps. This is what happened with the components that existed before he joined us, and it's been a really crazy amount of work to create a consistent framework that both supports the pre-existing components _and_ has internal consistency, solid design principles underpinning it etc.

I did talk about this in the FE roundtable on Thursday, and after a comment you made there @avirati I did describe that there is an existing process whereby anyone can add a request/suggestion to the DS backlog Jira board and Miguel will start the process of evaluating it and designing it if it should be in the DS. Do you know what the reason is that none of that has happened here?

I'm very conscious of discouraging people/teams from contributing to this library, which is the absolute _last_ thing I want to do, I promise. We need to follow processes for getting this stuff designed and implemented so that we can maintain the high quality of the library (and especially at the moment when our design resource is so stretched which makes the stakes of any one mistake higher) but we really do want to get as many people as possible contributing to the library! Miguel and I are both extremely happy to support pods/individuals in doing so, and to make it as smooth a process as possible for everyone involved, but we can only do so if pods follow the processes and involve us in the work.

I also want to make clear that there is no one individual who is at fault. It's unfortunate that PRs are always the first thing that raises these issues to our attention, because it means that this feedback always arrives first with engineers, but there's clearly a broader issue involving PMs, product design and and comms between _everyone_. 

We're going to start work today on a document to describe the rationale behind what we think should/shouldn't go into the design system library, because we've clearly not nailed comms on this so far. Perhaps we can also rethink how to communicate these processes more generally, and any feedback/suggestions you have on these topics would be very gratefully received. I appreciate it must be very frustrating to get a whole PR ready (the actual work for which looks great, by the way) and only then be told all this, and that's definitely something we want to avoid as much as possible by making sure that all the relevant people are involved early in DS work. 

For this specific component, I would suggest that your pod should discuss with Miguel whether it belongs in the design system library. Based just on the linked designs I wouldn't have thought that it does, but I'm conscious I could be missing a lot of context in terms of other in-progress designs using the same UI for different use cases, different apps etc.

lib/src/components/section/Section.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+interface ISectionProps extends React.ComponentProps<typeof Flex> {
+  heading?: string
+}
+
+export const Section: React.FC<ISectionProps> & {
+  Content: typeof Grid
+} = ({ heading, children, css }) => {
+  return (
+    <Flex
```
andoulla, line 15: can we use the abstraction that is used everywhere, where we do
```
const StyledContainer = styled(Flex, {....})
```

This allows grouping the styling and when we start adding variants it will be more readable.

lib/src/components/section/Section.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { Grid } from '../grid'
+import { Heading } from '../heading'
+
+interface ISectionProps extends React.ComponentProps<typeof Flex> {
+  heading?: string
+}
+
+export const Section: React.FC<ISectionProps> & {
+  Content: typeof Grid
+} = ({ heading, children, css }) => {
```
andoulla, line 13: 2 points here:
1. I think the heading should be part of the children if possible (so the parent has more ownership of what level it will be and if they need to skip it

2. doing `children, ...rest` will allow the parent to pass anything extra it needs without adding it to the signature of the component (think data-cy tags, aria-labels etc)

lib/src/components/section/SectionItem.mdx
```diff
@@ -0,0 +1,61 @@
+---
+title: SectionItem
+component: SectionItem,SectionItem.Icon,SectionItem.Heading,SectionItem.SubHeading
+description: The SectionItem component renders children in specific locations and with specific styles
+category: Content
+---
```
andoulla, line 6: I would add a note that the SectionItem needs to be used only within a section

lib/src/components/section/SectionItem.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+        <SectionItem.Icon is={CheckList} size="md" css={{ color: 'white' }} />
+        <SectionItem.Heading as="h4" css={{ fontFamily: '$body' }} size="sm">
+          A small heading
+        </SectionItem.Heading>
+        <SectionItem.SubHeading css={{ color: '$tonal300' }} size="sm">
+          A small subheading
+        </SectionItem.SubHeading>
+      </SectionItem>
+    )
+    expect(await axe(container)).toHaveNoViolations()
```
andoulla, line 26: We intentionally decided to separate the axe checks into their own tests, as we found that a) they are 2 different things and b) it masked that there was a snapshot failing if there were any accessibility issues. could we please split these

lib/src/components/section/SectionItem.tsx
```diff
@@ -0,0 +1,152 @@
...
+const slideRightAnimation = keyframes({
+  from: {
+    transform: 'translateX(-5px)'
+  },
+  to: {
+    transform: 'translateX(0)'
+  }
+})
+
+const containerCss: CSS = {
```
andoulla, line 42: Just to keep this consistent with everything else, there is not need to extract the CSS into its own prop. Also, this might be losing functionality that we get from stitches where you can define local CSS variables

avirati:
Closing this PR in favour of https://github.com/Atom-Learning/atom-core/pull/3132
As it was decided that this component is not a good candidate for DS components

PR #413, opened 2022-10-26
feat: add the Avatar component

D7Torres:
## Description
This PR adds a new component: `Avatar`

It follows [this design](https://www.figma.com/file/KjgQh7eTwEfJmH6e56k9mX/DS-Documentation?node-id=6280%3A45064)

[JIRA ticket](https://atomlearningltd.atlassian.net/browse/DS-44)

This PR includes all the states of the component, but not the group of avatars or the button with the down caret to the side.

## Screenshots of the docs
image
image
image
image
image
image
image
image
image
image


lib/src/components/avatar/Avatar.tsx
```diff
@@ -0,0 +1,130 @@
...
+  name
+}) => {
+  if (src) {
+    return (
+      <img src={src} alt={name ? `${name}'s avatar` : 'avatar'} width="100%" />
+    )
+  }
+
+  if (children) {
+    return <Box>{children}</Box>
```
D7Torres, line 86: I wanted to have `role="image"` but I got this error:

```
Elements with ARIA roles must use a valid, non-abstract ARIA role  jsx-a11y/aria-role
```

lib/src/components/avatar/Avatar.tsx
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  if (name) {
+    return <Box>{name[0].toUpperCase()}</Box>
```
D7Torres, line 90: I wanted to have `role="image"` but I got this error:

```
Elements with ARIA roles must use a valid, non-abstract ARIA role  jsx-a11y/aria-role
```

lib/src/components/avatar/Avatar.tsx
```diff
@@ -0,0 +1,130 @@
...
+  name,
+  disabled = false,
+  onClick,
+  ...rest
+}) => {
+  return (
+    <StyledAvatar
+      size={size}
+      disabled={disabled}
+      onClick={disabled ? null : onClick}
```
D7Torres, line 119: I need help here, I'm getting a TS error in `onClick` SyledAvatar's prop.
```
No overload matches this call.
```

If I change the `null` to `undefined` the error disappears but the same error appears in the `disabled` prop instead. I'm not sure what's going on here.

michalrozenek: I checkout to your branch and got this error

```
Type '(MouseEventHandler<HTMLDivElement> & (() => void)) | null | undefined' is not assignable to type 'MouseEventHandler<HTMLDivElement> | undefined'.

Type 'null' is not assignable to type 
 ```
 
 I changed to `undefined` and all is good. Disabled does not throw me any error.
 
 Second option just in case is maybe this way of onClick declaration
 
 ```
 onClick={() => !disabled && onClick?.()}
 ```

File: lib/src/components/avatar/Avatar.tsx [hunk 4]
```diff
@@ -0,0 +1,130 @@
...
+        size: '$7'
+      }
+    }
+  }
+})
+
+type AvatarContentProps = {
+  src?: string
+  children?: React.ReactNode
+  name?: string
```
michalrozenek, line 66: Ignore the `children` type (it's just for playground purposes)

Zrzut ekranu 2022-10-27 o 13 29 15

michalrozenek: Try

```
type AvatarContentProps =
    { src?: string, name?: never, children?: never } |
    { src?: never, name?: string, children?: never } |
    { src?: never, name?: never, children?: JSX.Element, }
```

lib/src/components/avatar/Avatar.tsx
D7Torres: Are you sure `disabled` does not cause an error? To me, it took time to appear.
image

The error:
```
No overload matches this call.
  Overload 1 of 3, '(props: Omit<Omit<Pick<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, "key" | keyof HTMLAttributes<...>> & { ...; }, "css"> & TransformProps<...> & { ...; }, "size" | ... 1 more ... | "as"> & TransformProps<...> & { ...; }): ReactElement<...> | null', gave the following error.
    Type '{ children: Element; color?: string | undefined; translate?: "yes" | "no" | undefined; slot?: string | undefined; style?: CSSProperties | undefined; title?: string | undefined; ... 252 more ...; onClick: (MouseEventHandler<...> & (() => void)) | undefined; }' is not assignable to type 'IntrinsicAttributes & Omit<Omit<Pick<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, "key" | keyof HTMLAttributes<...>> & { ...; }, "css"> & TransformProps<...> & { ...; }, "size" | ... 1 more ... | "as"> & TransformProps<...> & { ...; }'.
      Property 'disabled' does not exist on type 'IntrinsicAttributes & Omit<Omit<Pick<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, "key" | keyof HTMLAttributes<...>> & { ...; }, "css"> & TransformProps<...> & { ...; }, "size" | ... 1 more ... | "as"> & TransformProps<...> & { ...; }'.
  Overload 2 of 3, '(props: Omit<Omit<Pick<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, "key" | keyof HTMLAttributes<...>> & { ...; }, "css"> & TransformProps<...> & { ...; }, "size" | ... 1 more ... | "as"> & TransformProps<...> & { ...; }): ReactElement<...> | null', gave the following error.
    Type '{ children: Element; color?: string | undefined; translate?: "yes" | "no" | undefined; slot?: string | undefined; style?: CSSProperties | undefined; title?: string | undefined; ... 252 more ...; onClick: (MouseEventHandler<...> & (() => void)) | undefined; }' is not assignable to type 'IntrinsicAttributes & Omit<Omit<Pick<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, "key" | keyof HTMLAttributes<...>> & { ...; }, "css"> & TransformProps<...> & { ...; }, "size" | ... 1 more ... | "as"> & TransformProps<...> & { ...; }'.
      Property 'disabled' does not exist on type 'IntrinsicAttributes & Omit<Omit<Pick<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, "key" | keyof HTMLAttributes<...>> & { ...; }, "css"> & TransformProps<...> & { ...; }, "size" | ... 1 more ... | "as"> & TransformProps<...> & { ...; }'.
  Overload 3 of 3, '(props: Omit<Omit<Pick<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, "key" | keyof HTMLAttributes<...>> & { ...; }, "css"> & TransformProps<...> & { ...; }, "size" | "css"> & TransformProps<...> & { ...; }): ReactElement<...> | null', gave the following error.
    Type '{ children: Element; color?: string | undefined; translate?: "yes" | "no" | undefined; slot?: string | undefined; style?: CSSProperties | undefined; title?: string | undefined; ... 252 more ...; onClick: (MouseEventHandler<...> & (() => void)) | undefined; }' is not assignable to type 'IntrinsicAttributes & Omit<Omit<Pick<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, "key" | keyof HTMLAttributes<...>> & { ...; }, "css"> & TransformProps<...> & { ...; }, "size" | "css"> & TransformProps<...> & { ...; }'.
      Property 'disabled' does not exist on type 'IntrinsicAttributes & Omit<Omit<Pick<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, "key" | keyof HTMLAttributes<...>> & { ...; }, "css"> & TransformProps<...> & { ...; }, "size" | "css"> & TransformProps<...> & { ...; }'.ts(2769)
```

lib/src/components/avatar/Avatar.tsx
D7Torres: We actually want `name` even if we get the `src` for the alt of the image. But I got the point, thanks!

andoulla:
@D7Torres is this ready for review or still in draft?

D7Torres:
@andoulla I think I need to fix [this](https://github.com/Atom-Learning/components/pull/413#discussion_r1007768268) before setting it as "read for review".

I could use some help

andoulla:
> @andoulla I think I need to fix [this](https://github.com/Atom-Learning/components/pull/413#discussion_r1007768268) before setting it as "read for review".
> 
> I could use some help

oh thanks, have you posted in the ts channel in slack?

lib/src/components/avatar/Avatar.tsx
D7Torres: Hey, in the end I refactored the whole component so it doesn't exactly apply here, but I actually had to apply it in another place 👍

lib/src/components/avatar/Avatar.tsx
D7Torres: As I refactored the whole component, this error no longer applies

lib/src/components/avatar/Avatar.tsx
```diff
@@ -0,0 +1,221 @@
...
+  is: typeof Icon
+}
+
+const AvatarIcon: React.FC<TAvatarIconProps> = ({ is }) => {
+  const rootContext = React.useContext(AvatarRootContext)
+  const { size } = rootContext
+
+  return (
+    <Icon
+      size={avatarSizeToIconSize[size]}
```
D7Torres, line 157: Here it complains about `size` possibly being undefined, but I don't know how to fix it.

I tried to do above
```
let { size } = rootContext
if (!size) size = 'lg'
```

(I know, ugly) And still complained...

lib/src/components/avatar/Avatar.tsx
```diff
@@ -0,0 +1,220 @@
...
+import { User } from '@atom-learning/icons'
+import { styled } from '~/stitches'
+import { focusVisibleStyleBlock } from '~/utilities'
+import { overrideStitchesVariantValue } from '~/utilities/override-stitches-variant-value/overrideStitchesVariantValue'
+import { Box } from '../box'
+import { Icon } from '../icon'
+import { Image } from '../image'
+
+const avatarRootStyles = {
+  display: 'flex',
```
D7Torres, line 12: @LimeWub Your version had `display: inline-flex`. What's the benefit?

lib/src/components/avatar/Avatar.tsx
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  return (
+    <StyledDiv size={size}>
+      <AvatarRootProvider name={name} size={size}>
+        {children}
+      </AvatarRootProvider>
+    </StyledDiv>
+  )
+}
```
D7Torres, line 137: Here I took the decision of wrapping with the Provider only the minimum necessary, but I repeated the use of the Provider.

Would it be better to write it like this?

```
}) => (
  <AvatarRootProvider name={name} size={size}>
    {onClick ? (
      <StyledButton
        size={size}
        disabled={disabled}
        onClick={disabled ? undefined : onClick}
        css={{ cursor: disabled ? 'auto' : 'pointer' }}
      >
        {children}
      </StyledButton>
    ) : (
      <StyledDiv size={size}>{children}</StyledDiv>
    )}
  </AvatarRootProvider>
)
```

lib/src/components/avatar/Avatar.mdx
```diff
@@ -0,0 +1,100 @@
+---
+title: Avatar
+component: Avatar
+description: Show a thumbnail representation of an individual or entity.
+category: Media
```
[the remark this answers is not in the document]
D7Torres: He's happy with it, there's not a better one at the moment, unfortunately.

lib/src/components/avatar/Avatar.tsx
```diff
@@ -0,0 +1,220 @@
...
+      xxl: {
+        fontSize: '$lg',
+        size: '$7'
+      }
+    }
+  }
+}
+
+const StyledDiv = styled('div', avatarRootStyles)
+const StyledButton = styled('button', {
```
LimeWub, line 51: FYI:
We had a conversation with @andoulla and @mzeenkala23 about Avatar not being a `button` but rather always being a `div`/non-interactable and only getting interact styles when it's used `asChild` in some sort of trigger ie.

```
<Dropdown.Trigger asChild>
    <Avatar />
</Dropdown.Trigger>
```

```
// I don't think Button supports `asChild` yet but perhaps it should at some point
<Button asChild>
    <Avatar />
</Button>
```

I don't know if this is actually needed or if the button option is actually better. I don't know how these `asChild` combinations actually work - if they pass an `onClick` down or if they just attach events.

I don't think you need to make any changes for this in this PR but FYI this is something I've been thinking about and it *miiiiiight* lead to tweaks in the future (it will be simplifying this, if anything). 
Just mentioning - it might never change and the way it's done rn is fairly clean with the style reuse ^^

lib/src/components/avatar/Avatar.tsx
```diff
@@ -0,0 +1,220 @@
...
+      />
+    </Box>
+  )
+}
+
+const AvatarInitial: React.FC<Record<string, never>> = () => {
+  const rootContext = React.useContext(AvatarRootContext)
+  const { name } = rootContext
+
+  if (name) {
```
LimeWub, line 181: /Opinion:/
It's a bit weird that we have the non-optimal case be the last return. Super minor but could you reverse this to:
`if (!name) return <AvatarPlaceholder />`
so the actual optimal return of the component is the last return?


It's just something that has been mentioned in some other PRs and I've bought into ^^





On another note: Are we sure we want these to be falling back to other options when they are missing data? Is this fallback pattern part of the design?

lib/src/components/avatar/Avatar.tsx
```diff
@@ -0,0 +1,220 @@
...
+type TAvatarImageProps = typeof StyledImage & {
+  src: string
+}
+
+const AvatarImage: React.FC<TAvatarImageProps> = ({ src }) => {
+  const rootContext = React.useContext(AvatarRootContext)
+  const { name } = rootContext
+
+  if (src) {
+    return <StyledImage src={src} alt={name ? `${name}'s avatar` : 'avatar'} />
```
LimeWub, line 202: We shouldn't have copy directly in the DS.

I'd suggest we let the implementation pass `alt` as well so they can pass an appropriate translation. 

That means we can remove `name` from the context as well.
Edit: Actually if you are keeping the fallbacks in you probably do want name on the top level so the `Initial` fallback works...

LimeWub (commented):
Good stuff ^^ Couple of comments and opinions but all in all 👍

lib/src/components/avatar/Avatar.tsx
andoulla: after taking time to think about it, I think Myrto is right, the avatar should only be concerned with displaying the information and then we can simply extend it by wrapping it with a link or a button or whatever else we need (in the student app needs to be used as a modal trigger in at least 2 placeS)

lib/src/components/avatar/Avatar.tsx
```diff
@@ -148,17 +140,24 @@ type TAvatarIconProps = {
...
+const toIconSize = {
+  xs: 'sm',
+  sm: 'sm',
+  md: 'md',
+  lg: 'md',
+  xl: 'lg',
+  xxl: 'lg'
+}
+
 const AvatarIcon: React.FC<TAvatarIconProps> = ({ is }) => {
```
andoulla, line 152: very small but can we call this icon?

lib/src/components/avatar/Avatar.mdx
```diff
@@ -0,0 +1,146 @@
...
+```jsx preview
+<Avatar name="Alice Smith">
+  <Avatar.Image src="" />
+</Avatar>
+```
+
+When the name is also missing it will fallback to the placeholder icon.
+
+```jsx preview
+<Avatar name="">
```
andoulla, line 40: it would be nice not to have to pass an empty string for name and src to beable to use the fall back

lib/src/components/avatar/Avatar.mdx
```diff
@@ -0,0 +1,146 @@
...
+```
+
+Instead of an image, an icon can be passed.
+
+```jsx preview
+<>
+  <Avatar size="xs">
+    <Avatar.Icon is={BatteryMedium} />
+  </Avatar>
+  <Avatar size="sm">
```
andoulla, line 80: I think we only need one example here as the sizes are described just below

lib/src/components/avatar/Avatar.mdx
```diff
@@ -0,0 +1,146 @@
...
+  </Avatar>
+  <Avatar size="xxl">
+    <Avatar.Icon is={BatteryMedium} />
+  </Avatar>
+</>
+```
+
+## Size
+
+The available `size`s for this component are: `xs`, `sm`, `md`, `lg`, `xl` and `xxl`. The default is `lg`.
```
andoulla, line 100: Why is the default `lg`? wouldn't `md` or sm make more sense? in most places we go with md to be honest

lib/src/components/avatar/Avatar.mdx
```diff
@@ -0,0 +1,146 @@
...
+  <Avatar size="xl" name="Alice Smith">
+    <Avatar.Image src="https://thispersondoesnotexist.com/image" />
+  </Avatar>
+  <Avatar size="xxl" name="Alice Smith">
+    <Avatar.Image src="https://thispersondoesnotexist.com/image" />
+  </Avatar>
+</>
+```
+
+```jsx preview
```
andoulla, line 125: I think this section is redundant and probably will confuse readers with our a description

lib/src/components/avatar/Avatar.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+    const { container } = render(
+      <Avatar name="a name">
+        <Avatar.Image src="path-to-image" />
+      </Avatar>
+    )
+
+    expect(container).toMatchSnapshot()
+  })
+
+  it('has no programmatically detectable a11y issues', async () => {
```
andoulla, line 20: Can we please split these into separate tests? We don't save anything by merging them, and actually, we are impeding troubleshooting as if it fails on the first one, it stops running. it is better to know you have issues for each variant separately.

lib/src/components/avatar/Avatar.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+  })
+
+  it('does not call the onClick function passed when click if disabled', () => {
+    const onClick = jest.fn()
+    const { container } = render(
+      <Avatar disabled onClick={onClick}>
+        <Avatar.Icon is={BatteryMedium} />
+      </Avatar>
+    )
+    fireEvent.click(container.firstChild as ChildNode)
```
andoulla, line 119: good test here!

lib/src/components/avatar/Avatar.tsx
```diff
@@ -0,0 +1,220 @@
...
+  name,
+  size
+}) => {
+  const value = React.useMemo<TAvatarRootContext>(
+    () => ({ name, size }),
+    [name, size]
+  )
+
+  return (
+    <AvatarRootContext.Provider value={value}>
```
andoulla, line 102: Do we really need a provider for a simple component like this? Do we even need to track these values in a context? The avatar just displays some information.

lib/src/components/avatar/Avatar.tsx
```diff
@@ -0,0 +1,220 @@
...
+      },
+      xxl: {
+        fontSize: '$lg',
+        size: '$7'
+      }
+    }
+  }
+}
+
+const StyledDiv = styled('div', avatarRootStyles)
```
andoulla, line 50: Why can we have one StyledContainer with all the styles and use the polymorphic as that stitches offer to render it as a button?

it would remove the need for having different wrapping containers, and simplify the code so much

lib/src/components/avatar/Avatar.tsx
```diff
@@ -0,0 +1,220 @@
...
+  const { size } = rootContext
+  const iconSize = React.useMemo(
+    () => overrideStitchesVariantValue(size, (s) => toIconSize[s]),
+    [size]
+  )
+
+  return <Icon size={iconSize} is={is} css={{ color: '$tonal400' }} />
+}
+
+const AvatarPlaceholder: React.FC<Record<string, never>> = () => {
```
andoulla, line 163: I think these children component should go into their own files. This file here is getting very loaded and there is no cost to having them split out

lib/src/components/avatar/Avatar.tsx
```diff
@@ -0,0 +1,220 @@
...
+    </Box>
+  )
+}
+
+const AvatarInitial: React.FC<Record<string, never>> = () => {
+  const rootContext = React.useContext(AvatarRootContext)
+  const { name } = rootContext
+
+  if (name) {
+    return <Box>{name[0].toUpperCase()}</Box>
```
andoulla, line 182: Is the decision to show only the first name initial? from the documentation, I understood that it was going to be showing both initials (ie: MS for mariandi stylianou) Btw this is a pattern that I have seen in most recent online tools like (JIRA, Notion and some I use privately

lib/src/components/avatar/Avatar.tsx
andoulla: Do we need the alt? Does it add value to a vision-impaired user to tell them this is your avatar in case the avatar is not clickable? just a minor thought as I think we should only if it is clickable

lib/src/components/avatar/Avatar.tsx
D7Torres: Thank you! As I'm making more changes, I'll do this too

lib/src/components/avatar/Avatar.tsx
D7Torres: People are waiting for this component, so for time's sake I'll leave it be for now.

Funny enough, initially I did it without using buttons, but that brought the TS error that sparked your first review

lib/src/components/avatar/Avatar.tsx
D7Torres: Regarding the order, yes I think it's a preference. My goal usually is to avoid unnecessary negations in conditionals.

But the goal you mention is legit too, and if that's what we are doing in Atom at the moment I'll happily change it for consistency

lib/src/components/avatar/Avatar.tsx
D7Torres: oh nooooo! 😂 

I'm screaming because I called it `icon` during the whole development and I changed it to `is` for consistency with the component `<Icon>`.

What's the logic? if we call it `icon` here... Why not call it `icon` also in the `<Icon>` component?

lib/src/components/avatar/Avatar.mdx
D7Torres: oh, it's not needed! It's just to illustrate the example where you might pass the name coming from the BE and if it's empty, it will fall back.

So it's to indicate that you don't need it to be missing for the fallback to take place.

D7Torres: So, to be clear:
`<Avatar name="">` will fall back

`<Avatar>` will fall back just the same

lib/src/components/avatar/Avatar.mdx
D7Torres: Just followed the documentation from Miguel

lib/src/components/avatar/Avatar.mdx
andoulla: shall we update the documentation the as this is a bit misleading when reading this the first time?

lib/src/components/avatar/Avatar.mdx
D7Torres: Initially, I only had one, but I had mistakes with the icon size and I added all to see them.

As there are only 3 icon sizes, that have to map to the 6 avatar sizes, I think is useful to see that mapping visually.

That's the same reason that below, in the size documentation I show the images, but also the initials in all the possible sizes.

If after the explanation you still want me to display only one size, please let me know

lib/src/components/avatar/Avatar.tsx
andoulla: hmm, it didn't even cross my mind that that was the association with the Icon component. I just didn't get it. 

If we are going for consistency then we should just go with the same logic as the ActionIcon. The component takes a child of type Icon and just readapts the size to the correct size.

lib/src/components/avatar/Avatar.tsx
D7Torres: That was my exact first reaction too!! :D

I didn't take the full Myrto's code in advance because I didn't agree with that. So I started taking up only the parts that made sense.

But once I paired with Gary he explained the advantages of making the API composable. And that led to take the context, which was a Myrto suggestion.

As there are some props that are in the parent, and the different possible children and fallbacks might need them there are two choices: Context, or drilling down.

But drilling down generically, to `{children}` can bring problems when expanding the component to take other childrens too. Imagine that we drill down `size` and then we get allow another child component to be used, which has it's own `size` prop. Then you would have a problem which you would not have with Context.

lib/src/components/avatar/Avatar.mdx
andoulla: if you are adding an explanation above this section then happy to keep them. I still think it is a bit heavy-handed but when it comes to documentation that is ok. with out the description there is nothing to tell the reader what they are seeing why it is different from the other sections

lib/src/components/avatar/Avatar.tsx
D7Torres: I don't know how to do that. Can we pair?

andoulla: you are in luck today is my guild day and I have lots of time. lets pair after standup

lib/src/components/avatar/Avatar.tsx
D7Torres: Ok. One file with all the children components or one file each?

andoulla: I would mirror other components like Carousel which it is one per file

lib/src/components/avatar/Avatar.tsx
D7Torres: I asked the same question to Miguel the 1st day I worked on this. Having one initial is a conscious decision. If I recall correctly, the reason was that in some of the sizes two initials would not fit

D7Torres:
@andoulla for some reason, GitHub doesn't allow me to reply to this comment
> Do we need the alt? Does it add value to a vision-impaired user to tell them this is your avatar in case the avatar is not clickable? just a minor thought as I think we should only if it is clickable

Honestly, I'm not very knowledgeable on a11y, so one rule I follow is to add `alt` to all the images that are content.

Now, regarding your question: imagine a case where there are 3 sections with info from a school, and each section has the avatar of the school instead of a heading with the name of the school. I think in that case `alt` would be useful even if they are not clickable. (btw, Miguel mentioned the case where Avatars would be used for schools)

image

lib/src/components/avatar/Avatar.tsx
andoulla: ok fair enough. Feel free to skip this comment. I was just worried that when we stack the avatars in groups showing one letter will not be helpful but we can cross that bridge when we get to it

andoulla:
> @andoulla for some reason, GitHub doesn't allow me to reply to this comment
> 
> > Do we need the alt? Does it add value to a vision-impaired user to tell them this is your avatar in case the avatar is not clickable? just a minor thought as I think we should only if it is clickable
> 
> Honestly, I'm not very knowledgeable on a11y, so one rule I follow is to add `alt` to all the images that are content.
> 
> Now, regarding your question: imagine a case where there are 3 sections with info from a school, and each section has the avatar of the school instead of a heading with the name of the school. I think in that case `alt` would be useful even if they are not clickable. (btw, Miguel mentioned the case where Avatars would be used for schools)
> 
> image

ok fair enough. in this context it makes sense completely

lib/src/components/avatar/Avatar.tsx
D7Torres: one moment... Miguel told me that he doesn't want the hover, focus, active styles to apply unless you can click it, so the Avatar component needs to know about those, doesn't it?

Or are we pulling all that outside too?

LimeWub: @andoulla Just to clarify: I am not actually sure what the better solution is. 

I think we should do some manual tests to decide if it needs to be a button itself or if some sort of setup/combination with other buttons works for what it needs to do. 
I can't think - off the top of my head - how it would combo with Button and Link for example (both of which don't currently support `asChild` prop, so I don't think they can be used without affecting styles atm) or with `DropdownMenu.Trigger` (which supports `asChild` but I don't know how it would affect the actual styling of the avatar on focus-visible etc). 

I think leaving this as both button and non-button might actually end up being the more legit solution. But it would be good to do a couple of tests to see how it actually interacts w other elements.

lib/src/components/avatar/Avatar.tsx
LimeWub: @andoulla  I don't think the way ActionIcon works is necessarily the best way - as it involves cloning the child to pass a prop through.
This is not necessary here and an `Avatar.Icon` is just a styling (sizing) extension of the `Icon`. I think the way it's done (including `is`) is fine.

lib/src/components/avatar/Avatar.mdx
D7Torres: So you saw this: `<Avatar name="">` and you didn't know if this `<Avatar>` would fall back.

But if we write this: `<Avatar>` some readers might wonder whether this `<Avatar name="">` would fall back

lib/src/components/avatar/Avatar.tsx
andoulla: @LimeWub @D7Torres , what I meant with my last comment was that we can delay the dependent work to figure out the correct solution. I rather merge this knowing we have closed this discussion than let it linger after it is in and used. 


so this is the time to investigate this and get it sorted. 

However, if we need the non-clickable version to not have the different states, I don't see a clean way of avoiding it with the polymorphic "as".

lib/src/components/avatar/Avatar.mdx
andoulla: I made the same thought for the reverse. When I saw it I thought I need to pass the name as empty to fall back

lib/src/components/avatar/Avatar.mdx
andoulla: @MiguelGDLR Just trying to understand the reasoning behind it. is it lg because it is the middle of the scale?

MiguelGDLR: Hey! Sorry I don't remember defining a default for Avatars. I don't think that always using `md` as default will work with all components but I don't have a preference for Avatars yet. So I'm ok with switching to `md` if you think it's better

andoulla: I'm just trying to think of the most common use case, which will be either in tables or/and the navbars. so I think it should be something that works with those

MiguelGDLR: unfortunately, I'm not able to decide yet until I have more usage context. I think lg/md is fine for now. We use those sizes a lot in our designs for cards / profile pages etc

andoulla: ok

lib/src/components/avatar/Avatar.mdx
D7Torres: Then no matter what I write, it is possible to be confused about it, right?

Should I write the 2 versions, so it's clear that both `<Avatar name="">` and `<Avatar>` will fall back? Although it will show the same avatar twitce so visually will be a bit weird.

lib/src/components/avatar/Avatar.mdx
D7Torres: In the end, I removed that block

lib/src/components/avatar/Avatar.mdx
D7Torres: Happy to apply whatever I'm told here.

Just to clarify: I had a look (I didn't remembered...) and the documentation doesn't say explicitely "lg is the default" but in the documentation "lg" is used everywhere except when talking about different sizes.

lib/src/components/avatar/Avatar.mdx
D7Torres: I'll remove it

lib/src/components/avatar/Avatar.tsx
D7Torres: Also, it would be better if parent components don't need to know how the Avatar should look when it's clickable. It seems like an Avatar's responsibility

LimeWub (approved):

andoulla (approved):

PR #415, opened 2022-10-27
feat: ColorScheme alpha version

LimeWub:
This is an alpha version of a colour scheme mechanism which has been shown to design (aka @MiguelGDLR ). 

This functionality has stemmed from the [CMS work](https://github.com/Atom-Learning/components/pull/398). 
It was introduced to tackle the major trouble I was having with the tabs component which was targeting direct children for theming and making the entire website $primary colour. It has been used in a couple of components and the documentation site itself.

I believe this implementation is the best I can do with the [information I have at this early stage](https://www.figma.com/file/UzIMkPUWQVFghjRgkp0RiT/COLOUR?node-id=521%3A11159) but I fully expect it to change in the future when the rules are properly considered.

## README

`color-schemes/`, is a **local only** package in an alpha stage (usable but not finalised) `@atom-learning/color-schemes` which introduces a `ColorScheme` component. It has been worked on heavily in terms of considering the required options and logic from dev but is **not finished and needs more design input**.

The problem that this component was introduced to solve is the difficult way we have approached theming so far in our components.
In the past we have heavily relied on props drilling and passing down theme related stitches variants to children, which means _(1)_ the colours can easily break when the expected children are nested further than directly under the 'theme providing' parent (The component which has a `theme` prop and tries to prop-drill theme). Moreover, _(2)_ we have inconsistent naming for our themes and _(3)_ inconsistent implementations on active components since _(4)_ the theming need to be documented and coded separately for each new component.

This package introduces a solution via the `<ColorScheme />` component which allows for properties to pass a `base` (`var(--baseX)`) and `accent` (`var(--accentX)`) theme property, as well as an `interactive` contrast mode to affect all interactable components.

The props are:

- `base` is used for the base colours of the wrapped component: different tones of background and text. It currently only supports the `slate` (grey) colour but can be extended for different base colors.
- `accent` is used for highlighted elements and interactive elements of the color scheme. It can be used directly but should be used with an interactive theme for interactive elements - see `<Button />`, `Accordion.Trigger` etc. Currently allowed accents are `slate` and `blue`.
- `interactive` supports 4 versions `loContrast1` and `loContrast2` which on light color schemes are lighter and on a dark mode they would be darker (so low contrast in comparison to the background), and `hiContrast1`, `hiContrast2` which are used for the opposite.

All the color scheme configurations setup used in the component can be found in `color-scheme/src/stitches.colorscheme.config.ts`

The ColorScheme itself works by re-declaring the needed CSS variables any time it's used. It is heavily inpired by both [stitches theming guidelines](https://stitches.dev/docs/theming) and particularly by [the radix colors project](https://www.radix-ui.com/colors) of which it's almost a copy of but with our own colours defined and some aliasing to match the current design use requirements.

Currently in use by a couple of components in library and the full documentation site. This package is subject to change when the Atom Learning colour theming patterns are properly established and we can make final decisions on the colours and API. **There will be changes to this code in the future.**


color-scheme/src/stitches.colorscheme.config.ts
```diff
@@ -0,0 +1,107 @@
...
+    colors[`${tokenPrefix}${i}`] = `$${colorName}${i}`
+    colors[`${tokenPrefix}A${i}`] = `$${colorName}A${i}`
+  }
+  return colors
+}
+
+const white = 'white'
+const black = '#191919'
+
+export const { styled, createTheme } = createStitches({
```
hollg, line 19: Do we anticipate any issues with calling `createStitches` here separately to where it's called in the lib? I remember you saying that some styles were appearing twice in snapshots. Could that be related to having two calls to this func?

LimeWub: No this just makes a new theme basically.
Moving it here (being unable to use lib's own theme due to dependency hell) actually solved the issue of all the library/regular theme styles being included in the `ColorScheme` so now it only includes the colours I've asked it to.

thomasdigby:
Hey @LimeWub, thanks for putting this together, definitely a problem that needs solving so good to see some energy has been put into finding a solution.

General notes:
- I think the use of custom properties to override more localised themes is the right approach, it's ultimately how Stitches works so it's good to align with their conventions.
- Generally naming conventions across the board are great. Really like the use of `hi/loContrast` naming, and its use in combination with a "base" colour to define which colour scale you want to use. This is how we should theme things.
- The "mode" and ability to reference items from a colour scale to define the theme is also great. e.g. "hiContrast" gets you colours from the end of the scale.

My questions about the implementation:
- Is there any reason that we're not storing these new colour values in the theme? Ideally, we should store all references to theming, whether that's colour, spacing, etc. within a single source.
- On the theme itself, can we not also use the `base`, `accent`, and `mode` in our naming conventions for our colours and allow components the ability to reference them directly? I'm concerned that we're creating multiple conventions for keys when styling components now, even using the same names, e.g. `$tonal3` when we already have `$tonal300`.
- I assume we want to use `ColorScheme` as a `styled` component to avoid adding an extra `div` to the rendered element? e.g. `<ColorScheme as={StyledTriggerList} />`. Is there a way we could render a child component and apply the custom properties without requiring that extra DOM node? 🤔 I don't think there is, just asking!
- Do we want to consider naming `interactive1`/2/3 so they're more aligned with actual usage? I think the benefit of grouping colours into a named theme is that we can start to alias the keys to be more suitable for their use. e.g. `hover`, `active`. Actually just digging through the usage it seems it's not super consistent so I don't know if my comment is useful/relevant.

I have more thoughts but keen to get an understanding on your approach first.

The thing sticking out to me most at the moment is that we're introducing a way to style components that doesn't fit in with our current `@atom-learning/theme` usage, so I'd like to hear how you consider updating that package in line with the changes here.

LimeWub:
@thomasdigby 
> Is there any reason that we're not storing these new colour values in the theme? Ideally, we should store all references to theming, whether that's colour, spacing, etc. within a single source.

Not really. They are stored in this local-package cause it was significantly easier for me to add, edit and use impromptu in the components repo - rather than getting PRs in for theme and asking for mock releases etc etc. 
Off the top of my head including the theme package from colorscheme and using colours from there rather than declaring them in this colorscheme itself, shouldn't be a problem (as there's no circular dependencies with theme).


> On the theme itself, can we not also use the base, accent, and mode in our naming conventions for our colours and allow components the ability to reference them directly? I'm concerned that we're creating multiple conventions for keys when styling components now, even using the same names, e.g. $tonal3 when we already have $tonal300.

So you mean:
`tonal` -> `base`
`accent` -stays- `accent`
`interactive` -stays- `interactive`
`mode`: I'm reserving this word for whenever we decide we might want light/dark mode

I'm happy to be renaming `tonal` to `base` and I think it makes sense with the naming of the CS component itself.

> I assume we want to use ColorScheme as a styled component to avoid adding an extra div to the rendered element? e.g. <ColorScheme as={StyledTriggerList} />. Is there a way we could render a child component and apply the custom properties without requiring that extra DOM node? 🤔 I don't think there is, just asking!

I'm not 100% I'm understanding this? Can you clarify?
If we use `<ColorScheme as={Button} />` it will apply the classnames to the button - there won't be any extra div around said button.
We could also employ the `asChild` pattern from `radix` for the same solution but I've used that only a little and didn't see the need for it here (plus I wonder if it'd lose the classnames or something - it'd need to be tested)

> Do we want to consider naming interactive1/2/3 so they're more aligned with actual usage? I think the benefit of grouping colours into a named theme is that we can start to alias the keys to be more suitable for their use. e.g. hover, active. Actually just digging through the usage it seems it's not super consistent so I don't know if my comment is useful/relevant.

Yes! I am happy to rename them (there were stages which I had renamed them) but as you said sometimes the usage is a bit off and I didn't want to have to worry too much about their naming for all the contexts they might be used in, which is why I went for numbers. Even though `1 === default`, `2 === hover`, `3 === active/pressed` usually I wasn't confident enough about that always being the case. If that's agreed on that in interactive elements that's always how it works I'd be all for the explicit naming.


> The thing sticking out to me most at the moment is that we're introducing a way to style components that doesn't fit in with our current @atom-learning/theme usage, so I'd like to hear how you consider updating that package in line with the changes here.

Does this mean adding 20 theme colours per colour (10 for the steps and 10 for the alphas)? Happy to move the colour files in theme, add them to the main theme to be used directly, and reference them for the colour schemes (mapping `blue1` to `tonal1`/`base1`) from there.

hollg:
@LimeWub 

> Not really. They are stored in this local-package cause it was significantly easier for me to add, edit and use impromptu in the components repo - rather than getting PRs in for theme and asking for mock releases etc etc.
Off the top of my head including the theme package from colorscheme and using colours from there rather than declaring them in this colorscheme itself, shouldn't be a problem (as there's no circular dependencies with theme).

Are there any blockers to this in terms of getting confirmation from Miguel about colours? I think @thomasdigby's point about splitting colours between the theme and this internal package is a good one, but I know one of the things we saw as a benefit of this approach was that it "hid" decisions we weren't 100% sure about from the public API.

LimeWub:
@hollg 
> Are there any blockers to this in terms of getting confirmation from Miguel about colours? I think @thomasdigby's point about splitting colours between the theme and this internal package is a good one...

Yes and that blocker is Miguel's time? Do we think this is in a good enough condition in terms of code to bug Miguel about finalising colours?

This is all a broader conversation and the entire reason this was kept as a different package and not part of the DS is because it's not finalised in design. Are we saying to block this until it's finalised in design? .-. 

> but I know one of the things we saw as a benefit of this approach was that it "hid" decisions we weren't 100% sure about from the public API.
^ this

thomasdigby:
> Not really. They are stored in this local-package cause it was significantly easier for me to add, edit and use impromptu in the components repo - rather than getting PRs in for theme and asking for mock releases etc etc.
Off the top of my head including the theme package from colorscheme and using colours from there rather than declaring them in this colorscheme itself, shouldn't be a problem (as there's no circular dependencies with theme).

Awesome, there shouldn't be any circular deps as there's a one-way dependency chain. I really think this can be a game-changer for how we theme our components not only in this context, but across the whole of atom-core as well. That's why I'm keen for us to tighten this up in the initial release so we have a strong base to build on.

> So you mean:
tonal -> base
accent -stays- accent
interactive -stays- interactive
mode: I'm reserving this word for whenever we decide we might want light/dark mode

> I'm happy to be renaming tonal to base and I think it makes sense with the naming of the CS component itself.

Yeah so we have a single set of categories or groups for our different types of colour. I was thinking about this yesterday and I wonder whether a consistent set of keys as well as a prefix could help the usability of the new theme but also differentiate between the two use-cases clearly enough. e.g.

```
// root/base theme (deprecated)
$colorPrimary;

// root/base theme, but aliased (the name might be wrong here, would it be `$colorAccent2`?), may need a postfix integer as per your current usage
$colorBase;

// uses local theme token
$themeColorBase;
```

This is maybe a bad example, and I reckon we probably ought to chat in person about it as you'll have a much better idea about the edge-cases and even if I've missed the point entirely, but the shared name "color base" and the prefix "theme" might help to tie the two methods of theming together.

> I'm not 100% I'm understanding this? Can you clarify?
If we use <ColorScheme as={Button} /> it will apply the classnames to the button - there won't be any extra div around said button.
We could also employ the asChild pattern from radix for the same solution but I've used that only a little and didn't see the need for it here (plus I wonder if it'd lose the classnames or something - it'd need to be tested)

Yep that's all good, I think that's the right way to do it in the circumstances. I _like_ the idea of wrapping a component as opposed to `ColorScheme as`, because I don't think there's quite the same relationship within the JSX as the CSS, e.g. a wrapping div that provides the tokens to child elements. TBH it has no impact on the output at all so no need to worry 👍 

> Yes! I am happy to rename them (there were stages which I had renamed them) but as you said sometimes the usage is a bit off and I didn't want to have to worry too much about their naming for all the contexts they might be used in, which is why I went for numbers. Even though 1 === default, 2 === hover, 3 === active/pressed usually I wasn't confident enough about that always being the case. If that's agreed on that in interactive elements that's always how it works I'd be all for the explicit naming.

Yeah that was my concern as well, it would be _awesome_ to have a consistent naming convention that aligned with colour usage like you suggested, `default`, `hover`, `active`, but it's almost certainly going to be used differently to that. I'd like to settle on something good now though, before we introduce this to the rest of the components lib as I think it will inform how we theme stuff everywhere in future. Again probably one to chat through in person.

> Does this mean adding 20 theme colours per colour (10 for the steps and 10 for the alphas)? Happy to move the colour files in theme, add them to the main theme to be used directly, and reference them for the colour schemes (mapping blue1 to tonal1/base1) from there.

That would be good, but it was also about the naming conventions that add an element of doubt about what colour will be used. Updating to an aligned naming convention for aliased tokens, and moving those tokens to the theme (just for the selected colours for now) should be enough to make it work 👍

LimeWub:
@thomasdigby @hollg 
I've:
* renamed `tonal` to `base`
* added an experimental warning to ColorScheme

Is there anything additional to do for this bit of the code? I don't have anything else in my notes from our meeting but just in case u remember anything else?

Otherwise this is ready to re-review.

thomasdigby:
> @thomasdigby @hollg I've:
> 
>     * renamed `tonal` to `base`
> 
>     * added an experimental warning to ColorScheme
> 
> 
> Is there anything additional to do for this bit of the code? I don't have anything else in my notes from our meeting but just in case u remember anything else?
> 
> Otherwise this is ready to re-review.

Thanks @LimeWub, I'll re-review today

color-scheme/src/blue.ts
```diff
@@ -0,0 +1,16 @@
+import { generateAlphaColors } from './generateAlphaColors'
+
+export const blue = {
```
hollg, line 3: I think one of the things we talked about in the meeting with @thomasdigby was storing these colour configs in a JSON file, but I don't remember details of the conversation now so I wonder whether these `.ts` files achieve the same thing? I think it _maybe_ adds something more specific to what we're communicating with the code structure if we move the actual map of names and hex codes to a JSON file, but happy to be overruled if you both disagree.

color-scheme/rollup.config.js
```diff
@@ -0,0 +1,36 @@
...
+  input: 'src/index.ts',
+  external: externalDeps,
+  output: [
+    { file: pkg.main, format: 'cjs' },
+    { dir: 'dist', format: 'esm', preserveModules: true }
+  ],
+  plugins: [
+    summary({ showMinifiedSize: false, showBrotliSize: false }),
+    commonjs(),
+    alias({ entries: hq.get('rollup', { format: 'array' }) }),
```
thomasdigby, line 28: As we're not using aliases within this project, can we remove this and the relevant packages? I'd recommend removing as much of this custom configuration as possible really, to match what's needed from the `color-scheme` package only. That _might_ include the `commonjs` plugin as well... but I think `react` might still be CJS so maybe not. Worth a try.

color-scheme/jest-setup.ts
```diff
@@ -0,0 +1,5 @@
+import 'jest-axe/extend-expect'
+import '@testing-library/jest-dom'
+import './test/mocks/ResizeObserver'
+import './test/mocks/DOMRect'
+import './test/patches/cssom'
```
thomasdigby, line 5: Are these available to use? I can't see them in this PR. If we're going to add tests before the initial release, let's make sure they're included only if needed.

color-scheme/tsconfig.json
```diff
@@ -0,0 +1,22 @@
...
+    "lib": ["es6", "dom"],
+    "types": ["node", "jest"],
+    "baseUrl": "./src",
+    "skipLibCheck": true,
+    "paths": {
+      "~/*": ["./*"]
+    },
+    "declaration": true
+  },
+  "include": ["./src/*", "types/global.d.ts"],
```
thomasdigby, line 20: I'd simplify this `tsconfig.json` a touch as well, remove the missing `types/global.d.ts`, and the `paths` config as we don't need it

color-scheme/src/stitches.colorscheme.config.ts
```diff
@@ -0,0 +1,107 @@
...
+  theme: {
+    colors: {
+      background: white,
+      foreground: black,
+      foreground6plus: white,
```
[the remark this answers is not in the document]
LimeWub: I don't remember if we've discussed it but on the experimental scales on step 6 out of 10 the color (background) becomes significantly more saturated/darker which means the font colour over the 6 scale needs to be flipped (from white to black).

color-scheme/jest.config.js
```diff
@@ -0,0 +1,14 @@
...
+  testEnvironment: 'jsdom',
+  testPathIgnorePatterns: ['/node_modules/'],
+  setupFilesAfterEnv: ['./jest-setup.ts'],
+  moduleNameMapper: {
+    '~/(.*)': '<rootDir>/src/$1' // matches the path alias in tsconfig.json
+  },
+  transform: {
+    '^.+\\.(ts|tsx)$': 'es-jest'
+  }
+}
```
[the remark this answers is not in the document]
LimeWub: I actually don't think I'll be writing tests as there's nothing significant to test - other than adding some classnames (?). So I might just remove this (?)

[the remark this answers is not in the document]
thomasdigby: Yeah that's fine for now, worth removing it all and then we can add in when we need it 👍

color-scheme/src/stitches.colorscheme.config.ts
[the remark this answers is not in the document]
thomasdigby: Right, got it, thanks. Could we use the values directly in the new themes instead of creating tokens and referencing them? e.g.

```ts
colorSchemes['interactive-hiContrast2'] = createTheme(
  'interactive-hiContrast2',
  {
    colors: {
      interactiveForeground: 'white',
      interactive1: '$accent8',
      interactive2: '$accent9',
      interactive3: '$accent10'
    }
  }
)
```

Are we expecting `$foreground6plus` to be used by the dev directly? Or is it just used in `color-scheme`?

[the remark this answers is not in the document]
LimeWub: Ye it'd be used by the dev directly if they use `base6` or over for a background

color-scheme/README.md
```diff
@@ -0,0 +1,20 @@
...
+- `interactive` supports 4 versions `loContrast1` and `loContrast2` which on light color schemes are lighter and on a dark mode they would be darker (so low contrast in comparison to the background), and `hiContrast1`, `hiContrast2` which are used for the opposite.
+
+All the color scheme configurations setup used in the component can be found in `color-scheme/src/stitches.colorscheme.config.ts`
+
+The ColorScheme itself works by re-declaring the needed CSS variables any time it's used. It is heavily inpired by both [stitches theming guidelines](https://stitches.dev/docs/theming) and particularly by [the radix colors project](https://www.radix-ui.com/colors) of which it's almost a copy of but with our own colours defined and some aliasing to match the current design use requirements.
+
+Currently in use by a couple of components in library and the full documentation site. This package is subject to change when the Atom Learning colour theming patterns are properly established and we can make final decisions on the colours and API. **There will be changes to this code in the future.**
```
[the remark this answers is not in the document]
LimeWub: Updated - lmk if the examples are okay

LimeWub:
@thomasdigby Comments addressed I believe

color-scheme/src/stitches.colorscheme.config.ts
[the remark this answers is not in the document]
thomasdigby: Ok cool, I think I still have a few questions about the naming conventions used and how they're utilised within the child components, but I'm really keen to unblock this so I think as a v1 this is looking good and we can think about what "v2" looks like if we open this package up to the main apps as well 🙂

color-scheme/package.json
```diff
@@ -0,0 +1,54 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/Atom-Learning/components.git"
+  },
+  "author": "",
+  "license": "ISC",
+  "devDependencies": {
+    "@rollup/plugin-node-resolve": "^11.2.0",
+    "@testing-library/react": "^11.2.3",
+    "@testing-library/user-event": "^12.8.1",
```
thomasdigby, line 24: Can we remove these unused packages?

LimeWub: Ah. Missed them when I did a search for 'jest'.

color-scheme/src/stitches.colorscheme.config.ts
```diff
@@ -0,0 +1,109 @@
...
+  const colors = {}
+  for (let i = 1; i <= 10; i++) {
+    colors[`${tokenPrefix}${i}`] = `$${colorName}${i}`
+    colors[`${tokenPrefix}A${i}`] = `$${colorName}A${i}`
+  }
+  return colors
+}
+
+const white = 'white'
+const black = '#191919'
```
thomasdigby, line 19: Can we reference `$tonal600` here instead?

LimeWub: Wouldn't that just be confusing?

LimeWub: Also, in terms of colour, this isn't `tonal600` (`1F1F1F`) - it's `slate9` and this one is darker than all slate colours. Perhaps it should be `1F1F1F` which is tonal600 but I don't want to reference tonals directly in this package (as I'm not referencing primary/Mid/Dark for the blues.

LimeWub: Better? https://github.com/Atom-Learning/components/pull/415/commits/c3efe632a930f3c5a644d6775d74ea9dfd6c59f8

thomasdigby (approved):

color-scheme/src/stitches.colorscheme.config.ts
thomasdigby: Yeah, I think that's better, good to reference the same token. Although it feels like `slate` is baked into the defaults for `ColorScheme` now, but I guess we can update that when we add more base themes 👍

hollg (approved):

PR #418, opened 2022-10-31
feat: navigation - active state and focus fix, dropdown item title a…

michalrozenek:
### Added new NavigationMenu elements:
- NavigationMenuDropdownItemTitle
- NavigationMenuDropdownItemSubtitle

both items added to makes navigation dropdown item elements consistent across applications.

Zrzut ekranu 2022-10-31 o 13 58 25
Zrzut ekranu 2022-10-31 o 13 58 35

### Fixes for Dropdown `active` state:
- hover
- active
- focus

+ added text, icon (img) proper colors on active state.

https://user-images.githubusercontent.com/23363033/199012856-92da5a65-d1c2-4e59-b23d-e17026080e0d.mov
Zrzut ekranu 2022-10-31 o 11 35 50

### Fix for focus state:
Before:
Zrzut ekranu 2022-10-31 o 11 32 14

After:
Zrzut ekranu 2022-10-31 o 11 35 35

### DIFF
--- a/lib/src/components/navigation/NavigationMenu.test.tsx
+++ b/lib/src/components/navigation/NavigationMenu.test.tsx
@@ -3,6 +3,7 @@ import userEvent from '@testing-library/user-event'
 import { axe } from 'jest-axe'
 import * as React from 'react'
 
+import { Flex } from '../flex'
 import { NavigationMenu } from '.'
 
 const ExampleNav = () => (
@@ -11,13 +12,28 @@ const ExampleNav = () => (
     <NavigationMenu.Dropdown title="Theme">
       <NavigationMenu.DropdownContent>
         <NavigationMenu.DropdownItem href="https://app.atomlearning.co.uk/colours">
-          Colours
+          <Flex
+            css={{
+              flexDirection: 'column'
+            }}
+          >
+            <NavigationMenu.DropdownItemTitle bold>
+              Colours
+            </NavigationMenu.DropdownItemTitle>
+            <NavigationMenu.DropdownItemSubtitle>
+              Atom Learning color palette
+            </NavigationMenu.DropdownItemSubtitle>
+          </Flex>
         </NavigationMenu.DropdownItem>
         <NavigationMenu.DropdownItem href="https://app.atomlearning.co.uk/effects">
-          Effects
+          <NavigationMenu.DropdownItemTitle>
+            Effects
+          </NavigationMenu.DropdownItemTitle>
         </NavigationMenu.DropdownItem>
         <NavigationMenu.DropdownItem href="https://app.atomlearning.co.uk/icons">
-          Icons
+          <NavigationMenu.DropdownItemTitle>
+            Icons
+          </NavigationMenu.DropdownItemTitle>
         </NavigationMenu.DropdownItem>
       </NavigationMenu.DropdownContent>
     </NavigationMenu.Dropdown>
--- a/lib/src/components/navigation/NavigationMenu.test.tsx
+++ b/lib/src/components/navigation/NavigationMenu.test.tsx
@@ -4,6 +4,7 @@ import { axe } from 'jest-axe'
 import * as React from 'react'
 
 import { Flex } from '../flex'
+import { Text } from '../text'
 import { NavigationMenu } from '.'
 
 const ExampleNav = () => (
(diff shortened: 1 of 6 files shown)


elliotlaws (approved):

Jdrazong (approved):

lib/src/components/navigation/NavigationMenu.test.tsx
```diff
@@ -11,13 +12,28 @@ const ExampleNav = () => (
...
-          Effects
+          <NavigationMenu.DropdownItemTitle>
+            Effects
+          </NavigationMenu.DropdownItemTitle>
         </NavigationMenu.DropdownItem>
         <NavigationMenu.DropdownItem href="https://app.atomlearning.co.uk/icons">
-          Icons
+          <NavigationMenu.DropdownItemTitle>
+            Icons
+          </NavigationMenu.DropdownItemTitle>
```
thomasdigby, line 36: Is `NavigationMenu.DropdownItemTitle` necessary now for all menus? Or can you just use text within `NavigationMenu.DropdownItem` as before?

lib/src/components/navigation/NavigationMenuItem.tsx
```diff
@@ -81,6 +89,21 @@ const StyledLink = styled(NavigationMenuPrimitive.Link, itemStyles, {
   }
 })
 
+export const NavigationMenuDropdownItemTitle = styled(Text, {
+  color: '$tonal500',
+  variants: {
+    bold: {
```
thomasdigby, line 95: This `bold` prop doesn't fit with our styling API used elsewhere, is there a way we could avoid using it? Are we against setting the title to be bold by default?

lib/src/components/navigation/NavigationMenu.test.tsx
michalrozenek: It was discussed with @MiguelGDLR and @elliotlaws.
This is basically better to keep navigations consistent. 
If we do not have it, people can add different styles to navigation dropdown item. I know it's not styles heavy, but guarantee more consistency at some point :)

lib/src/components/navigation/NavigationMenuItem.tsx
```diff
@@ -71,7 +72,14 @@ const StyledLink = styled(NavigationMenuPrimitive.Link, itemStyles, {
...
         '&[data-active]': {
           background: '$primaryLight',
-          color: '$primary'
+          '*': {
+            color: '$primary'
+          },
+          '&:hover': { background: '$tonal50' },
+          '&:active': { background: '$tonal100' },
+          '&:focus-visible': {
+            boxShadow: `0 0 0 2px ${theme.colors.primary}`
```
thomasdigby, line 81: We can access the theme colours in every CSS key be prefixing with the "group" that they're in; in this case `$colors`. 

```suggestion
            boxShadow: `0 0 0 2px $colors$primary`
```

lib/src/components/navigation/NavigationMenuItem.tsx
michalrozenek: Maybe I can consider different naming for `DropdownItemTitle` then.
We have two variants in navigation title (bold and normal).

Maybe we can go with `title/heading` as a bold one, and `text` instead of `subtitle` as a `normal` one.. 🤔

lib/src/components/navigation/NavigationMenuItem.tsx
michalrozenek: Thanks, did not notice this

lib/src/components/navigation/NavigationMenu.test.tsx
thomasdigby: It seems unnecessary for simple menus, it would be great if you could use it only if you need multiple levels of info in an nav item, like your examples in the description. I think it would be beneficial to retain the ability to create a simpler nav without the title, this is already fairly verbose as it is.

```tsx
<NavigationMenu>
  <NavigationMenu.Dropdown title="Theme">
    <NavigationMenu.DropdownContent>
      <NavigationMenu.DropdownItem href="/" active>
        Example title
    </NavigationMenu.DropdownContent>
  </NavigationMenu.Dropdown>
</NavigationMenu>
```

Could we not default to the current styles as we have, and then use `NavigationDropdown.DropdownItemTitle` for the bold and stylised text as in the design?

lib/src/components/navigation/NavigationMenuItem.tsx
thomasdigby: I think we're having the same conversation across two threads, this should be resolved by my answer in the other thread!

lib/src/components/navigation/NavigationMenu.test.tsx
michalrozenek: That's fine.
Subtitle was not styles heavy, so maybe it's enough to have a `title` ready to use and add the rest by hand.
Let's see how devs will use it.

thomasdigby (approved):

PR #420, opened 2022-11-01
fix stepper state updates

kerryambt:
### Description

https://atomlearningltd.atlassian.net/browse/DS-143
Note: part of that ticket requires a change on core once this work has been done and DS version on core updated

This fixes a number of issues with the stepper component and the various states that is has:

- The stepper track was not correctly showing as blue for completed items, this has been fixed
- Viewed (but not completed) items now show a darker than normal colour to separate them from default state
- Added a prop to allow showing tick icons for completed steps instead of their step number
- Changed the default state colour to be slightly darker so it shows up better on light backgrounds
- Moved the interaction states (hover, focus) to only appear when allow skip prop is present, this is because otherwise there would be no interaction so hover states were superfluous and confusing to the user
- Updated tests and snapshots

### Screenshots

image

### DIFF
--- a/lib/src/components/stepper/StepperSteps.tsx
+++ b/lib/src/components/stepper/StepperSteps.tsx
@@ -30,7 +30,8 @@ export const StepperSteps: React.FC<IStepperStepsProps> = ({ css }) => {
     allowSkip,
     direction,
     hideLabels,
-    completedSteps
+    completedSteps,
+    showCompletedIcons
   } = useStepper()
 
   const getBulletStatus = (index: number) => {
@@ -57,8 +58,12 @@ export const StepperSteps: React.FC<IStepperStepsProps> = ({ css }) => {
       return Status.SUCCESS
     }
 
-    if (bulletStatus === Status.VIEWED || index < Math.max(...viewedSteps)) {
+    if (bulletStatus === Status.COMPLETED || index < Math.max(...viewedSteps)) {
       return Status.ACTIVE
+    }
+
+    if (bulletStatus === Status.VIEWED) {
+      return Status.VIEWED
     }
 
     return Status.DEFAULT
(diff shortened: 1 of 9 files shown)


MiguelGDLR:
really great work as always Kerry, thank you!

elliotlaws (approved):

LimeWub (approved):

lib/src/components/stepper/StepperSteps.tsx
```diff
@@ -99,7 +105,12 @@ export const StepperSteps: React.FC<IStepperStepsProps> = ({ css }) => {
                   allowSkip && viewedSteps.includes(index) ? 'pointer' : 'auto'
               }}
             >
-              {step.status === 'success' ? <Icon is={Ok} /> : index + 1}
+              {step.status === Status.SUCCESS ||
+              (showCompletedIcons && bulletStatus === Status.COMPLETED) ? (
```
thomasdigby, line 109: Is there a need to show the number on a completed step? Is that a variant that we want to use on the site?

I'd probably lean towards always using icons on completed steps, but obviously happy to go with whatever has been designed

kerryambt: one for @MiguelGDLR  I think

MiguelGDLR: It was originally designed by Sara with only numbers, but then Jess suggested using an icon instead. If it's already developed to provide both options I would leave it like that until we see more real use cases.

thomasdigby: Ok cool, I'm wary of introducing a prop for a visual change as opposed to using component composition, so just wanted to see if we could simplify this a touch. Happy to go ahead with this as proposed 👍

PR #422, opened 2022-11-04
Feat: server-side pagination for `DataTable`

hollg:
This demonstrates a way forward for `DataTable`s with externally controlled pagination. The obvious use case is server-side pagination.

The basic idea is that a new `DataTable.Remote` is added, which acts as an alternative to the `DataTable` provider. Rather than a `data` prop, it takes a `fetcher`, which must be an async function that returns the table data. `fetcher` is provided with the current pagination and sorting config.

Once the features are all added and kinks ironed out (see below), it may be possible to combine `DataTable` and `DataTable.Remote` into one component that takes _either_ `data` or `fetcher`.

- [x] `fetcher` runs when pagination or sort is updated
- [x] loading state for `fetcher`
- [x] pre-existing `DataTable` components mostly work with `DataTable.Remote` context
- [x] `DataTable.Pagination` updates page index when pagination changes (the issue here seems to be that some internal state in `@tanstack/react-table` isn't getting updated)

Loading:

https://user-images.githubusercontent.com/11061661/205672772-ff2f4b3b-e4a6-4197-9e46-71b244779646.mov

Error state: 
image


lib/src/components/data-table/DataTable.test.tsx
```diff
@@ -120,28 +127,28 @@ describe('DataTable.Pagination component', () => {
...
 
-    const options = screen.getAllByRole('option')
-    userEvent.click(options[1])
+    userEvent.selectOptions(
+      screen.getByRole('combobox'),
+      screen.getByRole('option', { name: '2' })
+    )
 
-    waitFor(async () => expect(screen.getByRole('combobox')).toHaveValue('1'))
+    await waitFor(() => expect(screen.getByRole('combobox')).toHaveValue('1'))
```
Marcdup1, line 148: @hollg we updated some of the tests using `waitFor` since it returns a promise was not awaited the tests would pass regardless

lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -48,8 +43,28 @@ export const DataTableProvider = ({
     setIsPaginated(true)
   }, [])
 
+  const applyDragAndDrop = React.useCallback(() => {
```
Marcdup1, line 46: @hollg can we remove the drag-and-drop logic from the PR?

lib/src/components/data-table/RemoteDataTableContext.tsx
```diff
@@ -0,0 +1,160 @@
...
+      doFetchData
+    }),
+    [table, applyPagination, getTotalRows, isSortable]
+  )
+
+  const isLoading = apiQueryStatus === ApiQueryStatus.PENDING
+
+  return (
+    <Box css={{ position: 'relative' }}>
+      {isLoading && <DataTableLoading />}
```
Marcdup1, line 141: We kept the loader outside of the HTML table, because we were not sure if it is semantically correctly to include it inside.

It would be nice to be able to add it as a subcomponent to the data table but then chances are the user might not include it, should we have that flexibility for a loader?

lib/src/components/data-table/DataTableContext.tsx
hollg: Oh god, sorry, that is some poor git hygeine on my part. Yes, you can remove anything that seems specific to drag-and-drop **except** the change on line `26` to hold the `data` in state rather than passing it straight into `useReactTable` from the prop. 

I think merging the different feature branches will become a pain down the line otherwise.

Marcdup1: no problem! removed it here [e905835](https://github.com/Atom-Learning/components/pull/422/commits/e9058356f777725e26a793e595d6de649272415b) so we can refer to this in case you don't have anywhere else

lib/src/components/data-table/RemoteDataTableContext.tsx
Marcdup1: >But I would also include it in DataTable.Table, which should always be the "I just want all the features with their default config" option for our consumers. (And these context files should only be in charge of the state logic; not rendering anything.)

Good point! added it here [cd776ad](https://github.com/Atom-Learning/components/pull/422/commits/cd776ad95aa6310c8f30507088e128330d9a1005). I also simplified the loading a bit and removed the provider wrapper to add the blur an added it directly to the <table>

lib/src/components/data-table/RemoteDataTableContext.tsx
```diff
@@ -0,0 +1,148 @@
...
+    try {
+      setApiQueryStatus(ApiQueryStatus.PENDING)
+      const newData = await fetcher(
+        pageIndex,
+        pageSize,
+        sorting[0]?.id,
+        sorting[0] ? (sorting[0].desc ? 'desc' : 'asc') : null
+      )
+
+      setData(newData)
```
thomasdigby, line 88: This assumes that the response matches the required data schema right? What do we do if `results` or `total` aren't returned, or the object layout for the data differs? Do we document that the response must be transformed before being passed into the fetcher? e.g.

```tsx
// api returns results directly, e.g. `[{}, {}, {}]`
<DataTable.Remote
  fetcher={(index, size) => {
    return fetch('api?current=${index}&total=${size}')
      .then(res => res.json)
      .then(res => ({ results: res, total: res.length }))
  }}
/>
```

Marcdup1: Omitting `result` would result in a crash, and omitting `total` would negatively affect the navigation data that is displayed.

I updated the docs to include a bit of information and example usage.

lib/src/components/data-table/RemoteDataTableContext.tsx
```diff
@@ -0,0 +1,148 @@
+import * as React from 'react'
```
hollg, line 1: Now that the features have been figured out, is there anything stopping us combining the two context providers into one component, which has to take either a `data` or `fetcher` prop? I think having one component with an option of two props feels like a lower-friction API than needing to use two different  components in different cases (and to understand that the other subcomponents work with either one).

hollg:
Has @MiguelGDLR  been involved in the design of the error state? To me the spacing of the error message and button looks too small and I'm not sure about the theme/size of the retry button, or its position relative to the other content. Maybe it should be underneath the copy that tells you there's an error, since you need to read that before you know to retry?

lib/src/components/data-table/RemoteDataTableContext.tsx
```diff
@@ -0,0 +1,148 @@
...
+  results: Array<Record<string, unknown>>
+}
+
+type TTableProviderProps = {
+  columns
+  defaultPageSize?: number
+  defaultSort?: { column: string; direction: 'asc' | 'desc' }
+  children: React.ReactNode
+  css?: CSS
+  fetcher: (
```
hollg, line 31: I'd be interested to hear other people's thoughts on this but when we're providing the inputs for a callback, I think it can be nice to give them all on an object and let the consumer destructure the ones they need, rather than making them list every input when they might not need them all. But maybe that's mitigated somewhat here by having the values that the consumer is most likely to need at the start of the list?

MiguelGDLR: Marc asked me and told me: 
"Error: The component is basically an **empty component** that renders whatever the user of the component wants to. So that error message and button is **not built into the error** component. The error component adds a retry function that the children can use if they want. When an error occurs we hide the body of the table and the pagination"
I understood this was just an example as it would be an empty canvas, if this is something that's going to be displayed we can improve it then, let me know.

MiguelGDLR:
> Has @MiguelGDLR been involved in the design of the error state? To me the spacing of the error message and button looks too small and I'm not sure about the theme/size of the retry button, or its position relative to the other content. Maybe it should be underneath the copy that tells you there's an error, since you need to read that before you know to retry?

Marc asked me and told me:
"Error: The component is basically an empty component that renders whatever the user of the component wants to. So that error message and button is not built into the error component. The error component adds a retry function that the children can use if they want. When an error occurs we hide the body of the table and the pagination"
I understood this was just an example as it would be an empty canvas, if this is something that's going to be displayed we can improve it then, let me know.

lib/src/components/data-table/RemoteDataTableContext.tsx
hollg: Maybe we could use `invariant` here to throw an error if the data isn't all included/formatted correctly? Or is TypeScript clever enough to get grumpy if the return type isn't right?

thomasdigby: I think we ought to add a runtime catch as well, `invariant` is a good shout, minimally intrusive schema validation.

lib/src/components/data-table/pagination/Pagination.tsx
```diff
@@ -47,6 +49,9 @@ export const Pagination: React.FC<PaginationProps> = ({
   const recordsCountFrom =
     paginationState.pageIndex * paginationState.pageSize + 1
   const recordsCountTo = recordsCountFrom + getRowModel().rows.length - 1
+  const isLoading = apiQueryStatus === ApiQueryStatus.PENDING
+
+  if (apiQueryStatus === ApiQueryStatus.FAILED) return null
```
thomasdigby, line 54: What will `apiQueryStatus` be if not wrapped in a `DataTable.Remote`? I assume `undefined`? Do we want a bit more control as to when we do these checks? e.g. only when rendering within a `DataTable.Remote` provider

lib/src/components/data-table/__snapshots__/DataTable.test.tsx.snap
```diff
@@ -410,6 +1242,12 @@ exports[`DataTable component renders 1`] = `
   }
 }
 
+@media  {
+  .c-PJLV-illIiIE-css {
+    mb: var(--4);
```
thomasdigby, line 1247: Huh, what's the deal with this CSS? 🤔 I can't see where it gets applied or what has changed to break the key and custom property

lib/src/components/data-table/RemoteDataTableContext.tsx
thomasdigby: I'm not sure there are any downsides really; I'll plus one your suggestion.

lib/src/components/data-table/DataTableTable.tsx
```diff
@@ -12,14 +15,29 @@ type DataTableTableProps = Omit<
   > &
   Partial<Pick<React.ComponentProps<typeof Table.Body>, 'striped'>>
 
+const TableStyled = styled(Table, {
+  variants: {
+    isLoading: {
+      true: {
+        filter: 'blur(3px)'
```
thomasdigby, line 22: I personally don't love this blur, and I haven't really seen us use this style visually across any of our other apps. Is this a preview of what's to come? Or is it unique to the table?

I'd recommend knocking back the contrast and overlaying the data with a click blocker and loader instead, which keeps the visual change relatively simple but clear that it's "disabled" and non-interactive while loading.

Obviously, this is personal preference, @MiguelGDLR I assume you've 👍 this approach?

lib/src/components/data-table/DataTableLoading.tsx
```diff
@@ -0,0 +1,30 @@
...
+  if (apiQueryStatus !== ApiQueryStatus.PENDING) return null
+
+  return (
+    <Loader
+      css={{
+        position: 'absolute',
+        marginBottom: '$4',
+        left: '50%',
+        transform: 'translate(-50%, 0)',
+        bottom: '$9',
```
thomasdigby, line 24: I understand why we're positioning it relative to the pagination, but something feels a bit off about the position of it relative to the height of the table. Do you think placing it centrally would render it off-screen most of the time?

I also wonder if this takes into account the loading state for sorting columns? I assume we're making a request to the API for any changes to columns, so the loader might not even be visible if you click on a column heading

lib/src/components/data-table/DataTableTable.tsx
MiguelGDLR: This was Marc's idea, I haven't studied a general pattern yet so I was ok with his proposal, your improvements Tom seem good to me.

lib/src/components/data-table/DataTableError.tsx
```diff
@@ -0,0 +1,35 @@
...
+    <Flex
+      css={{
+        flexDirection: 'column',
+        alignItems: 'center',
+        p: '$5',
+        bg: '$background',
+        ...css
+      }}
+    >
+      {children(doFetchData)}
```
thomasdigby, line 32: I wonder if we could mirror the `fetcher` API here and pass in the relevant arguments for the retried request. The current behaviour is a "try the exact same request again" approach, but it might be best to give a bit more control to the dev to update/tweak the request if there's something they're able to change to increase the chance of it working. e.g. an extra search param to bust a cache 🤷

lib/src/components/data-table/RemoteDataTableContext.tsx
D7Torres: We definitely have a type for it
```
type TFetcherResult = {
  total: number
  results: Array<Record<string, unknown>>
}
```

and TS complains about it when missing from the return of the fetcher function you pass.

I'll add the `invariant` nonetheless. But I think I can do it only for result, as we can use our fallback to zero for `total` instead of throwing. Thoughts?

lib/src/components/data-table/__snapshots__/DataTable.test.tsx.snap
D7Torres: We probably copied the style from this existing test
Screenshot 2022-11-29 at 17 13 01

D7Torres: Just to clarify, as I'm not sure I understand "what has changed to break the key":

This is a new snapshot.

lib/src/components/data-table/RemoteDataTableContext.tsx
D7Torres: done, but I could use some advice regarding the logging

lib/src/components/data-table/pagination/Pagination.tsx
D7Torres: It looks like we are going to combine the 2 context providers as asked [here](https://github.com/Atom-Learning/components/pull/422#discussion_r1028117665)

So we'll have to do the above depending on `fetcher` or `data` being passed.

lib/src/components/data-table/DataTable.types.ts
```diff
@@ -0,0 +1,21 @@
...
+  setIsSortable: React.Dispatch<React.SetStateAction<boolean>>
+  applyPagination: () => void
+  getTotalRows: () => number
+  isSortable: boolean
+  apiQueryStatus?: ApiQueryStatus
+  doFetchData?: () => Promise<void>
+}
+
+export const DataTableContext =
+  React.createContext<DataTableContextType<unknown> | null>(null)
```
D7Torres, line 21: This is not a type, it should probably be defined in the place where it's imported

lib/src/components/data-table/__snapshots__/DataTable.test.tsx.snap
thomasdigby: So the issue here is that the CSS that is output to the DOM (as shown in the snapshot) contains invalid CSS. `mb` is an invalid CSS name and `var(--4)` doesn't exist in our custom properties from stitches. It should be `margin-bottom` and `var(--space-4)`.

I'm not entirely sure how this has happened though, it is probably something to do with the introduction of a `StyledTable` component in `components/data-table/DataTableTable.tsx` 🤔

lib/src/components/data-table/RemoteDataTableContext.tsx
```diff
@@ -0,0 +1,156 @@
...
+        pageIndex,
+        pageSize,
+        sortBy: sorting[0]?.id,
+        sortDirection: sorting[0] ? (sorting[0].desc ? 'desc' : 'asc') : null
+      })
+
+      invariant(
+        Array.isArray(newData?.results),
+        'The fetcher function must return an object with a property `result` which must be an array'
+      )
```
thomasdigby, line 94: I wonder if we can validate the `total` key as well with this assertion. Is it equally as important as the results array?

```ts
invariant(
  Array.isArray(newData?.results) || typeof newData?.total !== 'number',
  `The fetcher function must return an object with a shape:
  {
    results: Object[]
    total: number
  }`
)
```

thomasdigby: Sorry, didn't see your message in the other thread. I think it might be worth requiring `total` as well just to ensure that pagination is working as expected if we have some results from the API

lib/src/components/data-table/RemoteDataTableContext.tsx
thomasdigby: Logging should be sorted with the `invariant` in DEV, on PROD the error state and network tab should give enough insight. Although I wonder if we should test catching the error in the fetcher before passing into the `DataTable` so we can log errors in Sentry but still handle the error state within the table. e.g.

```tsx
<DataTable.Remote
  columns={columns}
  fetcher={(index, size) => async {
    try {
      const resp = await fetch('api?current=${index}&total=${size}')
      const data = resp.json()
      return data
    } catch (err) {
      // does this catch prevent `DataTable` from handling the error as well? Or does it bubble up here?
      // I need to learn how async/await and catch works! 😅 
      Sentry.captureException(err)
    }
  }}
>
  <DataTable.Table />
</DataTable.Remote>
```

lib/src/components/data-table/DataTable.types.ts
```diff
@@ -0,0 +1,21 @@
+import type { Table } from '@tanstack/react-table'
```
LimeWub, line 1: Why are we making this separate file? I've seen it used before but to me it's more confusing than having these where they are used.

lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -54,11 +46,13 @@ type TableProviderProps = {
 
 export const DataTableProvider = ({
   columns,
-  data,
+  data: dataProp,
   defaultSort,
   initialState = undefined,
   children
 }: TableProviderProps): JSX.Element => {
+  const [data, setData] =
```
LimeWub, line 54: Do we need a remote paginated table to store data internally? 
Can it not just be calling a data update callback outside of it - in implementation? //thinking about it//

lib/src/components/data-table/DataTableError.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { Flex } from '../flex'
+import { ApiQueryStatus } from './DataTable.types'
+import { useDataTable } from './RemoteDataTableContext'
+
+type TDataTableErrorProps = Omit<React.FC, 'children'> & {
+  css?: CSS
+  children: (retry?: () => Promise<void>) => React.ReactNode
+}
+
+export const DataTableError: React.FC<TDataTableErrorProps> = ({
```
LimeWub, line 14: Why is this not just using the `ErrorState` component?

lib/src/components/data-table/DataTableError.tsx
LimeWub, line 32: What is `children(doFetchData)` - children is a function?

lib/src/components/data-table/DataTableHeaderCell.tsx
```diff
@@ -31,26 +32,27 @@ export const DataTableHeaderCell: React.FC<DataTableHeaderProps> = ({
   ...props
 }) => {
   const sortDirection = header.column.getIsSorted()
-  const { isSortable: isSortableTable } = useDataTable()
+  const { isSortable: isSortableTable, apiQueryStatus } = useDataTable()
   // false for display columns, e.g. "Actions"
   const isDataColumn = header.column.getCanSort()
-
+  const isLoading = apiQueryStatus === ApiQueryStatus.PENDING
+  const isSortEnabled = isSortableTable && !isLoading
```
LimeWub, line 39: We clearly are disabling a bunch of logic based on `ApiQueryStatus` in multiple components - I wonder if there's a way to do this "cleaner" rather than doing checks in each of them //thinking about it - maybe not//

lib/src/components/data-table/DataTableLoading.tsx
LimeWub: So - is this a loader on the table or a loader on the pagination only?

lib/src/components/data-table/RemoteDataTableContext.tsx
```diff
@@ -0,0 +1,156 @@
...
+      pageSize: defaultPageSize
+    })
+
+  const applyPagination = React.useCallback(() => {
+    setIsPaginated(true)
+  }, [])
+
+  const getTotalRows = () => data?.total ?? 0
+
+  const doFetchData = React.useCallback(async () => {
```
LimeWub, line 81: Erm - does DataTable need to do the fetching in itself?
Couldn't there be an `onPaginationChange` function or something to do the fetching in atom-core and replace data in the table from outside?

I'm a bit wary of doing fetching logic in a DS component.

LimeWub (commented):
Hey soz I haven't spent time reviewing this prior to now. I've noticed we are doing data fetching in the DS and I'm not convinced this is the right approach so I've left some comments.

I think we should maybe reconsider this bit of the API? I'll do another pass and try to recommend an alternative - just putting this comment down to flag this.

FYI tagging @hollg @thomasdigby since u have both already reviewed this quite a bit.

LimeWub:
So the way I've imagined this would generally work for the user in atom-core with remote fetching would be something like the following:

```tsx
const { data: myTableData, status: myTableDataRequestStatus } = useApi(
  GET_WHATEVER_DATA_URL,
  {},
  { page, filterDataHere: "if we make a remote filter component too" },
  refetchKey, // If we need one. Assuming we dont if state and api props have changed but if we do we make a refetch random key function
  undefined,
  MicroservicesName.WHATEVER
)

const pageCount = myTableData.meta?.pageCount

<DataTable columns={columns} data={data} loading={myTableDataRequestStatus === ApiQueryStatus.PENDING} error={myTableDataRequestStatus === ApiQueryStatus.ERROR}>
  <DataTable.Table sortable css={{mb: '$4'}}/>
  <DataTable.Pagination pageCount={pageCount} onPageChange={(newPage) => setPage(newPage)} />
</DataTable>
```

This:
* keeps the DS data agnostic
* all fetching happens outside of the DS table so its clear to end user what is being fetched and when
* we can use APIs with a different format to the table and pass in data that way
* we don't include ApiQueryStatus and data fetch constants in the DS
* DataTable context remains relatively simple

What we are keeping that's useful from this PR:
`DataTable.Loading`, `DataTable.Error` both for when someone wants to build their own custom table setup and needs to access the components directly
What we drop:
`DataTable.Remote` as a separate context, including all fetching etc BUT we might need tweaks to Pagination to force page count etc - but shouldn't be too many

I haven't been deeply involved with this PR - at least not prior to today but is there any reason why my above suggested approach wouldn't work OR why this current approach with the internal fetching is preferred?
I'm high key concerned about the internal fetching in DS (which is why I took the time to write this) but if there's a solid reason for it to exist in here I could be convinced.

I'd be happy to join a call about this or continue convo here.

lib/src/components/data-table/DataTableError.tsx
D7Torres: done

lib/src/components/data-table/DataTable.mdx
```diff
@@ -1,6 +1,6 @@
 ---
 title: DataTable
-component: DataTable, DataTable.Table, DataTable.Head, DataTable.HeaderCell, DataTable.Body, DataTable.Row, DataTable.DataCell, DataTable.GlobalFilter
+component: DataTable, DataTable.Remote, DataTable.Head, DataTable.HeaderCell, DataTable.Body, DataTable.Row, DataTable.DataCell, DataTable.GlobalFilter, DataTable.Remote, DataTable.Loading, DataTable.Error
```
D7Torres, line 3: Did we removed `DataTable.Table` and have `DataTable.Remote` twice?

lib/src/components/data-table/__snapshots__/DataTable.test.tsx.snap
Marcdup1: Nice catch! we fixed it here: [d543339](https://github.com/Atom-Learning/components/pull/422/commits/d5433391d30d14a8f325eae088b93f2e585fc987)

lib/src/components/data-table/DataTableTable.tsx
Marcdup1: Updated the loading state based on the suggestion: 
https://user-images.githubusercontent.com/11061661/205672772-ff2f4b3b-e4a6-4197-9e46-71b244779646.mov

lib/src/components/data-table/DataTable.types.ts
Marcdup1: Done

lib/src/components/data-table/DataTable.types.ts
Marcdup1: We kept the types in a separate file as many of them are imported into different files

lib/src/components/data-table/DataTableError.tsx
Marcdup1: There is an `ErrorState` component?

lib/src/components/data-table/pagination/Pagination.tsx
Marcdup1: We have 4 `AsyncDataState` states: `NONE`, `PENDING`, `FULFILLED` & `REJECTED`, by default it will be `NONE`

lib/src/components/data-table/DataTableContext.tsx
Marcdup1: We discussed offline - but for visibility, we are keeping the current API.

lib/src/components/data-table/DataTableError.tsx
Marcdup1: Anyway, we are letting the user pass whatever they want so they have that flexibility.

lib/src/components/data-table/RemoteDataTableContext.tsx
Marcdup1: We discussed offline - but for visibility, we are keeping the current API.

lib/src/components/data-table/DataTable.mdx
Marcdup1: Great catch! fixed it with you 😉 [4ea5901](https://github.com/Atom-Learning/components/pull/422/commits/4ea5901ca80011a0d304acfd32a0e0ead4c72566)

lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -46,22 +47,56 @@ type InitialState = Partial<
...
+
+  const [paginationState, setPagination] = React.useState<
+    PaginationState | undefined
+  >(
+    asyncData
+      ? {
+          ...defaultPaginationState,
+          ...initialState?.pagination
+        }
+      : initialState?.pagination
```
hollg, line 98: What's the reason for not using `initialState.pagination` if we have `asyncData`?

lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -46,22 +47,56 @@ type InitialState = Partial<
...
   defaultSort?: { column: string; direction: 'asc' | 'desc' }
   children: React.ReactNode
   initialState?: InitialState
-}
+  css?: CSS
+} & (
+  | { data: Array<Record<string, unknown>>; asyncData?: never }
+  | {
+      data?: never
+      asyncData: (
```
hollg, line 58: Apologies if I've missed a conversation about this name but `asyncData` feels off to me, because this is a func that fetches the data, rather than the data itself. I would call it `fetcher` or similar personally

lib/src/components/data-table/pagination/Pagination.tsx
```diff
@@ -30,23 +27,22 @@ export const Pagination: React.FC<PaginationProps> = ({
...
-  }, [setPageSize, pageSize])
-
   const { pagination: paginationState } = getState()
 
   const recordsCountFrom =
     paginationState.pageIndex * paginationState.pageSize + 1
   const recordsCountTo = recordsCountFrom + getRowModel().rows.length - 1
+  const isPending = asyncDataState === AsyncDataState.PENDING
+
+  if (asyncDataState === AsyncDataState.REJECTED) return null
```
thomasdigby, line 45: We probably want to maintain some form of context for the user if the `asyncData` rejects, hiding both content and pagination feels a bit over the top to me. Maybe we can disable the controls similar to the `isPending` state? I think it's important that we get this right, we're not great at error handling and resolution in the core app

lib/src/components/data-table/pagination/Pagination.tsx
```diff
@@ -2,26 +2,23 @@ import * as React from 'react'
...
-  pageSize?: number
-}
+type PaginationProps = React.ComponentProps<typeof StyledNav>
 
 /** Applies pagination to parent DataTableProvider and renders UI to switch pages etc */
-export const Pagination: React.FC<PaginationProps> = ({
-  pageSize = 10,
-  ...props
-}) => {
+export const Pagination: React.FC<PaginationProps> = ({ ...props }) => {
```
thomasdigby, line 21:
```suggestion
export const Pagination: React.FC<PaginationProps> = (props) => {
```

lib/src/components/data-table/DataTableLoading.tsx
```diff
@@ -0,0 +1,29 @@
...
+    <Loader
+      css={{
+        position: 'absolute',
+        left: '50%',
+        top: '50%',
+        transform: 'translate(-50%, -50%)',
+        zIndex: 1,
+        ...css
+      }}
+    />
```
thomasdigby, line 27: Could we update this to be a `styled` component? It would remove the need for the custom CSS type and `css` prop e.g.

```tsx

import { Loader } from '../loader'
import { AsyncDataState } from './DataTable.types'
import { useDataTable } from './DataTableContext'

const PendingState = styled(Loader, {
  position: 'absolute',
  left: '50%',
  top: '50%',
  transform: 'translate(-50%, -50%)',
  zIndex: 1
)

export const DataTableLoading: React.FC<React.ComponentProps<typeof PendingState>> = (props) => {
  const { asyncDataState } = useDataTable()

  if (asyncDataState !== AsyncDataState.PENDING) return null

  return <PendingState {...props} />
}
```

lib/src/components/data-table/DataTable.mdx
```diff
@@ -133,6 +133,70 @@ Note also that `useDataTable` can only be called by a child component of `DataTa
...
+  asyncData={async ({ pageIndex, pageSize, sortBy, sortDirection }) => {
+    const nextPage = pageIndex + 1
+    const data = await fetch(
+      `https://your-api?page=${nextPage}&pageSize=${pageSize}&order=${sortBy}&dir=${sortDirection}`
+    )
+    const results = await data.json()
+
+    return {
+      results,
+      total: results.total
```
thomasdigby, line 179: We should change this example as `total` needs to represent the total number of items, not just the number of items for this page. I also wonder if we could tweak the way we construct the search params to be clearer about the relationship between the data available through the `asyncData` function. e.g.

```ts
asyncData={async ({ pageIndex, pageSize, sortBy, sortDirection }) => {
  const params = new URLSearchParams({
    page: pageIndex,
    pageSize,
    order: sortBy,
    dir: sortDirection
  })
  const response = await fetch(`https://your-api?${params.toString()}`)
  const { results, total } = await data.json()

  return { results, total }
}}
```

lib/src/components/data-table/DataTable.mdx
```diff
@@ -133,6 +133,70 @@ Note also that `useDataTable` can only be called by a child component of `DataTa
...
+      return (
+        <>
+          <InlineMessage>Error occured</InlineMessage>
+          <Button onClick={() => retry?.({ pageIndex: 3, pageSize: 10 })}>
+            Retry
+          </Button>
+        </>
+      )
+    }}
+  </DataTable.Error>
```
thomasdigby, line 195: It would be good to simplify this example a touch and write more on the `retry` function params in a separate part of the docs. I think before we have an `ErrorState` component to recommend in the docs, we should keep it simple to avoid unintentionally recommending certain components to use, e.g. `InlineMessage`.

```tsx
<DataTable.Error>
  {(retry) => <Button onClick={retry}>Try again</Button>}
</DataTable.Error>
```

lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -46,22 +47,56 @@ type InitialState = Partial<
...
 }: TableProviderProps): JSX.Element => {
+  const [data, setData] = React.useState<TAsyncDataResult>({
+    results: dataProp ?? [],
+    total: dataProp?.length ?? 0
+  })
+
   const [isPaginated, setIsPaginated] = React.useState<boolean>(
     !!initialState?.pagination
   )
+  const [asyncDataState, setApiQueryStatus] = React.useState<AsyncDataState>(
```
thomasdigby, line 86:
```suggestion
  const [asyncDataState, setAsyncDataState] = React.useState<AsyncDataState>(
```

lib/src/components/data-table/DataTableError.tsx
thomasdigby: I'd actually remove the custom CSS we have in this component and defer all the styling to the user, there are some very specific styles here that may need to be overridden. We don't have an `ErrorState` component at the moment, but when we create one, I can imagine we'll need different default styling here anyway, e.g. vertical/horizontal centering.

```tsx
return children(runAsyncData)
```

lib/src/components/data-table/DataTable.mdx
thomasdigby: This example has actually made me consider whether we should provide a way to cancel the rejected async data call and render the previous data from before, e.g. click next page, it fails, click "cancel" instead of "try again", current page re-renders.

```tsx
<DataTable.Error>
  {(retry, cancel) => (
    <Stack>
      <Button onClick={retry}>Try again</Button>
      <Button onClick={cancel} appearance="outline">Cancel</Button>
    </Stack>
  )}
</DataTable.Error>
```

lib/src/components/data-table/DataTableBody.tsx
```diff
@@ -12,7 +14,9 @@ export const DataTableBody: React.FC<DataTableBodyProps> = ({
   striped = false,
   ...props
 }) => {
-  const { getRowModel } = useDataTable()
+  const { getRowModel, asyncDataState } = useDataTable()
+
+  if (asyncDataState === AsyncDataState.REJECTED) return null
```
thomasdigby, line 19: Based on similar feedback for pagination and on the `DataTable.Error` documentation, I think we ought to try to render something if the async call is rejected, can we retain the current page information?

lib/src/components/data-table/DataTableContext.tsx
Marcdup1: We still use `initialState.pagination` if `asyncData` is present but we first want to ensure the `defaultPaginationState` is applied (page = 0, pageSize = 10) in case no initial state was passed.

lib/src/components/data-table/pagination/Pagination.tsx
Marcdup1: Okay that makes sense - we weren't 100% sure whether to add or remove the pagination component

Marcdup1: [4b862fd](https://github.com/Atom-Learning/components/pull/422/commits/4b862fdce2c514a74bb8d5335935334a3f27af05)

lib/src/components/data-table/pagination/Pagination.tsx
Marcdup1: Updated

lib/src/components/data-table/DataTableLoading.tsx
Marcdup1: > It would remove the need for the custom CSS type and css prop e.g.

Oooh okay, makes sense!

lib/src/components/data-table/DataTableBody.tsx
Marcdup1: I think we do currently retain the current page's info (Although it's technically the previous page). Will make the change

Marcdup1: [43fab2f](https://github.com/Atom-Learning/components/pull/422/commits/43fab2f937319527729bfa6da3fe61325d36e63c)

lib/src/components/data-table/DataTable.mdx
Marcdup1: Oh yes nice catch,  and +1 for the search params

lib/src/components/data-table/DataTable.mdx
Marcdup1: I simplified the docs and added a bit more information regarding the retry: [2d19566](https://github.com/Atom-Learning/components/pull/422/commits/2d19566a612914aedb1b59cae5d898df46867d4f)

I think with implementing the "cancel" the only tricky part would be restoring the pagination state to the previous pageIndex - might be easy but didn't check, but we have the previous data set so that wont be a problem as we don't clear it if it fails

lib/src/components/data-table/DataTableContext.tsx
Marcdup1: I think @thomasdigby might explain the reasoning better but I think `fetcher` caused confusion bit by conveying that the datatable component was doing the data fetching. I'm happy to come to a compromise as I agree that `asyncData` does not convey much about it, perhaps `asyncDataFetcher` 🤷‍♂️

lib/src/components/data-table/DataTableBody.tsx
Marcdup1: Important to note, now the error will be displayed either below or above the table depending on where the user mounts it

lib/src/components/table/Table.tsx
```diff
@@ -51,6 +51,12 @@ const StyledTable = styled('table', {
         }
       },
       square: {}
+    },
+    isPending: {
+      true: {
+        opacity: 0.5,
```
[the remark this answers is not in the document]
thomasdigby: `pointer-events: none` also disables interaction with any user-rendered buttons/links within the table contents, e.g. table actions like "Edit".

I'd assume that in a loading/pending state we wouldn't want them to be able to use those either

lib/src/components/data-table/DataTableContext.tsx
thomasdigby: @LimeWub raised a valid point about our API usage leaking into the design system and muddying the waters of what's acceptable to use in that context, so we leant more into the fact that it's just "promisified data" rather than anything to `fetch`. It essentially returns `data` as per the other prop but as a promise for `DataTable` to resolve.

Something I didn't consider though is that it does call a _function_ to return that promise, rather than passing the promise in directly. I'm not super concerned about the naming here, so happy to run with whatever either of you decide 👍

lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -79,39 +114,102 @@ export const DataTableProvider = ({
     setIsPaginated(true)
   }, [])
 
-  const getTotalRows = () => data.length
+  const runAsyncData = React.useCallback(
```
thomasdigby, line 117: Something I just considered, would it be worth pulling this `runAsyncData` method into a separate file? I think it adds a bit too much complexity inline here... just a thought 🤔 

```ts
// just getting my thoughts down in code
const runAsyncData = React.useCallback(
  async (params) => {
    try {
      setAsyncDataState(AsyncDataState.PENDING)

      const data = getDataFromSeparateFunction(params)
  
      setData(data as TAsyncDataResult)
      setAsyncDataState(AsyncDataState.FULFILLED)
    } catch (error) {
      setAsyncDataState(AsyncDataState.REJECTED)
    }
  },
  [asyncData, paginationState?.pageIndex, paginationState?.pageSize, sorting]
)
```

lib/src/components/data-table/RemoteDataTableContext.tsx
thomasdigby: Sorry, it's early for me still 😅, will it throw within the `fetcher` as above and call `Sentry`?

Marcdup1: @thomasdigby yep it will, if something happens here:
```tsx
      const resp = await fetch('api?current=${index}&total=${size}')
      const data = resp.json()
      return data
```
it will be caught

lib/src/components/data-table/DataTableContext.tsx
Marcdup1: @hollg we changed the name to `getAsyncData` [c00e204](https://github.com/Atom-Learning/components/pull/422/commits/c00e204cc214a85a2c9c484a12b866b72a0adb74). This at least indicates that its a function and coincides with the naming for some methods in `useReactTable`.

lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -79,39 +111,82 @@ export const DataTableProvider = ({
...
+      asyncDataState,
+      runAsyncData
     }
   }, [table, applyPagination, getTotalRows, isSortable])
 
   return (
-    <DataTableContext.Provider value={value}>
-      {children}
-    </DataTableContext.Provider>
+    <Box css={{ position: 'relative', ...css }}>
```
hollg, line 184: Sorry, I think I missed this in my previous review. I don't think `DataTableContext` should render anything. Can we move `DataTableLoading` to `DataTableTable` instead? Or has that already been tried and not worked?

The reason is that if the context provider doesn't render anything then the consumer has total freedom to structure their code, e.g. you can put the provider at the top level of a page and give components outside of the actual table sight of the data as it updates. We lose that ability if we start rendering things like the loader in here, because in that case the whole page would get the loader instead of just the table.

Marcdup1: @hollg oh wow valid point! I moved the loader to `DataTableTable` [bc23383](https://github.com/Atom-Learning/components/pull/422/commits/bc23383c2912143d523a1d7821d7741f2490bac0)

hollg: Nice, thanks! Will take another look this afternoon but I think that was the last thing I could see to comment on. I actually can't approve though because I opened the PR 😩

Marcdup1: No problem thank you! Oh yes of course 😅 I can approve it on behalf of you once you are happy with everything

thomasdigby (approved):

lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -79,32 +109,72 @@ export const DataTableProvider = ({
...
+      } catch (error) {
+        setAsyncDataState(AsyncDataState.REJECTED)
+      }
+    },
+    [
+      getAsyncData,
+      paginationState?.pageIndex,
+      paginationState?.pageSize,
+      sorting
+    ]
```
hollg, line 137: Should this dependency array include the value from `DataTable.GlobalFilter` too?

Marcdup1: I completely forgot about the global filter! I think it will be nice to add and we can include it as a param to `getAsyncData`. I feel it would be confusing when using `getAsyncData` and the global filter does nothing.

lib/src/components/data-table/DataTable.mdx
```diff
@@ -168,35 +168,28 @@ You can use `DataTable.Error` to display your own error component when the `asyn
...
-        </>
-      )
-    }}
+    {(retry) => <Button onClick={retry}>Try again</Button>}
   </DataTable.Error>
   <DataTable.Pagination />
 </DataTable>
 ```
 
+`DataTable.Error` provides a `retry` function to the children which allows you to recall the `asyncData` function. The `retry` function can be called with all the paginated parameters as an optional object. If no parameters are provided, `retry` will be called with the last paginated options.
```
D7Torres: Very minor: a code example with all the options in use would be good

D7Torres (approved):

lib/src/components/data-table/DataTable.mdx
Marcdup1: Added [dc9e087](https://github.com/Atom-Learning/components/pull/422/commits/dc9e087bbf6594b32c4d547fa2bbbcbc41ccfeae)

lib/src/components/data-table/DataTableContext.tsx
Marcdup1: @hollg  I add the global filter support here: [504cd7e](https://github.com/Atom-Learning/components/pull/422/commits/504cd7ea84f3cf822d6e18c30a285aba169353b0)

hollg: Awesome! Feel free to approve on my behalf now 😄 

Also, thanks for linking to the commits where you made the changes on all these threads. Makes it very easy to follow up!

Marcdup1 (approved):

PR #426, opened 2022-11-10
feat: navigation-menu custom trigger component

michalrozenek:
This PR is about to introduce a custom trigger component.

## Solution
This solution is really flexible. It let us keep the whole Dropdown Trigger functionality like content toggling with arrows up/down indication as normal Trigger does. All states with css (hover/active/focus) are kept as well.

I added new element called `NavigationMenu.Trigger` which is optional.
`NavigationDropdown` expects specific children. It should be `NavigationMenu.DropdownTrigger`, `NavigationMenu.DropdownContent` or both.

If `NavigationMenu.Trigger` is passed as a part of children, it will be rendered next to the `arrow`.

In this example we render Icon as a Trigger

```
const TriggerComponent = () => (
  <Flex
    css={{
      alignItems: 'center',
      border: '2px solid $tonal500',
      borderRadius: '50%'
    }}
  >
    <Icon is={Person} />
  </Flex>
)

<NavigationMenu.DropdownTrigger>
    <TriggerComponent />
</NavigationMenu.DropdownTrigger>
```

Zrzut ekranu 2022-11-14 o 17 15 08

If  `<NavigationMenu.DropdownTrigger>` is not declared, we render `title` as a `Trigger`, just like we did before this PR.

Zrzut ekranu 2022-11-14 o 17 15 18

❌ I ALSO ADDED ERROR IF PASSED CHILDREN IS NOT EXPECTED COMPONENT

Zrzut ekranu 2022-11-14 o 14 05 20
Zrzut ekranu 2022-11-14 o 14 05 43

ℹ️ ⚠️ 
THIS PR INCLUDES LOT OF CLEANUP. I MOVED COMPONENTS WITH THEIR STYLES AND CONSTANTS TO SEPARATE FILES. ⚠️  ℹ️ 

## Screenshots and videos:

https://user-images.githubusercontent.com/23363033/201135899-8c0351cd-d220-4f78-9511-3ae3481f30b4.mov
Zrzut ekranu 2022-11-10 o 14 55 47
Zrzut ekranu 2022-11-10 o 14 55 52

After re-write to `NavigationMenu.Trigger`.

https://user-images.githubusercontent.com/23363033/201710865-62507a84-45f4-45a3-b3c4-fe78658cc899.mov


Mhoog (approved):

lib/src/components/navigation/NavigationMenuDropdown.tsx
```diff
@@ -33,7 +36,7 @@ export const NavigationMenuDropdown: React.FC<NavigationMenuDropdownProps> = ({
         {...props}
         ref={(node: HTMLButtonElement) => onNodeUpdate(node, title)}
       >
-        {title}
+        {triggerComponent || title}
```
LimeWub, line 39: Rather than doing this, is there no way to export `NavigationDropdown.Trigger` itself?

It will need to be a forwardRef component for the ref bit but otherwise should be fine to do?

thomasdigby: Yeah I'd recommend the same 👍

lib/src/components/navigation/NavigationMenu.mdx
```diff
@@ -34,7 +34,7 @@ category: Navigation
 
 #### Example using React Router
 
-```tsx
+```tsx preview
```
thomasdigby, line 37: Does this preview work on the docs site? I'd assume it wouldn't as you can only return JSX components within a `tsx preview` block.

lib/src/components/navigation/NavigationMenuDropdown.tsx
elliotlaws: I have some code stashed from a while ago that enabled this kind of API, but it did require a lot more complexity  (mapping and cloning children etc.)

image

elliotlaws: [changes.patch.zip](https://github.com/Atom-Learning/components/files/9990201/changes.patch.zip)

I've exported the stash so anybody can take a look. Never done this before but you should be able to see this on your local by following the instructions from [here](https://www.vasos-koupparis.com/how-to-share-a-git-stash-with-another-developer-or-yourself-to-another-machine/)

lib/src/components/navigation/NavigationMenu.mdx
```diff
@@ -119,3 +119,31 @@ DropdownItem gives a lot of flexibility. It's an easy to compose it for own purp
   </Grid>
 </NavigationMenu.DropdownItem>
 ```
+
+## Customizable Dropdown Trigger
+
+NavigationMenu.Dropdown gives you the way to exchange the default Trigger text with a custom component.
+Be aware that 'title' property is required in any case to satisfy underlying operations.
```
andoulla, line 126: is title used for accessibility reasons? it might be worth pointing this out as if it is it needs to be passing a translated string

andoulla (approved):

lib/src/components/navigation/NavigationMenuDropdown.tsx
michalrozenek: Thanks @elliotlaws .
As I shared with you on slack - I already did something really similar on Friday.
Thanks anyway!

lib/src/components/navigation/NavigationMenu.mdx
michalrozenek: You're right

lib/src/components/navigation/NavigationMenu.constants.ts
```diff
@@ -0,0 +1,43 @@
+export const navigationMenuActiveParentItemStyles = {
```
michalrozenek, line 1: NOTE FOR REVIEWERS:
This is moved from `NavigationMenuItem.tsx` to own file

lib/src/components/navigation/NavigationMenuDropdownContent.tsx
```diff
@@ -0,0 +1,25 @@
+import * as NavigationMenuPrimitive from '@radix-ui/react-navigation-menu'
```
michalrozenek, line 1: NOTE FOR REVIEWERS:
This is moved from NavigationMenuItem.tsx to own file

lib/src/components/navigation/NavigationMenuDropdownItem.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
```
michalrozenek, line 1: NOTE FOR REVIEWERS:
This is moved from NavigationMenuItem.tsx to own file

lib/src/components/navigation/NavigationMenu.constants.ts
LimeWub: Do these not have more relevant files to be added in?

They seem to be Item styles so why did they need to be moved? (Separate consts are fine if you need them exported for other components ?)

lib/src/components/navigation/NavigationMenu.mdx
LimeWub: What underlying operations?

lib/src/components/navigation/NavigationMenuDropdown.tsx
```diff
@@ -1,41 +1,81 @@
...
+import { NavigationMenuDropdownTrigger } from './NavigationMenuDropdownTrigger'
 
 type NavigationMenuDropdownProps = {
-  title: string
   active?: boolean
+  css?: CSS
+  title: string
 } & React.HTMLProps<HTMLButtonElement>
 
+const DefaultDropdown: React.FC<
```
LimeWub, line 16: Why is this component needed? Aren't the NavigationMenuDropdown components exported individually anyway?

lib/src/components/navigation/NavigationMenuDropdown.tsx
```diff
@@ -1,41 +1,81 @@
...
 }) => {
   const { onNodeUpdate } = useNavigationMenuContext()
 
+  const hasCustomTrigger = React.Children.toArray(children).some(
+    (child) =>
+      React.isValidElement(child) &&
+      child.type === NavigationMenuDropdownTrigger
+  )
+
   return (
```
LimeWub, line 48: This is really complicated ;-;
Is this all done to pass `onNodeUpdate` with a node up to something to calculate the position or...?

lib/src/components/navigation/NavigationMenu.constants.ts
michalrozenek: Exactly that's the case - Few components uses these styles.
I was wondering how we should make them reusable in DS and I am not entirely sure. I thought .constant file might work. If you have some standard already or better idea i am keen to hear.

lib/src/components/navigation/NavigationMenu.mdx
michalrozenek: Zrzut ekranu 2022-11-15 o 06 46 18
This comment is about the code introduced by @elliotlaws . Do you have any suggestions how should I describe it better? Maybe we should mention that we need it to calculate position, but depends if this is important to library user, is it?

lib/src/components/navigation/NavigationMenuDropdown.tsx
michalrozenek: This is default Dropdown (trigger + content). This is rendered when developer did not pass `<NavigationMenu.DropdownTrigger ... /> to the `NavigationMenu.Dropdown children`.

michalrozenek: Please take a look at the code - we check if there's any `custom trigger` and render mentioned `default` if there's not

```
{!hasCustomTrigger ? (
    <DefaultDropdown title={title} onNodeUpdate={onNodeUpdate} {...props}>
        {children}
    </DefaultDropdown>
) : (
```

lib/src/components/navigation/NavigationMenuDropdown.tsx
michalrozenek: 1. We need to pass `onNodeUpdate` - yes,
2. We need to support both scenarios:
a) when user pass custom trigger - we need to map through children and search for `trigger` element, then we need to pass ref and props.

b) if used do not pass custom trigger, we need a little bit different approach to actually **render** `trigger` (because it's not passed as a children) with a `title`, pass ref to it. And render content below.

The most important is - We do not know if user passed `DropdownTrigger` component or not, that's why I need to detect if and cover both scenarios.

If we decide to basically **require** `NavigationMenu.DropdownTrigger` to be passed as a `children`, then we can simplify this component. But for now I made it the way that it's not mandatory, and we render fallback option with `title` as we had before my changes.

Mhoog (approved):

lib/src/components/navigation/NavigationMenuDropdownTrigger.tsx
```diff
@@ -0,0 +1,50 @@
...
+    justifyContent: 'space-between',
+    gap: '$1',
+    '&[data-state="open"]': {
+      background: '$tonal100'
+    },
+    variants: { active: { true: { ...navigationMenuActiveParentItemStyles } } }
+  }
+)
+
+export const NavigationMenuDropdownTrigger = React.forwardRef<
```
LimeWub, line 29: This is potentially an insane solution but I wrote one where `onNodeUpdate` is ran only on `data-state` attribute change of the trigger (and only gets passed in a ref of only that one trigger at the time).

https://codesandbox.io/s/navigation-menu-track-position-forked-okosyy?file=/src/App.js:0-3526

This means less calls to the functions and we can decouple the `title`/`value` needing to be passed down. No longer require a 'title', no longer need a forwardRef on this component and no need to check direct children and pass anything down to them from NavigationMenu.

...perhaps there's a cleaner solution even but I think it's worth considering the one in the above sandbox as it will lead to much less code (and much less complexity imo).

lib/src/components/navigation/NavigationMenuLink.tsx
```diff
@@ -0,0 +1,88 @@
...
+import React from 'react'
+
+import { CSS, styled } from '~/stitches'
+import {
+  navigationMenuActiveParentItemStyles,
+  navigationMenuDisabledStyles,
+  navigationMenuItemStyles
+} from './NavigationMenu.constants'
+
+const DisabledButton = styled('button', {
```
LimeWub, line 11: Why does this need to be its own component - can't the disabled styles be on the link directly?

lib/src/components/navigation/NavigationMenuLink.tsx
```diff
@@ -0,0 +1,88 @@
...
+  ...navigationMenuItemStyles,
+  ...navigationMenuDisabledStyles
+})
+
+const ListItem = styled('li', { unset: 'all' })
```
LimeWub, line 16: Rather than this - can it instead use `NavigationMenu.Item`?

lib/src/components/navigation/NavigationMenuLink.tsx
michalrozenek: Sorry, honestly I just copied a lot of code and moved to separate files for better readability. Did not consider if we need this button - it would take another extra time

michalrozenek: I thought about `disabled` state and it cannot be applied to the `link` (`a`) element.
https://www.w3schools.com/tags/att_disabled.asp

I will keep it as it is as we discussed yesterday, but I am wondering if we should support `disabled` state in that case. I don't see too many situations when we might need it.

LimeWub (approved):

elliotlaws (approved):

PR #428, opened 2022-11-11
feat: pass onClick prop onto anchor links

Mhoog:
## Issue
We were stripping the `onClick` prop when we rendered an `a` in `Link`.

In the main repo, we had several `Link` implementations with both an `href` and `onClick`. Because we had a wrapper component that rendered this `Link` component as a `react-router` `Link`, the `onClick` would still be fired (this component would render a `button` that would then be rendered as a `react-router` `Link` that rendered an `a`, retaining both the `href` and `onClick`.

_(kudos if you're still following me)_

To make sure `noCapsize` was only applied to `button` links, we had to make a change to this wrapper and make sure it rendered an `a` straight away when creating an internal link. This broke links with both `onClick` and `href` as we strip `onClick` in this `Link` component.

## Changes
- Pass on `onClick` to both `a` and `button` links

### DIFF
--- a/lib/src/components/link/Link.tsx
+++ b/lib/src/components/link/Link.tsx
@@ -41,19 +41,15 @@ type LinkProps = Override<
 >
 
 export const Link = React.forwardRef<HTMLAnchorElement, LinkProps>(
-  ({ size = 'md', onClick, href, ...remainingProps }, ref) =>
-    href ? (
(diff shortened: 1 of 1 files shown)


thomasdigby (approved):
😅

andoulla (approved):

PR #429, opened 2022-11-17
feat: sidedrawer - core components foundations

michalrozenek:
## Sidedrawer component

Zrzut ekranu 2022-11-28 o 13 52 42

Ticket: https://atomlearningltd.atlassian.net/browse/DS-49

### Implementation

This PR contains whole Sidedrawer implementation. 
- Sidedrawer component with all subcomponents
- Unit tests
- Documentation

More about the structure and usage in the next section

### Structure and usage

`Sidedrawer` exports components that combine to build complex side navigation.

## The main sidedrawer structure

Sidedrawer exposes few main elements to build foundations.
The core element is a `Sidedrawer` which wraps entire components composition.
`Sidedrawer` has a controlled state, that's why it requires `isOpen` and `onClose` properties.

- `isOpen` property determines if the whole `Sidedrawer` is visible or not.
- `onClose` prop should be a function which updates `isOpen` state. This function is triggered automatically when click on `Sidedrawer.Overlay` or press `Escape` key.

Other main elements:

- `Sidedrawer.Header` - It's based on `TopBar` component, so it plays really well with all `TopBar` subcomponents. This component is flexible, so you can render anything you need inside of it (for example app logo, Sidedrawer.Close etc.)
- `Sidedrawer.Close` - It's an `"close" ActionIcon` which requires `onClose` prop.
- `Sidedrawer.Content` - This is the wrapper for all `Sidedrawer.Item` items. The content is scrollable.
- `Sidedrawer.Footer` - Optional part of the `Sidedrawer`. It's fixed to the botton. You can pass whatever you need as it's children.
- `Sidedrawer.Overlay` - Full screen overlay positioned under the `Sidedrawer menu`. It inherits the `onClose` prop from `Sidedrawer`, so you can close the whole menu by clicking on `Overlay` area. This in an optional element.

```tsx
<Sidedrawer isOpen onClose={() => console.log('closed')}>
  <Sidedrawer.Header>
    <Sidedrawer.Close onClose={() => console.log('closed')} />
  </Sidedrawer.Header>
  <Sidedrawer.Content>Sidedrawer items and accordtions here</Sidedrawer.Content>
  <Sidedrawer.Footer>Footer items here</Sidedrawer.Footer>
  <Sidedrawer.Overlay />
</Sidedrawer>
```

## Sidedrawer items

Sidedrawer exposes `Item` component to build a navigation elements.
`Sidedrawer.Item` is only component you need to define `Accordion item`, `Link` or a `Button`.
This component lets you pass various composition of props which determines what kind of element you'll render.

- `href` prop determine a link render
- `onClick` prop determines a button render
- `value` determines accordion item render. Accordion expects few other elements which are described below in this section.

**Link item:**

```tsx
<Sidedrawer.Item href={'/'}>
  <Icon is={Apps} css={{ mr: '$2' }} />
  Dashboard
</Sidedrawer.Item>
```

**Button item:**

```tsx
<Sidedrawer.Item onClick={logout}>
  <Icon is={Exit} css={{ mr: '$2' }} />
  Log out
</Sidedrawer.Item>
```

**Accordion item:**

Accordion item requires few elements to work properly:

- `value` property is a unique identifier needed to expand accordion content. [read more](https://www.radix-ui.com/docs/primitives/components/accordion#item)
- `Sidedrawer.AccordionTrigger` - This component is a trigger which expand/collapse accordtion content. You can pass a text or components composition as its children.
- `Sidedrawer.AccordionContent` - This component contains whole accorion content. Most likely you would like to pass few another `Sidedrawer.Item` items inside. Accordion content is collapsed until you click the `Sidedrawer.AccordionTrigger` to expand it.

```tsx
<Sidedrawer.Item value="your_unique_identifier">
  <Sidedrawer.AccordionTrigger>
    <Icon is={New} css={{ mr: '$2' }} />
    Set work
  </Sidedrawer.AccordionTrigger>
  <Sidedrawer.AccordionContent>
    <Sidedrawer.Item href={'/practice'}>
      <Icon is={Feed} css={{ mr: '$2' }} />
      Practice
    </Sidedrawer.Item>
    <Sidedrawer.Item href={'/test'}>
      <Icon is={Inbox} css={{ mr: '$2' }} />
      Test
    </Sidedrawer.Item>
  </Sidedrawer.AccordionContent>
</Sidedrawer.Item>
```

## All Sidedrawer components composition

```tsx
<Sidedrawer isOpen onClose={() => console.log('closed')}>
  <Sidedrawer.Header>
    <Sidedrawer.Close onClose={() => console.log('closed')} />
    <TopBar.BrandLogo
      src={'https://atomlearning.co.uk/_next/static/media/logo.e9276284.svg'}
    />
  </Sidedrawer.Header>
  <Sidedrawer.Content>
    <Sidedrawer.Item href={'/'} active>
      Dashboard
    </Sidedrawer.Item>
    <Sidedrawer.Item value="1" disabled>
      <Sidedrawer.AccordionTrigger>Set work</Sidedrawer.AccordionTrigger>
      <Sidedrawer.AccordionContent>
        <Sidedrawer.Item href={'/'}>Set Practice</Sidedrawer.Item>
        <Sidedrawer.Item href={'/'}>Mock Tests</Sidedrawer.Item>
      </Sidedrawer.AccordionContent>
    </Sidedrawer.Item>
    <Sidedrawer.Item value="2">
      <Sidedrawer.AccordionTrigger>Resources</Sidedrawer.AccordionTrigger>
      <Sidedrawer.AccordionContent>
        <Sidedrawer.Item onClick={() => console.log('clicked!')}>
          I am a button
        </Sidedrawer.Item>
        <Sidedrawer.Item active href="/">
          I am a link
        </Sidedrawer.Item>
      </Sidedrawer.AccordionContent>
    </Sidedrawer.Item>
  </Sidedrawer.Content>
  <Sidedrawer.Footer>The footer content</Sidedrawer.Footer>
  <Sidedrawer.Overlay />
</Sidedrawer>
```

### Videos and Screenshots

Base state item (link/button)
Zrzut ekranu 2022-11-28 o 13 49 19

Base state Accordion item
Zrzut ekranu 2022-11-28 o 13 50 04

Active state (normal & disabled)
Zrzut ekranu 2022-11-28 o 13 50 30
Zrzut ekranu 2022-11-28 o 13 49 08

https://user-images.githubusercontent.com/23363033/204284269-726b9734-6d10-41f7-9e90-9cc61a0d80bb.mov
https://user-images.githubusercontent.com/23363033/204284347-d6f9f202-acf0-40b1-b41f-0ff419673bc1.mov


lib/src/components/sidedrawer/Sidedrawer.tsx
```diff
@@ -0,0 +1,85 @@
...
+    if (React.isValidElement(child) && child.type === SidedrawerOverlay) {
+      overlayChild = child as unknown as typeof SidedrawerOverlay
+    } else {
+      contentChildren.push(child)
+    }
+  })
+
+  return (
+    <>
+      <StyledBox open={isOpen}>{contentChildren}</StyledBox>
```
thomasdigby, line 71: Are we trapping focus within the sidedrawer when open? If not, (or even if we are) I'd recommend utilising Radix's Dialog component to trap focus and add controls like ESC to close.

elliotlaws:
Do you think it's worth animating the accordion dropdown?

michalrozenek:
> Do you think it's worth animating the accordion dropdown?

Why not?
I added simple `fadeIn` animation. Looks a little bit smoother.
Do you think we don't need it?

lib/src/components/sidedrawer/Sidedrawer.tsx
michalrozenek: Thanks @thomasdigby .
I will take a look at it.

elliotlaws:
> > Do you think it's worth animating the accordion dropdown?
> 
> Why not? I added simple `fadeIn` animation. Looks a little bit smoother. Do you think we don't need it?

I may be missing something but it looks like from the video, that the accordion content (which drops down) doesn't seem to be animated. I was just wondering if adding some animation here would make it look smoother.

michalrozenek:
> 

Thank you.
Maybe I can improve the animation or make it longer and more noticeable

michalrozenek:
> > > Do you think it's worth animating the accordion dropdown?
> > 
> > 
> > Why not? I added simple `fadeIn` animation. Looks a little bit smoother. Do you think we don't need it?
> 
> I may be missing something but it looks like from the video, that the accordion content (which drops down) doesn't seem to be animated. I was just wondering if adding some animation here would make it look smoother.

@elliotlaws you were right, I did not add `keyframe` function so animation did not work 🙈 I just fixed it.

https://user-images.githubusercontent.com/23363033/204017882-054d3329-9010-4a11-bca7-055cce7ad0e7.mov

elliotlaws:
As this will most likely be used for navigation, I'm wondering if we need to consider accessibility more with this component, for example:
- Make the root `Sidebar` component polymorphic so we can pass an `as` prop and render a semantic `nav` element. 
- Rendering `ul` and `li` elements where necessary (i.e. SideBar.Content and Sidebar.Item) 

Something similar to [this example](https://codesandbox.io/s/5oq37g?file=/demo.tsx) using Material UI.

lib/src/components/sidedrawer/Sidedrawer.tsx
```diff
@@ -0,0 +1,107 @@
...
+  React.Children.forEach(children, (child) => {
+    if (React.isValidElement(child) && child.type === SidedrawerOverlay) {
+      overlayChild = child as unknown as typeof SidedrawerOverlay
+    } else {
+      contentChildren.push(child)
+    }
+  })
+
+  return (
+    <>
```
elliotlaws, line 85: What do you think of wrapping `Sidedrawer` in context so that children can access `isOpen` and `onClose` etc?

michalrozenek: Honestly I was thinking about it and wanted to discuss with you guys.
These are only one prop (onClose) which `Close` and `Overlay` can use, so it's 50/50 worth it, but I can agree that it's clean option to go.

lib/src/components/sidedrawer/SidedrawerAccordion/SidedrawerAccordionContent.tsx
```diff
@@ -0,0 +1,44 @@
...
+const close = keyframes({
+  from: { height: 'var(--radix-accordion-content-height)' },
+  to: { height: 0 }
+})
+
+const StyledContent = styled(Content, {
+  bg: 'white',
+  overflow: 'hidden',
+  '@allowMotion': {
+    '&[data-state="open"]': { animation: `${open} 300ms ease-out` },
```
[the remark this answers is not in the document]
michalrozenek: I think I copied animation from `Accordion` component to make it the same, but I can double check

lib/src/components/sidedrawer/Sidedrawer.tsx
```diff
@@ -0,0 +1,107 @@
...
+    }
+  }
+
+  React.useEffect(() => {
+    document.addEventListener('keydown', handleCloseOnEscape)
+
+    return () => {
+      document.removeEventListener('keydown', handleCloseOnEscape)
+    }
+  }, [])
```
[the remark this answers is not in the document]
thomasdigby: We shouldn't even need this if we're using the `Dialog` correctly...

lib/src/components/sidedrawer/Sidedrawer.tsx
```diff
@@ -0,0 +1,107 @@
...
+    if (React.isValidElement(child) && child.type === SidedrawerOverlay) {
+      overlayChild = child as unknown as typeof SidedrawerOverlay
+    } else {
+      contentChildren.push(child)
+    }
+  })
+
+  return (
+    <>
+      <Dialog>
```
thomasdigby, line 86: Have we tested this? `@radix-ui/react-dialog` doesn't export a `Dialog` component. I'd have a look through the docs and compare against what we're using in `Sidedrawer` https://www.radix-ui.com/docs/primitives/components/dialog

We should be using `Dialog.Overlay`, `Dialog.Close`, potentially `Dialog.Title`/`Dialog.Description` if relevant, and maybe even `Dialog.Trigger` to ensure correct focus management from and to the initial trigger for the sidedrawer open/close events.

lib/src/components/sidedrawer/Sidedrawer.tsx
```diff
@@ -0,0 +1,107 @@
...
+        <StyledBox
+          role="navigation"
+          data-state={isOpen ? 'open' : 'closed'}
+          open={isOpen}
+        >
+          {contentChildren}
+        </StyledBox>
+      </Dialog>
+      {isOpen && <Box onClick={onClose}>{overlayChild}</Box>}
```
thomasdigby, line 95: We could simplify the `overlayChild` and `onClose` logic here by using `Dialog.Overlay` from Radix

lib/src/components/sidedrawer/Sidedrawer.tsx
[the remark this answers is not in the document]
michalrozenek: If you add a dependency to array it will also run `unmount callback` when prop changes.
I am not sure if this is going to work with your suggestion. I would like to create and remove listener once (mount and unmount).

lib/src/components/sidedrawer/Sidedrawer.tsx
```diff
@@ -0,0 +1,107 @@
...
+  variants: {
+    open: {
+      true: {
+        translate: '0 0'
+      }
+    }
+  }
+})
+
+export const Sidedrawer: React.FC<React.PropsWithChildren<SidedrawerProps>> &
```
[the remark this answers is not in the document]
michalrozenek: That's correct

lib/src/components/sidedrawer/Sidedrawer.tsx
```diff
@@ -0,0 +1,107 @@
...
+}
+
+const StyledBox = styled(Box, {
+  bg: 'white',
+  boxShadow: '$2',
+  position: 'fixed',
+  top: 0,
+  left: 0,
+  height: '100vh',
+  maxWidth: '304px',
```
thomasdigby, line 42: This is a very specific width, is there a reason for that? Should we consider rounding this and providing either a way to override it or a few variants? e.g. `sm`, `md`, `lg` 🤷

lib/src/components/sidedrawer/Sidedrawer.tsx
```diff
@@ -0,0 +1,107 @@
...
+      true: {
+        translate: '0 0'
+      }
+    }
+  }
+})
+
+export const Sidedrawer: React.FC<React.PropsWithChildren<SidedrawerProps>> &
+  SidedrawerSubComponents = ({ children, isOpen, onClose }) => {
+  let overlayChild: typeof SidedrawerOverlay | null = null
```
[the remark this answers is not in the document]
michalrozenek: How come it will be stronger? Imo it won't be.
How union of `React.ElementType<typeof SidedrawerOverlay>` and `null` can be stronger then precise type?

lib/src/components/sidedrawer/Sidedrawer.tsx
thomasdigby: Also, is there any consideration for a responsive layout? I think using a `width` value relative to the viewport on a mobile device would work well, e.g. `width: 90vw`, to prevent it from accidentally filling the viewport.

thomasdigby: Do you not think we should push back against specific magic numbers and aim for a bit more flexibility? Why `304px` exactly?

lib/src/components/sidedrawer/SidedrawerAccordion/SidedrawerAccordtion.tsx
```diff
@@ -0,0 +1,22 @@
+import { Root } from '@radix-ui/react-accordion'
+import React from 'react'
+import { styled } from '~/stitches'
+
+const StyledRoot = styled(Root, {
+  width: '100%'
+})
+
+interface SidedrawerAccordionProps {
+  type?: 'single' | 'multiple'
```
thomasdigby, line 10: Can we get the types from `StyledRoot` directly? That should contain the `type` attribute as well as `css` and the other props that `styled()` accepts

lib/src/components/sidedrawer/Sidedrawer.tsx
MiguelGDLR: I got 304 as a multiple of 16 and considering enough room to also show the overlay in small devices. I would be very happy with both a way to override it and also different sizes 
 - sm = 304 (default)
 - md = 336
 - lg = 352
 
thanks Tom

lib/src/components/sidedrawer/SidedrawerContent.tsx
```diff
@@ -0,0 +1,30 @@
+import React, { isValidElement } from 'react'
+
+import { Box } from '../box'
+import { SidedrawerAccordion } from './SidedrawerAccordion'
+
+export const SidedrawerContent: React.FC = ({ children }) => {
+  let hasAccordionItems = false
+
+  React.Children.forEach(children, (child) => {
```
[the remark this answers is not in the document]
thomasdigby: I agree with this, clearer component composition and reduced magic behind the scenes is a good thing. Plus, it improves the relationship between the parent `Sidedrawer.Item` (now `Sidedrawer.Accordion`) and its direct children being `Sidedrawer.AccordionTrigger` and `Sidedrawer.AccordionContent`.

lib/src/components/sidedrawer/SidedrawerFooter.tsx
```diff
@@ -0,0 +1,21 @@
...
+  <Flex
+    css={{
+      alignItems: 'center',
+      borderTop: '1px solid $tonal100',
+      bottom: 0,
+      boxShadow: '$3',
+      height: '$6',
+      justifyContent: 'center',
+      p: '$3',
+      position: 'absolute',
```
thomasdigby, line 15: Do we need to account for the height of the footer in the main content area? In this current state, I assume the content area falls behind the footer so if there's enough content to introduce a scrollbar you may not see the last item.

I wonder if the header/content/footer layout could be flexbox with the content area filling the available space instead of removing it from the flow.

lib/src/components/sidedrawer/SidedrawerFooter.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+
+import { Flex } from '../flex/Flex'
+
+export const SidedrawerFooter: React.FC = ({ children }) => (
```
thomasdigby, line 5: When we have a single component with css returned like this, we could instead export a `styled()` component directly. It will handle the types, pass all props down, handles children correctly, and results in simpler/reduced source code. e.g.

```ts
export const SidedrawerFooter = styled(Flex, {
  alignItems: 'center',
  borderTop: '1px solid $tonal100',
  bottom: 0,
  boxShadow: '$3',
  height: '$6',
  justifyContent: 'center',
  p: '$3',
  position: 'absolute',
  width: '100%'
})
```

lib/src/components/sidedrawer/SidedrawerOverlay.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { Box } from '../box'
+
+const fadeIn = keyframes({
+  from: {
+    opacity: 0
+  },
+  to: {
+    opacity: 0.6
+  }
+})
```
thomasdigby, line 15: Are there any animations within `Sidedrawer` that could be shared with our other standard animations here? https://github.com/Atom-Learning/components/blob/main/lib/src/utilities/style/keyframe-animations.ts

For `Dialog` and other overlays, we use `$alpha600` as the background colour, which will result in the same visual output, and we can use a shared keyframe animation.

lib/src/components/sidedrawer/Sidedrawer.tsx
thomasdigby: Could I recommend maybe rounding it out as `320px`? Still divisible by 16/32/64. We don't necessarily need to introduce sizes for now, but worth considering for future flexibility and usage, and I assume that we'll be able to use `css` to override this property if need be..?

@MiguelGDLR What are your thoughts on setting the max-width to `90vw`/`95vw` to ensure that it never fills the viewport width on mobile, and always gives you a little bit of background to the right? I think it would help on smaller devices to reinforce it as a "side-drawer" and not full screen overlay

lib/src/components/sidedrawer/SidedrawerFooter.tsx
michalrozenek: Makes sense !

lib/src/components/sidedrawer/Sidedrawer.tsx
MiguelGDLR: I'm not sure if having it full width is bad honestly. Although I see what you mean. 
I like fixed widths because it's easier to control and to replicate in Figma. Also I need to provide these different widths for when the Fixed Sidedrawer (desktop - no overlay). So I will review the sizes and get back to you in a bit.

MiguelGDLR: please review my proposal [here](https://www.figma.com/file/KjgQh7eTwEfJmH6e56k9mX/DS-Documentation?node-id=9235%3A81481&t=QCp9wxtSQ3OdqV2U-1) and leave comments on Figma

lib/src/components/sidedrawer/SidedrawerAccordion/SidedrawerAccordionItem.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { Item } from '@radix-ui/react-accordion'
+import { SidedrawerAccordionTrigger } from './SidedrawerAccordionTrigger'
+import { SidedrawerAccordionContent } from './SidedrawerAccordionContent'
+
+type SidedrawerAccordionItemProps = React.ComponentProps<typeof Item>
+
+export const SidedrawerAccordionItem: React.FC<
+  SidedrawerAccordionItemProps
+> = ({ children, value, ...remainingProps }) => (
+  <Item value={value} {...remainingProps}>
```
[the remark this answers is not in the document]
michalrozenek: I added it to inform the user of the library that they can pass only Trigger or Content, but yeah, I am fine with removing it if we believe it's unnecessary.

lib/src/components/sidedrawer/Sidedrawer.tsx
michalrozenek: Thank you @thomasdigby.

lib/src/components/sidedrawer/SidedrawerContent.tsx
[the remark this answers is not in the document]
michalrozenek: I think it's a good idea to separate the Accordion from an Item.

I did it this way because I think we need a single `Accordion.Root` rendered inside a `Sidedrawer.Content` only when we render `Accordtion.Item` inside of it. This is done in `Sidedrawer.Content`.
This is because single Root wraps all Accordion.Items and they work correctly together (expand one - you close another etc).

To follow this idea I thought we don't need an Item separation, and we can keep single component (item) to render any of it's type (link, button, accordion.item).

But...
I think we still can keep rendering `Accordion.Root` in `Sidedrawer.Content` when we detect an `Accorditon.Item` child to be sure that it's working properly + we can separate Sidedrawer.Accordtion to represents `Accordion.Item` and `Sidedrawer.Item` for `link` and `button`.

lib/src/components/sidedrawer/Sidedrawer.tsx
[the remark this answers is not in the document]
elliotlaws: Apologies, I've misread this, `React.ElementType<typeof SidedrawerOverlay>` is very much identical to `typeof SidedrawerOverlay`

It's down to personal preference on which one you choose.

lib/src/components/sidedrawer/SidedrawerAccordion/SidedrawerAccordionItem.tsx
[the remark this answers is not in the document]
elliotlaws: Yeah, I think considering that we're the consumers of the library and in most cases, Trigger and Content will be rendered I think this can be removed.

lib/src/components/sidedrawer/SidedrawerAccordion/SidedrawerAccordionContent.tsx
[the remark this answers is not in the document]
michalrozenek: Yeah, it's the same. Do you think we should update it in both places?
I think we can leave it for later and probably discuss with @MiguelGDLR as well.
At the moment It's too many more important things to cover.

lib/src/components/sidedrawer/SidedrawerContent.tsx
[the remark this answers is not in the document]
michalrozenek: To give more context about the root - if we don't want to do this behind the scenes which imo is nothing bad, the library user would need to do something like this
```
<Sidedrawer isOpen onClose={() => console.log('closed')}>
  <Sidedrawer.Header>
    <Sidedrawer.Close onClose={() => console.log('closed')} />
  </Sidedrawer.Header>
  // HERE IS IMPORTANT PART
  <Sidedrawer.Content>
    <Sidedrawer.Accordion>
        // it wraps all items in content because accordtion items are mixed with links and buttons and it is not always in order which allows us to wrap all accordtion items together.

        // normal link
       <Sidedrawer.Item href="/">Example</Sidedrawer.Item>

        // accordion item
       <Sidedrawer.AccordionItem value="2">
          <Sidedrawer.AccordionTrigger>Resources</Sidedrawer.AccordionTrigger>
          <Sidedrawer.AccordionContent>
            <Sidedrawer.Item onClick={() => console.log('clicked!')}>
              I am a button
            </Sidedrawer.Item>
            <Sidedrawer.Item active href="/">
              I am a link
            </Sidedrawer.Item>
          </Sidedrawer.AccordionContent>
        </Sidedrawer.AccordionItem>
      </Sidedrawer.Accordtion>
   </Sidedrawer.Content>
  <Sidedrawer.Footer>Footer items here</Sidedrawer.Footer>
  <Sidedrawer.Overlay />
</Sidedrawer>
```

To sum up, I would:
- keep the mapping through children in `Sidedrawer.Content` to check if we render at least a single AccordionItem, and then wrap children in `Accordtion.Root` or render `children` without any wrapper depends on the check.
- keep `Sidedrawer.Accordion` name for `radix Accordion.Item` to make naming simpler for our lib user. But in this case I am open to rename to `Sidedrawer.AccordionItem` if we all agreed on it.

lib/src/components/sidedrawer/SidedrawerAccordion/SidedrawerAccordionRoot.tsx
```diff
@@ -0,0 +1,14 @@
+import { Root } from '@radix-ui/react-accordion'
+import React, { ComponentProps } from 'react'
+import { styled } from '~/stitches'
+
+const StyledRoot = styled(Root, {
+  width: '100%'
+})
+
+export const SidedrawerAccordionRoot: React.FC<ComponentProps<typeof Root>> = ({
```
thomasdigby, line 9: Same as my other comment, this whole file could be:

```ts
import { Root } from '@radix-ui/react-accordion'

export const SidedrawerAccordionRoot = styled(Root, {
  width: '100%'
})
```

👌

lib/src/components/sidedrawer/SidedrawerContent.tsx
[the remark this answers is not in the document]
elliotlaws: I see your point about mixing links and accordion items. 

I suppose another reason for removing any instance to `Accordion` in `Sidebar.Content` is that it decouples `Content` from `Accordion`. 

IMO I think it might be better to be explicit and render `Accordion` separately, as it makes `Sidebar.Content` more agnostic of its children. 

But, interested to hear what others think about this. Any thoughts @thomasdigby?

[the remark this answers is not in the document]
michalrozenek: Thanks @elliotlaws for your opinion.
In this case I'd like to keep rendering of `Accordion Root` as a part of content responsibilities.
It knows if we need to wrap items in Accordion or not.
Thanks to that building navigation blocks is easier and users do not need to consider if they should wrap whole `SidedrawerContent children` in Accordion Root, or maybe just a single item.
Imo it's easier to use.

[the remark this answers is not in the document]
michalrozenek: My proposal is to have a discussion about it of course, but not stop merging because of this.

lib/src/components/sidedrawer/SidedrawerClose.tsx
```diff
@@ -1,21 +1,22 @@
...
-    <Icon
-      is={Close}
-      css={{
-        color: '$tonal300'
-      }}
-    />
-  </ActionIcon>
+export const SidedrawerClose: React.FC = () => (
+  <DialogClose asChild>
+    <StyledIcon aria-label="close" is={Close} />
```
thomasdigby, line 20: This still needs to be an `ActionIcon` as we're rendering `DialogClose` with `asChild`

lib/src/components/sidedrawer/Sidedrawer.tsx
michalrozenek: Can we take this discussion as a follow-up ticket?
I'd like to proceed with current implementation and be able to introduce this sidedrawer in parent app :)
We can iterate on it when we have clear agreement.

lib/src/components/sidedrawer/Sidedrawer.tsx
```diff
@@ -60,42 +68,29 @@ export const Sidedrawer: React.FC<SidedrawerProps> &
...
-        <StyledBox
+    <Dialog.Root open={isOpen} onOpenChange={onClose}>
+      <Dialog.Portal>
+        {overlayChild}
+        <StyledContent
           role="navigation"
-          data-state={isOpen ? 'open' : 'closed'}
-          open={isOpen}
+          onEscapeKeyDown={onClose}
+          onInteractOutside={onClose}
```
thomasdigby, line 88: Do we need these props if `<Dialog.Root onOpenChange={} />` is used?

lib/src/components/sidedrawer/Sidedrawer.tsx
```diff
@@ -60,42 +68,29 @@ export const Sidedrawer: React.FC<SidedrawerProps> &
...
     } else {
       contentChildren.push(child)
     }
   })
 
   return (
-    <>
-      <Dialog>
-        <StyledBox
+    <Dialog.Root open={isOpen} onOpenChange={onClose}>
```
thomasdigby, line 82: I assume we only want `onClose` to fire if it's closing, we will need to change `onOpenChange` to be `onOpenChange={(open) => !open && onClose()}`

lib/src/components/sidedrawer/Sidedrawer.tsx
```diff
@@ -60,42 +68,29 @@ export const Sidedrawer: React.FC<SidedrawerProps> &
...
-    return () => {
-      document.removeEventListener('keydown', handleCloseOnEscape)
-    }
-  }, [])
-
   React.Children.forEach(children, (child) => {
     if (React.isValidElement(child) && child.type === SidedrawerOverlay) {
-      overlayChild = child as unknown as typeof SidedrawerOverlay
+      overlayChild = React.cloneElement(child, {
+        'data-testid': 'sidedrawer_overlay'
```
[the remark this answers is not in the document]
michalrozenek: It's not the only reason.
The reason is to separate an overlay from the rest of children.
We don't want to render overlay inside of `Content`.

Zrzut ekranu 2022-11-29 o 14 25 38

Zrzut ekranu 2022-11-29 o 14 26 25

If I change it, we will be rendering overlay inside of `Dialog.Content` and we don't want it.
Zrzut ekranu 2022-11-29 o 14 26 09

Other option is to maybe get a prop which tells us if we want to render Overlay or not, so we can handle that internally, but it changes api aka components composition a little bit. As far as I know we do not want to follow this approach.

lib/src/components/sidedrawer/Sidedrawer.tsx
thomasdigby: Also, unrelated to my previous comment, but I wonder if we might want to also export a `SidedrawerTrigger` to allow focus to return correctly to the initial button used to open it.

```tsx
<Sidedrawer>
  <Sidedrawer.Trigger>
    Open
  </Sidedrawer.Trigger>
  // or
  // <Sidedrawer.Trigger asChild>
  //  <ActionIcon><Icon is={Menu} /></ActionIcon>
  // </Sidedrawer.Trigger>
  <Sidedrawer.Content>
    // rest of the stuff
  </Sidedrawer.Content>
</Sidedrawer>
```

lib/src/components/sidedrawer/SidedrawerClose.tsx
michalrozenek: Good point, ActionIcon introduces a button layer we want to use.

lib/src/components/sidedrawer/Sidedrawer.tsx
thomasdigby: Yep that's all good from me, just wanted to make sure we had good reasoning for the width value and that we were considering responsiveness. Happy to treat this as a follow up task 👍

lib/src/components/sidedrawer/Sidedrawer.tsx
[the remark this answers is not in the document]
thomasdigby: Cool ok, so I guess I'd maybe recommend a slightly different approach where we render the `Overlay` by default (without the user requiring to use a component). Then in future when we're considering the different types of `Sidedrawer`, we can use this root file as one of a few variants, as we won't want to render it within a `Dialog` if it's static.

I'd probably recommend for now just rendering the overlay by default, and then we can tackle in a follow up task the other sidedrawer variants. Do you agree?

lib/src/components/sidedrawer/Sidedrawer.tsx
michalrozenek: Doesn't it introduce some limitation about positioning the `trigger` button?
Imagine that we want to trigger the Sidedrawer from some place in the app, let's assume it's a topbar.
Does it mean that whole Sidedrawer should be declared inside the topbar to makes this working?
This is the requirement to place `trigger` inside of the `dialog` it triggers, right?

thomasdigby: Not necessarily, the `Dialog.Portal` allows Radix to dump the actual dialog contents into a portal on the page, separate to where the trigger is rendered. Worth a quick test as we can render the topbar in the component lib sandbox, but it should work fine.

lib/src/components/sidedrawer/Sidedrawer.tsx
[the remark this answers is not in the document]
michalrozenek: The purpose of doing that was a criteria that it cannot be hard coupled in sense we can use sidedrawer with or without an overlay. That's why I did it this way.
I know it introduces some additional logic and a little bit of light operations, but it let us achieve the requirement.

I think yes, we can render it by default at the moment, but we would need to address this requirement later if we find the first case to use it differently.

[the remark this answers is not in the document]
michalrozenek: Do you want to change it because of additional logic/complexity or the operations (the loop)?
This is a pure tradeof between the `additional complexity` vs `feature requirement`

lib/src/components/sidedrawer/Sidedrawer.tsx
michalrozenek: Yeah, I know that there's a Portal, but have another concerns. I'll try it first

lib/src/components/sidedrawer/Sidedrawer.tsx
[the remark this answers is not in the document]
thomasdigby: My main reasoning is that I think we can reduce the complexity and that ideally, we should try to avoid using `data-testid`s in our components as they'll propagate through to atom-core. But I understand the necessity for mapping through children for that purpose.

Also worth mentioning (based on my assumption that the requirement to not render the overlay is the same as rendering a static nav that doesn't show/hide) is that I don't think we're meeting the feature requirement of the static nav anyway, as in that particular context we wouldn't want to render the `Sidedrawer` in a `Dialog`.

I don't want to add much more work here, I'm wary that we're discussing fairly fundamental component API decisions after a chunk of work's already been completed, but I would probably treat the static nav as an additional piece of work to complete once this initial "dynamic nav" has been completed. There are a few bits of composition and naming that we can tweak to allow a "static nav" as well.

Happy to chat about this further in a call if that makes more sense?

lib/src/components/sidedrawer/SidedrawerContent.tsx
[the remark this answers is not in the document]
michalrozenek: I can prepare follow up ticket to everything we want to discuss and introduce as follow up.
I would appreciate if we are able to merge this big one and then work separately on `nice to have` changes :)

lib/src/components/sidedrawer/SidedrawerClose.tsx
```diff
@@ -5,18 +5,21 @@ import React from 'react'
...
+import { ActionIcon } from '../action-icon/ActionIcon'
 
-const StyledIcon = styled(Icon, {
-  color: '$tonal300',
-  cursor: 'pointer',
-  margin: '$3',
-  '&:hover': {
-    color: '$tonal500'
+const StyledActionIcon = styled(ActionIcon, {
+  '> svg': {
```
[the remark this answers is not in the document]
michalrozenek: Time to break for a dinner 🤯

lib/src/components/sidedrawer/SidedrawerContent.tsx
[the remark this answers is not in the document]
michalrozenek: https://atomlearningltd.atlassian.net/browse/DS-157

lib/src/components/sidedrawer/Sidedrawer.tsx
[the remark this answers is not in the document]
michalrozenek: Ok, got your point. Will render `Overlay` from `Sidedrawer` as default at the moment.

lib/src/components/sidedrawer/Sidedrawer.tsx
michalrozenek: Hey, there's same situation as we had with Overlay :)
I am rendering it in the `Dialog.Content`.
To achieve this trigger I would need to change the `Sidedrawer` composition.

And btw this is what I mean --> I need to move up `Sidedrawer` wrapper to wrap the Trigger as well.
Zrzut ekranu 2022-11-29 o 17 03 44

Because otherwise I have this error
Zrzut ekranu 2022-11-29 o 17 04 23
Zrzut ekranu 2022-11-29 o 17 04 28

This is what I mean in above comment (https://github.com/Atom-Learning/components/pull/429#discussion_r1034769264)

lib/src/components/sidedrawer/SidedrawerAccordion/SidedrawerAccordionContent.tsx
```diff
@@ -0,0 +1,44 @@
...
+  }
+})
+
+export const SidedrawerAccordionContent: React.FC = ({ children }) => (
+  <StyledContent>
+    {React.Children.map(children, (child) => {
+      return (
+        <Box
+          css={{
+            '> a': {
```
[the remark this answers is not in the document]
michalrozenek: I know how it looks like, and let me explain why we have it :)
First of all we're using `Sidedrawer.Item` on each level (main level of `Sidedrawer.Content` and in nested levels inside of `Sidedrawer.AccordionContent`. We have only one `Item` component which is common to both levels because it's basically the same thing (differs by padding only).

To make a proper paddings for nested item, we need to know it's inside of `Sidedrawer.AccordionContent`.
I have been thinking that it was a way to do this - Iterate on `children` inside of `Sidedrawer.AccordionContent` and add proper paddings to `Sidedrawer.Item` (a, button).

Zrzut ekranu 2022-11-29 o 17 14 35

Looking bad - padding issue
Zrzut ekranu 2022-11-29 o 17 16 29

Looking bad - no indentation
Zrzut ekranu 2022-11-29 o 17 22 37

Looking good (current solution)
Zrzut ekranu 2022-11-29 o 17 17 02


Maybe we can consider better idea later on?

lib/src/components/sidedrawer/Sidedrawer.tsx
thomasdigby: It would just need to wrap `Sidedrawer.Trigger` but the entire thing would be contained within `TopBar` (obviously in practice you can make a new component to render instead of rendering the entire chain of nested components as well)

```jsx
<TopBar>
 {/* this renders `Dialog` */}
  <Sidedrawer>
    {/* this renders `Dialog.Trigger` */}
    <Sidedrawer.Trigger>Open</Sidedrawer.Trigger>
   {/* this renders `Dialog.Portal` and `Dialog.Content` */} 
   <Sidedrawer.Content>
      {/* Header/Body/Footer */}
   </Sidedrawer.Content>
  </Sidedrawer>
</TopBar>
```

I know this is probably frustrating that we might have to reconstruct some of our component names/APIs, but I think it's important that we manage the focus state correctly. Without this, when the `Sidedrawer` closes, the browser will focus the document and the user will lose their place in the tab order of the page.

_However_, saying that, if we don't want to do this now, then in the short-term we could set `document.activeElement` in the `onClose` callback to ensure focus is passed back to the button that was used to open the sidedrawer, but it will be custom JS in atom-core and I think we'll eventually want to settle on the `Sidedrawer.Trigger` approach in future anyway.

If it's a timing thing and you just want to get it out of the door, I'm happy to put together a POC (or even just add a commit to this PR) that can get it working with the trigger? Or we can go with the custom JS in atom-core for now, up to you how you want to handle it. 🙂

lib/src/components/sidedrawer/SidedrawerAccordion/SidedrawerAccordionContent.tsx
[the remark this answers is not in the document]
thomasdigby: Yeah all good, I'm sure there will be a better way to apply these styles, but certainly not a blocker for now 👍 

Nested accordion items look good!

lib/src/components/sidedrawer/Sidedrawer.tsx
michalrozenek: Yes, that's right. This is what I mean - entire thing needs to be declared in the code in place when we want to have a trigger :P That's odd, but I understand the reasons and how Dialog api works. It's just looking weird from the implementation point of view.

I would say - merge what we have not and start `Trigger` work to close something big and do not add more things into. We can discuss the components/apis reconstruction to introduce Trigger (and possibly it opens the gate for `Overlay` again).

This is true about the time... Since few days I am alone FE in parent pod and have long list of todos (tickets of mine, Kyle's work, postponed guild day and still few peer reviews) and I am trying to do really great sidedrawer which is time consuming to discuss all of it, find a compromise etc. 
I think I should be able to do it by myself, just need to have an agreement about solution before I start to not waste more time + nice to have this PR merged before

lib/src/components/sidedrawer/SidedrawerItem.tsx
```diff
@@ -0,0 +1,50 @@
...
+const StyledLink = styled('a', sidedrawerItemStyles)
+const StyledButton = styled('button', sidedrawerItemStyles)
+
+type SidedrawerItemProps = {
+  active?: boolean
+  disabled?: boolean
+} & (
+  | {
+      href: string
+      onClick?: never
```
thomasdigby, line 16: I think we had this issue with the main `Link` component, in that we may need to bind an `onClick` to an anchor still, e.g. for event tracking. What do you think about simplifying this to type both `href` and `onClick` without requring either or?

lib/src/components/sidedrawer/SidedrawerHeader.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export const SidedrawerHeader: React.FC = ({ children }) => (
+  <Box
+    css={{
+      '> div': {
+        width: '100%'
+      }
+    }}
+  >
+    <TopBar
```
thomasdigby, line 14: A few questions on this, is this a valid and expected use of `TopBar`? I assumed `TopBar`s purpose was purely for an app-level navigation container. Also, is it necessary to include this additional complexity for the `SidedrawerHeader`? Are we expecting to next sub-components of `TopBar` within `SidedrawerHeader`?

lib/src/components/sidedrawer/SidedrawerClose.tsx
```diff
@@ -0,0 +1,22 @@
...
+export const SidedrawerClose: React.FC = () => (
+  <DialogClose asChild>
+    <ActionIcon theme="neutral" appearance="simple" label="close" size="md">
+      <Icon
+        css={{
+          color: '$tonal300',
+          '&:hover': {
+            color: '$tonal500'
+          }
+        }}
```
thomasdigby, line 17: Is this `css` necessary with the `ActionIcon` theme?

lib/src/components/sidedrawer/SidedrawerBody.tsx
```diff
@@ -0,0 +1,32 @@
...
+  React.Children.forEach(children, (child) => {
+    if (isValidElement(child) && child.props?.value) {
+      hasAccordionItems = true
+    }
+  })
+
+  return (
+    <Box
+      css={{
+        height: 'calc(100% - $6)',
```
thomasdigby, line 18: Could we replace this with `flex-grow: 1`? That should fill the space in-between the header/footer if the parent is a flex container (which I think it is)

lib/src/components/sidedrawer/SidedrawerAccordion/SidedrawerAccordionItem.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import { Item } from '@radix-ui/react-accordion'
+
+type SidedrawerAccordionItemProps = React.ComponentProps<typeof Item>
+
+export const SidedrawerAccordionItem: React.FC<
+  SidedrawerAccordionItemProps
+> = ({ children, value, ...remainingProps }) => (
+  <Item value={value} {...remainingProps}>
```
thomasdigby, line 9:
```suggestion
  <Item {...remainingProps}>
```

`value` can be passed through `remainingProps`

lib/src/components/sidedrawer/SidedrawerContent.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+const slideIn = keyframes({
+  '0%': { transform: 'translateX(-100%)' },
+  '100%': { transform: 'translateX(0)' }
+})
+
+const slideOut = keyframes({
+  '0%': { transform: 'translateX(0)' },
+  '100%': { transform: 'translateX(-100%)' }
+})
```
thomasdigby, line 17: Could we move these over to our shared animation utilities? `slideOutLeft`, `slideInLeft`, etc.

lib/src/components/sidedrawer/SidedrawerItem.tsx
michalrozenek: oh,
it's fair point to make both optional.

lib/src/components/sidedrawer/SidedrawerClose.tsx
michalrozenek: Yes, without it the icon is `dark blue`
Zrzut ekranu 2022-11-30 o 14 34 02

lib/src/components/sidedrawer/SidedrawerContent.tsx
michalrozenek: I moved it there initially but I was not sure about the values (-100%).
I will move it there one more time

lib/src/components/sidedrawer/SidedrawerClose.tsx
thomasdigby: Is that only in a focus state though? I think we should try to keep the close button styles consistent across our UI, at least within the component library. This is the same close button in the `Dialog` component https://github.com/Atom-Learning/components/blob/main/lib/src/components/dialog/DialogContent.tsx#L109-L117

michalrozenek: yes, in focus state, but also as a hover when it's not `focused`.
Ok, I will remove these styles at the moment, but I think that maybe `hover` state could be different

MiguelGDLR: We will improve the styling of ActionIcons in the future, I think it's better to not add any extra styling right now.

lib/src/components/sidedrawer/Sidedrawer.mdx
```diff
@@ -0,0 +1,159 @@
...
+- `Sidedrawer.Footer` - Optional part of the `Sidedrawer`. It's fixed to the botton. You can pass whatever you need as it's children.
+
+```tsx
+<Sidedrawer>
+  <Sidedrawer.Trigger asChild>
+    <Button>Open the navigation</Button>
+  </Sidedrawer.Trigger>
+  <Sidedrawer.Content>
+    <Sidedrawer.Header>
+      <Sidedrawer.Close onClose={() => console.log('closed')} />
```
thomasdigby, line 52: `onClose` no longer exists on this component. You can use `onOpenChange` instead if necessary for the docs?

lib/src/components/sidedrawer/Sidedrawer.mdx
```diff
@@ -0,0 +1,159 @@
...
+      <Icon is={Inbox} css={{ mr: '$2' }} />
+      Test
+    </Sidedrawer.Item>
+  </Sidedrawer.AccordionContent>
+</Sidedrawer.Accordion>
+```
+
+## Whole Sidedrawer components composition
+
+```tsx
```
thomasdigby, line 121: Do we want to add `preview` here to render it in the docs? Would be nice to have an interactive example

lib/src/components/sidedrawer/Sidedrawer.mdx
michalrozenek: I just updated docs :)

lib/src/components/sidedrawer/Sidedrawer.mdx
michalrozenek: added :)

lib/src/components/sidedrawer/SidedrawerHeader.tsx
MiguelGDLR: I will create a ticket

lib/src/components/sidedrawer/SidedrawerContent.tsx
```diff
@@ -0,0 +1,37 @@
...
+    '&[data-state="closed"]': {
+      animation: `${slideOutLeft} 250ms ease-out`
+    }
+  }
+})
+
+export const SidedrawerContent: React.FC = ({ children }) => (
+  <Portal>
+    <SidedrawerOverlay data-testid="sidedrawer_overlay" />
+    <StyledContent role="navigation">{children}</StyledContent>
```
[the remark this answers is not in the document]
michalrozenek: I think it should be the navigation. I believe we want to use it for navigating elements. Do you think that it might have different roles?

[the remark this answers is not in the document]
elliotlaws: Probably not. I was just curious 😄

elliotlaws (approved):

PR #430, opened 2022-11-18
fix: workaround to only render first page of rows

hollg:
Problem: paginated `DataTable`s render all of their rows on the initial render, not just the first page. So if any heavy-lifting happens in a row (e.g. API requests are made) then a big table can cause a browser crash.

Cause: `DataTable.Pagination` tells `DataTable` to paginate its data in a `useEffect`. `useEffect` fires after the render.

Short-term solution (this PR): Add an optional `pagination` prop to `DataTable` which will set pagination state for the first render

Long-term solution: overhaul the API

### DIFF
--- a/lib/src/components/data-table/DataTableContext.tsx
+++ b/lib/src/components/data-table/DataTableContext.tsx
@@ -23,7 +23,7 @@ type TableProviderProps = {
   data: Array<Record<string, unknown>>
   defaultSort?: { column: string; direction: 'asc' | 'desc' }
   children: React.ReactNode
-  pagination?: { pageIndex: 0; pageSize: number }
+  pagination?: { pageIndex: number; pageSize: number }
 }
 
 export const DataTableProvider = ({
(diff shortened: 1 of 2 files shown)


lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -23,15 +23,18 @@ type TableProviderProps = {
   data: Array<Record<string, unknown>>
   defaultSort?: { column: string; direction: 'asc' | 'desc' }
   children: React.ReactNode
+  pagination?: { pageIndex: 0; pageSize: number }
```
thomasdigby, line 26:
```suggestion
  pagination?: { pageIndex: number; pageSize: number }
```

?

lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -23,15 +23,18 @@ type TableProviderProps = {
...
 }
 
 export const DataTableProvider = ({
   columns,
   data,
   defaultSort,
-  children
+  children,
+  pagination = undefined
```
thomasdigby, line 34: I wonder if we just expose `initialState` and type it from what `react-table` expects. We can get around the temporary prop and allow a touch more flexibility for users?

`<DataTable initialState={{ pagination: { pageSize: 3, pageIndex: 0 } }} />`

hollg: That seems like a much better solution! Have updated. Fewer changes between this PR and a long-term API change now I reckon. Thanks!

thomasdigby (approved):
🎉

PR #433, opened 2022-11-22
feat: remove CSSWrapper from Stack so the elements its used as can be…

LimeWub:
… targetted with css

## Description

For the [CMS](https://github.com/Atom-Learning/components/pull/398) - I needed to use a `ul` for `Stack`. Because of how `Stack` is wrapped with a `CSSWrapper` the stack main element itself was becoming a `ul` as expected - but I could not target it with css to remove `ul` paddings.

I'm doubting the need for the `CSSWrapper` to exist in this component altogether so I've removed it. Now CSS gets passed directly onto `Stack`.

### DIFF
--- a/lib/src/components/stack/Stack.tsx
+++ b/lib/src/components/stack/Stack.tsx
@@ -2,7 +2,6 @@ import * as React from 'react'
 
 import { CSS, styled } from '~/stitches'
 import { createThemeVariants } from '~/utilities'
-import { CSSWrapper } from '~/utilities/css-wrapper'
 
 const StyledStack = styled('div', {
   display: 'flex',
@@ -67,7 +66,6 @@ export const Stack: React.ForwardRefExoticComponent<StackPropsType> =
   React.forwardRef(
     (
       {
-        css,
         gap = 2,
         direction = 'row',
         wrap = 'wrap',
@@ -78,21 +76,19 @@ export const Stack: React.ForwardRefExoticComponent<StackPropsType> =
       ref
     ) => {
       return (
-        <CSSWrapper css={css}>
-          <StyledStack
-            ref={ref}
-            direction={direction}
-            gap={gap}
-            wrap={wrap}
-            justify={justify}
-            align={
-              typeof align === 'undefined' && direction !== 'column'
-                ? 'center'
-                : align
-            }
-            {...remainingProps}
-          />
-        </CSSWrapper>
+        <StyledStack
+          ref={ref}
+          direction={direction}
+          gap={gap}
+          wrap={wrap}
+          justify={justify}
+          align={
+            typeof align === 'undefined' && direction !== 'column'
+              ? 'center'
+              : align
+          }
+          {...remainingProps}
+        />
       )
     }
   )


lib/src/components/stack/Stack.tsx
```diff
@@ -78,21 +76,19 @@ export const Stack: React.ForwardRefExoticComponent<StackPropsType> =
       ref
     ) => {
       return (
-        <CSSWrapper css={css}>
```
thomasdigby, line 81: So I _think_ `CSSWrapper` is needed for `Stack` as you can't apply `margin` to the container without affecting the contents. Negative `mt` and `ml` are set on the container to counter the children's margins. The extra wrapping `div` is needed to ensure both margins applied via `css` and internally within `StyledStack` are respected

LimeWub: So - I believe you are right - but Stack is bugged anyway when it comes to spacing. I thought it was bugged as in the CSS Wrapper wasn't doing anything, as I've used Stack in its current form (with wrapper) for a recent refactor I did and applying margins to Stack was still overriding its margin top/left negative CSS.

I'm now realising the bugged setup was only caused in nested `Stack` setups which were overriding children margins rather than parent margin.

Which is why using `gap` would had been so much cleaner as it would not cause any problems with `Stack` margins or nesting within another `Stack`.

Regardless - I prefer I remove it here then add a `Box` around all `Stacks` in core with the appropriate spacing.
That makes the behaviour clear, lets CSS target Stack elements directly and removed loads of nesting when it's not required to target the Stack with margins. Do you think this would work?

thomasdigby: > Stack is bugged anyway when it comes to spacing

Nested spacing it certainly is, which sucks, because there are a bunch of use-cases that would be much easier if we could stack within stack.

> I've used Stack in its current form (with wrapper) for a recent refactor I did and applying margins to Stack was still overriding its margin top/left negative CSS.

Are you sure? That shouldn't be the case... I just did a test on the DS site and the outer margins don't overwrite the negative top/left that's required for `Stack`.

Screenshot 2022-11-28 at 10 02 15
Screenshot 2022-11-28 at 10 01 01

> Regardless - I prefer I remove it here then add a Box around all Stacks in core with the appropriate spacing.

The problem is that we would have to fix up a bunch (50+) of `Stack` implementations in atom-core where `css` is applied and margins are added. What's the specific problem that you're trying to solve with this change? I bet we can do something less impactful that would still solve your use-case.

Separate but related to this, I'd be keen to chat about JS-only CSS feature detection. To continue to support Safari 12 we can't use flexbox gap, and frustratingly we can't even check for support within CSS and use it only for the newer browsers, _however_, we could potentially use some JS to detect `gap` support and then update the layout to use that in most cases, falling back to our current implementation if the browser/runtime lacks support... but then we're relying on JS for layout implementation 😬

LimeWub: >  I bet we can do something less impactful that would still solve your use-case.

Ye I will revert this and use `styled` on the ul that uses it to remove paddings b4 I pass it in so I don't need this change anymore.

>  I'd be keen to chat about JS-only CSS feature detection

If some of the browsers we support cannot support it - js gap detection won't really help us here I think. 
Because we'd still want the other browsers to not look broken and we'd be shipping extra divs/developing for 2 things/potentially shipping something broken (since it'll work with gap on our screen but we don't browser test for Saf 12) anyway. It's either all of them support gap and we use it out of the box or find and use some kinda polyfill maybe.
I think it'd be great if we could drop this extra div in terms of bloat and dom nesting too - which we'd be able to do with `gap`. + Gap would work fine with nested Stacks.
I don't mind having a discussion on this 🤷 

I know I didn't chase this last time, but dropping some browsers - based on our actual recorded usage - and then just doing it properly would be the best imo.

PR #434, opened 2022-11-22
feat: space out all children of StackContent

LimeWub:
In the [CMS project](https://github.com/Atom-Learning/components/pull/398)
I use StackContent to space out the contents of the documentation site and noticed it was not adding spacings to code elements. 
Thus, I made it more generic to add spacing to all direct children - this should not affect anything other than snapshots.


lib/src/components/stack-content/StackContent.tsx
```diff
@@ -10,18 +10,18 @@ import { StyledList } from '../list/List'
 import { StyledText } from '../text/Text'
 
 const StyledStackContent = styled('div', {
-  [`& > ${StyledHeading}`]: {
-    maxWidth: '65ch',
+  [`& > *`]: {
```
elliotlaws, line 13: does this still need to be wrapped in square brackets?

LimeWub: Ya know what - it probably doesn't. I just did a copy paste and change from the ones below that have them. I can remove the brackets and use simple quotes 👍

LimeWub: Done 👍

elliotlaws (approved):

PR #435, opened 2022-11-22
feat: remove hardcoded color from Text and Heading

LimeWub:
Text and Heading changed to be inheriting colour from parent rather than having a color hardcoded in them.
This means we will be able to override colour easier, without having to force colours via inline css which we usually have to do. 

ie.
**before** 
```
<Box><Text css={{color: 'white'}} /><Text css={{color: 'white'}} /><<Text css={{color: 'white'}} />/<Text css={{color: 'white'}} />Box>
```
**after** 
```
<Box css={{color: 'white'}}><Text /><Text /><Text /><Text /></Box>
```

This change will lead to significantly less code, hard colour overrides and less CSS being shipped to the client.


elliotlaws (approved):

thomasdigby (approved):
We have a couple of places where `color: inherit` is still used to override the set colour on `Text` and `Heading`, can we update these? I'd be keen to release this update soon so we can start migrating atom-core over 👍

LimeWub:
@thomasdigby do you mean in the component library itself? I can go through lib and remove hardcoded colours where not needed if that's what you mean? But it will add a bit more bloat to this PR, is that ok?

thomasdigby:
> @thomasdigby do you mean in the component library itself? I can go through lib and remove hardcoded colours where not needed if that's what you mean? But it will add a bit more bloat to this PR, is that ok?

Yeah in the component library. Hardcoded colours are fine where they're necessary, but there are a few unnecessary `color: inherit` now, so we should remove them. I found two, but worth having a deeper look across the components.

LimeWub:
@thomasdigby Done that btw~ will merge this and update in core tomo

thomasdigby (approved):

PR #436, opened 2022-11-22
refactor: use ColorScheme for Tabs and Accordion component

LimeWub:
## Description

Replaces Accordion and Tab theming to use `ColorScheme`. 
This was a required change to `Tabs` so it can be used without colour related issues from prop drilling colours when used more flexibly during the CMS build and an optional/experimental change to Accordion - which worked well thus I'm deciding to keep.

Included in the changes of `Tab` is a significant simplification of the component since it had complex props to do with scrolling arrows and scroll percentages that were used nowhere in core. As well as a removal of the uppercase option and lightening of the divider (since I'm in the file) [according to this Jira ticket.](https://atomlearningltd.atlassian.net/browse/DS-107)

Documentation has also been updated with the ColorScheme related content.

## Videos

https://user-images.githubusercontent.com/6905473/203793945-3ca3ffef-18eb-43f5-9996-16c1a00a1706.mov

https://user-images.githubusercontent.com/6905473/203796840-10189da1-6bd8-458d-9f56-7532c177f216.mov


lib/src/components/tabs/TabsTriggerList.tsx
```diff
@@ -1,205 +1,30 @@
...
+export const TabsTriggerList: React.FC<
+  React.ComponentProps<typeof StyledTriggerList> & {
+    colorScheme?: typeof ColorScheme
   }
-
+> = ({ children, colorScheme = {}, ...rest }) => {
   return (
-    <StyledTriggerList
-      theme={theme}
+    <ColorScheme
```
thomasdigby, line 19: Do you think there could be some value from moving the `ColorScheme` component up to the root `Tabs` element? It shouldn't change how the rest of the component/theme is implemented but could also allow use of these theme tokens in the content area as well. I can imagine some tab implementation in the main app using `loContrast` colour scheme in the content area

LimeWub: Ye I need to think about this one a bit - I might end up wrapping a CS around Content rather than wrapping the full thing in it to keep flexibility.

thomasdigby: Ok fair, I think as a user I would expect the color scheme to apply on a per-component basis, rather than to individual sub-components within, but I probably need to use it a bit before I have a particularly strong opinion on it 👍

LimeWub: When it's eventually used it would be used per section - rather than per component or subcomponent (and then nested as much as we want into more granular components). Adding it to the component/subcomponent here is a semi-temporary solution.

LimeWub: @thomasdigby 
Sooooo - I started adding a ColorScheme around `TabContent` to make it themable as well and I've decided against it for the following reasons:
* The way CS will eventually be used is not via a specific component but instead the same way as `Box` is: as a wrapper of sections from which the nested components would just get colours. Then the nested components do not receive any CS relevant props but the CSS just updates with the correct css vars cascading. So, adding a theme to Contents or to the full Tabs was being considered as a temporary solution anyway.
* As there is only 'slate' in the ColorScheme right now there is no change to `TabContent` regardless of what CS settings are applied to it. Accent colour changes don't do anything to Links and Buttons right now so nothing would be affected in its children yet. Thus there's no benefit to adding it to Content as a temporary solution either.
* Tab **did** use to have a 'dark' theme so I considered re-adding that for a bit. ColorScheme currently doesn't support dark mode and we don't have any dark based themes so that could be an issue. **However,** I checked across the core website and nothing is using the theming in Tabs - so this shouldn't be an issue as the functionality may be regressing but it was unused so doesn't affect anything. If it's needed again in the future (hopefully more progress with CS until then?) We can consider and re-add it then. If CS has made enough progress to be usable in core then we wouldn't need to add it anyway as we would be applying the CS as a wrapper - rather than on the specific component ^^

**Why not apply it on the Tab/Accordion root?**
* The theming does nothing for the Tab Content or Accordion Content - I think it'd be confusing to apply it to the full component (even temporarily) if it currently will have no effect for the entire component 🤷 
* Tabs can be many nesting levels above where the Tab Links are in the DOM - which makes me feel a bit icky about applying a scheme to it. For the purposes of the use in design system's site it kinda doesn't matter where the CS is added as I can override (and I am overriding in the page Header for example with another CS anyway)... but basically adding a CS to the Tabs doesn't mean it'd be cascading down enough without being overwritten before it reaches where its actually useful. Again, this kinda doesn't matter as the embedded CS in these components is a temporary solution anyways but it doesn't feel right to put the scheme much further up from where it's actually used.

lib/package.json
```diff
@@ -166,7 +167,7 @@
   "husky": {
     "hooks": {
       "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
-      "pre-commit": "bash scripts/prevent-commit.sh && lint-staged"
+      "pre-commit": "bash ../scripts/prevent-commit.sh && lint-staged"
```
LimeWub, line 170: This change was needed to unblock the pre-commit hook from dying like so:
Screenshot 2022-12-21 at 12 29 30

It worked, but let me know if it's wrong 🤷

lib/src/components/tabs/TabsTriggerList.tsx
hollg: @LimeWub thank you for writing that detailed response — the original question wasn't from me, but I learned a lot reading it and now understand this PR much better!

Your explanation of the reasoning for _not_ putting the `ColorScheme` at the root of these components makes sense to me, but the fact that the logic relies so much on "this is only a short-term solution anyway" makes me a little nervous. I do think it seems like the right approach, just that we need to make sure we don't let this linger too long. I think you're already talking to Miguel lots about next steps for the `ColorScheme` component though, which seems like the relevant next step, so I don't think there are any new actions here from me.

hollg (approved):

thomasdigby (approved):

README.md
```diff
@@ -31,7 +31,7 @@ You can read more about the components included in `@atom-learning/components` a
 
 We can run the documentation site locally, with hot-reloading triggered by changes in the library. This offers a low-friction dev environment where we can see the effects of any local changes when applied to all existing components. (Note: hot reloading is currently only supported for changes to Typescript/.tsx files, not markdown files.)
 
-First, run `yarn build:lib && yarn build:docs`. `yarn build:docs` collects the markdown documentation files for each component in `lib` into its `dist` directory, where the documentation site can find them. `yarn build:docs` compiles the documentation and extract the component props.
+First, run `yarn build:docs && yarn dev:lib`. `yarn build:docs` collects the markdown documentation files for each component in `lib` into its `dist` directory, where the documentation site can find them. `yarn dev:lib` compiles the React component library and watches for changes.
```
LimeWub, line 34: Just doing a test rn to see if the changes version of this is working correctly still (red version) if it is I'll revert this override and keep `yarn build:lib && yarn build:docs`

lib/src/components/tabs/TabsContent.tsx
```diff
@@ -0,0 +1,11 @@
+import { Content } from '@radix-ui/react-tabs'
+import { styled } from '~/stitches'
+
+const StyledTabsContent = styled(Content, {
+  flexGrow: 1,
+  fontFamily: '$body'
+})
+
+export const TabsContent = StyledTabsContent
```
thomasdigby, line 9: Can we export the `styled` component directly? Is there any benefit to avoiding that? e.g.

```ts
export const TabsContent = styled(Content, {
  flexGrow: 1,
  fontFamily: '$body'
})
```

lib/src/components/tabs/TabsTrigger.tsx
```diff
@@ -0,0 +1,55 @@
...
+  extends React.ComponentProps<typeof StyledTabsTrigger> {
+  value: string
+}
+
+export const TabsTrigger: React.FC<TabsTriggerProps> = ({
+  children,
+  ...rest
+}) => (
+  <StyledTabsTrigger asChild {...rest}>
+    <Text size="sm" as="button">
```
thomasdigby, line 49: What's the reason for using `Text` here instead of styling the `TabTrigger`?

lib/src/components/tabs/TabsTriggerList.tsx
```diff
@@ -54,43 +20,40 @@ const StyledTriggerList = styled(List, {
...
-  const scrollTriggerListTo = useCallback((direction: 'left' | 'right') => {
-    const triggerList = triggerListRef.current
-    if (triggerList) {
-      const { scrollWidth, scrollLeft, offsetWidth } = triggerList
-      const scrollAmount = Math.round(scrollWidth * (scrollPercentage / 100))
+const StyledChevronActionIcon = styled(ActionIcon, {
+  height: '100% !important',
+  position: 'absolute',
+  top: '50%',
+  transform: 'translateY(-50%)',
```
thomasdigby, line 30: Is this change to the logic of positioning the `<` and `>` arrows relevant to this PR? It would be great to just see the theme/color-scheme changes here

lib/src/components/tabs/TabsTriggerList.tsx
```diff
@@ -54,43 +20,40 @@ const StyledTriggerList = styled(List, {
...
-    }
-  }
+  scrollbarWidth: 'none'
 })
 
-export const TriggerListWrapper: React.FC<ListProps> = ({
-  children,
-  theme,
-  appearance,
-  enableTabScrolling,
```
thomasdigby, line 80: Why are we removing the `enableTabScrolling` prop? I think we should reinstate it alongside any other logic related to the previous way of scrolling through tab triggers. It would be good to move this to your separate PR if you can

lib/src/components/tabs/TabsTrigger.tsx
LimeWub: It uses the correct font size and family from Text. Better than repeating the same css in StyledTabsTrigger right?

lib/src/components/tabs/TabsTriggerList.tsx
LimeWub: I've removed the props which are not used anywhere.

lib/src/components/tabs/TabsContent.tsx
LimeWub: I think there's a benefit to setting it to a const as I can give it a displayName.
What's the benefit of exporting the styled component directly?

lib/rollup.config.js
```diff
@@ -36,6 +37,7 @@ export default {
       target: ['chrome80', 'edge91', 'firefox78', 'ios12.2', 'safari12.1']
     }),
     resolve({ extensions: ['.ts', '.tsx'] }),
-    isDebug && visualizer()
+    isDebug && visualizer(),
+    json()
```
hollg, line 41: Is this to process `slate.json` (and other similar future files)?

LimeWub: yes

lib/src/components/tabs/TabsTrigger.tsx
hollg: My 2c is that  really like the `asChild` pattern for stuff like this. I wonder whether we run the risk  the case of causing any confusion in this case by having a big block of styles defined for `StyledTabsTrigger` that _doesn't_ contain some of the styles which are always applied when we render it, though.

@LimeWub do you know how much styling we'd be repeating from `Text` if we applied it directly on `StyledTabsTrigger` / whether any of it could be generated using a util rather than writing it out manually?

LimeWub: No I don't know how much stuff would be repeated but why would we want to be repeating and applying it directly on `StyledTabsTrigger` when we can use the component with `asChild` and pass the size prop?

I don't see how leaving the sizing and text related things to `Text` - it is clearly using `Text` so 🤷

lib/src/components/top-bar/TopBar.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 
 import { CSS, styled } from '~/stitches'
-import { useScrollPosition } from '~/utilities/hooks/useScrollPosition'
+import { useWindowScrollPosition } from '~/utilities/hooks/useWindowScrollPosition'
```
hollg, line 4: What's the reason for these changes? `TopBar` isn't related to `Tabs` or `Accordion`, right?

LimeWub: It's a rename for a file that has been deleted and readded in the subsequent PR for the tabs scrolling. There is 0 difference in this change it's only a file rename.

hollg: There's also a change _inside_ the `useWindowScrollPosition` file too, to use a different function from `throttle-debounce`. If these changes are related to the other PR, please can you remove them from this one and add them to that one? I think there's another similar thread somewhere about styling something in the tab buttons. Hopefully it's not too complex to move them across since I think they'll all be going into the same PR.

This PR should be reviewable by someone who doesn't know anything about `ColorScheme` before they open it. Changes like this are confusing because someone who doesn't already fully understand this work will be trying to understand how this change relates to the other stuff they're looking at. If it doesn't, it's just noise here and we should move it to a PR where it is helping people to see the full picture of a change, rather than obscuring it.

I get the _the changes are going in either way, so why does it matter which PR it happens in?_ feeling, but well-scoped PRs are an easy to come back to later to understand where changes came from/how they relate to each other, and poorly scoped PRs are not. Similarly, PRs are an important tool for getting our less experienced (either in general, or with regard to specific areas of the codebase) colleagues up-to-speed on what's going on. So leaving these unrelated changes in a PR like this because we know that they will be merged somewhere else anyway is really not a no-downside situation.

LimeWub: there: https://github.com/Atom-Learning/components/pull/457/files

lib/src/components/tabs/TabsContent.tsx
hollg: Is React not able to infer the name `TabsContent` in this context? I've never felt particularly strongly either way on this one but if React can't pick up the name then making manual `displayName` the convention seems like a good idea to me.

lib/src/components/tabs/TabsTrigger.tsx
hollg: > why would we want to be repeating and applying it directly

I don't think repetition is particularly evil in and of itself, it's all just about how it's balanced against other values. In this case, I was just wondering whether avoiding repetition conflicts with code clarity. Maybe not though — as you say,  the `Text` isn't exactly hidden once you're in this file. And I think we are using `Text` pretty consistently throughout the codebase, so breaking that pattern might introduce its own kind of confusion here. I reckon `Text` is probably fine.

lib/src/experiments/color-scheme/ColorScheme.tsx
```diff
@@ -0,0 +1,52 @@
+import * as React from 'react'
```
hollg, line 1: I like `experiments/` 👍🏻

lib/src/utilities/hooks/useWindowSize.ts
```diff
@@ -0,0 +1,30 @@
+import { useEffect, useState } from 'react'
+import { debounce } from 'throttle-debounce'
+
+interface Size {
+  width: number | undefined
+  height: number | undefined
+}
+
+export const useWindowSize = ({ delay = 500 } = {}) => {
```
hollg, line 9: I can't get my head around `{ delay = 500 } = {}` 😅. I get that we're defaulting `delay` to 500ms, but what's the second `=` doing?

LimeWub: It's because I've not used typescript properly so it gives me a curly underline unless I give the object a default value. Tbf I think this one is an easy enough fix - I found out how to do it with ts in a later hook.

hollg: Oh, does it not just need a type for the input along the lines of 

```ts
{ delay?: number }
```

LimeWub: You know what I actually do need the = {} because all the properties are option so you can use the hook like this:
`useHook()` instead of having to do `useHook({})`.
Went to remove them and noticed. Another way would be to do something like `<Optional<type>>` but that doesn't work in components (I don't know why). The equal empty object does the same thing for infering type.

hollg: Would it work without the object at all? i.e.

```ts
export const useWindowSize = (delay: number = 500)) => {
```

LimeWub: It would I just like objects cause they make it clear which property is for what and they don't depend on order.

hollg: I am definitely v pro config objects when the alternative is passing a long list of argument (**cough** useAPI **cough**) but imo that is overkill for a function with a single param. TypeScript should provide the name and type of the param and we can add a JSDoc comment to describe it too if it's helpful to explain _what_ exactly is delayed etc.

LimeWub: I'm keeping the same pattern for all the hooks I write and I think if it sometimes is an object and sometimes it isn't it'll b confusing 👀

lib/src/components/tabs/TabsContent.tsx
LimeWub: In a roundtable near the end of last year we agreed to add `.displayName` to our components as they do seem to not be getting picked up correctly in the react inspector. This is what I'm going by when adding this here.

fwiw I was debugging the react components last week or so and I had no clue where I was in the tree based on names - so ye - I don't think it's getting inferred correctly most times.

I think exporting `Styled` directly and forwardRefs seem to be ones to blame:
Screenshot 2023-01-26 at 11 07 14

Anyway it's been agreed to try to use `displayName` as much as we can (when we don't forget to basically)

hollg: Oh great, if a group decision has already been reached on that then 👍🏻. Also lol that tree is a mess, so anything we can do to improve it ...

lib/src/utilities/hooks/useWindowSize.ts
LimeWub: I've added the type for it which was missing but left the defaulting in.

hollg: _ohhhh_ I now see what you were describing earlier. Makes sense! Not sure where I come down on passing objects vs numbers in general, but v happy for this to go in and then maybe we can discuss it in an FE roundtable and see how people feel generally. I also never know whether to wrap the return values in objects, because it feels unnatural not to destructure the value after calling the hook 🙃

PR #437, opened 2022-12-02
Feature: Upgrade `@atom-learning` packages

thomasdigby:


site/package.json
```diff
@@ -60,8 +60,5 @@
     "react-docgen-typescript": "^1.21.0",
     "type-fest": "^0.20.2",
     "typescript": "^4.1.3"
-  },
-  "engines": {
-    "node": "^14.20"
```
thomasdigby, line 65: This `node` engine version is causing a bunch of issues locally, and I'm certain it was only added as an attempt to work around DO deployment issues. If anyone thinks otherwise let me know and I'll keep this in

LimeWub: I think it was a work around for something `site` related, was it the deployment or something? (Didn't you add it?)
In my CMS PR I've also removed it as it's no longer needed for it (`site` entirely dropped). I dunno what effect it'll have if it's removed early.

LimeWub (approved):

site/package.json
thomasdigby: I think I was trying to work around a TS/Node issue when deploying the docs site one night while everyone was having drinks in the office... so I'm pretty sure it's a test thing that I didn't remove. It'll be easy to resolve if it does break the site 😅

andoulla (approved):

PR #439, opened 2022-12-05
Feat/drag drop table rows

hollg:
## Drag-and-drop table rows
Adds `DataTable.DragAndDropTable`, enabling the user to reorder its rows via drag-and-drop. An `onDragAndDrop` prop can synchronise external state / remote data with those changes.

## Implementation
I used `@dnd-kit/core` and `@dnd-kit/sort` since we've already used them in `atom-core` for the exam config builder. We should bring some of the components used there across to this library, and during that work figure out which aspects of these two features can use shared abstractions.

`DataTable.DragAndDropTable` renders a `DragAndDropContainer` around the table, which sets up all the logic for the drag-and-drop functionality and for keeping track of the order of the rows.`DataTable.Row` is replaced with a specific `DraggableRow` here, which includes a draggable handle and hooks into the relevant contexts from the `@dnd-kit` packages by setting a ref on the `tr` element. 

The `@dnd-kit/sort` requires that each row has a unique ID. The `DataTable.DraggableRow` will expect that id to be in an `id` property on the items in `data` by default, but this can be overridden by providing an `idColumn` prop. E.g. we'll probably use things like `idColumn="id_user"` a lot. This should prevent consumers from having to map or mutate their data every time they want to display it in a table.

## Conflict with other features
Drag-and-drop reordering conflicts with column sorting (enabled by passing `sortable` prop to `DataTable.Table`). I think the correct way to handle this problem is to clearly document it and tell consumers not to enable both at the same time. In fact, we probably want to encourage consumers to put the drag-and-drop functionality behind some sort of toggle, which could also toggle off the column sorting. 

Similarly, if pagination is enabled then the user can't use drag-and-drop properly. We should warn against this too.

## Test issues
The `@dndkit` libraries rely on browser APIs that aren't available in the JSDom environment, so I haven't found a way to write meaningful unit tests for this functionality yet. Instead I've exported some should-really-be-private functions from `DragAndDropContainer`, and am testing that they are passed the right data when the `onDragAndDrop` change event is triggered.

## Future work
In the future we'll build some more generic drag-and-drop utils (likely when we move the drag-and-drop components used in the exam config builder over to the library), at which point some of the logic here can probably be simplified by making use of the generic stuff.

## Examples

https://user-images.githubusercontent.com/21319237/207274783-2d5425ec-2ca0-451a-9ad6-8d98fe1af27f.mov


lib/src/components/data-table/DataTable.mdx
```diff
@@ -57,12 +57,12 @@ There are basically two ways to use `DataTable` to build a table from this confi
 The following two examples are exactly equivalent in their output. The second example is included to demonstrate what `DataTable`'s subcomponents do: they bundle up UI components and logic to provide useful defaults. They exist at various levels of abstraction, e.g. `DataTable.Table` renders `DataTable.Body`, which renders `DataTable.Row`. This means that you can use whichever component provides useful functionality for your use case while still being low-level enough to let you combine it with your own custom logic.
 
 ```tsx
-<DataTable columns={columns} data={data}>
+<DataTable columns={columns} data={data} idColumn='userId'>
```
thomasdigby, line 60: I thought this was only required for drag & drop tables? Do we want to pass that prop in here if it's not relevant to the example?

lib/src/components/data-table/DataTable.mdx
```diff
@@ -154,3 +154,11 @@ If `DataTable`'s `isSortable` state is `true`, then `DataTable.Header` will be c
 ### Pagination
 
 `DataTable.Pagination` can be passed as a child to `DataTable` to render the pagination UI and configure the parent `DataTable` to paginate its data. Note: there is currently a bug where `DataTable` will render all rows on the initial render, even when paginated. There is no visible effect, but it can cause performance issues on large tables. The workaround for this is to pass an `initialState` prop to `DataTable`. E.g. `initialState={pagination: {pageSize: 5, pageIndex: 0}}`. This will force pagination to apply properly on the initial render. Make sure the `pagination` prop matches the config you use in `DataTable.Pagination`, as the latter will take precedence after the initial render.
+
+### Drag and dropp
```
thomasdigby, line 158:
```suggestion
### Drag and drop
```

lib/src/components/data-table/DataTable.mdx
```diff
@@ -154,3 +154,11 @@ If `DataTable`'s `isSortable` state is `true`, then `DataTable.Header` will be c
...
+`DataTable` accepts a `dragAndDrop` prop which you can use to allow users to reorder table rows via drag and drop. You'll normally use it like this: `dragAndDrop={ {active: true, onChange: (oldIndex, newIndex, newData) => console.log (`row ${oldIndex} moved to ${newIndex})} }`. If you would prefer
```
thomasdigby, line 159:
```suggestion
`DataTable` accepts a `dragAndDrop` prop which you can use to allow users to reorder table rows via drag and drop. You'll normally use it like this: `dragAndDrop={ {active: true, onChange: (oldIndex, newIndex, newData) => console.log ('row ${oldIndex} moved to ${newIndex}')} }`. If you would prefer
```

lib/src/components/data-table/DataTable.mdx
hollg: Yeah that's a good point, I should document it in the prose rather than just putting it here in the generic example! Though maybe it's worth pointint out that since we can't do hooks conditionally, we'll still record an `order` array in state, whether drag and drop is being used or not. It's just likely to be full of `undefined`.

lib/src/components/data-table/DataTableRow.tsx
```diff
@@ -2,14 +2,60 @@ import type { Row } from '@tanstack/react-table'
...
+      css={{
+        transform: CSS.Transform.toString(transform),
+        // Online examples apply the transition to all rows, but here it causes a bug where
+        // the displaced rows move a second time on drag end
+        // transition,
+        zIndex: isDragging ? 5 : undefined
+      }}
+    >
+      {row.getVisibleCells().map((cell, i) => {
+        if (isDragAndDrop && i === 0) {
```
thomasdigby, line 42: Do you think it might be worth creating a separate `DataTableDraggableRow` component to keep the logic separate? We're including `css` for all tables with a transform/z-index that may not be necessary, and the duplicated code will be fairly minimal.

lib/src/components/table/__snapshots__/Table.test.tsx.snap
```diff
@@ -453,31 +453,31 @@ exports[`Table component renders with square corners 1`] = `
...
+  .c-cwQMhQ-codIxD-corners-round [object Object]:last-child .c-cCZzXk:last-child {
     border-bottom-right-radius: var(--radii-0);
   }
 
   .c-PJLV-jEDwTF-theme-primaryDark .c-GSaiK {
     background: var(--colors-primaryDark);
   }
 
-  .c-PJLV-gdcSSw-striped-true .c-hjAYDb:nth-child(odd) {
+  .c-PJLV-gbDJOS-striped-true [object Object]:nth-child(odd) {
```
thomasdigby, line 476: Do you know what css/component might be affecting this selector?

lib/src/components/table/TableRow.tsx
```diff
@@ -1,7 +1,14 @@
...
 
-export const TableRow = styled('tr', {
+const StyledRow = styled('tr', {
   bg: 'unset'
 })
 
+export const TableRow = React.forwardRef<
+  HTMLTableRowElement,
+  React.ComponentProps<typeof StyledRow>
+>((props, ref) => <StyledRow {...props} ref={ref} />)
```
thomasdigby, line 12: Does the `styled()` component not accept a `ref` by default?

lib/src/components/data-table/DataTableRow.tsx
hollg: Certainly happy to!

lib/src/components/data-table/drag-and-drop/DragAndDropContainer.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const { idColumn, onDragAndDrop, order, setData } = useDataTable()
+
+  const sensors = useSensors(
+    useSensor(PointerSensor),
+    useSensor(KeyboardSensor, {
+      coordinateGetter: sortableKeyboardCoordinates
+    })
+  )
+
+  function handleDragEnd(event) {
```
thomasdigby, line 26: Super minor, just for consistency

```suggestion
  const handleDragEnd = (event) => {
```

lib/src/components/data-table/drag-and-drop/DragAndDropContainer.tsx
```diff
@@ -0,0 +1,49 @@
...
+  useSensor,
+  useSensors
+} from '@dnd-kit/core'
+import { restrictToVerticalAxis } from '@dnd-kit/modifiers'
+import { arrayMove, sortableKeyboardCoordinates } from '@dnd-kit/sortable'
+
+import * as React from 'react'
+
+export const DragAndDropContainer: React.FC = ({ children }) => {
+  const { idColumn, onDragAndDrop, order, setData } = useDataTable()
```
thomasdigby, line 17: Could we rename the `order` key? Feels a touch too "functiony", like "order this thing". `rowOrder`? 🤷

lib/src/components/data-table/DataTableTable.tsx
```diff
@@ -17,9 +18,23 @@ export const DataTableTable: React.FC<DataTableTableProps> = ({
...
-    <DataTable.Head theme={theme} sortable={sortable} />
-    <DataTable.Body striped={striped} />
-  </Table>
-)
+}) => {
+  const { isDragAndDrop } = useDataTable()
+
+  if (isDragAndDrop)
+    return (
+      <DragAndDropContainer>
```
thomasdigby, line 26: A minor point for now, but it would be fantastic to avoid bundling `@dnd-kit/*` packages if they're unused in a simple table. Might be a bit tricky as we're relying on a dynamic prop, but worth considering for future iterations if there's a component API change that might allow for this. https://bundlephobia.com/package/@dnd-kit/core@6.0.6

lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -75,12 +116,32 @@ export const DataTableProvider = ({
       : []
   )
 
+  React.useEffect(() => {
+    setData(dataProp)
+  }, [dataProp])
```
thomasdigby, line 121: What's the reason for including this effect? I can see that we're setting data when calling `moveRow`, is there another side-effect, e.g. syncing with external state, that requires this?

lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -49,21 +67,44 @@ type TableProviderProps = {
...
   children: React.ReactNode
-  initialState?: InitialState
+  dragAndDrop?:
+    | boolean
+    | {
+        active: boolean
+        onChange: (
+          oldIndex: number,
+          newIndex: number,
+          newData: Array<Record<string, unknown>>
```
thomasdigby, line 77: Might be useful to pull out a "TableData" type as we're using this in a few places, even if it's simple.

lib/sandbox/index.tsx
```diff
@@ -1,22 +1,69 @@
...
+          // }}
+          // dragAndDrop
+        >
+          {/* <DataTable.GlobalFilter /> */}
+          <DataTable.Table sortable css={{ mb: '$4' }} />
+          <DataTable.Pagination pageSize={5} />
+        </DataTable>
+      </Box>
+    </Flex>
+  </Tooltip.Provider>
```
thomasdigby, line 66: Can we reset this file?

lib/src/components/data-table/DataTable.mdx
LimeWub, line 60: Do we need to have an 'idColumn'? Can we just require `id` to be passed into the data? (so: 
`{ name: 'chrissy', hobby: 'bare-knuckle boxing', userId: 1 }` => `{ name: 'chrissy', hobby: 'bare-knuckle boxing', id: 1 },` )

And just assume `idColumn` is always `id`?

LimeWub: Is passing `idColumn` better for performance?

lib/src/components/data-table/DataTableBody.tsx
```diff
@@ -12,13 +13,31 @@ export const DataTableBody: React.FC<DataTableBodyProps> = ({
   striped = false,
   ...props
 }) => {
-  const { getRowModel } = useDataTable()
+  const { isDragAndDrop } = useDataTable()
+
+  if (isDragAndDrop)
+    return (
+      <Table.Body {...props} striped={striped}>
+        <SortableContainer>
```
LimeWub, line 21: Why not just export the 'SortableContainer' like `DataTable.DragAndDrop` and not require any props and put everything drag and drop related on this?

I think it'd be much cleaner and we can add more things without too much complexity/repeating code.

I really don't like it if we start to add a billion props to the DataTable itself and do if checks in there. That's something I thought we wanted to fix in DT as opposed to the complexity of EnhancedTable.

Lets catch up and discuss this?

lib/src/components/data-table/DataTableBody.tsx
```diff
@@ -12,13 +13,31 @@ export const DataTableBody: React.FC<DataTableBodyProps> = ({
...
+
+const Rows = () => {
+  const { getRowModel } = useDataTable()
+  return (
+    <>
       {getRowModel().rows.map((row) => {
         return <DataTable.Row row={row} key={row.id} />
       })}
-    </Table.Body>
+    </>
```
LimeWub, line 41: You don't need a fragment here (but I also don't think this setup with the prop which causes the need for `Rows` is good anyway so this all could potentially be removed/reverted anyway 🤷 )

lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -6,6 +6,9 @@ import {
   getSortedRowModel,
```
LimeWub, line 6: This file is starting to get a bit too long. Can we move some functionality in hooks or similar to keep it a bit neater? What do u think?

lib/src/components/data-table/DataTableTable.tsx
LimeWub, line 26: Can the `DragAndDropContainer` and `SortableContainer` be applied on the same level or does `SortableContainer` NEED to be in `Table.Body`?

lib/src/components/table/__snapshots__/Table.test.tsx.snap
```diff
@@ -47,31 +47,31 @@ exports[`Table component renders 1`] = `
...
     border-top-left-radius: var(--radii-0);
   }
 
-  .c-cwQMhQ-fofZjp-corners-round .c-GSaiK:last-of-type {
+  .c-cwQMhQ-codIxD-corners-round .c-GSaiK:last-of-type {
     border-top-right-radius: var(--radii-0);
   }
 
-  .c-cwQMhQ-fofZjp-corners-round .c-hjAYDb:last-child .c-cCZzXk:first-child {
+  .c-cwQMhQ-codIxD-corners-round [object Object]:last-child .c-cCZzXk:first-child {
```
LimeWub, line 58: `[object Object]`

uh oh

lib/src/components/data-table/DataTableRow.tsx
```diff
@@ -2,14 +2,60 @@ import type { Row } from '@tanstack/react-table'
...
+          return (
+            <Table.Cell key={cell.id}>
+              {/* TODO: decide on label */}
+              <Handle
+                {...attributes}
+                {...listeners}
+                isDragging={isDragging}
+                css={{ display: 'inline-block' }}
+                label="drag and drop"
```
D7Torres, line 51: I've been discouraged from having copy in DS components in the past.
Instead, I was told it should be passed as a prop so that the appropriate translation could be passed.

I wonder if, in some cases like this, we should make an exception to avoid complicating an already big API.

lib/src/components/data-table/drag-and-drop/DragAndDropContainer.tsx
```diff
@@ -0,0 +1,49 @@
+import { useDataTable } from '../DataTableContext'
+import {
+  closestCenter,
+  DndContext,
+  DragOverlay,
```
D7Torres, line 5: It looks like `DragOverlay` is not being used

lib/src/components/data-table/drag-and-drop/Handle.tsx
```diff
@@ -0,0 +1,33 @@
+import { ActionIcon } from '../../action-icon'
+import { Icon } from '../../icon'
+import { styled } from '../../../stitches'
+import { DragHandle, User } from '@atom-learning/icons'
```
D7Torres, line 4: `User` not in use

lib/src/components/data-table/DataTable.mdx
```diff
@@ -154,3 +154,11 @@ If `DataTable`'s `isSortable` state is `true`, then `DataTable.Header` will be c
...
+
+### Drag and drop
+`DataTable` accepts a `dragAndDrop` prop which you can use to allow users to reorder table rows via drag and drop. You'll normally use it like this: `dragAndDrop={ {active: true, onChange: (oldIndex, newIndex, newData) => console.log ('row ${oldIndex} moved to ${newIndex}')} }`. If you would prefer
+to define the `onChange` behaviour elsewhere, by subscribing to `useDataTable` and watching for changes in the `data` property, you can simply pass `dragAndDrop` to `DataTable` as a boolean instead.
+
+Note that column sorting conflicts with drag and drop behaviour. In any context where you allow drag and drop reordering, you probably want to disable column sorting (see User Sorting above). Similarly, you should probably disable pagination.
```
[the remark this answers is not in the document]
hollg: That's exactly the reason! But I'll definitely add something to state that explicitly here. Thanks @D7Torres

lib/src/components/data-table/drag-and-drop/DragAndDropContainer.tsx
hollg: Great call. Will do!

lib/src/components/data-table/DataTable.mdx
```diff
@@ -154,3 +154,11 @@ If `DataTable`'s `isSortable` state is `true`, then `DataTable.Header` will be c
...
+
+### Drag and drop
+`DataTable` accepts a `dragAndDrop` prop which you can use to allow users to reorder table rows via drag and drop. You'll normally use it like this: `dragAndDrop={ {active: true, onChange: (oldIndex, newIndex, newData) => console.log ('row ${oldIndex} moved to ${newIndex}')} }`. If you would prefer
+to define the `onChange` behaviour elsewhere, by subscribing to `useDataTable` and watching for changes in the `data` property, you can simply pass `dragAndDrop` to `DataTable` as a boolean instead.
```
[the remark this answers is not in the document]
hollg: I generally try to avoid those prefixes for boolean _props_ (as opposed to variables) because doing so keeps them consistent with HTML attributes, which are used in the same way. E.g. it's `<input disabled/>` rather than `<input isDisabled>`. That's always felt more intuitive to me, but I can totally see why someone might find the inconsistency between boolean variables and boolean props more confusing than the alternative inconsistency between HTML attributes and custom React props, so I'm happy to be led by the group!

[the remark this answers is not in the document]
D7Torres: Ah, thanks for letting me know. As far as we are consistent it seems reasonable.

lib/src/components/data-table/DataTableBody.tsx
hollg: The reason I didn't (only) make it a separate export that can be composed with the other subcomponents by the consumer is that it would mean anyone who wants to use drag-and-drop rows has to build the whole table from scratch. I.e. they can't use `DataTable.Table`. 

I do think this is a big enough feature that we should include it within the "easy-to-use defaults" category but am happy to chat about it!

lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -75,12 +116,32 @@ export const DataTableProvider = ({
...
   const getTotalRows = () => data.length
 
+  const moveRow = (oldIndex: number, newIndex: number) => {
+    const maxIndex = data.length - 1
+    const minIndex = 0
+    if ([oldIndex, newIndex].some((i) => i > maxIndex || i < minIndex)) {
+      throw new Error(`You can't move a row to or from an index that doesn't exist`)
+    }
+
+    const dataCopy = JSON.parse(JSON.stringify(data))
```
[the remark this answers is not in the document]
hollg: Ooh, I hadn't thought about that at all. Nothing silly in pointint it out! V happy to make that change.

lib/src/components/data-table/DataTableRow.tsx
hollg: This is a really good point and it's something that we have a ticket in the DS backlog to tackle properly. 

Since we've progressed from only having _super low-level_ components (e.g. `Button`) in the library to having more complex and feature-rich ones like this, it has become much harder to stick to a strict rule of never having copy in the lib. In this example, we really want the consumer of the library not to have to build their own table row, but if they don't then they don't currently have direct access to this label prop to pass their own value in. So I think what we need in the medium — long term is a consistent, systematic way of specifying default copy in the library and for the consumer to overwrite it. I'm keen to prioritise that heavily in the DS backlog and just accept in the meantime that there's gonna be a few examples like this that make us feel a bit 🤢.

It might be that we can actually re-use a lot of the same logic that we have in `atom-core`'s `i18n` util. Maybe we could document string names for complex components like this, and let the consumer use them to overwrite as they please.

lib/src/components/data-table/DataTableContext.tsx
hollg: Good call. I've rearrange some stuff and pulled some logic into a couple of hooks. Let me know if you think there's anything else that should move!

lib/src/components/table/TableRow.tsx
hollg: No 😩

lib/src/components/data-table/DataTableTable.tsx
hollg: Nope, they can be combined into one 🤯 thanks!

lib/src/components/table/__snapshots__/Table.test.tsx.snap
hollg: 😨

lib/src/components/table/__snapshots__/Table.test.tsx.snap
hollg: I thought it was the extra layer of abstraction in `Table.Row` with my changes there. But when I'm targetting the underlying stitches component, this persists. Looking into it atm!

hollg: That _was_ the problem, I just needed to update references to that component from within the table body, as well as the table 🙃

D7Torres (approved):

lib/package.json
```diff
@@ -127,6 +127,10 @@
   },
   "dependencies": {
     "@atom-learning/theme": "1.1.1",
+    "@dnd-kit/core": "^6.0.5",
+    "@dnd-kit/modifiers": "^6.0.0",
+    "@dnd-kit/sortable": "^7.0.1",
```
thomasdigby, line 132: Are all these versions compatible? I assume they're not all published with the same major version number

lib/src/components/data-table/DataTable.test.tsx
```diff
@@ -114,6 +114,16 @@ describe('DataTable component', () => {
     expect(screen.getByText('xena')).toBeVisible()
     expect(screen.queryByText('agatha')).not.toBeInTheDocument()
   })
+
+  it('Renders drag handles for draggable table rows', () => {
+   const {container} =  render(
+        <DataTable columns={columns} data={data} dragAndDrop>
```
thomasdigby, line 120: Do we still need this `dragAndDrop` prop?

package.json
```diff
@@ -2,10 +2,13 @@
   "private": true,
   "name": "design-system",
   "workspaces": [
+    "color-scheme",
     "lib",
     "site"
   ],
   "scripts": {
+    "dev:color-scheme": "cd color-scheme && yarn dev",
+    "build:color-scheme": "cd color-scheme && yarn build",
```
thomasdigby, line 11: Did you say this was an unnecessary change in our call?

lib/src/components/data-table/DataTable.test.tsx
hollg: 😮 I must not have run those tests again after the most recent change, but this doesn't work any more! Thanks, updated now 😅

package.json
hollg: Yep! Gone now.

lib/src/components/data-table/DataTableTable.tsx
```diff
@@ -22,7 +22,7 @@ export const DataTableTable: React.FC<DataTableTableProps> = ({
   css,
   ...props
 }) => {
-  const { asyncDataState } = useDataTable()
+  const { asyncDataState, isDragAndDrop } = useDataTable()
```
thomasdigby, line 25: Is `isDragAndDrop` used here?

lib/package.json
hollg: I just installed them with `yarn add {packageName}`. I'll double check whether they have any docs about mixing and matching major versions, but they definitely work currently, at least. Could always pin a specific version if it does seem iffy.

lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -148,6 +105,24 @@ export const DataTableProvider = ({
 
   const getTotalRows = () => data.total
 
+  const moveRow = (oldIndex: number, newIndex: number) => {
```
thomasdigby, line 108: Could we move this out of the context provider? I think it might make more sense co-locating it with the rest of the DnD logic/components. Similar maybe in approach to the asyncData stuff, e.g. there's stuff still here that's necessary, but anything that could be pulled out into a separate file has probably improved the readability a bit

hollg: Yeah you're right, I should have moved this with the other logic! Will move now.

hollg: In fact, no I'll delete it — it isn't used any more! :-)

lib/src/components/data-table/DataTableRow.tsx
```diff
@@ -2,14 +2,22 @@ import type { Row } from '@tanstack/react-table'
...
-    {row.getVisibleCells().map((cell) => (
-      <DataTableDataCell key={cell.id} cell={cell} />
-    ))}
-  </Table.Row>
-)
+
+export const DataTableRow: React.FC<DataTableRowProps> = ({ row }) => {
+  const { isDragAndDrop } = useDataTable()
+
+  if (isDragAndDrop) return <DraggableRow row={row} />
```
thomasdigby, line 14: Have you tested whether the conditional here is enough to stop DnD from being added to the bundle for a standard table? I have a suspicion that it's not... because `isDragAndDrop` is a runtime value that can't be evaluated when bundling

yarn.lock
```diff
@@ -1274,6 +1274,52 @@
...
+    tslib "^2.0.0"
+
+"@dnd-kit/utilities@^3.2.0":
+  version "3.2.0"
+  resolved "https://registry.yarnpkg.com/@dnd-kit/utilities/-/utilities-3.2.0.tgz#b3e956ea63a1347c9d0e1316b037ddcc6140acda"
+  integrity sha512-h65/pn2IPCCIWwdlR2BMLqRkDxpTEONA+HQW3n765HBijLYGyrnTCLa2YQt8VVjjSQD6EfFlTE6aS2Q/b6nb2g==
+  dependencies:
+    tslib "^2.0.0"
+
+"@dnd-kit/utilities@^3.2.1":
```
thomasdigby, line 1316: We have a duplicate package here, can we downgrade to `3.2.0` to sync up with the expected versions from the other packages?

yarn.lock
```diff
@@ -10953,6 +11004,13 @@ rollup-plugin-esbuild@^4.8.2:
     joycon "^3.0.1"
     jsonc-parser "^3.0.0"
 
+rollup-plugin-json@^4.0.0:
```
thomasdigby, line 11007: Do we know where this is coming from? 🤔

lib/src/components/data-table/drag-and-drop/DragAndDropContainer.tsx
```diff
@@ -0,0 +1,81 @@
...
+  children
+}) => {
+  const { data, setData, setIsDragAndDrop } = useDataTable()
+
+  const rowOrder = React.useMemo(
+    () =>
+      data.results.map((row) => {
+        const id = row[idColumn]
+        if (id === undefined)
+          console.error(
```
thomasdigby, line 37: Is it worth `throw`ing here? I assume this wouldn't allow drag and drop, and it would silently fail? I'm on the fence... but if we have a good Error Boundary wrapping each table then it should limit the impact 🤔

yarn.lock
hollg: I assume one of the `@dnd-kit` packages, because another `rollup-plugin-` dep changed when I switched the version of their `utilities` package just now

thomasdigby (commented):
Do we want to add some specific DnD tests to ensure the behaviour is working as expected? I appreciate that it could seem like we're testing the underlying package functionality (that is already well tested no doubt), but we've also written a lot of 1st party code that ought to be tested alongside that.

lib/src/components/data-table/drag-and-drop/DragAndDropContainer.tsx
hollg: Weirdly, if some-but-not-all rows are missing IDs then it can still work. But that's a risky thing to rely on and I imagine it's likely that every row will be given its ID the same way and therefore if any are missing one, they all will be. Will throw and add a recommendation to use an `ErrorBoundary` in the docs.

hollg:
@thomasdigby 
> Do we want to add some specific DnD tests to ensure the behaviour is working as expected? I appreciate that it could seem like we're testing the underlying package functionality (that is already well tested no doubt), but we've also written a lot of 1st party code that ought to be tested alongside that.

Testing the actual UI behaviour seemed to be a non-starter because it relies on browser APIs that aren't available in the test environment. I looked into it a bit, expecting to be able to mock them like we do with `ResizeObserver`, but (for reasons I can't remember any more) that didn't seem to be possible. 

But thinking about it now, I should probably be able to write a test that renders the different context providers, manually calls the event handlers and then asserts that the table state has been updated in the expected way. Will do that this afternoon!

lib/src/components/data-table/DataTableRow.tsx
hollg: 🤦🏻 So I guess the way to exclude it would be to have a `DataTable.DragAndDropTable` which is equivalent to `DataTable.Table` except it uses `DraggableRow`?

lib/src/components/avatar/Avatar.tsx
```diff
@@ -74,7 +74,7 @@ export const AvatarRootProvider: React.FC<TAvatarProps> = ({
   name,
   size
 }) => {
-  const value = React.useMemo<TAvatarRootContext>(
+  const value = React.processDragEndEvent<TAvatarRootContext>(
```
LimeWub, line 77: What dis?

hollg: _what_

lib/src/components/data-table/DataTable.test.tsx
```diff
@@ -204,7 +283,7 @@ describe('DataTable.Pagination component', () => {
   })
 })
 
-describe('DataTable Search component', () => {
+describe.only('DataTable Search component', () => {
```
thomasdigby, line 286:
```suggestion
describe('DataTable Search component', () => {
```

PR #445, opened 2023-01-03
Feat: Datatable FE pagination state

Marcdup1:
#### Description

There are two issues that this PR addresses:
- The internal data array is not updated when new data is passed in.
     - If the data is being manipulated outside of the table (such as filtering) the data is passed back into the component. I believe we need to maintain this flexibility.
- Unable to retrieve the pagination state using `useDataTable` from the FE implementation of DataTable
    - Since we provide a managed pagination state, onPaginationChange and a page count to`useReactTable` we need to ensure our `paginationState` is not undefined.

#### Implementation
- Adds a `useEffect` which will update the internal data state when the data prop changes
- Defaults the pagination state for both `getAsyncData` & `dataProp` implementations

### DIFF
--- a/lib/src/components/data-table/DataTable.test.tsx
+++ b/lib/src/components/data-table/DataTable.test.tsx
@@ -63,11 +63,7 @@ describe('DataTable component', () => {
   it('renders', () => {
     const { container } = render(
       <Wrapper>
-        <DataTable
-          columns={columns}
-          data={data}
-          initialState={{ pagination: { pageIndex: 0, pageSize: 5 } }}
-        >
+        <DataTable columns={columns} data={data}>
           <DataTable.GlobalFilter label="User search" css={{ mb: '$4' }} />
           <DataTable.Table sortable css={{ mb: '$4' }} />
           <DataTable.Pagination />
--- a/lib/src/components/data-table/DataTableContext.tsx
+++ b/lib/src/components/data-table/DataTableContext.tsx
@@ -21,7 +21,6 @@ import type {
   PaginationState,
   SortDirection
 } from '@tanstack/react-table'
-import { CSS } from '~/stitches'
 
 import {
   DataTableContextType,
@@ -50,7 +49,6 @@ type TableProviderProps = {
   defaultSort?: { column: string; direction: SortDirection }
   children: React.ReactNode
   initialState?: InitialState
-  css?: CSS
 } & (
   | { data: Array<Record<string, unknown>>; getAsyncData?: never }
   | { data?: never; getAsyncData: TGetAsyncData }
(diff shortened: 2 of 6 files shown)


D7Torres (approved):

lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -146,6 +139,12 @@ export const DataTableProvider = ({
     runAsyncData({})
   }, [runAsyncData])
 
+  React.useEffect(() => {
+    if (!dataProp) return
+
+    setData({ results: dataProp, total: dataProp.length })
+  }, [dataProp])
```
thomasdigby, line 146: Do we need to deep compare this `dataProp` value? https://github.com/kentcdodds/use-deep-compare-effect

I'm not certain that we do, but if `data` reference changes even if the values stay the same, then we probably don't want to rerender.

lib/src/components/data-table/DataTable.test.tsx
```diff
@@ -63,11 +63,7 @@ describe('DataTable component', () => {
   it('renders', () => {
     const { container } = render(
       <Wrapper>
-        <DataTable
```
thomasdigby, line 66: What's the reason for changing this test? Are we checking that the default pagination state is applied in this example?

lib/src/components/data-table/DataTableContext.tsx
hollg: I agree. We either need to deep compare here, or clearly instruct consumers to memoize the value they pass into this component. The second option seems like unneeded consumer overhead to me so I'd go for the deep compare.

I think we should probably also specifically highlight this change in the docs though, since updating the value of the `data` prop will now reset the table state in a way that people might not expect.

Marcdup1: Good call! thank you, I added that hook now and amended the docs

lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -63,12 +62,11 @@ const DataTableContext =
 
 export const DataTableProvider = ({
   columns,
-  data: dataProp,
+  data: dataProp = [],
```
Marcdup1, line 65: I had to do this because of:

https://github.com/kentcdodds/use-deep-compare-effect/blob/ec3c204ae3a29be7db8e30fc7ca29dd38d1c9441/src/index.ts#L17-L22

the alternative would be 

```tsx
useDeepCompareEffect(() => {
    if (!dataProp) return

    setData({ results: dataProp, total: dataProp.length })
  }, [dataProp || []]) // 🤷‍♂️
```

`dataProp` will be null if we are using `getAsyncData`

thomasdigby (approved):

PR #448, opened 2023-01-11
feat(data-table): adding support to sticky column

viniciusvolpe:
As an effort to bring the data table up to feature parity with the old enhanced table I'm adding the support for a single sticky column in the data table.

### Implementation
- created `Table.StickyColumnsContainer` to wrap the table and add the scroll bar
- changed the `Table.Header` and `Table.Body` to add the sticky CSS props when the number of sticky columns is 1
- created `DataTable.StickyColumnsContainer` to wrap the data table and set the number of sticky columns in the context
- added tests

### Screenshots

https://user-images.githubusercontent.com/11462265/211869945-56f900de-378a-418a-bfa8-b667e10cb888.mov


lib/src/components/data-table/DataTableStickyColumnsContainer.tsx
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export const DataTableStickyColumnsContainer: React.FC<
+  IDataTableStickyColumnsContainerProps
+> = ({ children, css }) => {
+  const { numberOfStickyColumns, setNumberOfStickyColumns } = useDataTable()
+
+  React.useEffect(() => {
+    // Temporarily hardcoded as 1 and it will be a prop once we add support to multiple sticky columns
+    setNumberOfStickyColumns(1)
```
viniciusvolpe, line 20: preparing for the multiple sticky columns support

lib/src/components/table/TableBody.tsx
```diff
@@ -14,17 +14,41 @@ const StyledTableBody = styled('tbody', {
         }
       },
       false: {
-        bg: 'white'
+        [`${TableRow}`]: {
+          bg: 'white'
+        }
```
viniciusvolpe, line 19: Changing to set the table row bg as white instead of the table body as it was, this will make setting the sticky column bg easier

lib/src/components/table/TableBody.tsx
```diff
@@ -14,17 +14,41 @@ const StyledTableBody = styled('tbody', {
...
+    <StyledTableBody
+      striped={striped}
+      css={{
+        ...(numberOfStickyColumns === 1 && {
+          [`& td:nth-of-type(${numberOfStickyColumns})`]: {
+            bg: 'inherit',
+            position: 'sticky',
+            left: '0',
+            zIndex: '2'
+          }
```
viniciusvolpe, line 45: `numberOfStickyColumns === 1` will be changed when adding multiple sticky columns

lib/src/components/table/TableStickyColumnsContainer.tsx
```diff
@@ -0,0 +1,41 @@
...
+    <Box
+      onScroll={handleScroll}
+      role="scrollbar"
+      css={{
+        overflow: 'auto',
+        ...(isScrolling && {
+          [`& th,td:nth-of-type(${numberOfStickyColumns})`]: {
+            boxShadow: '$colors$alpha200 -2px -3px 9px 1px',
+            clipPath: 'inset(0px -10px 0px 0px)'
+          }
```
viniciusvolpe, line 32: Since this component controls the scroll bar I had to add the shadows here by selecting the last sticky column (either th or td)

lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -102,6 +102,9 @@ export const DataTableProvider = ({
       : []
   )
 
+  const [numberOfStickyColumns, setNumberOfStickyColumns] =
```
hollg, line 105: I'm not totally sure how I feel about this, so I'm just thinking out loud here really:

I had pictured pretty much all the code for sticky columns living within `Table` rather than `DataTable`, since it's only relevant to display rather than the data itself. What would you say we gain by storing this state here? It would allow us to change the number of sticky columns from the nested `DataTable` components, but I'm not sure I can see a use case for that. Is that something your pod has a need for?

viniciusvolpe: My intention was to pass the number of sticky columns to one component only, the structure to use the data table with sticky columns would be:
```
    <DataTable.StickyColumnsContainer numberOfStickyColumns={2}>
      <DataTable.Table />
    </DataTable.StickyColumnsContainer>
</DataTable>
```

Then I set the `numberOfStickyColumns` in the context to get it in the `DataTable.Table` component.

The alternative would be not to store the value in the context and pass it to both `DataTable.StickyColumnsContainer` and `DataTable.Table` but that would be weird since the users of the data table would need to know they need to send the prop to both components.

We can have a different approach if you think it's better, the structure would be: 
```
    <DataTable.Table numberOfStickyColumns={2} />
</DataTable>
```
and inside the `DataTable.Table` add a check if `numberOfStickyColumns` is greater than zero we add the sticky columns container to the table and pass the value down to both `Table` and `StickyColumnsContainer` components, not needing to store in the context anymore, how do you feel about that?

hollg: > We can have a different approach if you think it's better, the structure would be ...

I really like that API. Anything we can do to keep it simple from the perspective of the consumer is a plus and that approach makes it super straightforward and intuitive imo. Also a (tiny) performance improvement in avoiding setting state etc.

But I still think the decision about whether to render a `StickyColumnsContainer` should happen inside `Table` itself, rather than `DataTable`, because it's purely presentational rather than relating to data or any interactive functionality. So the way I'd see it working would be to have `DataTable.Table` pass its `numberOfStickyColumns` prop straight down to `Table`, which would then decide whether to render the `StickyColumnsContainer` _and_ a `table` or just the latter.

Does that make sense?

viniciusvolpe: hey @hollg I made the changes we discussed in this thread, let me know your thoughts, my only concern is now if we are using just the table components (without the data table) we will need to pass the number of sticky columns to 3 different components, Table, Table.Head and Table.Body, is it a problem? I guess it can be if different numbers are passed to each component, not sure how we can prevent that

hollg (approved):

lib/src/components/data-table/DataTableTable.tsx
```diff
@@ -30,6 +31,7 @@ export const DataTableTable: React.FC<DataTableTableProps> = ({
       <DataTableLoading />
       <Table
         {...props}
+        numOfStickyColumns={numOfStickyColumns}
```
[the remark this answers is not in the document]
LimeWub: I think it can even be `numberOfStickyColumns` and it will read most clearly :)
It's not like we need to keep variable names short for any reason and why spend the processing power of the brain of whoever reads it to decipher ;p 

(I don't have a problem with what this is named as rn and prefer it in comparison to this suggestion as it's clearer)

lib/src/components/table/TableBody.tsx
```diff
@@ -14,17 +14,41 @@ const StyledTableBody = styled('tbody', {
...
+  numOfStickyColumns = 0,
+  css,
   ...rest
-}) => <StyledTableBody striped={striped} {...rest} />
+}) => {
+  return (
+    <StyledTableBody
+      striped={striped}
+      css={{
+        ...(numOfStickyColumns === 1 && {
```
LimeWub, line 39: since you're already targetting the td from the container why not do all the relevant css from there and leave `TableBody` and `TableHeader` agnostic as to the numberOfColumns effects?

lib/src/components/table/Table.tsx
```diff
@@ -55,13 +57,26 @@ const StyledTable = styled('table', {
...
 export const Table: React.FC<TableProps> & TableSubComponents = ({
   size = 'md',
   corners = 'round',
+  numOfStickyColumns = 0,
   ...rest
-}: TableProps) => <StyledTable size={size} corners={corners} {...rest} />
+}: TableProps) => {
+  const tableComponent = <StyledTable size={size} corners={corners} {...rest} />
+
+  return numOfStickyColumns ? (
```
LimeWub, line 72: minor but can this be written as an early return rather than ternary. I think early return is a bit more readable (opinion)

```
if (numOfStickyColumns) return (
<TableStickyColumnsContainer numOfStickyColumns={numOfStickyColumns}>
      {tableComponent}
</TableStickyColumnsContainer>
)

return tableComponent
```

Or like this which will also make it easier to stack multiple components in the future if we choose to, ie.

```
let tableComponent = <StyledTable size={size} corners={corners} {...rest} />
if (numOfStickyColumns){
tableComponent = (
<TableStickyColumnsContainer numOfStickyColumns={numOfStickyColumns}>
      {tableComponent}
</TableStickyColumnsContainer>
)
}
/* Maybe in future:
if (scrollable){
tableComponent = (
<TableScrollContainer>
      {tableComponent}
</TableScrollContainer>
)
}
*/

return tableComponent
```

(assuming we do need multiple nested containers at some point :eyes:)

This is not blocking but is an easy change and I think potentially good food for thought for the future :)

lib/src/components/table/TableBody.tsx
LimeWub: This one is the comment I'm most interested in addressing in this PR - the other comments have been added as opinions or to start a discussion about them.

lib/src/components/data-table/DataTable.test.tsx
```diff
@@ -673,3 +674,31 @@ describe('DataTable server-side', () => {
...
+        <DataTable columns={columns} data={data}>
+          <DataTable.Table numOfStickyColumns={1} />
+        </DataTable>
+      </Wrapper>
+    )
+
+    fireEvent.scroll(screen.getByRole('scrollbar'), {
+      target: { scrollY: 100 }
+    })
+    expect(await screen.findByText('name')).toBeVisible()
```
LimeWub, line 702: Could you clarify what this is testing? Is name not visible before we scroll? If it isn't visible could we assert that in this test as well so it's clearer?

LimeWub: I'm happy with the test being removed if it's not actually testing anything haha :)
I don't know how I would test this either tbh

PR #450, opened 2023-01-16
feat: split Sortable out of DataTable into a reusable primitive

LimeWub:
Fully functional split of Sortable primitive from DataTable.
Written in a way to support implementation in other components - such as core's Sortable.

## Video
(Still working use in DataTable + 2 custom implementations)

https://user-images.githubusercontent.com/6905473/212680877-970c26e7-63f2-4012-b2e9-22d4e957280c.mov

PR #451, opened 2023-01-16
Fix: read DataTable custom idColumn correctly

hollg:
Fixes a problem where custom `idColumn` values didn't work correctly.

My original drag-and-drop row implementation had a bug where `DraggableRow` assumed the consumer was providing IDs for each row in an `id` property and `DragAndDropContainer` tried to read the id `@dnd-kit/sortable` provided dynamically using the `idColumn` value. This was the wrong way round!

Instead, `DraggableRow` now uses the provided `idColumn` find the value to provide as `@dnd-kit/sortable`'s `id` and `DragAndDropContainer` correctly reads the `id` from the objects that come out of `@dnd-kit/sortable`.

### DIFF
--- a/lib/src/components/data-table/drag-and-drop/DragAndDropContainer.tsx
+++ b/lib/src/components/data-table/drag-and-drop/DragAndDropContainer.tsx
@@ -18,6 +18,20 @@ import {
 } from '@dnd-kit/sortable'
 import * as React from 'react'
 
+const DragAndDropTableContext = React.createContext<{ idColumn: string }>({
+  idColumn: 'id'
+})
+export const useDragAndDropTable = () => {
+  const context = React.useContext(DragAndDropTableContext)
+
+  if (!context)
+    throw new Error(
+      'useDragAndDropTable can only be called within a DragAndDropContainer'
+    )
+
+  return context
+}
+
 type DragAndDropContainerProps = {
   idColumn?: string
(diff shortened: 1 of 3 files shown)


LimeWub (approved):

lib/src/components/data-table/drag-and-drop/DragAndDropContainer.tsx
```diff
@@ -33,8 +47,8 @@ export const processDragEndEvent = (
   onChange?: (oldIndex: number, newIndex: number, newData: TableData) => void
 ) => {
   const { active, over } = event
-  const oldIndex = rowOrder.indexOf(active[idColumn])
```
LimeWub, line 36: Hey we may need to chat to make sure this still gets implemented in this bit of the refactor: https://github.com/Atom-Learning/components/pull/450/files#diff-a16b309577ec69cc7245526f317f674ef2acecd36f3ef15f08b62037d3ef0669R44

as I'm not sure I fully understand it.

But I trust it does fix the issue - hence the approve :)

LimeWub: It might not even need to be implemented and I may have fixed it by accident 🤷

hollg: What you have in `SortableRoot` looks right to me!

D7Torres (approved):

PR #452, opened 2023-01-17
Fix: broken list theme variant

andoulla:
This PR re-introduces the theme variant for the list component, as we need it for the DS migration of the list https://atomlearningltd.atlassian.net/browse/FG-197. I have only added the primary variant since this is what we need only now.

image

https://atomlearningltd.atlassian.net/jira/software/projects/DS/boards/27?selectedIssue=DS-283

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ In the future, you only need to rerun `yarn build:docs` after you either delete 
 
 ### Running the sandbox
 
-The library offers sandbox capabilities to ease the development process. This run with the following command
+The library offers sandbox capabilities to ease the development process. This runs with the following command
 
 `yarn start:sandbox`
 
--- a/README.md
+++ b/README.md
@@ -31,11 +31,17 @@ You can read more about the components included in `@atom-learning/components` a
 
 We can run the documentation site locally, with hot-reloading triggered by changes in the library. This offers a low-friction dev environment where we can see the effects of any local changes when applied to all existing components. (Note: hot reloading is currently only supported for changes to Typescript/.tsx files, not markdown files.)
 
-First, run `yarn build:docs && yarn dev:lib`. `yarn build:docs` collects the markdown documentation files for each component in `lib` into its `dist` directory, where the documentation site can find them. `yarn dev:lib` compiles the React component library and watches for changes.
+First, run `yarn build:lib && yarn build:docs`. `yarn build:docs` collects the markdown documentation files for each component in `lib` into its `dist` directory, where the documentation site can find them. `yarn build:docs` compiles the documentation and extract the component props.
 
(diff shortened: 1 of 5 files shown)


viniciusvolpe (approved):

README.md
```diff
@@ -31,12 +31,18 @@ You can read more about the components included in `@atom-learning/components` a
 
 We can run the documentation site locally, with hot-reloading triggered by changes in the library. This offers a low-friction dev environment where we can see the effects of any local changes when applied to all existing components. (Note: hot reloading is currently only supported for changes to Typescript/.tsx files, not markdown files.)
 
-First, run `yarn build:docs && yarn dev:lib`. `yarn build:docs` collects the markdown documentation files for each component in `lib` into its `dist` directory, where the documentation site can find them. `yarn dev:lib` compiles the React component library and watches for changes.
+First, run `yarn build:lib && yarn build:docs`. `yarn build:docs` collects the markdown documentation files for each component in `lib` into its `dist` directory, where the documentation site can find them. `yarn build:docs` compiles the documentation and extract the component props.
```
LimeWub, line 34: Hey I'm pretty sure `lib` compiles and extracts component props.
`docs` just builds the site (nextJs) to view the lib components in the format they are on on the public website.

also `dev:lib` is important if you're developing.

For what it's worth [this will all hopefully change soon](https://github.com/Atom-Learning/components/tree/feat/site-cms-content#readme) so if we're not certain about this documentation better to leave it as it was - otherwise we'll just be confusing ourselves.

andoulla: oh I was trying to infer from the command without looking at the scripts. if you are ok with not updating then I will just leave as is

LimeWub: Ye lets leave as is for now. I think flipping the lib and docs commands around (like you've done) will probably be okay but not 100% certain so maybe just leave...
Or test with deleting the dist folders and see if lib then docs works fine 🤷

README.md
```diff
@@ -31,12 +31,18 @@ You can read more about the components included in `@atom-learning/components` a
...
 
 In another terminal, run `yarn dev:site` to run the documentation site at `http://localhost:3000`, taking the output as the previous commands as its input.
 
 In the future, you only need to rerun `yarn build:docs` after you either delete the content of `lib/dist` or you add/update some markdown files inside `lib`.
 
+### Running the sandbox
+
+The library offers sandbox capabilities to ease the development process. This run with the following command
```
alexandrunemes, line 42:
```suggestion
The library offers sandbox capabilities to ease the development process. This runs with the following command
```

alexandrunemes (approved):
One minor typo (at least I *think* it's a typo).

PR #453, opened 2023-01-24
fix: remove isRounded button variant

andoulla:
We decided to stop using rounded variants for buttons a while ago, but they were never removed from the DS. 
this PR will remove the option from the button component
https://atomlearningltd.atlassian.net/browse/DS-284

image


thomasdigby (approved):
🙏🎉

hollg (approved):

LimeWub (approved):

PR #454, opened 2023-01-24
Remove wildcard Stitches imports

Mhoog:
Removed any wildcard imports from `@stitches/react`. This fix prevents the following TS error in our main repo:

```
Exported variable '...' has or is using name 'Stitches' from external module "..." but cannot be named.
```

More info:
- https://github.com/stitchesjs/stitches/issues/1014
- https://github.com/microsoft/TypeScript/issues/5711#issuecomment-157793294


lib/src/components/number-input/__snapshots__/NumberInput.test.tsx.snap
```diff
@@ -6,6 +6,10 @@ exports[`NumberInput component renders number input  1`] = `
     display: flex;
   }
 
+  .c-jVLVTy {
```
andoulla, line 9: I find these new z-indexes suspicious, as we have been burned by them before. Lets remember to test in atom-core once this is updated

andoulla (approved):

lib/src/components/number-input/__snapshots__/NumberInput.test.tsx.snap
Mhoog: The `z-index`es were there before, they're only re-ordered in this file.

PR #455, opened 2023-01-25
Simplify the scrolling tab list logic

LimeWub:
### Description
Simplifies (and fixes) the scrolling tab list logic.
This was originally done as part of the colour scheme update but has been moved to its own PR as per request.

### Videos

**before:**

https://user-images.githubusercontent.com/6905473/214551640-82d405d8-87b9-4d00-913f-9668aa953e13.mov

**after:**

https://user-images.githubusercontent.com/6905473/214550531-0110c909-6254-4547-9690-ca339ed8bd04.mov


lib/src/components/tabs/TabsTriggerList.tsx
```diff
@@ -35,76 +36,42 @@ const StyledChevronActionIcon = styled(ActionIcon, {
...
-const SCROLL_PERCENTAGE = 10
 export const TabsTriggerList: React.FC<
   React.ComponentProps<typeof StyledTriggerList> & {
     colorScheme?: typeof ColorScheme
   }
 > = ({ children, colorScheme = {}, ...rest }) => {
   const [listRef, setListRefCallback] = useCallbackRefState()
 
-  const { width: screenWidth } = useWindowSize()
+  const { width } = useWindowSize()
```
Mhoog, line 46: Instead of triggering on a window resize, it would be good to trigger when the size of the ref element changes. That way we also catch cases where the list changes size because of interactions in the app.

Bonus is that you wouldn't have to 'lie' about `width` being a dependency in the `canScrollXAxis` `useMemo`.

`ResizeObserver` could be a good way to wait for size changes.

LimeWub: Hey it's no longer as simple but you're right so I switched it over to useResizeObserver for the size :)

lib/src/utilities/hooks/useResizeObserver.ts
```diff
@@ -0,0 +1,39 @@
...
+  } catch (error) {
+    return null
+  }
+}
+
+export const useResizeObserver = ({
+  delay = 500,
+  elements,
+  onResize
+}: {
```
hollg, line 16: This is minor but could you extract the type for this object into a named type? Once there's more than one or two values, I think it makes things much easier to read

lib/src/utilities/hooks/useScrollPosition.ts
```diff
@@ -0,0 +1,31 @@
+import { useEffect, useState } from 'react'
+import { debounce } from 'throttle-debounce'
+
+export const useScrollPosition = ({
+  element,
+  delay = 500
+}: {
```
hollg, line 7: Also here, there's a series of very similar-looking lines which all do slightly different things and I think we could tidy it up by extracting types for the input and output of the hook

hollg:
Couple of comments from me but nothing blocking. Happy to approve once the conflict is resolved 👍🏻

lib/src/utilities/hooks/useScrollPosition.ts
LimeWub: addressed

lib/src/utilities/hooks/useResizeObserver.ts
LimeWub: addressed

hollg (approved):

D7Torres:
in the PR description: I can see what this is about with the videos, so no need for a description of the problem, but some implementation details would be useful.

lib/src/utilities/hooks/useResizeObserver.ts
```diff
@@ -0,0 +1,43 @@
...
+  onResize: (entries: ResizeObserverEntry[]) => void
+  delay?: number
+}
+
+type TUseResizeObserverOutput = ResizeObserver | null
+
+const createResizeObserver = (callback: () => void) => {
+  try {
+    return new ResizeObserver(callback)
+  } catch (error) {
```
D7Torres, line 15: super minor: you can remove `error` if you are not going to use it.

Question: do we have as a rule not to log any error in the Design System?

lib/src/utilities/hooks/useResizeObserver.ts
```diff
@@ -0,0 +1,43 @@
...
+  React.useEffect(() => {
+    elements.forEach((element) => {
+      if (element) {
+        observer?.observe(element)
+      }
+    })
+
+    return () => {
+      elements.forEach((element) => element && observer?.unobserve(element))
+    }
```
D7Torres, line 39: I'm sure it's my fault but I don't get it.

Are we observing and immediately unobserving?

D7Torres: Never mind, I got it: it returns a callback that it will be called when unmounting or something.

lib/src/components/tabs/TabsTriggerList.tsx
```diff
@@ -113,29 +80,29 @@ export const TabsTriggerList: React.FC<
...
-      {showLeftScroller && (
+      {canScrollLeft && (
         <StyledChevronActionIcon
           label="scroll left"
           size="md"
           css={{
             left: 0
           }}
-          onClick={() => scrollTriggerListTo('left')}
+          onClick={() => scrollList(-0.8)}
```
D7Torres, line 90: can we store the magic number 0.8 in a const?

(I would store 0.8 only and then use `SCROLL_STEP` or `-SCROLL_STEP` rather than using 2 variables)

D7Torres (approved):
Approved but some comments, could you do at least the one about the constant?

lib/src/utilities/hooks/useResizeObserver.ts
andoulla: I think we need to throw the error using invariant

lib/src/utilities/hooks/useResizeObserver.ts
LimeWub: yup

lib/src/utilities/hooks/useResizeObserver.ts
LimeWub: I'm pretty sure this is copy/pasta from the ResizeObserver in core. Can you clarify what, if anything, you think should be changed here?

Screenshot 2023-02-15 at 11 14 04

D7Torres:
```
catch (error) {}
```
or
```
catch (error) {
 return null
}
```

a.k.a. silencing the exception, is not considered a good practice. It's better to do something with the `error`
In core, it's being logged with Sentry.

If in `components` we don't log things, what I was suggesting is removing the `error` variable from the `catch()`

I'm not sure we want to make the component using this error if creating the ResizeObserver fails 🤔 , so I don't know if I agree with throwing the error using invariant @andoulla. But if you wanted to throw the error, why capture it in the first place? you could just remove the try-catch

PR #456, opened 2023-01-25
feat: do not open NavigationMenuDropdown via pointer events

LimeWub:
Jira: https://atomlearningltd.atlassian.net/browse/DS-165

Requirement: Have the navigation menu dropdown open only on click and not on hover.

Solution: https://github.com/radix-ui/primitives/issues/1630

### Videos

https://user-images.githubusercontent.com/6905473/214565843-ca1e4113-5b36-43df-a7d9-978fdaa572d2.mov

### DIFF
--- a/lib/src/components/navigation/NavigationMenuDropdownContent.tsx
+++ b/lib/src/components/navigation/NavigationMenuDropdownContent.tsx
@@ -1,5 +1,6 @@
 import * as NavigationMenuPrimitive from '@radix-ui/react-navigation-menu'
 import React from 'react'
+import { preventHover } from './preventHover'
 
 import { styled } from '~/stitches'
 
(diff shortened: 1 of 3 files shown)


hollg (approved):
Wow, you weren't kidding about it being a simple solution! LGTM, let's just get @MiguelGDLR to confirm that video is showing the expected behaviour / whether there are any other edge cases he knows of that we ought to check too.

MiguelGDLR:
> Wow, you weren't kidding about it being a simple solution! LGTM, let's just get @MiguelGDLR to confirm that video is showing the expected behaviour / whether there are any other edge cases he knows of that we ought to check too.

looks good to me, only edge case I can think of its make sure it work with tap? which I assume it works

LimeWub:
@MiguelGDLR I haven't tested it on mobile/touch but the events it blocks is `PointerEnter` and `PointerLeave` which wouldn't block tapping afaik (that would be `PointerDown` or `Click`)

D7Torres (approved):

PR #457, opened 2023-01-25
refactor: rename useScrollPosition to useWindowScrollPosition so it l…

LimeWub:
…eaves space for element useScrollPosition and replace throttle with debounce which is the correct function to use

**Reason for rename:** [I want a `useScrollPosition` hooks which works with elements and not just window](https://github.com/Atom-Learning/components/pull/455/files#diff-16f0a7cc24d6832829eb9761b9ff87ade094cdd5d1efcc88a29fbe2553080053)
**Reason for changing throttle to debounce:** Throttle runs a billion times at the end of the delay while debounce runs once. I assume this is a bug (**Edit**: I assumed wrong) and we've always wanted debounce but @elliotlaws feel free to correct me if you did mean it to be throttle for some reason.

**Edit!:** I've reverted the change from throttle to debounce and currently trying to allow for it to be overwritten to be a debounce if needed. Throttle was indeed the correct use case for the NavigationMenu! However, throttle runs too many times for any more complex code (like what I needed to do with the scrolling calculations in Tabs) which is why I'm a bit hesitant to leave as throttle only.


hollg:
Thanks for pulling this out into its own PR. Could you add something to the description to show the before/after behaviour (which I presume, from the user's perspective, should be identical)?

LimeWub:
@hollg FYI I'm running the sandbox to make a video on this and noticing why throttle was used rather than debounce now. It calls many more times than debounce but the shadow is appearing on the header is smoother. I will revert the change to throttle.

I'm considering whether having the debounce method as a prop would be beneficial because in my use case with the arrows (regular scroll in element) I used debounce for more performance as it doesn't matter but in the case of a topbar throttle is correct. But now the 2 effects are using a different delay method and it doesn't feel right...

LimeWub:
**throttle**
https://user-images.githubusercontent.com/6905473/214799681-13286e3a-670f-4ff5-8d2c-aa70982b0b35.mov


**debounce**
https://user-images.githubusercontent.com/6905473/214799654-e91b4b85-84eb-45a2-98a1-4da3afad1730.mov

LimeWub:
What do you think about this? https://github.com/Atom-Learning/components/pull/457/commits/fb544c99d358ea9f4d0873447a0c788fa9736783

lib/src/utilities/hooks/useWindowScrollPosition.ts
```diff
@@ -1,12 +1,14 @@
 import { useEffect, useState } from 'react'
-import { throttle } from 'throttle-debounce'
+import { debounce, throttle } from 'throttle-debounce'
 
-export const useScrollPosition = (): { x: number; y: number } => {
+export const useWindowScrollPosition = ({ delay = 500, delayMethod = throttle }: { delay?: number, delayMethod?: throttle | debounce } = {}): {
```
hollg, line 4: I'm into this enum prop — way better than the boolean alternative imo where you default to one but switch to the other if a truthy value is passed. But does this work, with the direct reference to the functions from the lib? I didn't think TS would be able to recognise that an import of `debounce` in the file _calling_ this hook would be the same `debounce` mentioned in this type.

If it doesn't, we could always have an enum of strings `'throttle' | 'debounce'` and and map them to the functions here.

LimeWub: Ye I don't know if it works with references outside the lib cause I haven't tested it but I'm very happy to go ahead and change it to the string enum. I also think it's probably cleaner with the strings like u suggested and the user wouldn't need to import any extra functions to pass in either.

LimeWub: @hollg this has now been addressed - lmk if it's better

hollg: 👍🏻 looks good to me! I'd suggest defining a type for that object separately though so that the syntax of the function definition itself is a bit easier to read though (I actually thought I had commented this already, but I either totally imagined that or was thinking of another review all together!)

LimeWub: haha ye you commented that on another PR and I was already on it (was doing it for consistency with the other PR) - so that's also done in the latest commit :p

D7Torres (approved):

PR #458, opened 2023-01-26
DS-286: show by default a tooltip for ActionIcon

andoulla:
**Description**
Add tooltip functionality in action icons (a boolean prop) and show the tooltip by default. 

**Implementation**
- Added the tooltip conditionally and added the ability to hide the tooltip
- updated all the tests that use the ActionIcon to have a tooltip. provider 

To release this change I will need to do an extra migration PR into core where I update the components that use the action icon to import it from here instead of the `_design-system` folder (about 50 instances in 23 files) and also update their tests


lib/src/components/action-icon/ActionIcon.tsx
```diff
@@ -195,12 +196,24 @@ const StyledButton = styled('button', {
   ]
 })
 
+const ConditionallyWrapWithTooltip = ({ hideTooltip, label, children }) => {
```
LimeWub, line 199: Instead of false conditionals can we have `showTooltip` with a default being false?

andoulla: The default would be to show the tooltip not hide it. I had it like that until I had to use it to hide some of the tooltips and I believe that 
`<ActionIcon hideTooltip><Icon/></ActionIcon>` reads better than `<ActionIcon showTooltip={false}><Icon/></ActionIcon>`

LimeWub:
My 2c (which you probably already know cause I've commented on that b4 but here I go again):
I'm concerned this change will be more trouble than it's worth.
We are saving a repeating pattern which adds 4 lines (the tooltip wrapping) but now needs extra nesting and a check on every ActionIcon.
On one hand I get that action icons themselves don't always convey the meaning of the action so I get why a tooltip is good. (A problem with action Icons always, anyway - so perhaps we should think about when we use icons vs small buttons in design)

But I'm also wondering what happens when we want to show different labels around ActionIcons when they are disabled? How will that work - do we just nest Tooltips over Tooltips? Does this still work?

andoulla:
So there are a few points here:
> My 2c (which you probably already know cause I've commented on that b4 but here I go again): I'm concerned this change will be more trouble than it's worth. We are saving a repeating pattern which adds 4 lines (the tooltip wrapping) but now needs extra nesting and a check on every ActionIcon. On one hand I get that action icons themselves don't always convey the meaning of the action so I get why a tooltip is good.

- It isn't that much effort to add the tooltip or do the migration now, however waiting will only make it harder so I think this is a good thing to have here. Now that it has a tiny bit more nesting and an extra check, I don't think it has a big enough impact on readability or performance to be a concern if you weigh it against the fact that this is going to improve the usability of the component everywhere in one go

> (A problem with action Icons always, anyway - so perhaps we should think about when we use icons vs small buttons in design)
After talking to Miguel I do believe that we shouldn't let people use the button component with just an icon, as this is why we have actionIcons. This move is step one to getting there


> But I'm also wondering what happens when we want to show different labels around ActionIcons when they are disabled? How will that work - do we just nest Tooltips over Tooltips? Does this still work?

Can't the parent just pass a different label to the component if it is disabled? I wouldn't wrap into extra Tooltips. 


Something else to say, this also addresses 2 more problems:
a) we keep having to tell people to use the TooltipActionIcon because they don't know it is there
b) allows us to remove the TooltipActionIcon completely

lib/src/components/action-icon/ActionIcon.tsx
LimeWub: I don't think I've ever seen a regular HTML element which has a negative attribute before which is why I don't like having negative attributes in our components. 
However, yes `showTooltip={false}`  isn't preferable, which is why I suggested it defaults to false.
But yes, then it doesn't default to `true` - so people will need to use it when they need it, so there will be a learning curve.
But is that really bad?

It really all comes back to the same thing of Tooltips being obfuscated and combined into ActionIcon in a way I'm not convinced about.
On the other hand - I do see how it will remove multiple lines of code from atom-core and developers wont even need to think about it so 🤷 .

LimeWub:
@andoulla 
I think I'm getting tripped up even from the notion that each and every ActionIcon is to have a Tooltip rather than just adding them in cases where necessary only. (ie. if something has been disabled and we had no other space to add the message to -debatable as well-).

When I said 'use small buttons' above, I didn't mean using small buttons with just an icon but just using text (+ icon?) if the action icon makes no sense visually - and leads us to needing tooltips.

As for passing different labels when there's an error or it's disabled - it just feels wrong. The icon's label is about the one thing the action is supposed to be doing but we'd want to say why we can't do it on hover as an extra tip. The button doesn't get to have another label. (I dunno honestly maybe it makes sense, I've just never seen this done before, we don't change labels on things which are disabled)

I'd say this isn't the PR for me to review honestly as I can't get behind the premise even though I have the understanding as to why it's being done. 
My solution would just be to pester/educate people in PRs to use the correct component and nest Tooltips **only** if needed and only in the ActionIcons that need it but what we seem to be going with is to just use either way. Yes it's simpler but I personally can't get over this mixing of functionality and overuse of Tooltips on every icon.

lib/src/components/action-icon/ActionIcon.tsx
andoulla: So you have concerns on the way I have combined the 2? it would help if you could share what way you think it is better

LimeWub: Hey, I'll have another look at this tomorrow and have a think

hollg: The component we expose to the consumer (`ActionIcon`, not `ConditionallyWrapWithTooltip`) does default the bool to `false`, right?

I share @LimeWub's preference for false defaults on boolean props, and  while I think I have a slight preference for `showTooltip` over `hideTooltip`, I think this strikes a good balance. If the design decision is that tooltips should be the rule rather than the exception, facilitating that feels more important than worrying about whether `hide` or `show` feels more natural. (I also think that design decision is a more important factor than saving 4 lines of code anyway.)

I guess the "correct" way to do this in terms of component composition is probably to have the `TooltipActionIcon` from `atom-core` in here as a separate component, and try to make using that, rather than using `ActionIcon` our default behaviour. But for me personally I can't see a benefit to that which outweighs the faff.

kerryambt: if we definitely want to use `showTooltip` could you not set a default value for it and instead set it as an optional boolean, then in the component have something like `const shouldShowTooltip = showTooltip !== false` which would pick up `true` and `undefined`.

I do agree with Gary though I think there's a balance to be had, and that 'hideX` isn't abhorrent by default

andoulla: ok so we have 2 open issues to address:

- Should the Tooltip be part of the ActionIcon or have 2 components? 
Where I sit with this is that we have tried for 18 months now, and having two components isn't working, as people will use the wrong one. I feel that the separation isn't worth the pestering effort. 

I think that if we want all of the ActionIcons to show the tooltip (which is something will reconfirm with Miguel) then why offer the plain one and create the opportunity for misuse? We could have 2 of them and only expose for usage the one with the tooltip externally and use the one without internally for places like the Inputs etc, but again that will only depend if we don't want to show the tooltip there

andoulla: @kerryambt, sorry I kind of feel that `const shouldShowTooltip = showTooltip !== false` is more complex than just setting the default to true here

kerryambt:
few thoughts here: 

- do we definitely want these on all the time? are there any performance issues, or do our tooltips only make one portal per provider and simply pass the current tooltip up to that provider? we should maybe double check there's nothing noticeable with profiler before allowing this change.

- if we _do_ have them on by default, do we need to check to see if the label has been passed to the component? I know it should be there because errors but we probably have some ActionIcons in JS files or just errors that are being ignored so it's at least worth having another safeguard against an empty tooltip.

lib/src/components/action-icon/ActionIcon.tsx
MiguelGDLR: I find it rare to think of an icon or icon button that should not have a tooltip to explain its purpose, I think we should make it the default and consistent use

andoulla: Thanks for the feedback @MiguelGDLR,

@LimeWub @hollg @kerryambt Based on this, what would be the preferred solution now? Should I have two components and only expose the Tooltip wrapped one for external usage? or just have one

lib/src/components/action-icon/ActionIcon.test.tsx
```diff
@@ -4,52 +4,81 @@ import * as React from 'react'
 
 import { expectToThrow } from '../../../test/custom-assertions/expect-to-throw'
 import { Icon } from '../icon/Icon'
+import { Tooltip } from '../tooltip'
```
D7Torres, line 7: Are we missing tests for `showTooltip` = `true` vs `false?

lib/src/components/number-input/NumberInputStepper.tsx
```diff
@@ -57,6 +57,7 @@ export const NumberInputStepper: React.ForwardRefExoticComponent<NumberInputStep
               tabIndex={-1}
               appearance="outline"
               ref={forwardedRef}
+              hideTooltip
```
D7Torres, line 60: Is it weird that this didn't break any test?

D7Torres (changes requested):
Nice PR! It's exciting to see how much impact is going to cause across all parts of atom-core once is released.

Some comments, only the two about potentially missing tests are real request changes.

lib/src/components/action-icon/ActionIcon.tsx
andoulla: I'm going to be picking this up next week, but I'm thinking of going with `showTooltip` and then default it to true. I think this will address some of the naming concerns and keep the behaviour close to the real life need.

D7Torres: Regarding the name, I won't push it, but it doesn't address the naming concern.

`showTooltip` starts with a verb, which is usually used for functions. `hasTooltip` it starts with a word that says "this is a boolean"

PR #459, opened 2023-01-26
chore: contains only linting changes

andoulla:
These are all the linting changes that prettier and EsLint have auto fixed.


lib/src/components/data-table/__snapshots__/DataTable.test.tsx.snap
```diff
@@ -174,54 +174,6 @@ exports[`DataTable component Renders drag handles for draggable table rows 1`] =
     transition: all 100ms ease-out;
```
LimeWub, line 174: If it's just linting changes why has it affected CSS and snapshots? 👀

andoulla: no worries, I had a mistake when cherry picking and I have now fixed it in both PRs

LimeWub (approved):

lib/src/components/date-input/DateInput.tsx
```diff
@@ -76,6 +77,7 @@ export const DateInput = React.forwardRef<HTMLInputElement, DateInputProps>(
               label={updatedLabels.open}
               size={size}
               theme="neutral"
+              hideTooltip
```
hollg, line 80: I assume this shouldn't be in here? (Is also causing a failure in the GitHub Actions checks)

andoulla: Well, I'm not sure, to be honest. Do we want to show a tooltip while wrapping this in a popover? This could cause a situation where the popover is open and the tooltip is open simultaneously.

I guess I can switch this to `hideTooltip={calendarOpen}` only to show it when the calendar is closed. Any thoughts?

andoulla: Please ignore my comment, I mixed the 2 PR. Will sort out soon, thanks

andoulla: done

hollg (approved):

PR #460, opened 2023-01-26
fix: allow numbers to be rendered when checking and wrapping type che…

LimeWub:
…cked children

**Jira:** https://atomlearningltd.atlassian.net/browse/DS-161

In a couple of components we check children type to wrap with an element (to do spacing between children via the parent).
We have been using `typeof child === 'string'` but sometimes the child would be a number which meant it would never render. For the ToggleGroup button it would render but it would not be space appropriately from the Icon.
This PR fixes this bit; **numbers are now treated the same way as string children**.

Additionally, when nesting multiple unwrapped children of varying types ie. `{1}/{3}` this is being read as 3 children and not 1, which means we get spacing between them - as you would get spacing between text and an icon. This is not being resolved in this PR.
I started coding to 'fix' this but it makes the children rendering crazy and it gets really complicated so I don't think it's worth as a requirement and implementation should make sure they are passing a string if they don't want random spaces. 

I kinda think that `string` casting in implementation for all use cases was probably good enough to begin with but at least with this change if a number is used directly it will appear instead of not even showing up and if it's not mixed with other types present in the same way as a string.

### Screenshots
**before**
Screenshot 2023-01-26 at 13 52 11

**after**
Screenshot 2023-01-26 at 13 51 09

**before**
Screenshot 2023-01-26 at 13 54 23

**after**
Screenshot 2023-01-26 at 13 48 17

### But this is still happening and I'm proposing to not fix but handle in implementation:

Screenshot 2023-01-26 at 13 48 07

Screenshot 2023-01-26 at 13 24 20

**Just pass a string:**
Screenshot 2023-01-26 at 13 24 33

### DIFF
--- a/lib/src/components/badge/Badge.mdx
+++ b/lib/src/components/badge/Badge.mdx
@@ -7,7 +7,7 @@ category: Feedback
 
 The `Badge` component can be used to highlight a short piece of information, like a status.
 
-```tsx preview
+```tsx preview live
 <Badge>New Data</Badge>
 ```
 
--- a/lib/src/components/badge/Badge.tsx
+++ b/lib/src/components/badge/Badge.tsx
@@ -70,7 +70,7 @@ export const Badge: React.FC<BadgeProps> = ({
   return (
     <StyledBadge role="status" theme={theme} size={size} {...rest}>
       {React.Children.map(children, (child) => {
-        if (typeof child === 'string') {
+        if (typeof child === 'string' || typeof child === 'number') {
           return (
             <Box
               css={{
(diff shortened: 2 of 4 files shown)


elliotlaws (approved):

lib/src/components/badge/Badge.mdx
```diff
@@ -7,7 +7,7 @@ category: Feedback
 
 The `Badge` component can be used to highlight a short piece of information, like a status.
 
-```tsx preview
+```tsx preview live
```
hollg, line 10: 💯

PR #461, opened 2023-01-30
Chore: Update `@stitches/react`

thomasdigby:
Updating `@stitches/react` to fix a number of issues with the library with later versions of Typescript.

@Mhoog and I have been through the wringer over the past few weeks trying to resolve some type issues with `@atom-learning/components` and `@stitches` in particular, resulting in this need to update `@stitches` to the latest published version (a beta).

I've written some of the steps that we went through to provide a bit of context, but it included a fair amount of trial and error testing, patching local versions of the packages, whilst also trying to resolve local TS issues with our codebase, so please let me know if you want any more detail on why we did certain things, but hopefully, we concluded with a relatively simple fix.

The initial issue we hit was with how `@stitches` accessed types from modules that weren't exported from the module

```
Exported variable "Variable" has or is using name 'Stitches' from external module
```

This was flagged sporadically across `styled()` and standard `Box`, `Flex`, etc. components, and we've also had this a couple of times in our own codebase. @Mhoog fixed this in https://github.com/Atom-Learning/components/pull/454

The next issue was types missing altogether from `@atom-learning/components` & `@stitches`, which required a change to TS module resolution `"node16"` to `"node"`, alongside an upgrade to TS itself. There are a whole heap of weird compatibility issues with the various ways that TS resolves modules and the requirements that it has with each condition to return the correct files. So TS needed an update and the type imports needed to include the relevant extension. This was fixed in this PR to `@stitches` https://github.com/stitchesjs/stitches/issues/1084

The final issue, now we'd upgraded to TS 4.9 was with `@stitches/react`s use of symbols. https://github.com/stitchesjs/stitches/pull/1082

```
TS4118: The type of this node cannot be serialized because its property '[$$PropertyValue]' cannot be serialized.
```

Thankfully that fix went into the beta release, so the only remaining issue was our use of `inset` as a key in our `@stitches` `utils` 🤷 

```
Type 'CSS | undefined' is not assignable to type 'CSS<{ sm: string; md: string; lg: string; xl: string; reducedMotion: string; allowMotion: string; hover: string; }, { col...
```

_For some reason_, in this updated version of `@stitches`, the key `inset` produced this TS error. I can only assume that somewhere in the update they've changed how custom utils and standard CSS properties overlap, and produce an error if any key in `utils` matches a standard CSS property. https://developer.mozilla.org/en-US/docs/Web/CSS/inset

We don't use `inset` too much in the codebase, and unfortunately, we can't use it natively until we drop Safari 12/13, so I have just removed the custom util and replaced it inline with `top: 0; right: 0; bottom: 0; left: 0`.

I've probably forgotten a few things to mention, but that's broadly how we got here.

### DIFF
--- a/lib/package.json
+++ b/lib/package.json
@@ -149,7 +149,7 @@
     "@radix-ui/react-tooltip": "^1.0.0",
     "@radix-ui/react-visually-hidden": "^1.0.0",
     "@reach/combobox": "^0.16.1",
-    "@stitches/react": "^1.2.8",
+    "@stitches/react": "1.3.1-1",
     "@tanstack/react-table": "^8.5.11",
     "@types/react": "^17.0.30",
     "color2k": "^2.0.0",
--- a/lib/src/components/alert-dialog/AlertDialogContent.tsx
+++ b/lib/src/components/alert-dialog/AlertDialogContent.tsx
@@ -19,7 +19,10 @@ const slideOut = keyframes({
 
 const StyledAlertDialogOverlay = styled(Overlay, {
   backgroundColor: '$alpha600',
-  inset: 0,
+  top: 0,
+  right: 0,
+  bottom: 0,
+  left: 0,
   position: 'fixed',
   zIndex: DIALOG_Z_INDEX,
   '@allowMotion': {
--- a/lib/src/components/dialog/DialogContent.tsx
+++ b/lib/src/components/dialog/DialogContent.tsx
@@ -25,7 +25,10 @@ const slideOut = keyframes({
 
 const StyledDialogOverlay = styled(Overlay, {
   backgroundColor: '$alpha600',
-  inset: 0,
+  top: 0,
+  right: 0,
+  bottom: 0,
+  left: 0,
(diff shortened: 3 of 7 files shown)


Mhoog (approved):
That PR description brings back some traumas.. 😅

lib/package.json
```diff
@@ -149,7 +149,7 @@
     "@radix-ui/react-tooltip": "^1.0.0",
     "@radix-ui/react-visually-hidden": "^1.0.0",
     "@reach/combobox": "^0.16.1",
-    "@stitches/react": "^1.2.8",
+    "@stitches/react": "1.3.1-1",
```
hollg, line 152: Have you intentionall removed the `^` because you think it is unsafe to take minor releases automatically?

thomasdigby: Yeah, this just reduces the already tiny chance that we'll upgrade this package without due process. Not necessary for this update at all but thought it was best to fix it as I'm upgrading the version.

LimeWub (approved):

PR #462, opened 2023-02-01
feat: RadioButtonField allow controlled value

LimeWub:
RadioButtonField (Radio group) currently doesn't allow to be controlled. This causes bugs when trying to use it with `onValueChange` in core.
Other Field components do allow for the value to be controlled so this PR is to allow it fro RadioButtonField as well.

**eg. SelectField (allows for value controlling)**

https://user-images.githubusercontent.com/6905473/216050826-06d45412-c5c6-4b51-90e4-ff637f78a808.mov

### Videos
**before**

https://user-images.githubusercontent.com/6905473/216050958-8420ae71-c84b-4836-8ace-5d5ef5a1e9f4.mov

**after**

https://user-images.githubusercontent.com/6905473/216051013-713a4e78-fc1e-4b7d-aae7-3485a7af6b64.mov

### DIFF
--- a/lib/src/components/radio-button-field/RadioButtonField.mdx
+++ b/lib/src/components/radio-button-field/RadioButtonField.mdx
@@ -9,7 +9,7 @@ category: Form fields
 
 **Note**: a `RadioButtonField.Item`’s `value` **must** be a `string`.
 
-```tsx preview
+```tsx preview live
 <Form>
   <RadioButtonField name="options" label="Please select an option">
     <RadioButtonField.Item label="Option 1" value="1" />
--- a/lib/src/components/radio-button-field/RadioButtonField.tsx
+++ b/lib/src/components/radio-button-field/RadioButtonField.tsx
@@ -29,46 +29,48 @@ export const RadioButtonField: React.FC<RadioButtonFieldProps> & {
   label,
   name,
   validation,
-  onValueChange
+  onValueChange,
+  ...remainingProps
 }) => {
-  const { control } = useFormContext()
-  const { error } = useFieldError(name)
+    const { control } = useFormContext()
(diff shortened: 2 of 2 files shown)


lib/src/components/radio-button-field/RadioButtonField.tsx
```diff
@@ -29,46 +29,48 @@ export const RadioButtonField: React.FC<RadioButtonFieldProps> & {
   label,
   name,
   validation,
-  onValueChange
+  onValueChange,
+  ...remainingProps
```
hollg, line 33: Is passing `...remainingProps` down the only change here? (Hard to tell because of formatting). If so, what prop is on there that solves this?

lib/src/components/radio-button-field/RadioButtonField.mdx
```diff
@@ -9,7 +9,7 @@ category: Form fields
 
 **Note**: a `RadioButtonField.Item`’s `value` **must** be a `string`.
 
-```tsx preview
+```tsx preview live
```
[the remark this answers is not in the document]
D7Torres: ❤️

lib/src/components/radio-button-field/RadioButtonField.tsx
LimeWub: Yes - that's it exactly, btw u can hide whitespace changes w this:
Screenshot 2023-02-01 at 16 24 24

LimeWub: it allows `value` to be overwritten with whatever is being passed in. (making it controlled)

hollg (approved):

thomasdigby (approved):

PR #464, opened 2023-02-07
Feature: Add sourcemaps

thomasdigby:
As we're using [`source-map-loader`](https://webpack.js.org/loaders/source-map-loader/) in our webpack configuration in `core`, we can build and link sourcemaps from `@atom-learning/components` which will allow us to see more accurate stack traces and error reports.


D7Torres (approved):
Thank you Tom!

hollg (approved):

PR #466, opened 2023-02-10
fix: allow non-native fields used in react-hook-form to be controlled…

LimeWub:
… externally via props

**Follow up to this:** https://github.com/Atom-Learning/components/pull/462

The changes in the above PR were not enough. D: 

The component displayed as if it did have a controlled value set on it but the value was not _actually_ being registered in the form. 

This had to do with the fact these fields were non-native-html fields (radix) and had to be using the `Controller` from `react-hook-form` to register at all.  The `Controller` though seems to be handling anything relevant to value internally. For the value to register in the form the fields needed to be interacted with, manually, so that the `onChange` function would run internally in the `Controller`.

Which meant that when we were editing a form and not changing the field OR changing the values via something external the value was not updating in react-hook-form, validation wasn't passing and the form was not submitting.

To fix we need to tell the controller that the value has changed when it changes externally.

So, I did a bit of a refactor in the fields which use Controller to allow for this value to be controlled from outside as you'd expect. Hopefully bringing them to parity with the non-custom-component fields (which do not need to use `Controller`).

## Screenshots/Videos

**With this sandbox (notice highlight checked/value and NOT defaults):**
Screenshot 2023-02-10 at 20 01 48

**before:**

https://user-images.githubusercontent.com/6905473/218189360-d98063f8-9070-4d30-b30c-41c816aa8cc6.mov

**after:**

https://user-images.githubusercontent.com/6905473/218189430-d6c0f977-49ad-47e5-9ef8-75972d727c94.mov

### DIFF
--- a/lib/src/components/checkbox-field/CheckboxField.tsx
+++ b/lib/src/components/checkbox-field/CheckboxField.tsx
@@ -22,7 +22,7 @@ export const CheckboxField: React.FC<CheckboxFieldProps> = ({
   name,
   validation,
   description,
-  defaultChecked,
+  defaultChecked = false,
   checked,
   ...remainingProps
 }) => {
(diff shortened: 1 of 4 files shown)


hollg (approved):

D7Torres (approved):
Usually, it's very helpful to TDD the bugs (make a test that fails because of the bug before solving it)

PR #467, opened 2023-02-12
feat: allow SearchInput to be both controlled and uncontrolled

LimeWub:
**BREAKING CHANGE: will need to change all uses of `value` to `defaultValue` in core or use controlled version**

> SearchInput component is uncontrolled. The value prop only acts as defaultValue setting the initial state of inner component state and not being used after that.

This PR allow for use both with a defaultValue (uncontrolled) and value (controlled).

JIRA: https://atomlearningltd.atlassian.net/browse/DS-308

## Screenshots

**With this sandbox:**
Screenshot 2023-02-12 at 19 29 07

**before:**

https://user-images.githubusercontent.com/6905473/218333054-9badc578-a598-4cf0-99b7-6b7cb3e465d1.mov

**after:**

https://user-images.githubusercontent.com/6905473/218332838-67fa72c4-8dd3-44e6-9c7b-91a92011862a.mov

### DIFF
--- a/lib/src/components/data-table/DataTableGlobalFilter.tsx
+++ b/lib/src/components/data-table/DataTableGlobalFilter.tsx
@@ -38,7 +38,7 @@ export const DataTableGlobalFilter: React.FC<DataTableSearchProps> = ({
       </OptionallyVisuallyHiddenContainer>
       <SearchInput
         {...props}
-        value={globalFilter}
+        defaultValue={globalFilter}
         onChange={handleChange}
         name={label}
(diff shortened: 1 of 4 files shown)


Jdrazong (approved):

D7Torres (approved):
I don't remember if you added a test in the end for checking the controlling logic.

If you didn't, you could render the component, then update the prop, and check that it is selected/unselected correctly.

kerryambt (approved):

PR #469, opened 2023-02-13
Feat/field wrapper multiple errors

hollg:
# Feat: support multiple errors in form fields

This is the first step in supporting multiple inline messages _of different types_ in our form field components.

## Implementation
I've added the `FieldFeedback` component, which replaces the inline render of `<InlineMessage theme="error">` in the `FieldWrapper` when there is a validation error on the field. If multiple validation errors apply, `FieldFeedback` will either show the one that `react-hook-form` has prioritised (i.e. the existing behaviour) or all of them, depending on whether the `criteriaMode` prop is set to `'all'` or `'firstError'`. Errors are stacked in either a column or a row according to whether `feedbackDirection` is set to `column` or `row`. It defaults to `column`. 

`criteriaMode` has been added as an optional property on the object we pass as the `validation` prop of fields. `feedbackDirection` has been added as a separate prop because when we add the option to display other types of messages within the `FieldWrapper`, it will apply to all of them, not just validation error messages.

To enable the option to show multiple validation errors, `criteriaMode: 'all'` must be passed to `useForm()` inside the `Form` component. I've added this as a prop to `Form` and defaulted it to `'all'`. We can override it with `firstError` if needed, but as far as the user is concerned we'll get the same effect by setting the `criteriaMode` prop on individual fields.  

## Coming soon
Next (in a separate PR) I will add the ability to pass custom (unrelated to validation) messages to `FieldFeedback`, which will be able to either display them all together or hide non-error messages when errors are present.

## InlineFieldWrapper
Note: I haven't made any changes to `InlineFieldWrapper` because I discovered that validation messages don't actually seem to work in the components that use it. I'm not sure how much sense it makes to have validation errors on radio buttons / checkboxes anyway, but either way I think any fixes there should be a separate PR.

### Default behaviour
Single error, as prioritised by `react-hook-form`. 
Screenshot 2023-02-14 at 15 29 50

### All errors, default direction
Stacked in a vertical column
Screenshot 2023-02-14 at 15 29 30

### All errors, `direction='row'`
Note: this might cause the `FieldWrapper` container to grow in width, and our inputs take up 100% of the available width by default. So if you're going to use this option, make sure you set a sensible width on the field to begin with.
Screenshot 2023-02-14 at 15 30 22


LimeWub:
Did I miss something? Why is this closed 👀 ?

PR #474, opened 2023-02-21
Feat/control documentation via documentation content branch

LimeWub:
## Problem:

Because now the CMS tries to commit changes to content directly to the main branch we get this error for users without direct write access to main (most of us).

image

## This PR attempts to fix this problem by:

1. Pointing NetlifyCMS to use another branch (`documentation-content`) to update documentation in. This other branch is not protected like `main` is so anything goes and anyone can directly commit to it.
2. When a commit is made to `documentation-content`  which matches the `[NetlifyCMS]` commit pattern and affects the 2 directories which I'd expect Netlify to be editing (content or admin uploaded images):
           * checkout main as bot
           * make sure there are no files changed outside the 2 documentation directories. IF THERE ARE: Stop action and throw error as this is unsafe to merge
           * If everything same, merge the branch back into main to update the documentation
1.  When a commit is made into `main`; merge main back into `documentation-content`.

**WARNING: The workflows have not been tested and AFAIK to test them we just need to do it live!** 
image

Any input very much appreciated as this is my first time writing github actions 👀

### DIFF
--- a/.github/workflows/sync-from-documentation-to-main.yml
+++ b/.github/workflows/sync-from-documentation-to-main.yml
@@ -1,0 +1,49 @@
+name: Sync from documentation to main
+
+permissions:
+  contents: write
+
+on:
+  push:
+    branches:
+      - documentation-content
+    paths:
+      - documentation/content/**
+      - documentation/public/admin/images/**
+
+jobs:
+  sync:
+    if: "startsWith(github.event.head_commit.message, '[NetlifyCMS]')"
+    name: Sync only the documentation content relevant folders from documentation-content into the main branch
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+        with:
+          ref: main
+          fetch-depth: 0
+
+      - name: Get changed files excluding the docs
+        id: changed-files-excluded
+        uses: tj-actions/changed-files@v35
+        with:
+          base_sha: ${{ github.sha }}
+          files_ignore: |
+            documentation/content/**
+            documentation/public/admin/images/**
+
+      - name: Fail if files other than docs are changed
+        if: steps.changed-files-excluded.outputs.any_changed == 'true'
+        run: |
+          echo "Non-documentation files have been changed:"
+          for file in ${{ steps.changed-files-excluded.outputs.all_changed_files }}; do
+            echo "$file was changed"
+          done
+          echo "This action is exiting as a safety feature! Revert files before trying again!"
+          exit 1
+
+      - uses: everlytic/branch-merge@1.1.2
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          source_ref: "documentation-content"
+          target_branch: "main"
+          commit_message_template: "[Automated] Merged {source_ref} into target {target_branch}"
--- a/.github/workflows/sync-from-documentation-to-main.yml
+++ b/.github/workflows/sync-from-documentation-to-main.yml
@@ -1,7 +1,7 @@
 name: Sync from documentation to main
 
 permissions:
-  contents: write
+  contents: read
 
 on:
   push:
@@ -17,10 +17,9 @@ jobs:
     name: Sync only the documentation content relevant folders from documentation-content into the main branch
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@master
+      - uses: actions/checkout@v3
         with:
           ref: main
-          fetch-depth: 0
 
       - name: Get changed files excluding the docs
         id: changed-files-excluded
@@ -41,7 +40,7 @@ jobs:
           echo "This action is exiting as a safety feature! Revert files before trying again!"
           exit 1
 
-      - uses: everlytic/branch-merge@1.1.2
+      - uses: everlytic/branch-merge@1.1.5
         with:
           github_token: ${{ secrets.BOT_ACCESS_TOKEN }}
           source_ref: "documentation-content"
--- a/.github/workflows/sync-from-documentation-to-main.yml
+++ b/.github/workflows/sync-from-documentation-to-main.yml
@@ -43,7 +43,7 @@ jobs:
 
       - uses: everlytic/branch-merge@1.1.2
         with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
+          github_token: ${{ secrets.BOT_ACCESS_TOKEN }}
           source_ref: "documentation-content"
           target_branch: "main"
           commit_message_template: "[Automated] Merged {source_ref} into target {target_branch}"
--- a/.github/workflows/sync-from-main-to-documentation.yml
+++ b/.github/workflows/sync-from-main-to-documentation.yml
@@ -1,0 +1,22 @@
+name: Sync from main to documentation
+
+permissions:
+  contents: write
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  sync:
+    name: Sync everything from main into the documentation-content branch
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - uses: everlytic/branch-merge@1.1.2
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          source_ref: "main"
+          target_branch: "documentation-content"
+          commit_message_template: "[Automated] Merged {source_ref} into target {target_branch}"
--- a/.github/workflows/sync-from-main-to-documentation.yml
+++ b/.github/workflows/sync-from-main-to-documentation.yml
@@ -1,7 +1,7 @@
 name: Sync from main to documentation
 
 permissions:
-  contents: write
+  contents: read
 
 on:
   push:
@@ -13,8 +13,8 @@ jobs:
     name: Sync everything from main into the documentation-content branch
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@master
-      - uses: everlytic/branch-merge@1.1.2
+      - uses: actions/checkout@v3
+      - uses: everlytic/branch-merge@1.1.5
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           source_ref: "main"
--- a/documentation/public/admin/config.yml
+++ b/documentation/public/admin/config.yml
@@ -1,7 +1,7 @@
 backend:
   name: github
   repo: Atom-Learning/components
-  branch: main # Branch to update (optional; defaults to master)
+  branch: documentation-content # Branch to update (optional; defaults to master)
   base_url: https://atomlearning.design/ #Change to EXACT match for deployment url
   api_root: https://api.github.com # Shouldn't be necessary
   auth_endpoint: api/auth
--- a/documentation/public/admin/config.yml
+++ b/documentation/public/admin/config.yml
@@ -6,6 +6,14 @@ backend:
   api_root: https://api.github.com # Shouldn't be necessary
   auth_endpoint: api/auth
 
+  commit_messages:
+    create: "[NetlifyCMS] Create {{collection}} “{{slug}}”"
+    update: "[NetlifyCMS] Update {{collection}} “{{slug}}”"
+    delete: "[NetlifyCMS] Delete {{collection}} “{{slug}}”"
+    uploadMedia: "[NetlifyCMS] Upload “{{path}}”"
+    deleteMedia: "[NetlifyCMS] Delete “{{path}}”"
+    openAuthoring: "[NetlifyCMS] {{message}}"
+
 # local_backend: true
 
 media_folder: "documentation/public/admin/images" # Media files will be stored in the repo under images/uploads


.github/workflows/sync-from-documentation-to-main.yml
```diff
@@ -0,0 +1,26 @@
...
+  sync:
+    if: "contains(github.event.head_commit.message, '[NetlifyCMS]')"
+    name: Sync only the documentation content relevant folders from documentation-content into the main branch
+    steps:
+      - uses: actions/checkout@master
+        with:
+          token: ${{ secrets.BOT_ACCESS_TOKEN }}
+          ref: main
+
+      - run: git checkout documentation-content documentation/content documentation/public/admin/images
```
hollg, line 21: I wonder whether we could maintain the original git history by replacing this scoped checkout with the following logic:

1. checkout to `main`
2. `git diff` against most recent commit in `documentation-content` and check which directories have changed ([this SO question](https://stackoverflow.com/questions/50440420/git-diff-only-show-which-directories-changed) has some ideas about how to do that)
3. if only `documentation/content` changed, merge the commit into `main`

LimeWub: I'll try to do this...
I don't know how to do the if at all but maybe it can be one lined via bash 🤷

LimeWub: Wait this still would allow for the malicious first commit but healthy last commit issue we talked about.... >.<

hollg: Why/how? We'd only be merging the healthy commit

LimeWub: Oh you mean cherry pick?

LimeWub: I'll see if I can do that 👀

hollg: Ah, sorry, yes, I probably do!

LimeWub: Fixed/changed so we now keep the actual commits.

hollg (approved):
I'm approving mostly because I've seen you demo this working, rather than because I understand all the code!

thomasdigby:
@LimeWub Tagging @euanhunteratom as this is his world!

.github/workflows/sync-from-main-to-documentation.yml
```diff
@@ -0,0 +1,22 @@
...
+
+jobs:
+  sync:
+    name: Sync everything from main into the documentation-content branch
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - uses: everlytic/branch-merge@1.1.2
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
euanhunteratom, line 19: Don't think `secrets.GITHUB_TOKEN` is a valid secret/token. `github.token` is the default token for the github-actions user

.github/workflows/sync-from-documentation-to-main.yml
```diff
@@ -0,0 +1,49 @@
...
+        if: steps.changed-files-excluded.outputs.any_changed == 'true'
+        run: |
+          echo "Non-documentation files have been changed:"
+          for file in ${{ steps.changed-files-excluded.outputs.all_changed_files }}; do
+            echo "$file was changed"
+          done
+          echo "This action is exiting as a safety feature! Revert files before trying again!"
+          exit 1
+
+      - uses: everlytic/branch-merge@1.1.2
```
euanhunteratom, line 44: As noted on Slack, I _really_ think this should be a PR instead of a direct merge, even with the safety check. See how the [format-files workflow in atom-core does it for an example](https://github.com/Atom-Learning/atom-core/blob/0acb47851b70636244c770a074f10477a3e8ca85/.github/workflows/format-files.yml#L45)

.github/workflows/sync-from-main-to-documentation.yml
LimeWub: I tested this on my own repo as is and somehow it worked 👀 
I think it was supposed to be the single use token of the action itself.

I think they might be the same thing? https://stackoverflow.com/a/70858840

Happy to change to `github.token` tho if they are indeed so similar

euanhunteratom: Oh oops, looks like [it is indeed an alias](https://docs.github.com/en/actions/learn-github-actions/contexts#secrets-context:~:text=properties%20listed%20below.-,secrets.GITHUB_TOKEN,-string). This is fine then, no changes required.

.github/workflows/sync-from-documentation-to-main.yml
euanhunteratom: The main problems are this: as it stands, this won't work because the `github-actions` "user" (GITHUB_TOKEN) can't bypass branch protections either and can't be configured to; GitHub refuse to allow it on security grounds. So, the alternative is an Administrator Personal Access Token (PAT). But having one of those available is an even bigger security risk (anyone with write access to the repo can write a Workflow that then can perform Admin-level actions, effectively giving them admin access). There might be potential solutions to this, but none we've had the time to explore.

Is there a reason the docs site can't be build from the `documentation-content` branch and the code package from `main`? The branch protections are really there to protect the code, so a auto-sync main->docs-content is fine with an auto-PR to bring the (non-functional) changes back into docs-content->main

LimeWub: Ah I didn't know it wont work - in my repo I haven't set up the PR requirement permissions so I must've missed this.

The `auto-fix` workflow auto-commits into main using the `token: ${{ secrets.BOT_ACCESS_TOKEN }}` would that work instead of `GITHUB_TOKEN `? .-. It's already there and doing auto-commits to main as far as I can tell so we maybe already have this problem?

The workflows get checked before being merged (case in point: this one) so how can someone just 'write a workflow' which does whatever? It would need to get into main to even work right? And to get into main it would be caught in a PR.
As for automatically getting them in - a check like I have in this workflow here means it would also not get into main via bot/workflow.

The docs could be built via the `documentation-content` actually so that's an okay solution but **would we get a bunch of spam with PRs one for each change? which will need approvals and will be conflicting between themselves? Hopefully not - right it should just be a push for future changes? Does the link you shared do duplicate PRs or no?** and also **anyone editing the docs without the cms/manually while editing main will have old files until these are approved and merged**.

Unless there's a way to: 
* drop any old PRs about merging in the documentation the bot has opened before OR just check if there's a PR open already? (Somehow this must be possible) 
* require little to no approvals to merge them in (defeats the purpose again maybe? I suppose we could tag a couple of people but it really would be spamming their emails if the CMS is indeed used lots)
* but we still have the problem of people editing documentation locally (developers), then having to manually check through the PRs for the bot one to see if there have been changes to the documentation via the cms/causing conflicts to resolve etc.

All in all I do get where you're coming from but I'm trying to make this a sleek process, and as safe as I can.
I could try the PR solution but it will make things slower 👀 

The BOT token is already there so there's already a weakness if that's what it is; and this workflow itself is fairly safe due to the check. All in all: I get it, and I'm not against trying out the PR way but iunno if it's worth the time to edit and sluggishness in the process.

Can I have an opinion on how this would affect the process @thomasdigby @hollg ?

euanhunteratom: > The auto-fix workflow auto-commits into main using the token: ${{ secrets.BOT_ACCESS_TOKEN }} would that work instead of GITHUB_TOKEN ? .-. It's already there and doing auto-commits to main as far as I can tell so we maybe already have this problem?

Yeah, then it might already be a problem. Though not sure if that account is a Admin or can just bypass branch protections specifically (which I remembered is a setting, oops). If it's already being done, it's maybe okay for now.

> how can someone just 'write a workflow' which does whatever?
Certain events will trigger on feature branches as soon as they are pushed; no requirement to be in `main`. E.g. if the `on` is just `push` with no qualifier. So, there is no check step such as a PR before it runs

> PRs for each change
I'd suggest that the PR be `documentation-content` (head) to `main` (base) directly (no extra branches created). Then GitHub will only allow one PR to be created and update it when new commits are pushed to `documentation-content`. The link I shared does us a separate branch so it'd need some tweaking. And, unfortunately, yes, `main` would be out of date until the PR was approved/merged.

I get this is not ideal and also want the process to be as sleak as possible, but also very conscious of the security implications, particularly as this is technically a public repo. I would maybe suggest to use `BOT_ACCESS_TOKEN` for now, but with an explicit intent on both sides (yourselves and us, DevOps) to revisit once we've had a look at better ways to do branch bypass in an automated fashion.

LimeWub: I've replaced with `BOT_ACCESS_TOKEN` for now. Shall we make a ticket for investigating the security of this for the 2 workflows which use it? Which board should it go in?

euanhunteratom: Probably the DEVOPS board (you should be able to raise one there, if not I'll do it)

LimeWub: https://atomlearningltd.atlassian.net/jira/software/projects/DEVOPS/boards/15?selectedIssue=DEVOPS-304

LimeWub:
@euanhunteratom @thomasdigby 
I need a second review :)

thomasdigby:
> @euanhunteratom @thomasdigby I need a second review :)

I'd prefer to lean on @euanhunteratom for that second review if that's ok @LimeWub? I'm not sure if I can provide much useful feedback outside of the general flow of the action, and I think the stuff that needs reviewing is the permissions/tokens layer

euanhunteratom:
> @euanhunteratom @thomasdigby I need a second review/approve :)

I'll review this afternoon if that's okay?

LimeWub:
> > @euanhunteratom @thomasdigby I need a second review/approve :)
> 
> I'll review this afternoon if that's okay?

Sure~ !

.github/workflows/sync-from-documentation-to-main.yml
```diff
@@ -0,0 +1,49 @@
...
+jobs:
+  sync:
+    if: "startsWith(github.event.head_commit.message, '[NetlifyCMS]')"
+    name: Sync only the documentation content relevant folders from documentation-content into the main branch
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+        with:
+          ref: main
+          fetch-depth: 0
```
euanhunteratom, line 23:
```suggestion
      - uses: actions/checkout@v3
```
Suggestion: use `v3` instead of `master` so it's not using unreleased, potentially broken code

Suggestion: removed `fetch-depth: 0` is the default. 

Suggestion: `ref` defaults to the commit that triggered the run e.g. the commit just pushed. This is (probably) preferable to `main` as:
1. I think `tj-actions/changed-files` works best if the actual commit is checked out
2. The specific commit is better than a branch name (e.g. `documentation-content`) as it prevents race conditions e.g. multiple changes in quick succession

.github/workflows/sync-from-documentation-to-main.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: Get changed files excluding the docs
+        id: changed-files-excluded
+        uses: tj-actions/changed-files@v35
+        with:
+          base_sha: ${{ github.sha }}
```
euanhunteratom, line 29:
```suggestion
```
With the above suggested change to remove `ref`, this is the default. The list of files changed (minus the allowed ones) will thus be both changed in the specific commit to the `documentation-content` branch

.github/workflows/sync-from-documentation-to-main.yml
euanhunteratom, line 44:
```suggestion
      - uses: everlytic/branch-merge@1.1.5
```
Should probably use the latest version

euanhunteratom (approved):
A few suggestions. I'm approving to unblock but suggest resolving suggestions before merging, particularly as this is untested

.github/workflows/sync-from-documentation-to-main.yml
LimeWub: Just to clarify: I want to be diffing the latest `origin/documentation-content` sha with the last commit in `origin/main` - that's why I have this `base_sha` here because the checkout above is checking out main.
It seems wrong to me to then remove this from here as it's so important to checking the diff correctly - am I misunderstanding something?

euanhunteratom: Hmm, I'm not sure. My suggestion was for getting the changed files just in the just pushed commit but I guess branch wise might be safer? Issue there is making sure it doesn't detect unrelated code changes in `main` that haven't been merged to `documentation-content` yet. I think either way you'd want to checkout `documentation-content` and use `main` as the `base_sha` because that's the way around they are in the merge (docs in the "head" and main the "base" branch)

LimeWub: > My suggestion was for getting the changed files just in the just pushed commit but I guess branch wise might be safer?

Yep I'm doing branch wise in case there's rogue commits in there that are changing outside of documentation. :)

> Issue there is making sure it doesn't detect unrelated code changes in main that haven't been merged to documentation-content yet.

Main merges to documentation automatically so it shouldn't take too long post main change for documentation to be updated. Plus if one automatic merge from documentation fails (as main out of sync) and then main gets merged the next one will succeed so nbd.

> I think either way you'd want to checkout documentation-content and use main as the base_sha because that's the way around they are in the merge (docs in the "head" and main the "base" branch)

How would I get the main's latest sha without checking it out? I already have documentations since it's the one that started the action but no clue how I'd get main's :|

euanhunteratom: Fair, branch wise is probably better for extra security and failure recovery.

Re `main` SHA - run `git rev-parse origin/master` in a `run` step before and use an [output parameter](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter) to reference it here

.github/workflows/sync-from-documentation-to-main.yml
```diff
@@ -0,0 +1,49 @@
...
+          for file in ${{ steps.changed-files-excluded.outputs.all_changed_files }}; do
+            echo "$file was changed"
+          done
+          echo "This action is exiting as a safety feature! Revert files before trying again!"
+          exit 1
+
+      - uses: everlytic/branch-merge@1.1.2
+        with:
+          github_token: ${{ secrets.BOT_ACCESS_TOKEN }}
+          source_ref: "documentation-content"
```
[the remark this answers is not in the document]
LimeWub: Screenshot 2023-03-01 at 11 02 53


Are we sure `sha` will work here? They don't mention anything about `sha` in the documentation for this action.

In the example for this they use `source_ref: ${{ github.ref }}` - is that better?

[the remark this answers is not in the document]
euanhunteratom: Strictly a SHA is a ref from Git's PoV. But as above, branch-centric probably is better so ignore these comments about SHA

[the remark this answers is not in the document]
LimeWub: oh okay didn't realise I should ignore this 👀

PR #476, opened 2023-02-22
feat: add tooltip to ActionIcon

andoulla:
### Implementation
Changing the ActionIcon into showing the tooltip by default

### Screen Recordings
The first tooltip is the default state, and the second is with the tooltip turned off. 
The third tool tip illustrates how it will behave if an ActionIcon is wrapped in a tooltip (as in atom-core). Radix is smart enough only to trigger the tooltip within the ActionIcon and not the external one. This simplifies the migration as we can release this change and independently do a PR to remove the TooltipActionIcon from core.
https://recordit.co/KCE6Xout9N

These are only the core changes that add the functionality; once I have the go-ahead with this pattern, I will fix the broken tests and then release


LimeWub (approved):

hollg (approved):

lib/src/components/action-icon/ActionIcon.tsx
```diff
@@ -195,12 +197,24 @@ const StyledButton = styled('button', {
   ]
 })
 
+const ConditionallyWrapWithTooltip = ({ hasTooltip, label, children }) => {
```
hollg, line 200: Not a blocker, but can you add types for the props here?

andoulla: hey of course, this was just to highlight the changes/pattern of the component, not ready yet.

andoulla: done

PR #477, opened 2023-02-22
feat(data-table): adding support to multiple sticky columns

viniciusvolpe:
### Description
As a follow up for the sticky columns changes to the data table I'm now adding support for multiple sticky columns

### Implementation
- Created hook to generate the required CSS for multiple sticky columns based on the number that is set
- Used the hook in the sticky columns container and changed the shadow CSS to not override the positioning

### Screenshot

https://user-images.githubusercontent.com/11462265/221198613-1eb33685-7ed8-404a-aca4-5a0f56dda2ed.mov


lib/src/components/table/useStickyColumnsCss.ts
```diff
@@ -0,0 +1,51 @@
...
+): IUseStickyColumnsCss => {
+  const [columnsCss, setColumnsCss] = React.useState<CSS>({})
+
+  React.useLayoutEffect(() => {
+    if (!numberOfStickyColumns) return
+
+    let accWidth = 0
+    const newElements: HTMLTableCellElement[] = []
+    wrapperRef?.current
+      ?.querySelectorAll('th')
```
LimeWub, line 21: Are these `?` defo enough here or will there be an error for say using `querySelectorAll` of undefined?

lib/src/components/table/useStickyColumnsCss.ts
```diff
@@ -0,0 +1,51 @@
...
+  columnsCss: CSS
+}
+
+export const useStickyColumnsCss = (
+  numberOfStickyColumns: number,
+  wrapperRef?: React.RefObject<HTMLTableSectionElement>
+): IUseStickyColumnsCss => {
+  const [columnsCss, setColumnsCss] = React.useState<CSS>({})
+
+  React.useLayoutEffect(() => {
```
LimeWub, line 15: For my understanding: Why `useLayoutEffect` and not `useEffect`?

viniciusvolpe: hey, if I'm not wrong `useLayoutEffect` runs before the screen be updated, so I intended to reduce the visual effect that changing the columns' styles could cause.

lib/src/components/table/useStickyColumnsCss.ts
viniciusvolpe: I think I can change the prop to not be undefined in the type, but the `?` after `.current` will always need to be there because the ref value can be null.

lib/src/components/table/useStickyColumnsCss.ts
```diff
@@ -0,0 +1,51 @@
...
+    if (!numberOfStickyColumns) return
+
+    let accWidth = 0
+    const newElements: HTMLTableCellElement[] = []
+    wrapperRef?.current
+      ?.querySelectorAll('th')
+      .forEach((element) => newElements.push(element))
+    setColumnsCss(
+      newElements
```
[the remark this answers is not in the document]
viniciusvolpe: sure, I'll split it

lib/src/components/table/useStickyColumnsCss.ts
```diff
@@ -0,0 +1,51 @@
...
+              {
+                position: 'sticky',
+                left: `${accWidth}px`,
+                minWidth: `${column.offsetWidth}px`, // fixing width for sticky columns
+                zIndex: '2'
+              }
+          }
+
+          accWidth +=
+            wrapperRef?.current?.querySelectorAll('th')?.item(index)
```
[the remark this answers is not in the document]
viniciusvolpe: yes, will extract that to a variable

lib/src/components/table/useStickyColumnsCss.ts
```diff
@@ -0,0 +1,51 @@
...
+              .clientWidth || 0
+
+          return cssObject
+        }, {} as CSS)
+    )
+  }, [])
```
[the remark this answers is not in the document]
viniciusvolpe: Actually, I want this effect to be executed only once to avoid rerenders, since the `numberOfStickyColumns` isn't dynamic (it's hardcoded in the data table declaration) I don't think we need it here.

lib/src/components/table/useStickyColumnsCss.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  React.useLayoutEffect(() => {
+    if (!numberOfStickyColumns) return
+
+    let accWidth = 0
+    const tableHeaderElements: HTMLTableCellElement[] = []
+
+    const tableHeaderCells = wrapperRef.current?.querySelectorAll('th')
+
+    tableHeaderCells?.forEach((element) => tableHeaderElements.push(element))
```
LimeWub, line 23: Why do you need to re-assign `tableHeaderCells` to `tableHeaderElements` ?

Are you just trying to make an array out of them? In which case you could just do `Array.from()` or `[...(wrapperRef.current?.querySelectorAll('th') || [])]`

lib/src/components/table/useStickyColumnsCss.ts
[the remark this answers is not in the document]
LimeWub: If the variable never changes then the effect will only run once anyway right? So it shouldn't be a problem.
I think it's wrong to not have it as a dependency in the effect (and surprised it's not flagging in ts tbh)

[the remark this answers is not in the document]
hollg: I agree with @LimeWub — it sounds like, conceptually, the value _is_ a dependency of the effect. We only want it to run once _so long as the value doesn't change_, right?

lib/src/components/table/useStickyColumnsCss.ts
viniciusvolpe: Fixed, thanks for the help with TS complains.

lib/src/components/table/useStickyColumnsCss.ts
```diff
@@ -0,0 +1,51 @@
...
+          }
+      }
+
+      accWidth += tableHeaderCells?.item(index).clientWidth || 0
+
+      return cssObject
+    }, {} as CSS)
+
+    setColumnsCss(newColumnsCss)
+  }, [numberOfStickyColumns, wrapperRef])
```
LimeWub, line 46: Minor but I think because `wrapperRef` is a ref (useRef) this will never change and shouldn't be needed in this array BUT if something was flagging it should then it makes sense for it to be here. Doesn't hurt anything it's here but just FYI :)

LimeWub (approved):

hollg (approved):
LGTM!

PR #481, opened 2023-03-02
fix(fieldwrapper): allow prompt to have an onClick

andoulla:
This change allows the `FieldWrapper prompt` to have an `onClick` instead of just a link. This allows for the edge case where we need to trigger an action from the prompt that isn't a redirect.

The component looks exactly the same, and the prompt still will look like a link. 

This is needed to unblock https://atomlearningltd.atlassian.net/browse/FG-237.

### DIFF
--- a/lib/src/components/field-wrapper/FieldWrapper.tsx
+++ b/lib/src/components/field-wrapper/FieldWrapper.tsx
@@ -16,7 +16,7 @@ export type FieldWrapperProps = {
   error?: string
   fieldId: string
   label: string
-  prompt?: { link: string; label: string }
+  prompt?: { link?: string; label: string; onClick?: () => void }
   description?: string
   required?: boolean
   hideLabel?: boolean
@@ -53,7 +53,7 @@ export const FieldWrapper: React.FC<FieldWrapperProps> = ({
           {label}
         </Label>
         {prompt && (
-          <Link href={prompt.link} size="sm">
+          <Link href={prompt.link} onClick={prompt.onClick} size="sm">
             {prompt.label}
           </Link>
         )}
--- a/lib/src/components/field-wrapper/FieldWrapper.tsx
+++ b/lib/src/components/field-wrapper/FieldWrapper.tsx
@@ -53,7 +53,7 @@ export const FieldWrapper: React.FC<FieldWrapperProps> = ({
           {label}
         </Label>
         {prompt && (
-          <Link href={prompt.link} onClick={prompt.onClick} size="sm">
+          <Link href={prompt?.link} onClick={prompt?.onClick} size="sm">
             {prompt.label}
           </Link>
         )}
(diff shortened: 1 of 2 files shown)


lib/src/components/field-wrapper/FieldWrapper.tsx
```diff
@@ -53,7 +53,7 @@ export const FieldWrapper: React.FC<FieldWrapperProps> = ({
           {label}
         </Label>
         {prompt && (
-          <Link href={prompt.link} size="sm">
+          <Link href={prompt?.link} onClick={prompt?.onClick} size="sm">
```
hollg, line 56: I think we need to prevent the consumer from passing both a `link` and an `onClick`. I don't know what will happen if both are provided with the current implementation (maybe both things will trigger?) but I don't think it makes sense to support it.

If you pull the type for the `prompt` prop into its own type, we should be able to achieve this with something like the following:

```ts
type Prompt = { label: string } & ({ link: string, onClick: never } | { link: never, onClick: () => void })
```

andoulla: We did have something like this on `Link` when we started, and we took it away (I can't remember why). `<Link>` doesn't stop you from passing both (https://github.com/Atom-Learning/components/blob/main/lib/src/components/link/Link.tsx) 

my question is what if we need to trigger an event on a redirect, won't we need to pass both?

hollg: Ooft, good point. Didn't think of that! Have you tested that the `onClick` fires before the navigation happens with this implementation?

andoulla: I can't remember, but I will definitely test now and let you know

andoulla: it seems to work on the correct order https://recordit.co/xtkJdEavKP first fires the onClick and then redirects so I'm going to resolve this comment if that is ok

LimeWub (approved):
I didn't even know FieldWrapper has a prompt or what the prompt looks like. 
I don't think it's even documented other than in the props table.
But the code change makes sense so 👍

hollg (approved):

andoulla:
> I didn't even know FieldWrapper has a prompt or what the prompt looks like. I don't think it's even documented other than in the props table. But the code change makes sense so 👍

I didn't either as we only use it in the PasswordField in the DS (don't know if we use it for other things in core). Shall I update the doc for it? It is such an edge case I thought

LimeWub:
> Shall I update the doc for it? It is such an edge case I thought it might be clutter

It doesn't have to block this PR but if we have it and use it technically it should be in the doc.

We may also want similar functionality added to the InlineFieldWrapper (Elliot needed that yesterday) so maybe we should just have a look at how to make both of these labels match what they support and document then. Maybe just a ticket in JIRA will be good for now.

andoulla:
> > Shall I update the doc for it? It is such an edge case I thought it might be clutter
> 
> It doesn't have to block this PR but if we have it and use it technically it should be in the doc.
> 
> We may also want similar functionality added to the InlineFieldWrapper (Elliot needed that yesterday) so maybe we should just have a look at how to make both of these labels match what they support and document then. Maybe just a ticket in JIRA will be good for now.

https://atomlearningltd.atlassian.net/jira/software/projects/DS/boards/27?selectedIssue=DS-335

PR #483, opened 2023-03-08
Ds 73/add table xl size

D7Torres:
JIRA ticket

The `xl` size is a bit bigger than the `lg`. Initially, we were going to call it `lg` and move the 2 existing sizes (`md` and `lg`) 1 level down. But we did it this way to be the least disruptive possible to the current usages.

This PR also makes the vertical padding depending on the `size`. Otherwise, tables with different sizes could actually look the same, as the row height which is the only thing the size was affecting is just a minimum, it can grow with taller content.

image

### DIFF
--- a/documentation/content/components.content.table.md
+++ b/documentation/content/components.content.table.md
@@ -6,8 +6,8 @@ links:
   showReportAnIssue: true
 tabs:
   - content: >-
-      The `Table` component displays a collection of data grouped into rows.
-      Its structure mirrors that of a regular HTML table, with the
+      The `Table` component displays a collection of data grouped into rows. Its
+      structure mirrors that of a regular HTML table, with the
       smaller `Table.Body`, `Table.Cell`, `Table.Footer`, `Table.Header`, `Table.HeaderCell` and `Table
       Row` components corresponding to
       the `<tbody>`, `<td>`, `<tfoot>`, `<thead>`, `<th>` and `<tr>` tags,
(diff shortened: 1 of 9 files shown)


documentation/content/components.content.table.md
```diff
@@ -6,8 +6,8 @@ links:
   showReportAnIssue: true
 tabs:
   - content: >-
-      The `Table` component displays a collection of data grouped into rows.
-      Its structure mirrors that of a regular HTML table, with the
+      The `Table` component displays a collection of data grouped into rows. Its
+      structure mirrors that of a regular HTML table, with the
```
D7Torres, line 10: 🤷‍♂️ I don't think I did this change, but the CMS editor did it automatically.

documentation/content/components.content.table.md
```diff
@@ -106,30 +106,54 @@ tabs:
...
-            <Table.Cell>35</Table.Cell>
-          </Table.Row>
-          <Table.Row>
-            <Table.Cell>Evelyn</Table.Cell>
-            <Table.Cell>Smith</Table.Cell>
-            <Table.Cell>27</Table.Cell>
-          </Table.Row>
-        </Table.Body>
-      </Table>`} language={"tsx"} />
+      The `Table` component has 3 size variants that control the row height and cells padding. The three available sizes are `md`(default), `lg` and `xl`.
```
D7Torres, line 109: Here I removed information about the usage. Miguel told me he will add it to the Usage tab

lib/src/components/table/Table.tsx
```diff
@@ -32,12 +32,20 @@ const StyledTable = styled('table', {
...
         [`${TableCell}, ${TableHeaderCell}, ${TableFooterCell}`]: {
-          height: '$5'
+          height: '$5',
+          padding: '$2 $3'
+        }
+      },
+      xl: {
+        [`${TableCell}, ${TableHeaderCell}, ${TableFooterCell}`]: {
+          height: '$6',
+          padding: '$4 $3'
```
D7Torres, line 48: I could have separated vertical and horizontal padding, then the $3 would be in only one place (DRY).

But if no one wants to challenge this... I think is cleaner this way 🤷‍♂️

LimeWub (approved):

lib/src/components/table/Table.test.tsx
```diff
@@ -51,6 +51,29 @@ describe(`Table component`, () => {
     expect(container).toMatchSnapshot()
   })
 
+  it('renders with size set to xl', async () => {
```
andoulla, line 54: Is there a benefit to this test? Snapshots don't really assess visual changes, just the HTML structure and the structure remains the same when the size changes. Since we already have a test testing the size variant, I think we don't need it

andoulla (approved):

lib/src/components/table/Table.test.tsx
andoulla: To be honest, not a blocker for me. However, it is a bit inconsistent as we are not testing every option of each variant everywhere else

D7Torres: I didn't know. I saw a test for `lg` so I thought: let's add one for `xl` too!

+1 for consistency, I'll remove it.

PR #490, opened 2023-03-09
fix: cast to string in globalFilterFn to make it safe for non-strings

LimeWub:
I believe this solves the mystery problem I'm getting in atom-core when trying to filter banners with non-string values. 
Screenshot 2023-03-09 at 16 33 58

The message is not exactly the same but I've pinpointed the glitching happens with non-string values (0-1 number) so I think this thread I got inspired from is relevant: https://github.com/TanStack/table/issues/4280#issuecomment-1244878232 (but made it generic for all values rather than just numbers - can change it to only do it for numbers if you think this is better)

## Before and after
**with this sandbox:**

```
import { createColumnHelper } from '@tanstack/react-table'
import { Flex, globalCss, DataTable, Tooltip } from '../src'

const columnHelper = createColumnHelper<{
  name: string
  hobby: string,
  testNumber: number,
  testBoolean: boolean,
}>()

const columns = [
  columnHelper.accessor('name', {
    cell: (info) => info.getValue(),
    header: 'Name'
  }),
  columnHelper.accessor('hobby', {
    cell: (info) => info.getValue(),
    header: 'Hobby'
  }),
  columnHelper.accessor('testNumber', {
    cell: (info) => info.getValue(),
    header: 'Test Number'
  }),
  columnHelper.accessor('testBoolean', {
    cell: (info) => info.getValue(),
    header: 'Test Boolean'
  }),
  columnHelper.display({
    cell: (info) => <button>do something</button>,
    header: 'Actions'
  })
]

const data = [
  { name: 'chrissy', hobby: 'bare-knuckle boxing', testNumber: 1, testBoolean: true },
  { name: 'agatha', hobby: 'crossfit', testNumber: 2, testBoolean: false },
  { name: 'betty', hobby: 'acting', testNumber: 3, testBoolean: true },
  { name: 'denise', hobby: 'bare-knuckle boxing', testNumber: 4, testBoolean: false },
]

const App = () => {
  const [sortState, setSortState] = React.useState({
    column: 'name',
    direction: 'asc'
  })

  return (
    <Tooltip.Provider>
      <Flex
        css={{
          minHeight: '100vh',
          justifyContent: 'center',
          alignItems: 'center',
          flexDirection: 'column'
        }}
      >
        <DataTable columns={columns} data={data} defaultSort={sortState}>
          <DataTable.GlobalFilter
            css={{ mb: '$4', maxWidth: '300px' }}
            label="User search"
            placeholder="Search for a user"
            hideLabel
          />
          <DataTable.Table css={{ mb: '$4' }} theme="light" striped />
          <DataTable.Pagination />
        </DataTable>
      </Flex>
    </Tooltip.Provider>
  )
}

```

**before:**

https://user-images.githubusercontent.com/6905473/224087798-124cc957-79b9-41fc-9b9b-e7581dc1e272.mov

**after:**

https://user-images.githubusercontent.com/6905473/224088001-fc43577a-5fff-43d8-a90d-8c7899b3d6c9.mov


lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -126,7 +126,11 @@ export const DataTableProvider = ({
     getFilteredRowModel: getFilteredRowModel(),
     onPaginationChange: isPaginated ? setPaginationState : undefined,
     onSortingChange: setSorting,
-    onGlobalFilterChange: setGlobalFilter
+    onGlobalFilterChange: setGlobalFilter,
+    globalFilterFn: (row, columnId, filterValue) => {
+      const safeValue = String(row.getValue(columnId))
```
hollg, line 131: Are there any types which we should be able to filter on but which don't convert to a string like this? I can't immediately think of any way that we'd be losing functionality here but it feels possible?

LimeWub: Do you prefer I do the safeValue only for numbers?

hollg: Can we do it for numbers and bools? They can both be cast with `String()`

LimeWub: changed :)

hollg (approved):

lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -126,7 +126,17 @@ export const DataTableProvider = ({
     getFilteredRowModel: getFilteredRowModel(),
     onPaginationChange: isPaginated ? setPaginationState : undefined,
     onSortingChange: setSorting,
-    onGlobalFilterChange: setGlobalFilter
+    onGlobalFilterChange: setGlobalFilter,
+    globalFilterFn: (row, columnId, filterValue) => {
+      const search = filterValue.toLowerCase()
+
+      let value = row.getValue(columnId) as string
```
D7Torres, line 133: can't `row.getValue(columnId)` be other than a string (would it be a number?)

If so, should we cast it to string, instead of asserting it is a string?

lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -126,7 +126,17 @@ export const DataTableProvider = ({
...
+      const valueType = typeof value
+      if (valueType === 'number' || valueType === 'boolean')
+        value = String(value)
+
+      return value?.toLowerCase().includes(search)
```
D7Torres, line 138: Do you think we could have this instead?

```
const value = row.getValue(columnId)
const valueType = typeof value
const isValidType = (valueType === 'number' || valueType === 'boolean' || valueType === 'string')
const stringValue = String(value)

return isValidType && value?.toLowerCase().includes(search)
```

Advantages
* you lose the `let`
* you are not asserting something that is not true (in line 133, `value` is not a `string` always, even if eventually will always end up being a string)

Disadvantages
* You cast to String also when it's already a string. But this is likely a no-op.

lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -126,7 +126,17 @@ export const DataTableProvider = ({
     getFilteredRowModel: getFilteredRowModel(),
     onPaginationChange: isPaginated ? setPaginationState : undefined,
     onSortingChange: setSorting,
-    onGlobalFilterChange: setGlobalFilter
+    onGlobalFilterChange: setGlobalFilter,
+    globalFilterFn: (row, columnId, filterValue) => {
+      const search = filterValue.toLowerCase()
```
D7Torres, line 131: Pedantic (sorry), but should we call this `searchTerm` or something like that?

lib/src/components/data-table/DataTableContext.tsx
LimeWub: My original version had a simple cast to a `String()` for any value regardless of type. I think this version is taking us back to that but in a more verbose way.
I'd be happier to stick with what I have now as it's closest to the comment it has been copy/pasta'd from which appears to be working for people and reassess if we have more bugs with it in the future.

D7Torres: > My original version had a simple cast to a String() for any value regardless of type
and what is the problem with that?

The previous version disappeared while I was looking at it, so not 100% sure, but I think you would be cover becacuse of `isValidType` in the return.

So the problem was not casting everything, but using everything for the search. In what I proposed, you still only use the value if it is string, boolean or number

D7Torres (approved):

PR #491, opened 2023-03-10
feat: add Dos and Donts editor component

LimeWub:
Adds the DosAndDonts component for use in the CMS. This is a component designers use to indicate correct and incorrect usage of a component.

JIRA: https://atomlearningltd.atlassian.net/browse/DS-345

**Editor:**
Screenshot 2023-03-10 at 17 09 32

**Looks like this on the page:**
Screenshot 2023-03-10 at 17 07 54


documentation/components/markdown/editor-component/DosAndDonts.tsx
```diff
@@ -0,0 +1,52 @@
...
+} from '@atom-learning/components'
+import * as React from 'react'
+
+type TDosAndDontsItemProps = { image: string, type: 'do' | 'avoid' | 'dont', description: string }
+
+const DosAndDontsItem: React.FC<TDosAndDontsItemProps> = ({ image, type, description }) => {
+  const typeBorderColor = type === 'do' ? '$success' : type === 'dont' ? '$danger' : '$warning'
+  const typeText = type === 'do' ? 'Do' : type === 'dont' ? "Don't" : 'Avoid'
+  return (
+    <Box as="li" css={{ listStyle: 'none' }}>
```
hollg, line 16: Is this actually `li` being rendered inside a list?

hollg (approved):
Two small comments from me but otherwise 🚀

documentation/components/admin/editor-component/DosAndDontsEditorComponent.tsx
```diff
@@ -0,0 +1,39 @@
...
+      name: "items",
+      label: "Items",
+      widget: "list",
+      summary: "{{fields.type}}: {{fields.description}}",
+      fields: [
+        { label: "Image", name: "image", widget: "image" },
+        {
+          label: "Type", name: "type", widget: "select", options: [
+            { label: "Do", value: "do" },
+            { label: "Avoid", value: "avoid" },
```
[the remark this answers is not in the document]
MiguelGDLR: **Avoid** is more cautionary, we recommended not to but you actually can for edge cases...
**Don't** is just a no-no. More obligatory

do you know if we can use a better word?

documentation/components/markdown/editor-component/DosAndDonts.tsx
LimeWub: oopsie doopsie! that's where the copy/pasta got me :p will add the ul

documentation/components/markdown/editor-component/DosAndDonts.tsx
```diff
@@ -0,0 +1,52 @@
...
+      <Box css={{
+        aspectRatio: "332/192",
```
MiguelGDLR, line 18: shouldn't it be 16/9?

LimeWub: I got the number from the design + adding the border - 16/9 here was including the border - giving the image itself the wrong aspect (slightly too short)

documentation/components/admin/editor-component/DosAndDontsEditorComponent.tsx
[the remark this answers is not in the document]
LimeWub: I think if the words have been used in the designs for a while they make sense for designers. I can't think of a better/different set of words - but I'd be assuming whatever set we go with is pretty standard in design systems.

documentation/components/markdown/editor-component/DosAndDonts.tsx
D7Torres: I wonder if [box-sizing](https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing) css attribute can help here, to include / not include the border + retaining a more descriptive 16/9.

If it doesn't help, could you add your explanation above as a comment, please?

LimeWub: it doesn't `content-box` messes up the padding logic. `border-box` is the default.
`padding-box` is sadly now deprecated (bring it back T_T)

documentation/components/admin/editor-component/DosAndDontsEditorComponent.tsx
```diff
@@ -0,0 +1,39 @@
...
+      ],
+    }
+  ],
+  pattern:
+    /^<DosAndDonts items={(.*?)} \/>$/ms,
+  fromBlock: function (match) {
+    return {
+      items: JSON.parse(match[1])
+    };
+  },
```
D7Torres, line 30: for my benefit, can you remind me what `fromBlock` does?

documentation/components/markdown/editor-component/DosAndDonts.tsx
```diff
@@ -0,0 +1,52 @@
...
+  Grid,
+  Box,
+  Heading
+} from '@atom-learning/components'
+import * as React from 'react'
+
+type TDosAndDontsItemProps = { image: string, type: 'do' | 'avoid' | 'dont', description: string }
+
+const DosAndDontsItem: React.FC<TDosAndDontsItemProps> = ({ image, type, description }) => {
+  const typeBorderColor = type === 'do' ? '$success' : type === 'dont' ? '$danger' : '$warning'
```
D7Torres, line 13: I've always been told not to nest ternary conditionals, as it makes the code less readable quickly. Maybe we should move this logic into a small function with early returns.

D7Torres: Also, should `'do'`, `'dont'` and `'avoid'` be an enum or something? or am I overthinking it?

documentation/components/markdown/editor-component/DosAndDonts.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const typeText = type === 'do' ? 'Do' : type === 'dont' ? "Don't" : 'Avoid'
+  return (
+    <Box as="li" css={{ listStyle: 'none' }}>
+      <Box css={{
+        aspectRatio: "332/192",
+        background: '$base1',
+        display: 'flex',
```
D7Torres, line 20: Should we use `<Flex>` instead of `<Box>` with display flex? I'm not precious about it, I would like to understand myself what's the best thing to do for when I have to do it.

LimeWub: eh ye tbh - could be a flex

documentation/components/admin/editor-component/DosAndDontsEditorComponent.tsx
LimeWub: It is reading from the `.md` -> gets the matches inside the block (text) using the regex above this bit -> then in this bit it makes the properties/values it got usable to be populates the admin editor component.

documentation/components/markdown/editor-component/DosAndDonts.tsx
D7Torres: But in the line we are commenting you use again a literal `'do'`.

I think my point would be to have them as a
```
enum ItemType {
  DO: 'do',
  DONT: 'dont',
  AVOID: 'avoid'
}
```

So then you can use it like `type === ItemType.DO`

But maybe I **am** overthinking it.

D7Torres (approved):
Approved pending to add the ratioAspect comment

PR #493, opened 2023-03-17
[CMS] Cards component

LimeWub:
Adds the **`Cards`** component. Basically a Grid of Cards which have Image/Text data. 
They are taggable and support filtering via these tags as well as searching them via a search bar.
The original use case is for easy discoverability by placing them on the landing page (see Design) but the component is _**fully**_ reusable for any sort of similar data.

**JIRA:** 
v1: https://atomlearningltd.atlassian.net/browse/DS-322
v2: https://atomlearningltd.atlassian.net/browse/DS-323
v3: https://atomlearningltd.atlassian.net/browse/DS-251

## Design
Screenshot 2023-03-17 at 17 10 48

## NEW WIDGET: creatableSelect
This PR also introduces the *`CreatableSelectWidget`* which is a custom NetlifyCMS select field using [`Creatable from 'react-select/creatable'`](https://react-select.com/creatable) and heavily based on the original `SelectWidget` which lives in NetlifyCMS internals. 
It expands on the `select` by allowing for the creation of new tags and, via using `persist_options: true`, also allows sharing options between fields on the page and persisting them until page refresh upon deletion.

It was added to this PR as it's necessary for the tagging functionality of the `Cards`.

## Videos
(Watermark because I sped it up to spare you 😛  )

https://user-images.githubusercontent.com/6905473/225970435-bc6ac1a9-fe15-40c0-af77-9a389c27fd5d.mov


documentation/components/markdown/editor-component/IconTokenList.tsx
```diff
@@ -84,7 +84,7 @@ export const IconTokenList: React.FC<IconTokenListProps> = ({ icons: specificIco
         name="icon-search"
         placeholder="Search for an icon"
         css={{
-          maxWidth: 300,
```
LimeWub, line 87: This is a minor change to match this search width with the new search. I just think it's nice if they match and totally not a big difference :)

documentation/components/markdown/editor-component/DosAndDonts.tsx
```diff
@@ -23,11 +23,9 @@ const DosAndDontsItem: React.FC<TDosAndDontsItemProps> = ({ image, type, descrip
         borderColor: typeBorderColor,
```
LimeWub, line 23: Changes in here are to match to `Cards` which is a super similar component.

Have talked w Miguel about the aspect-ration hack numbers and he's said it's nbd and since we'll be introducing aspectRatio tokens I just went ahead and changed it to the "16/9" value - which it will be when using the token.

maxHeight on the image fixes a bug where aspect-ratio isn't being respected

documentation/components/markdown/editor-component/Cards.tsx
```diff
@@ -0,0 +1,154 @@
...
+const CardsItem: React.FC<TCardsItemProps> = ({ image, heading, description, link }) => {
+  const isLink = !!link?.href;
+  const elementSpecificProps = isLink
+    ? {
+      as: 'a',
+      href: link.href,
+      target: link.isExternal ? '_blank' : '_self',
+      rel: link.isExternal ? 'noopener noreferrer nofollow' : false
+    }
+    : { as: 'div' }
```
hollg, line 27: Doesn't `Box` default to a `div`?

documentation/components/markdown/editor-component/Cards.tsx
```diff
@@ -0,0 +1,154 @@
...
+            mx: 'auto',
+            overflowX: 'auto',
+          }}>
+          <ChipToggleGroup
+            wrap="nowrap"
+            type="multiple"
+            value={selectedTags}
+            onValueChange={handleSelectedTagsValueChange}
+          >
+            <ChipToggleGroup.Item value="">All</ChipToggleGroup.Item>
```
hollg, line 129: What's the reason for this item having an empty string for its `value`? I.e. could it be `"all"` instead? I think it would make the logic in `handleSelectedTagsValueChange` a little easier to parse, because currently it's not clear from that function what an empty string has to do with selecting all.

documentation/components/markdown/editor-component/Cards.tsx
```diff
@@ -0,0 +1,154 @@
...
+    // "All" unselects all other tags
+    const hasJustToggledAll = newSelectedTags[newSelectedTags.length - 1] === ''
+    if (hasJustToggledAll) return setSelectedTags([''])
+
+    // Always keep "All" selected as a fallback if no value selected
+    if (newSelectedTags.length === 0) return setSelectedTags([''])
+
+    // Remove "All" from the tags array if other values have been selected
+    if (newSelectedTags.length > 1) return setSelectedTags(newSelectedTags.filter(Boolean))
+    setSelectedTags(newSelectedTags)
```
hollg, line 88: I think a space here will make it clearer that the final `setSelectedTags` call isn't related to this conditional
```suggestion
    if (newSelectedTags.length > 1) return setSelectedTags(newSelectedTags.filter(Boolean))
    
    setSelectedTags(newSelectedTags)
```

documentation/components/markdown/editor-component/Cards.tsx
```diff
@@ -0,0 +1,154 @@
...
+      <Grid
+        as={StyledUl}
+        css={{
+          width: '100%',
+          gap: '$5 $4',
+          gridTemplateColumns: '1fr',
+          '@sm': { gridTemplateColumns: '1fr 1fr' },
+        }}>
+        {items.map((props) => {
+          if (!checkIfMatchingTags(props.tags, selectedTags) || !checkIfMatchingSearch(props, searchValue)) return null;
```
hollg, line 143: Is this `return null` to avoid iterating over the array twice, once for a filter and then a second for a map? I don't think I've seen this done before. Seems neat!

documentation/components/markdown/editor-component/Cards.tsx
LimeWub: That's how we've done default tabs/toggle-items in internal-tools apps so I just went with the same patter of 'all' or 'no filter' being an empty string ^^

documentation/components/markdown/editor-component/Cards.tsx
LimeWub: Ah it's just for convenience tbh. I could had moved the filtering outside - maybe even put them in a memo. I didn't overthink it so to me it was just the easiest way to do the filter :p

documentation/components/markdown/editor-component/Cards.tsx
```diff
@@ -0,0 +1,154 @@
...
+}
+
+const checkIfMatchingSearch = (item, searchValue) => {
+  const itemStringified = JSON.stringify(item)
+  return itemStringified.toLowerCase().includes(searchValue.toLowerCase())
+}
+
+export const Cards: React.FC<CardsProps> = ({ showSearch, showTagsFilter, items }) => {
+  const [searchValue, setSearchValue] = React.useState('')
+  const handleSetSearchValue = React.useCallback(debounce(500, (value) => setSearchValue(value.toLowerCase())), [])
```
D7Torres, line 74: Sorry, I wished I could tell you how to fix it 🤷‍♂️ 

image

documentation/components/markdown/editor-component/Cards.tsx
D7Torres: I agree that it would make `handleSelectedTagsValueChange` easier to understand

D7Torres (commented):
Great PR, just some suggestion/pickiness comments

documentation/components/admin/editor-component/CardsEditorComponent.tsx
```diff
@@ -0,0 +1,72 @@
...
+    /^<Cards showSearch={(.*?)} showTagsFilter={(.*?)} items={(.*?)} \/>$/ms,
+  fromBlock: function (match) {
+    return {
+      showSearch: match[1] === 'true',
+      showTagsFilter: match[2] === 'true',
+      items: JSON.parse(match[3])
+    };
+  },
+  toBlock: function (data) {
+    const itemsString = data?.items ? JSON.stringify(data.items) : JSON.stringify([]);
```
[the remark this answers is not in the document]
LimeWub: huh - ye you're right. I'm just doing it too many times in these components to replace 😅 

Screenshot 2023-03-22 at 12 45 03

Will leave as is if it's no biggie - but ye you're right. Maybe we can replace all of them separately.

documentation/components/markdown/editor-component/Cards.tsx
```diff
@@ -0,0 +1,154 @@
...
+  return (
+    <Box {...elementSpecificProps} css={{ color: 'inherit', textDecoration: 'none', display: 'block' }}>
+      <Box css={{
+        background: '$base1',
+        display: 'flex',
+        alignItems: 'center',
+        justifyContent: 'center',
+        mb: '$2',
+        p: '$2',
+        aspectRatio: "16/9"
```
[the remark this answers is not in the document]
LimeWub: 👀 Maybe my prettier setup?

documentation/components/markdown/editor-component/Cards.tsx
```diff
@@ -0,0 +1,154 @@
...
+      }}>
+        {image && <Image src={image} alt='' css={{ maxHeight: '100%' }} />}
```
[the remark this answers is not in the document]
LimeWub: Yes it's intentional as I'm assuming the image content will just be repeated in the heading or description - so I haven't added an `alt` field for the image in the CMS.

documentation/components/markdown/editor-component/Cards.tsx
LimeWub: Erm - that's weird. It's actually not flagging for me 🤷

documentation/components/admin/editor-component/CardsEditorComponent.tsx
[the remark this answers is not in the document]
D7Torres: 🤔 why is it a problem to replace it in all of those places in one go?

documentation/components/markdown/editor-component/Cards.tsx
```diff
@@ -0,0 +1,154 @@
...
+  Grid,
+  Box,
+  Heading,
+  Stack,
+  ChipToggleGroup,
+  SearchInput,
+  styled
+} from '@atom-learning/components'
+import { intersection } from "lodash";
+import * as React from 'react'
```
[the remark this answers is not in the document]
D7Torres: If you remove it from everywhere then things are consistent too.

I guess the benefit is being less verbose, like when destructuring.

`useCallback`, instead of `React.useCallback`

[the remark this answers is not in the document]
LimeWub: Sometimes doing `useCallback` and sometimes doing `React.useCallback` is what I mean by being inconsistent.

documentation/components/admin/editor-component/CardsEditorComponent.tsx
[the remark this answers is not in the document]
LimeWub: Because it's out of scope for this PR and not necessary? I also think it's a bit more clear it's doing a fallback to a `[]` but ye - maybe could compromise to save some computational power. I just don't want to replace it in this PR.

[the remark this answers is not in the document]
D7Torres: Sure, it was a non-blocking suggestion anyway

documentation/components/markdown/editor-component/Cards.tsx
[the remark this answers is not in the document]
D7Torres: oh, in the case we got rid of `import * as React` we could kill all of the `React.` so we would be consistent

D7Torres (approved):
If you're happy with the debounce warning, maybe we should add a eslint ignore comment for that line.

That way, if some day we take the task of eliminating all the warnings, we don't worry about that one.

PR #496, opened 2023-03-22
feat: Tile

LimeWub:
**Tile** is a fundamental layout component used as a generic container to build the base of panels, cards, lists and other content components that group information. 

**TileInteractive** is the button variation which allows for href or onClick handlers.

**TileToggleGroup** is the... toggle group version - which allows for selecting the tiles.

All built wrapped with `ColorScheme` to allow for multiple colour palette setups.

JIRA: https://atomlearningltd.atlassian.net/browse/DS-352
Figma: https://www.figma.com/file/KjgQh7eTwEfJmH6e56k9mX/DS-Documentation?node-id=6096-44675&t=bvcqbBYJGTw9NesH-0

## Screenshots/Videos

**Tile (base)**
Screenshot 2023-03-22 at 12 32 20
Screenshot 2023-03-22 at 12 32 29

**TileInteractive**

https://user-images.githubusercontent.com/6905473/226906008-fdb8ac88-e7f3-48ce-949c-96e59256f86e.mov

**TileToggleGroup**

https://user-images.githubusercontent.com/6905473/226906353-c404a84d-0791-47fe-b1f6-e9dda3969877.mov

### DIFF
--- a/lib/src/components/tile-interactive/TileInteractive.test.tsx
+++ b/lib/src/components/tile-interactive/TileInteractive.test.tsx
@@ -1,0 +1,21 @@
+import { render } from '@testing-library/react'
+import { axe } from 'jest-axe'
+import * as React from 'react'
+
+import { TileInteractive } from '.'
+
+const TileInteractiveImplementation = () => (
+  <TileInteractive css={{ m: 'auto', size: 100 }}>A</TileInteractive>
+)
+
+describe(`TileInteractive component`, () => {
+  it('renders', async () => {
+    const { container } = render(<TileInteractiveImplementation />)
+    expect(container).toMatchSnapshot()
+  })
+
+  it('has no programmatically detectable a11y issues', async () => {
+    const { container } = render(<TileInteractiveImplementation />)
+    expect(await axe(container)).toHaveNoViolations()
+  })
+})
--- a/lib/src/components/tile-interactive/TileInteractive.tsx
+++ b/lib/src/components/tile-interactive/TileInteractive.tsx
@@ -1,0 +1,60 @@
+import React from 'react'
+import { styled } from '~/stitches'
+import { Tile } from '~/components/tile'
+import { focusVisibleStyleBlock } from '~/utilities'
+import { NavigatorActions } from '~/types'
+
+const StyledTileInteractive = styled(Tile, {
+  '&[data-disabled]': {
+    opacity: 0.3,
+    cursor: 'not-allowed'
+  },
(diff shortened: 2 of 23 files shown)


MiguelGDLR:
I think i forgot to mention in my documentation. Can the toggle be set up to do multi or single select?

LimeWub:
@MiguelGDLR  Yes it's a full fledged toggle group :)

lib/src/components/tile-interactive/TileInteractive.test.tsx
D7Torres, line 1: I wonder if we should add a test here to test the clicking

lib/src/components/tile-toggle-group/TileToggleGroupItem.tsx
```diff
@@ -0,0 +1,46 @@
...
+  '&[data-state="on"]': {
+    bg: '$interactive1',
+    borderColor: '$accent8',
+    '&::before': {
+      content: '',
+      position: 'absolute',
+      inset: -1, // account for 1px Tile border
+      boxShadow: 'inset $colors$accent8 0px 0px 0px 2px',
+      borderRadius: 'inherit',
+      zIndex: 1 // so it's over anything nested which touches sides
```
D7Torres, line 27: Hey, I understand the "so it's over anything nested", but I don't understand "which touches sides". Why does it matter if something touches or doesn't touch the side?

LimeWub: Because it's a boxShadow used as a border and I want it to be over any image or full width element nested inside of it so it appears above it. But it only matters for elements which touch the sides of the container (as they will be over the border/shadow)
Actually I think I can replace 'boxShadow' with 'border' in this pseudo-element and keep the same functionality (zIndex + comment will still remain though)

lib/src/components/tile-interactive/TileInteractive.tsx
```diff
@@ -0,0 +1,60 @@
...
+> & { type?: string } & NavigatorActions
+
+export const TileInteractive = React.forwardRef<
+  HTMLButtonElement,
+  TTileInteractiveProps
+>(({ onClick, href, type = 'button', ...rest }, ref) => {
+  const isLink = !!href
+  const elementSpecificProps = isLink
+    ? {
+        asWorkaround: 'a' as React.ElementType,
```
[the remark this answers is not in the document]
D7Torres: nvm! found the comment

hollg (approved):

PR #497, opened 2023-03-22
Update ColorScheme with new confirmed theme colours and for new design requirements

LimeWub:
- Updates colours which are set up within ColorScheme to use [the new colours added in theme](https://github.com/Atom-Learning/theme/pull/23)
- Removes all mentions on `slate` in favour of `grey`
- Moves distinction of `loContrast1` (starts from white/base colour) vs `loContrast2` (starts from lightest colour in the accent palette) to the pallets themselves. Having 2 palettes per colour ie. `grey` -> `grey1` (starts from white, finishes one colour before last), `grey2` (starts from `grey100`, includes the full palette). [This is done to accommodate designs like `Pagination` which do 3 colour stepping for both `base` and `accent` elements depending on the background colour they are placed on.](https://www.figma.com/file/KjgQh7eTwEfJmH6e56k9mX/DS-Documentation?node-id=5093-38072&t=bvcqbBYJGTw9NesH-0)
- Changes available options for **interactive**  from `loContast1`, `loContast2`, `hiContrast1`, `hiContast2` to just `loContrast`, `hiContrast`
- Changes available options for **base** from just `slate` to `grey1`, `grey2`, `blue1`, `blue2`, `purple1`, `purple2` 
- Changes available options for **accent** from `slate`, `blue` to `grey1`, `grey2`, `blue1`, `blue2`, `purple1`, `purple2` 
- Updates colours where needed to map correctly to the new palette which has 13 total colours (including base) from the old palette which had 10 total colours.
- Updates rest of the colours to the-old-colour + 1 step to look right when using a `{colour}1` palette (which starts from base "white" - or "black" if we ever add dark mode themes)
- Removes usage of alpha versions of colours in `<TabsTrigger />` in favour of a low opacity pseudo element as the new palette does not provide alpha colours.
- Updates documentation for everything changed

**Visually this should not have any /noticeable/ changes but will need snapshots updated for the affected components in core**


LimeWub:
I don't understand **why** Test & validate fails 👀 
It seems to be for this`lib/src/experiments/color-scheme/index.ts` but afaik it's legit?

Any help appreciated!

documentation/content/components.media.avatar.md
```diff
@@ -236,19 +236,19 @@ tabs:
...

+      **Initial**
 
 
       The user's name initial appear in the component's center with white background and border.
 
 
-      **Icon**

+      **Icon**
```
D7Torres, line 251: I'd love to learn someday how GitHub comes up with these diffs

D7Torres (approved):

hollg (approved):

PR #498, opened 2023-03-31
feat: add emptystate to DataTable

andoulla:
This adds the option to add an EmptyState to the DataTable. 
It requires the consumer to implement the state but at the moment this is all we need to unblock all the tickets for the table migrations as part of these 

The next iteration would be to implement a default empty state, however, this is blocked by the bigger discussion about how the DS handles generic illustrations so leaving this out of scope for this PR.

image

**Note: Under WIP to get feedback and I need some time to fix a few broken tests**

### DIFF
--- a/lib/src/components/data-table/DataTable.test.tsx
+++ b/lib/src/components/data-table/DataTable.test.tsx
@@ -12,6 +12,7 @@ import { Button } from '../button'
 import { Text } from '../text'
 import { Tooltip } from '../tooltip'
 import { DataTable } from '.'
+import { EmptyState } from '../empty-state'
 
 const columnHelper = createColumnHelper<{
   id: number
@@ -134,6 +135,24 @@ describe('DataTable component', () => {
 })
 
 describe('DataTable.Pagination component', () => {
+  it("doesn't render if table is empty", () => {
+    render(
+      <Wrapper>
+        <DataTable
+          columns={columns}
+          data={[]}
+          initialState={{ pagination: { pageIndex: 0, pageSize: 5 } }}
+        >
+          <DataTable.Table sortable />
+          <DataTable.Pagination />
+        </DataTable>
+      </Wrapper>
+    )
+
+    const nextPageButton = screen.queryByLabelText('Next page')
+    expect(nextPageButton).not.toBeInTheDocument()
+  })
+
   it('Displays the correct page number', async () => {
     render(
       <Wrapper>
@@ -221,7 +240,21 @@ describe('DataTable.Pagination component', () => {
   })
 })
 
-describe('DataTable Search component', () => {
+describe('DataTable GlobalFilter component', () => {
+  it("doesn't render if data is empty", () => {
+    render(
+      <Wrapper>
+        <DataTable columns={columns} data={[]}>
+          <DataTable.GlobalFilter label="Search" />
+          <DataTable.Table sortable />
+        </DataTable>
+      </Wrapper>
+    )
+
+    const nextPageButton = screen.queryByLabelText('Next page')
+    expect(nextPageButton).not.toBeInTheDocument()
+  })
+
   it('Filters table based on any column', async () => {
     render(
       <Wrapper>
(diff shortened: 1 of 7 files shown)


lib/src/components/data-table/DataTableEmptyState.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+type DataTableEmptyStateProps = React.ComponentProps<typeof EmptyState>
+
+export const DataTableEmptyState: React.FC<DataTableEmptyStateProps> = ({
+  children,
+  ...rest
+}) => {
+  const { asyncDataState, getTotalRows } = useDataTable()
+
+  if (asyncDataState !== AsyncDataState.PENDING && getTotalRows() !== 0)
```
hollg, line 15: Does this mean that we'll show the empty state while async data is loading? I thought we had a loading state for that already 🤔

andoulla: oh I see what I did wrong, I wanted to make it not show the empty state if the component is loading. 

this needs to be 
```if ((asyncDataState !== AsyncDataState.PENDING && getTotalRows()==0) || getTotalRows() !== 0)```
Good catch

Comment (andoulla) [reply lib/src/components/data-table/DataTableEmptyState.tsx]: done!

File: lib/src/components/data-table/DataTableTable.tsx [hunk 2]
```diff
@@ -23,9 +23,11 @@ export const DataTableTable: React.FC<DataTableTableProps> = ({
   numberOfStickyColumns = 0,
   ...props
 }) => {
-  const { asyncDataState } = useDataTable()
+  const { asyncDataState, getTotalRows } = useDataTable()
   const isPending = asyncDataState === AsyncDataState.PENDING
 
+  if (getTotalRows() === 0 && !isPending) return null
```
LimeWub, line 29: Can this be the same exact if as the empty state one? (Maybe with a bang "!" for opposite)

`const isNotEmpty = rowCount > 0 || isPending`
`const isEmpty = !isPending && rowCount === 0`

bit easier to understand/map in our brain its the same type of check.

lib/src/components/data-table/DataTable.test.tsx
```diff
@@ -221,7 +240,21 @@ describe('DataTable.Pagination component', () => {
...
+      <Wrapper>
+        <DataTable columns={columns} data={[]}>
+          <DataTable.GlobalFilter label="Search" />
+          <DataTable.Table sortable />
+        </DataTable>
+      </Wrapper>
+    )
+
+    const nextPageButton = screen.queryByLabelText('Next page')
+    expect(nextPageButton).not.toBeInTheDocument()
```
D7Torres, line 255: I think this is a copy+paste mistake. It should target the global filter

lib/src/components/data-table/DataTableTable.tsx
andoulla: will go for isEmpty

lib/src/components/data-table/pagination/Pagination.tsx
```diff
@@ -37,6 +37,9 @@ export const Pagination: React.FC<PaginationProps> = (props) => {
 
   const { pagination: paginationState } = getState()
 
+  if (asyncDataState !== AsyncDataState.PENDING && getTotalRows() === 0)
```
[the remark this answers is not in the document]
andoulla: I would rather keep it conceptually just for the empty state if that is ok

lib/src/components/data-table/DataTable.test.tsx
andoulla: thanks, good catch false positves are easy to miss

lib/src/components/data-table/DataTableEmptyState.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export const DataTableEmptyState: React.FC<DataTableEmptyStateProps> = ({
+  children,
+  ...rest
+}) => {
+  const { asyncDataState, getTotalRows } = useDataTable()
+  const rowCount = getTotalRows()
+  const isPending = asyncDataState === AsyncDataState.PENDING
+
+  if ((!isPending && rowCount !== 0) || isPending) return null
```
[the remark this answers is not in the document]
andoulla: done

LimeWub (approved):

lib/src/components/data-table/DataTableGlobalFilter.tsx
```diff
@@ -16,9 +17,18 @@ export const DataTableGlobalFilter: React.FC<DataTableSearchProps> = ({
...
+  const {
+    setGlobalFilter,
+    getState,
+    resetPagination,
+    getTotalRows,
+    asyncDataState
+  } = useDataTable()
   const { globalFilter } = getState()
 
+  if (asyncDataState !== AsyncDataState.PENDING && getTotalRows() === 0)
```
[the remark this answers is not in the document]
andoulla: done

D7Torres (approved):

lib/src/components/data-table/pagination/Pagination.tsx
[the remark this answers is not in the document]
andoulla: Yes I would be more confident to run that by them, so Ithink it can happen when the work for adding the default happens

PR #499, opened 2023-04-14
fix: empty page when navigating from a tab that is not the first

D7Torres:
[JIRA](https://atomlearningltd.atlassian.net/browse/DS-346)

## Description
When navigating from a component tab that was not the first tab to a component that only has one tab an empty screen was shown. This has been fixed

## Implementation
The problem is that the current tab was stored in the tab provider, and was not reset when navigating from one component to other.

~Going from un uncontrolled approach to a controlled approach for the tabs and resetting it when the `slug` changes did the trick.~

After some review comments, I tried usin `key={slug}` so the tabs re-render. It doesn't work in the `<Tab>` component directly, so I did it in the parent component.

### Before

https://user-images.githubusercontent.com/4931116/231988038-dff1abd6-3d19-4f08-981d-023c63d68d91.mov

### After

https://user-images.githubusercontent.com/4931116/231988085-3bf3e9b9-0c8a-4eb2-a3aa-da5fbe06c98f.mov

### DIFF
--- a/documentation/components/page/Layout.tsx
+++ b/documentation/components/page/Layout.tsx
@@ -3,30 +3,43 @@ import { Heading, Tabs, Box, StackContent } from '@atom-learning/components'
 import Head from 'next/head'
 import { Container, Header, Footer, Links, Pagination } from '.'
 import { components } from '~/components/markdown/index'
-import { MDXRemote } from "next-mdx-remote";
+import { MDXRemote } from 'next-mdx-remote'
 import { TDynamicPage } from '~/lib/types/DynamicPage'
+import { useState, useEffect } from 'react'
 
 export const Layout: React.FC<TDynamicPage> = (props) => {
-  const { links, slug, tabs, title } = props;
-
+  const { links, slug, tabs, title } = props
   const tabsLength = tabs?.length
-  const hasContent = !!tabsLength;
+  const hasContent = !!tabsLength
+
+  const [activeTab, setactiveTab] = useState('tab0')
+
+  useEffect(() => {
+    setactiveTab('tab0')
+  }, [slug])
 
   return (
-    <Tabs defaultValue="tab0">
+    <Tabs value={activeTab} onValueChange={setactiveTab}>
       <Head>
-        <title>
-          {title ? `${title} | ` : ''}Atom Learning Design System
-        </title>
+        <title>{title ? `${title} | ` : ''}Atom Learning Design System</title>
       </Head>
-      <Box as="article" css={{ minHeight: '100vh', display: 'flex', flexDirection: 'column' }}>
+      <Box
+        as="article"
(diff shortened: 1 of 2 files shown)


documentation/components/page/Layout.tsx
```diff
@@ -3,30 +3,43 @@ import { Heading, Tabs, Box, StackContent } from '@atom-learning/components'
...
+
+  const [activeTab, setactiveTab] = useState('tab0')
+
+  useEffect(() => {
+    setactiveTab('tab0')
+  }, [slug])
 
   return (
-    <Tabs defaultValue="tab0">
+    <Tabs value={activeTab} onValueChange={setactiveTab}>
```
thomasdigby, line 22: Could we use a `key` to force a re-render? e.g. `<Tabs key={slug}>`?

D7Torres: I'm so glad you're suggesting that, because it's the first thing I tried. For some reason didn't work. But it should, right?? Do you want to pair a little bit to try to figure out why it doesn't work?

hollg: Not sure which of us you're asking but I can spend half an hour or so on this with you this morning if that's useful!

D7Torres: I was asking Tom, because GitHub no longer refreshes for me when I get new comments (so annoying). But I'm happy to pair with any of you :)

D7Torres: That re-render didn't work, it looks like the way radix tabs work is with a Context provider, and the value of the selected tab must persist between rerenders.

However using the `key={slug}` in the parent component (Layout) did the trick.
0a98ac67e4057e503750639e4637c7ec18ec571b

MiguelGDLR: Hey, i've noticed the word 'slug' here so just wanted to point out that I have a [ticket request](https://atomlearningltd.atlassian.net/browse/DS-339?atlOrigin=eyJpIjoiN2ZmM2ZhNjNmY2Q3NDdlM2I0YWQyM2QyZjk3MjViNTAiLCJwIjoiaiJ9) for slugs because that would be needed to link to specific pages.  In case you want to kill 2 birds with 1 stone

D7Torres: Thanks for raising it! In this case, if I'm not mistaking that's going to be quite a bigger change, so it makes sense to keep it separated

thomasdigby: Do we still need this change if we have the `key` in `PageLayout`?

D7Torres: Good point, of course, we don't.
7a1ca4adf79da407a2d9a0b14131d145f268c785

thomasdigby (approved):

hollg (approved):

LimeWub (approved):

PR #501, opened 2023-04-19
Feat: Promo Banner (Regular)

avirati:
https://atomlearningltd.atlassian.net/browse/DS-255
https://atomlearningltd.atlassian.net/browse/STU-1306

image


lib/src/components/banner/BannerButton.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { Button } from '../button'
+import { useBannerContext } from './BannerContext'
+
+export const BannerButton: React.FC<React.ComponentProps<typeof Button>> = ({
+  appearance,
+  ...props
+}) => {
+  const { emphasis, size } = useBannerContext()
+
+  return appearance === 'outline' ? (
```
LimeWub, line 12: Do you need this ternary here? 
Doesn't passing in `appearance` just work (even when undefined?)

if not can we maybe just default the appearance to `solid` (or whatever the regular is)

lib/src/components/banner/BannerActions.tsx
```diff
@@ -0,0 +1,68 @@
+import invariant from 'invariant'
```
LimeWub, line 1: Is this a new thing we are introducing?

lib/src/components/banner/BannerActions.tsx
```diff
@@ -0,0 +1,68 @@
...
+  invariant(
+    React.Children.count(children) <= MAX_ALLOWED_CHILDREN,
+    INVALID_CHILDREN_MESSAGE
+  )
+
+  return (
+    <StyledActionsContainer containerSize={size} {...props}>
+      {React.Children.map(children, (child, index) => {
+        if (!React.isValidElement(child)) {
+          throw new Error(INVALID_CHILDREN_MESSAGE)
```
LimeWub, line 46: The message here actually doesn't say what it's really about.

lib/src/components/banner/BannerActions.tsx
```diff
@@ -0,0 +1,68 @@
...
+})
+
+export const BannerActions: React.FC<React.ComponentProps<typeof Flex>> = ({
+  children,
+  ...props
+}) => {
+  const { size } = useBannerContext()
+  const INVALID_CHILDREN_MESSAGE = `A maximum of ${MAX_ALLOWED_CHILDREN} ${BannerButton.displayName} component(s) are permitted as children of ${BannerActions.displayName}`
+
+  invariant(
```
LimeWub, line 37: I don't think it's important to test this tbh - worst case scenario we get a third button in `outline` style...

And this test here isn't really checking just valid children anyway (if not valid element child or non-button it still counts it...)

lib/src/components/banner/BannerImage.tsx
```diff
@@ -0,0 +1,35 @@
...
+  if (size === 'sm') return null
+
+  return (
+    <Flex
+      css={{
+        alignItems: 'center',
+        justifyContent: 'center'
+      }}
+    >
+      <Box
```
LimeWub, line 21: Why not just use Image + object-fit/positioning? Rather than Box

lib/src/components/banner/BannerRegular.tsx
```diff
@@ -0,0 +1,138 @@
...
+  '& > :first-child': {
+    flex: 2
+  },
+  '& > :last-child': {
+    flex: 'auto'
+  },
+  variants: {
+    size: {
+      sm: {
+        maxWidth: '736px'
```
LimeWub, line 39: Why maxWidth? and why 2 different ones... by 1 px?

lib/src/components/banner/BannerRegular.tsx
```diff
@@ -0,0 +1,138 @@
...
+})
+
+const Dismiss = styled(ActionIcon, {
+  position: 'absolute',
+  top: '12px',
+  right: '12px',
+  variants: {
+    containerSize: {
+      sm: {
+        '&:not(:disabled)': {
```
LimeWub, line 55: Why does it matter if it's disabled?

LimeWub: Also come to think of it. Why would it even be disabled? I'm assuming if the banner is non-dismissible the cross will not be there?

lib/src/components/banner/BannerRegular.tsx
```diff
@@ -0,0 +1,138 @@
...
+  colorScheme,
+  emphasis,
+  dismissible,
+  ...props
+}) => {
+  const content = findChildByType(children, BannerContent)
+  const image = findChildByType(children, BannerImage)
+  return (
+    <Box css={{ position: 'relative' }}>
+      <Dismissible value="dismiss-banner-regular">
```
LimeWub, line 99: value should probably be a prop - so that implementation can catch the dismiss with something meaningful onDismiss - also onDismiss should be getting passed in to this since it's not benefiting from the `...rest`

lib/src/components/banner/BannerRegular.tsx
```diff
@@ -0,0 +1,138 @@
...
+              appearance="solid"
+            >
+              <Icon is={Close} />
+            </Dismiss>
+          </Dismissible.Trigger>
+        )}
+        <BannerContext.Provider
+          value={{ colorScheme, size, emphasis, type: 'regular' }}
+        >
+          <ColorScheme {...colorScheme}>
```
LimeWub, line 119: Can you have the colourscheme be the outtermost element on this?

Just will be easier to remove if we ever just let CS be used as its own flexible wrapper component. Plus it can replace the wrapping Box for the position relative if you do actually need this on the box and it can't just go on `Dismissible` itself.

lib/src/components/banner/BannerRegular.tsx
```diff
@@ -0,0 +1,138 @@
...
+            <Container size={size} css={{ ...css }} {...props}>
```
LimeWub, line 120: css doesn't need spreading here and also doubt if that's where it would need to be passed in (?) Why does it need to be passed to the container?

lib/src/components/banner/BannerText.tsx
```diff
@@ -0,0 +1,39 @@
...
+    }
+  }
+})
+
+export const BannerText: React.FC<React.ComponentProps<typeof Text>> = ({
+  css,
+  ...props
+}) => {
+  const { emphasis, size } = useBannerContext()
+  const color = resolveEmphasis('description', emphasis)
```
LimeWub, line 27: imo (very imo) it'd be clearer if you just had the variants explicitely in this file rather than having the resolveEmphasis helper.

it's like 6 lines and it's obfuscating it less.

lib/src/components/banner/BannerButton.tsx
avirati: You're right, removed the redundant code 👍

lib/src/components/banner/BannerContent.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { BannerText } from './BannerText'
+
+const Container = styled(Flex, {
+  variants: {
+    size: {
+      sm: {
+        p: '$4'
+      },
+      md: {
+        p: 'calc($4 + $2)',
```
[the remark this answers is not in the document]
avirati: Yep 💯

lib/src/components/banner/BannerContent.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const actions = findChildByType(children, BannerActions)
+
+  return (
+    <Container
+      size={size}
+      css={{ flexDirection: 'column', background, ...css }}
+      {...props}
+    >
+      <Stack direction="column" css={{ pr: 'calc($5 + $4)' }}>
+        {heading}
```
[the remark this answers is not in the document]
avirati: Derived from our previous conversations, I thought the design itself wants us to restrict users from deviating from the design. Hence the added restrictions (to make sure things look good at any given time)

lib/src/components/banner/BannerActions.tsx
avirati: It has been there since a quite some time and used in a couple of components

lib/src/components/banner/BannerActions.tsx
avirati: Good spot, fixed 👍

lib/src/components/banner/BannerActions.tsx
avirati: The restrictions are in place as per the specs. Happy to discuss this further but I am in favour of putting restrictions as without it, we risk unintended use of the component

lib/src/components/banner/BannerImage.tsx
avirati: Personal preference, do you have any strong reasons for not using it ?

lib/src/components/banner/BannerRegular.tsx
avirati: This is as per design document.
https://www.figma.com/file/KjgQh7eTwEfJmH6e56k9mX/DS-Documentation?node-id=10641-95059&t=R2hkXpCanUQTM1xn-4

lib/src/components/banner/BannerRegular.tsx
avirati: This is done to avoid `!important`. Background passed here gets overridden by the component's css unless we provide a more specific selector (happy to use other alternative if you have a suggestion).

lib/src/components/banner/BannerRegular.tsx
avirati: Makes sense 👍

lib/src/components/banner/BannerRegular.tsx
avirati: Yep, great suggestion !

lib/src/components/banner/BannerRegular.tsx
avirati: You're right, this is left over code that I forgot to remove 👍

lib/src/components/banner/BannerText.tsx
avirati: I don't agree with you on this. At present we have 3 elements and 3 emphasis, a total of 9 colours for different use cases. This will blow up exponentially if we throw in a couple more emphasis / elements. Having these in one place is the right call imo. Happy to discuss this further.

lib/src/components/banner/BannerActions.tsx
andoulla: yes with Avi, the only restriction of this type we agreed for V1 was this one that they can have 0to 2 buttons

lib/src/components/banner/BannerActions.tsx
```diff
@@ -0,0 +1,70 @@
...
+            `Child passed to ${BannerActions.displayName} is not a valid element`
+          )
+        }
+
+        invariant(
+          child.type === BannerButton,
+          `Children of type ${child?.type} aren't permitted. Only an ${BannerButton.displayName} component is allowed in ${BannerActions.displayName}`
+        )
+
+        // Override button appearance - make the second button outlined
```
andoulla, line 56: This one we had agreed that don't need for V1 but since you did it lets not waste it

lib/src/components/banner/BannerActions.tsx
```diff
@@ -0,0 +1,70 @@
...
+        return index === 0
+          ? child
+          : React.cloneElement(
+              child as React.ReactElement<
+                React.ComponentProps<typeof BannerButton>
+              >,
+              { appearance: 'outline' }
```
andoulla, line 63: for highContrast,  a button can be only neutral or primary is it ok to set it to outline here?

lib/src/components/banner/Banner.utils.ts
```diff
@@ -0,0 +1,27 @@
+import { TBannerElements, TBannerEmphasis } from './Banner.types'
+
+const elementEmphasisMap: Record<
+  TBannerEmphasis,
+  Record<TBannerElements, string>
+> = {
+  hiContrast: {
```
andoulla, line 7: can we use the full name, please? `highContrast`, `lowContrast`

lib/src/components/banner/BannerContent.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const text = findChildByType(children, BannerText)
+  const actions = findChildByType(children, BannerActions)
+
+  return (
+    <Container
+      size={size}
+      css={{ flexDirection: 'column', background, ...css }}
+      {...props}
+    >
+      <Stack direction="column" css={{ pr: 'calc($5 + $4)' }}>
```
andoulla, line 44: $5+$4 = 48px which we have a token ($7). can we use that, please?

lib/src/components/banner/BannerImage.tsx
andoulla: Actually, our image component has some extra features so it might be good to use it

lib/src/components/banner/BannerRegular.tsx
andoulla: I think that might just be an oversight. I think we can use the same one. it is only a pixel nobody will notice

avirati: I'll confirm with @AtanasKovachev

AtanasKovachev: From my understanding, we decided not to go with different banner layouts for each breakpoint, but we still wanted to have the different layouts for desktop and mobile resolutions (one with the image on the side, one without). 

So that's how we defined the 'md' and 'sm' sizes of the banner - 'sm' is used only for resolutions below 736px (max-width of the banner on a 800px screen with 32px padding on each side, e.g. a typical dashboard layout) and 'md' is used only when the width is 737px + for a desktop layout. That ensures we'd never have the image-less mobile banner on a desktop layout and we'd never have the big desktop banner on a narrow mobile screen.

lib/src/components/banner/BannerActions.tsx
```diff
@@ -0,0 +1,70 @@
...
+      sm: {
+        flexDirection: 'column',
+        gap: '$2',
+        '& > button': {
+          width: '100%'
+        }
+      },
+      md: {
+        flexDirection: 'row',
+        gap: 'calc($4 + $1)'
```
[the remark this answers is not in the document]
MiguelGDLR: Is this for the gap between buttons? I think it should be 16px ($4), we had it wrong in the docs sorry!
@AtanasKovachev correct me if I'm wrong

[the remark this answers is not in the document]
avirati: Updated to `$4`

lib/src/components/banner/BannerActions.tsx
avirati: `primary` button does not look good as it kind of mixes in with the background. This was done after discussing with @AtanasKovachev

lib/src/components/banner/Banner.utils.ts
avirati: Done 👍

lib/src/components/banner/BannerContent.tsx
avirati: Good catch. Updated.

lib/src/components/banner/BannerActions.tsx
andoulla: thanks for confirming

andoulla (approved):

lib/src/components/banner/BannerImage.tsx
avirati: Okay so I spent some time into this and came to a conclusion that using `img` without fixed sized parent container will not yield the desired results. Fixed size containers are not going to work because the banner is supposed to have a flexible height / width. Given such restrictions, if you have a way out, do let me know and I shall implement that. Until then, I am sticking with this approach ☝️ 

https://user-images.githubusercontent.com/1936119/233560774-88319fca-ee28-4f53-b438-5bb05863ad10.mov

lib/src/components/banner/BannerRegular.tsx
avirati: I can them both `736px` if that helps. Since we are not using media query breakpoints, 1px difference can be avoided. Shall I proceed ?

lib/src/components/banner/BannerRegular.tsx
LimeWub: Honestly, just do `!important` if you need to override something which the ActionIcon or whichever component can't do - and maybe flag in a DS ticket that we have this use case so we can look at the component which doesn't allow it properly separately.

lib/src/components/banner/BannerRegular.tsx
```diff
@@ -0,0 +1,133 @@
...
+        maxWidth: '736px'
+      },
+      md: {
+        minWidth: '737px'
+      }
+    }
+  }
+})
+
+const Dismiss = styled(ActionIcon, {
```
LimeWub, line 52: I'd be tempted to move this component to its own little file to keep this one short and just about the BannerRegular component.

LimeWub (commented):
I want to pull this down in a bit to check some of my suggestions - went back to u for some of the things but would actually like to have a look at some things with the code running for a bit - soz for the delays.

lib/src/components/banner/BannerContent.tsx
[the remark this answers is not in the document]
hollg: I think I'm lacking context on the previous conversations that Avi refers to, but @LimeWub is right about the principle we want to stick to; we can't effectively enforce design decisions through code and things get messy when we try.

Wherever possible, we want the low-level components to be flexible, accepting arbitrary `children` which allow the consumer to to compose the components together. There should be design review steps in place to make sure that we're not using this flexibility to break any design guidelines. We can then also add higher-level components that package the low-level components together into useful defaults that require less boilerplate to use, and are more opinionated. But generally when we've tried to make the low-level stuff opinionated, we end up needing breaking changes to APIs when the design guidelines inevitably change.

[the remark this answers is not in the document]
D7Torres: I love restrictions in design, but I've not been here long enough
So, I've not seen this yet
> we can't effectively enforce design decisions through code and things get messy when we try.
but I believe you, and we agreed that we want to make components the way Gary and Myrto just described.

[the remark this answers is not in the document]
avirati: Thanks for the insight, I understand your point.

@AtanasKovachev if we were to remove restrictions like suggested above, we would lose this padding. The developer would be required to make visual adjustments as required. Is this acceptable ?

Screenshot 2023-04-24 at 05 05 16

lib/src/components/banner/BannerActions.tsx
avirati: Docs here seem to suggest that `invariant` throws different kind of messages depending on `environment`. I'd say that it is slightly more sophisticated than usual `throw new Error(...)`.

lib/src/components/banner/BannerRegular.tsx
avirati: Hmm, that would make it lose `:active` `:hover` `:focus` states unless I added css for that too. Too much of a hassle ?

lib/src/components/banner/Banner.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  it('renders dismissible sm variant', () => {
+    const onDismiss = jest.fn()
+    const { container } = render(
+      <BannerRegularComponent
+        colorScheme={{ base: 'purple1' }}
+        emphasis="highContrast"
+        size="sm"
+        type="regular"
+        dismissible={{
```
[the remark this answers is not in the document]
avirati: I did it to reduce number of props but sure, I can remove it

lib/src/components/banner/Banner.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+  })
+
+  it('renders dismissible sm variant', () => {
+    const onDismiss = jest.fn()
+    const { container } = render(
+      <BannerRegularComponent
+        colorScheme={{ base: 'purple1' }}
+        emphasis="highContrast"
+        size="sm"
+        type="regular"
```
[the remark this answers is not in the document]
avirati: Redundant code. Removing 👍

lib/src/components/banner/BannerActions.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+const MAX_ALLOWED_CHILDREN = 2
+
+const StyledActionsContainer = styled(Flex, {
+  variants: {
+    containerSize: {
+      sm: {
+        flexDirection: 'column',
+        gap: '$2',
+        '& > button': {
```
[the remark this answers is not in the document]
avirati: Good idea !

lib/src/components/banner/BannerRegular.tsx
avirati: Yep, sounds good 👍

lib/src/components/banner/BannerContent.tsx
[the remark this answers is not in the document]
avirati: Removed restrictions. @AtanasKovachev please review the screenshots in PR description

lib/src/components/banner/BannerRegular.tsx
avirati: Both are `736px` now.

lib/src/components/banner/Banner.test.tsx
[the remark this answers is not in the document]
avirati: Removed. Please check the PR description for updated info

lib/src/components/banner/BannerActions.tsx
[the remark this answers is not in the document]
avirati: Done

lib/src/components/banner/BannerRegular.tsx
avirati: Done

documentation/content/components.content.banner.md
```diff
@@ -0,0 +1,114 @@
...
+
+
+      The `Banner` component ships with multiple building blocks which can be put together in a composable manner to get the desired result.
+
+
+      ## Anatomy
+
+
+      The root `Banner` component manages the banner's configuration and exposes it via the React Context API. This state can be accessed by any child components by calling `useBannerContext` hook. The sub components call this hook to refer to the core configuration and render accordingly.
+
```
andoulla, line 21: can we add a note that this component might also be referred to as Promo Banner? it will help bridge the terminology between design and tech

documentation/content/components.content.banner.md
```diff
@@ -0,0 +1,114 @@
...
+        <BannerRegular.Dismiss />
+      </BannerRegular>
+
+      `} language={"tsx"} />
+
+
+      ## API Reference
+
+
+      <ComponentProps component="BannerRegular" />
```
andoulla, line 107: can we please make sure that we add the props for the sub components like in https://atomlearning.design/components/content/toggle-group

andoulla (changes requested):

documentation/content/components.content.banner.md
```diff
@@ -0,0 +1,114 @@
+---
```
LimeWub, line 1: Could you copy/paste a bit in here about the use of ColorScheme in this component so its a bit more clear? 

ie. like the bit about CS in here: https://atomlearning.design/components/layout/accordion

LimeWub, line 1: I'd also name this file "Banner Regular" since there will be a separate one for "Banner Slim" but we can work on renames separately and it's okay for now since it's the only one.

documentation/content/components.content.banner.md
LimeWub: Are the designers okay with this rename to simple "Banner" btw? Have they been informed? 

If they have and they are cool with it, they should change the name in their Figma and then we don't need to bridge this between design and tech in this documentation here.

documentation/content/components.content.banner.md
avirati: I added the following - 

```jsx
      <ComponentProps component="BannerRegular" />

      <ComponentProps component="BannerRegular.Content" />

      <ComponentProps component="BannerRegular.Heading" />

      <ComponentProps component="BannerRegular.Text" />

      <ComponentProps component="BannerRegular.Actions" />

      <ComponentProps component="BannerRegular.Image" />

      <ComponentProps component="BannerRegular.Button" />

      <ComponentProps component="BannerRegular.Dismiss" />
```

The docs do not seem to generate the API details for any other component. I traced it to 

```typescript
const getComponentProps = (name) =>
  docgen
    .filter(Boolean)
    .find((component) => component.displayName === name || component.displayName === pascalcase(name))
```

Seems like the displayName does not match, please suggest if I should rename all the sub components (and displayName) as `BannerRegular*` instead of `Banner*` (Seems to work)

documentation/content/components.content.banner.md
avirati: @AtanasKovachev could you please review the documentation (PR description) ?

documentation/content/components.content.banner.md
avirati: Done

avirati: Done

documentation/content/components.content.banner.md
```diff
@@ -0,0 +1,114 @@
...
+      The `Banner` component ships with multiple building blocks which can be put together in a composable manner to get the desired result.
+
+
+      ## Anatomy
+
+
+      The root `Banner` component manages the banner's configuration and exposes it via the React Context API. This state can be accessed by any child components by calling `useBannerContext` hook. The sub components call this hook to refer to the core configuration and render accordingly.
+
+
+      ### Examples
```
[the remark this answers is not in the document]
avirati: Thanks, have brought it to the same level as 'Anatomy'. Will make note of the other changes for future reference 👍

documentation/content/components.content.banner.md
```diff
@@ -0,0 +1,114 @@
...
+            <BannerRegular.Button>Secondary CTA</BannerRegular.Button>
+          </BannerRegular.Actions>
+        </BannerRegular.Content>
+      </BannerRegular>
+
+      `} language={"tsx"} />
+
+      #### `md` Variant
+
+      <CodeBlock live={false} preview={true} code={`
```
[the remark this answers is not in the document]
avirati: Done

documentation/content/components.content.banner.md
LimeWub: yep i'd do the rename

avirati: Done ✅

lib/src/components/banner/BannerContainer.tsx
```diff
@@ -0,0 +1,31 @@
...
+        color: '$foreground7plus'
+      },
+      midContrast: {
+        background: '$base3'
+      },
+      lowContrast: {
+        background: '$base1'
+      }
+    },
+    size: {
```
andoulla, line 24:
```suggestion
// Designs available, need to be build in next iterations
    size: {
```

andoulla (approved):

lib/src/components/banner/BannerContainer.tsx
avirati: Discussed separately

LimeWub (approved):

documentation/content/components.content.banner.md
avirati: Discussed offline, all good

PR #502, opened 2023-04-21
Ds 339 cms make slugs for tabs

LimeWub:
JIRA: https://atomlearningltd.atlassian.net/browse/DS-339

Adds query params to the url whenever a tab is selected - so that we can now link people to specific tabs.

## Implementation
* Adds `useQueryParams` hook - very similar to the `atom-core` hook but modified to work with next
* moves `kebabcase` to be a regular dependency as we now need it for the query param values
* Added logic for tabs in Layout.tsx

## Video

https://user-images.githubusercontent.com/6905473/233659515-fa0ec9e5-eb0d-4bce-b017-ec14ebbfabff.mov


documentation/components/page/Layout.tsx
```diff
@@ -22,7 +22,7 @@ export const Layout: React.FC<TDynamicPage> = (props) => {
 
   const isShowingTabs = tabs.length > 1
   useEffect(() => {
-    if (isShowingTabs) updateQueryParams({ tab }, { method: 'replace', debounce: 0 })
+    if (isShowingTabs) updateQueryParams({ tab }, { method: 'replace' })
```
D7Torres, line 25: I think your commit message should be a comment here as well

documentation/utilities/hooks/useQueryParams.ts
D7Torres, line 1: Did you make any changes to the file? or just copied from `atom-core`?

LimeWub: I kept the structure the same but the way we get the query params and set them is different in next router so there's changes to this file.

documentation/utilities/hooks/useQueryParams.ts
```diff
@@ -0,0 +1,54 @@
...
+  const queryParamsInternal = useRef({})
+  useEffect(() => {
+    queryParamsInternal.current = queryParams
+  }, [queryParams])
+
+  const debouncedUpdateQueryParams = useRef<() => unknown>()
+  const updateQueryParams = useCallback(
+    (
+      newSearchParams,
+      { method: updateMethod = 'push', debounce: debounceTicks = 500, shallow = true } = {}
```
D7Torres, line 26: The default object was `{ method: 'push' }` and you changed it to `{}`, why is that?

documentation/utilities/hooks/useQueryParams.ts
```diff
@@ -0,0 +1,54 @@
...
+import { debounce } from 'throttle-debounce'
+
+const configurableDebounce = (ticks: number, cb: () => unknown) =>
+  ticks ? debounce(ticks, cb) : cb
+export const useQueryParams = <T>() => {
+  const router = useRouter()
+  const routerRef = useRef<typeof router>()
+  routerRef.current = router; // (! Hack) router updates on *every* push/replace which leads to infinite loops; switching to a ref means it doesn't have to be a hook dependency
+
+  const { slug: _slug, ...queryParams } = router.query
```
D7Torres, line 12: I'm a bit confused on why this util needs to know about slug, instead of offering whatever queryParams the URL has to the consumer.

documentation/components/page/Layout.tsx
```diff
@@ -5,14 +5,28 @@ import { Container, Header, Footer, Links, Pagination } from '.'
...
+import { useEffect, useState } from 'react'
+import kebabCase from 'lodash.kebabcase'
 
 export const Layout: React.FC<TDynamicPage> = (props) => {
   const { links, slug, tabs, title } = props
   const tabsLength = tabs?.length
   const hasContent = !!tabsLength
+  const { queryParams, updateQueryParams } = useQueryParams()
+
+  const [tab, setTab] = useState(kebabCase(tabs?.[0]?.title))
```
D7Torres, line 18: Just some picky thing: if `tabs` exist, aren't we sure that they have a title?

i.e.: can't we do `tabs?.[0].title`?

documentation/utilities/hooks/useQueryParams.ts
LimeWub: Because when I'd pash in a different object ie. `{debounce: 100}` without the method we'd no longer have default value for the method. The format I've changed it to always has default values for the arguments even if we pass in an incomplete settings object (which is legit use)

documentation/utilities/hooks/useQueryParams.ts
LimeWub: it doesn't need to know about slug but slug is included in the url params in the next router - I'm just extracting it out so I don't append it to the url again as if it is queryParams. I'm not using it, I'm just using everything other than it.

documentation/components/page/Layout.tsx
LimeWub: What if tabs is an empty array?
The way it's now is safer.

documentation/utilities/hooks/useQueryParams.ts
D7Torres: That's not what I meant, but I think I got confused with the syntax, I thought there was a possibility of getting `{}` as default. But I've checked now that it's not the case. So please ignore me

D7Torres (approved):
Approved, only adding the comment I mentioned left.

hollg (approved):

PR #503, opened 2023-04-21
Ds 370 cms replace landing pages cards with interactive tile component build

LimeWub:
Makes `Card`'s filter chips stack rather than overflow auto and replaces `Card`'s container with the new `Tile`.

Slight tweak made to `Tile`'s export so I can use the `NoOverflowContainer` easier - technically I could just export the `NoOverflowContainer` itself; rather than making it a `Tile` thing and it would be more useful. **What do you think?**

JIRA 1: https://atomlearningltd.atlassian.net/browse/DS-370
JIRA 2: https://atomlearningltd.atlassian.net/browse/DS-372

## Images

Screenshot 2023-04-27 at 11 04 29


documentation/components/app/Header.tsx
```diff
@@ -104,7 +104,7 @@ export const Header: React.FC = (props) => {
       <HeaderTrigger onClick={() => setMenuOpen(true)} />
       <StyledHeader ref={ref} open={menuOpen} {...props}>
         <NextLink href="/">
-          <Image src={logo.src} width={80} alt="" css={{ mb: '$7' }} />
+          <Image src={logo.src} width={80} alt="" css={{ mb: '$7', cursor: 'pointer' }} />
```
hollg, line 107: Does this not inherit from `NextLink`? Or does `NextLink` not set it by default? 🤔

hollg (approved):

documentation/components/app/Header.tsx
LimeWub: I dunno why it's not cursor pointer but it wasn't....
I didn't do too in-depth debugging but I think maybe NextLink doesn't set it by default. (Yes it's weird - I also thought it did) - easiest fix was just this bit of css anyway.

PR #510, opened 2023-04-25
fix: auto-fix bad yml

LimeWub:
I think this is fine now? Have tried running locally to make sure but it's proving to be PITA so I'm just assuming the problem was the `:` all along and this is fine :o 

input welcome


thomasdigby (approved):

D7Torres (approved):

euanhunteratom:
Yes, it will have been the ':' since that's the field separator. In future, can wrap whole value in single/double quotes to escape the ':'

PR #511, opened 2023-04-26
chore: fix atomTheme typescript errors

LimeWub:
Fixes https://atom-learning.slack.com/archives/G01G926F9DM/p1682496172202549

Copied solution from `atom-core` uses of `atomTheme`

### DIFF
--- a/documentation/components/markdown/editor-component/ColorTokenList.tsx
+++ b/documentation/components/markdown/editor-component/ColorTokenList.tsx
@@ -1,5 +1,6 @@
 import { Box, Flex, Text } from '@atom-learning/components'
 import { hsl, parseToHsl } from 'polished'
+import type { Theme } from '@atom-learning/theme'
 import * as atomTheme from '@atom-learning/theme'
 import { TokenList } from './token-list'
 import * as React from 'react'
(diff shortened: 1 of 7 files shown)


thomasdigby (approved):

documentation/components/markdown/editor-component/ColorTokenList.tsx
```diff
@@ -39,7 +40,7 @@ export const ColorTokenList: React.FC<ColorTokenListProps> = ({
     <TokenList
       gap={3}
       direction="column"
-      allTokens={atomTheme.colors}
+      allTokens={(atomTheme as Theme).colors}
```
D7Torres, line 43: if it works, great. I'll approve to unblock. But I'd check with someone that knows more TS: wouldn't be possible to export/import atomTheme in a way that is already well typed?

D7Torres (approved):

documentation/components/markdown/editor-component/ColorTokenList.tsx
LimeWub: Ye I have no idea why this is needed to begin with, i do think it's really weird but alas. Same solution has been used in `Question` in `atom-core` by Marty too 🤷 

Probably something up with the theme repo (?)

PR #516, opened 2023-04-27
fix: only add margin-right to BannerRegular Text and Heading when Dis…

LimeWub:
…miss is used

## Description
My bad - in the tweaks I made I added margin-right to `Text` and `Heading` so there's space for the `Dismiss` - forgetting that the `Dismiss` is optional so they don't actually always need this additional spacing.
This PR fixes it by storing that there's a `Dismiss` or nah in the `BannerContext` so we can check it and add the margin right spacer style optionally.

## Videos
(red background added via inspect to showcase issue)

**before**

https://user-images.githubusercontent.com/6905473/234849025-3285b1f1-fd0e-469b-ad85-4ac25f11c541.mov

**after**

https://user-images.githubusercontent.com/6905473/234849042-6a1d1ca8-703f-482b-ba15-0e546f380606.mov


avirati (approved):

lib/src/components/banner/banner-regular/BannerRegularText.tsx
```diff
@@ -24,20 +22,33 @@ const StyledText = styled(Text, {
...
+      true: {}
+    }
+  },
+  compoundVariants: [
+    {
+      containerSize: 'sm',
+      hasDismiss: true,
+      css: {
+        mr: '$6'
+      }
```
D7Torres, line 36: 🤯 I didn't know about compoundVariants

lib/src/components/banner/BannerContext.tsx
```diff
@@ -12,10 +20,23 @@ export const useBannerContext = (): Partial<TBannerContextValue> => {
...
+}) => {
+  const [hasDismiss, setHasDismiss] = React.useState(false)
+  const value = React.useMemo<TBannerContextValue>(
+    () => ({ emphasis, size, hasDismiss, setHasDismiss }),
+    [emphasis, size, hasDismiss, setHasDismiss]
+  )
+  return (
+    <BannerContext.Provider value={value}>{children}</BannerContext.Provider>
+  )
+}
```
D7Torres, line 43: I'm a bit confused. How come we didn't have this block before? we were not using `hasDismiss` and `setHasDismiss`, but we were using `size`:

```
const { size } = useBannerContext()
```

LimeWub: It was using `BannerContext.Provider` directly.

D7Torres (approved):

PR #517, opened 2023-05-04
Ds 349/add aspect ratio tokens

D7Torres:
[JIRA](https://atomlearningltd.atlassian.net/browse/DS-349)

After [adding the aspect ratio tokens in the theme repo](https://github.com/Atom-Learning/theme/pull/25), this PR simply uses the new version and adds the necessary `themeMap` config so we can use aspect ratio tokens in the components library

## Screenshots
Here you can see an ugly box with the $16-9 token used in all its glory
image


thomasdigby (approved):

hollg (approved):

yarn.lock
```diff
@@ -33,7 +33,12 @@
     chalk "^4.1.0"
     css "^3.0.0"
 
-"@atom-learning/theme@1.2.0", "@atom-learning/theme@^1.2.0":
+"@atom-learning/theme@1.3.0":
```
LimeWub, line 36: Why do we have both versions? Could you check if you haven't updated it somewhere? (ie. documentation)

D7Torres: Thanks for catching it!
done

PR #519, opened 2023-05-05
fix: positioning of tick icon on Checkbox

LimeWub:
## The problem this is solving

Checkbox has no `position: relative` - so `position: absolute` Icon glitches when nested in scrollable parent.

https://user-images.githubusercontent.com/6905473/236504880-154ebfca-c44b-4e09-bdba-76ffd93eb729.mov

## Solution
add pos relative to checkbox; and since I'm there use a bit more solid css for positioning the icon and drop a div by using `asChild` for the indicator.

## Screenshots
**BOTH BEFORE AND AFTER (1 video - notice url change)**

https://user-images.githubusercontent.com/6905473/236507988-74752674-598e-4ad4-85d0-ad4b765b9660.mov

### DIFF
--- a/lib/src/components/checkbox-field/__snapshots__/CheckboxField.test.tsx.snap
+++ b/lib/src/components/checkbox-field/__snapshots__/CheckboxField.test.tsx.snap
@@ -8,15 +8,17 @@ exports[`CheckboxField component renders 1`] = `
     margin: 0;
   }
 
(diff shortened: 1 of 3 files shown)


thomasdigby (approved):
Boy, that scroll thing is weird. Nice fix 👍

D7Torres (approved):

PR #520, opened 2023-05-10
fix: Set line height on time input type to fix alignment in Safari

chris-atomlearning:
When input was set to time the text was misaligned in Safari.
To fix I've set a line height. I followed the other examples of line height and used a numerical value, 2 for `md` and 1.7 for `sm`

**BEFORE**:
image

image

**AFTER**:
image

image

**Normal text input still works:**
image

### DIFF
--- a/lib/src/components/input/Input.tsx
+++ b/lib/src/components/input/Input.tsx
@@ -33,11 +33,13 @@ const StyledInput = styled('input', {
     size: {
       sm: {
         height: '$3',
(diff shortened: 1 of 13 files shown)


lib/src/components/input/Input.tsx
chris-atomlearning, line 1: This is the main change, the others are just snapshots

elliotlaws (approved):

alexandrunemes (approved):

PR #526, opened 2023-05-30
SP-1473 add selectable rows in DataTable

alexandrunemes:
This PR adds a row selection feature to DataTable. If `allowRowSelection` is passed to the `DataTable` provider, each row (including the header) gets a checkbox. 
When at least one checkbox is `checked`, the `BulkActionsBar` appears, with a default `Cancel` action which deselects every selected row.

In order to have additional functionality within the BulkActionsBar, you need to call the component from inside a wrapper component, to be able to access `useDataTable`. You can then use `getSelectedRowModel()` to get all the currently selected rows. Then it's business as usual. You can get access to the data, manipulate it, etc.

```
const TableHead = () => {
  const { getSelectedRowModel } = useDataTable()
  console.log(getSelectedRowModel().rows)
  
  return (
    <DataTable.BulkActions>
      <Button onClick={() => console.log('clicked')}>Test</Button>
    </DataTable.BulkActions>
  )
}

const TableComponent = () => {
  <DataTable columns={columns} data={data} allowRowSelection>
     <TableHead />
     <DataTable.Table />
  </DataTable>
}
```

Screenshots:

Screenshot 2023-06-06 at 12 37 25
Screenshot 2023-06-06 at 12 37 31
Screenshot 2023-06-06 at 12 37 36
Screenshot 2023-06-06 at 12 37 41

JIRA ticket: https://atomlearningltd.atlassian.net/browse/SP-1473


lib/src/components/data-table/DataTableHead.tsx
```diff
@@ -26,6 +28,11 @@ export const DataTableHead: React.FC<DataTableHeadProps> = ({
       {getHeaderGroups().map((headerGroup) => {
         return (
           <Table.Row key={headerGroup.id}>
+            {allowRowSelection && (
+              <Table.HeaderCell css={{ width: '$4' }}>
```
andoulla, line 32: Kind reminder to add this example and prop into the documentation please?

andoulla (commented):
A few questions to cross-check my understanding of the code:
- How does this work when a user is filtering or sorting and a selection might not be visible? it preserves it right?
- If the table is empty, the component works it just doesn't show the bar as you can't select anything.

alexandrunemes:
> A few questions to cross-check my understanding of the code:
> 
> * How does this work when a user is filtering or sorting and a selection might not be visible? it preserves it right?
> * If the table is empty, the component works it just doesn't show the bar as you can't select anything.

On the filtering bit, here's how it behaves:

https://github.com/Atom-Learning/components/assets/9009155/e3640a39-9359-419e-b58a-c047ac2e270a

Do you think the selection should be cleared?


---------------------

On the empty state, yeah, I suppose if the table is empty, there's nothing to display (apart from the empty state). Here's what it looks like:

https://github.com/Atom-Learning/components/assets/9009155/8234f59e-57b0-4035-855b-4b48e3f8b0b5

lib/src/components/data-table/DataTableBulkActions.tsx
```diff
@@ -0,0 +1,52 @@
...
+  copy?: {
+    cancel?: string
+  }
+}
+
+export const DataTableBulkActions: React.FC<DataTableBulkActionsProps> = ({
+  children,
+  copy = {
+    cancel: 'Cancel'
+  }
```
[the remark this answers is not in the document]
alexandrunemes: Must've left it as an object because at first, i started building what's now the `DataTable.Metadata` component from scratch, inside the bulk actions bar. So all those labels needed to be customizable. But definitely not the case any longer. Good catch!

lib/src/components/data-table/DataTableBulkActions.tsx
```diff
@@ -0,0 +1,48 @@
...
+        mb: '$2',
+        justifyContent: 'space-between',
+        alignItems: 'center'
+      }}
+    >
+      <DataTable.MetaData />
+      <Flex css={{ justifyContent: 'flex-end', alignItems: 'center' }}>
+        {children}
+        {React.Children.count(children) > 0 && (
+          <Box css={{ width: '1px', height: '$2', bg: 'black', mx: '$2' }} />
```
viniciusvolpe, line 40: isn't it a divider? if yes you could use the DS divider component right?

lib/src/components/data-table/DataTableBulkActions.tsx
```diff
@@ -0,0 +1,48 @@
...
+      css={{
+        p: '$2',
+        bg: '$primaryLight',
+        width: '100%',
+        mb: '$2',
+        justifyContent: 'space-between',
+        alignItems: 'center'
+      }}
+    >
+      <DataTable.MetaData />
```
viniciusvolpe, line 36: does it mean the metadata component will be shown twice if added to the DataTable children? how will it look with both?

lib/src/components/data-table/DataTableBulkActions.tsx
alexandrunemes: 🤣  you're right!

lib/src/components/data-table/DataTableBulkActions.tsx
alexandrunemes: Yeah, you're right, it would double it up. I'm trying to figure out what the best solution for this would be here. Maybe add an "alwaysShow" flag to the bulk actions bar, so that when it's true, the action bar is always visible, alongside the metadata. or take out the metadata from the bulkactionsbar and have users add it as a child. but then they'd have to build the entire inner structure of the actionsbar themselves, and worry about positioning and stuff.

alexandrunemes: What do you think about this?

https://github.com/Atom-Learning/components/assets/9009155/f1be9b64-ef83-4f7f-931f-7cd8d90b7419

MiguelGDLR: hi, what is the problem with having 2 metadata components? aren't they communicating 2 different things? 
The bulk actions bar should appear on top of the actions for the table. You can see a good example [here](https://carbondesignsystem.com/components/data-table/code/) if you select the variant with batch actions

alexandrunemes: I think Vini's concern was that with the MetaData component being manually passed in, when the bulk actions bar would be toggled, we'd end up with 2 MetaData components on the screen.

I think with this approach (as seen in the video above), there's an added benefit of the table not moving up/down on the screen due to the actions bar getting toggled on/off. We could even do something similar to the example you shared, Miguel, where we include the table filter box in the bulk actions bar, and display it only when there's no active row selection.

MiguelGDLR: The blue background helps calling the attention of the user to show that there's a temporary special mode 'selected' and those actions in the box are applied to the selection.  So we can't have that blue box by default if nothing is selected as it will lose its intended meaning.

alexandrunemes: Should we then use a different background color for when  the selection mode is toggled on? or have the background transparent in "non-selection" mode, and switch to the light blue when selection mode is on?

alexandrunemes: @MiguelGDLR how's this?

https://github.com/Atom-Learning/components/assets/9009155/f71df6f5-5849-4b70-9a26-42c3b13ccda2

alexandrunemes: So we're good on the design side? @viniciusvolpe , what do you think about the API?

D7Torres:
Would it make sense to add something similar to the description of this PR to the documentation page of the DataTable component?

alexandrunemes:
> Would it make sense to add something similar to the description of this PR to the documentation page of the DataTable component?

For sure! I just wasn't certain what the current pattern for updating the documentation is. I was thinking that maybe we need to deploy the components lib, make sure it's all working properly, and then update the documentation.

viniciusvolpe (approved):

lib/src/components/data-table/DataTableRow.tsx
```diff
@@ -1,18 +1,35 @@
...
 export type DataTableRowProps = React.ComponentProps<typeof Table.Row> & {
   row: Row<Record<string, unknown>>
 }
 
+const StyledRow = styled(Table.Row, {
+  bg: 'initial',
+  variants: {
+    isSelected: {
+      true: {
+        [`${StyledTableBody} &, ${StyledTableBody} &:nth-child(2n),
```
andoulla, line 19: this made my head spin, would you consider adding a comment about which elements it is capturing? up to you to be honest

lib/src/components/data-table/bulk-actions/DataTableBulkActions.tsx
```diff
@@ -0,0 +1,62 @@
...
+        bg: '$primaryLight'
+      }
+    }
+  }
+})
+
+export const DataTableBulkActions: React.FC<DataTableBulkActionsProps> & {
+  DefaultActions: typeof BulkActionsDefaultActions
+  SelectedRowActions: typeof BulkActionsSelectedRowActions
+} = ({ children, customStyling }) => {
```
andoulla, line 45: minor, but we tend to pass styling overwrites as CSS (just for consistency)

you will find lots of places that we just do css={css}

andoulla (approved):

lib/src/components/data-table/DataTable.types.ts
```diff
@@ -5,7 +5,10 @@ import type {
   ExpandedTableState,
   FiltersTableState,
   GroupingTableState,
+  OnChangeFn,
```
LimeWub, line 8: Are you using `OnChangeFn` ? I don't see it in this file.

lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -113,12 +121,15 @@ export const DataTableProvider = ({
...
       globalFilter,
-      pagination: paginationState
+      pagination: paginationState,
+      rowSelection
     },
     manualPagination: getAsyncData && isPaginated,
     manualSorting: getAsyncData && isPaginated,
     enableSorting: asyncDataState !== AsyncDataState.PENDING,
     enableGlobalFilter: !getAsyncData,
+    enableRowSelection: allowRowSelection,
```
LimeWub, line 131: Minor: It's interesting there's already a name for this prop of `enableRowSelection`, do you think it's worth us matching it rather than using `allowRowSelection` for our prop name - which is very very similar but slightly different. Is `allow` coming from a similar prop or so?

lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -113,12 +121,15 @@ export const DataTableProvider = ({
...
+    onRowSelectionChange: setRowSelection,
```
LimeWub, line 132: Q: How do we hook into this `onRowSelectionChange` from implementation?

lib/src/components/data-table/DataTableSelectAllRowsCheckbox.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+  const updateCheckedState = () => {
+    if (getIsSomePageRowsSelected()) {
+      return toggleAllPageRowsSelected(false)
+    }
+    return toggleAllPageRowsSelected(!getIsAllPageRowsSelected())
+  }
+
+  return (
+    <Checkbox
```
LimeWub, line 27: This checkbox doesn't have a label. Could it have a label (Can wrap in VisuallyHidden) for accessibility purposes?

lib/src/components/data-table/DataTableRowSelectionCheckbox.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react'
+
+import { Checkbox } from '../checkbox'
+
+export const DataTableRowSelectionCheckbox = ({
+  isChecked,
+  onCheckedChange
+}) => {
+  return <Checkbox checked={isChecked} onCheckedChange={onCheckedChange} />
```
LimeWub, line 9: This checkbox doesn't have a label. Could it have a label (Can wrap in VisuallyHidden) for accessibility purposes?

LimeWub, line 9: Would it be worth just passing down props to the checkbox? 
Why are we renaming `checked` to `isChecked`?

lib/src/components/data-table/bulk-actions/DataTableBulkActions.tsx
LimeWub: Can even just spread {...rest} as we're not doing anything custom with the css anyway as far as I can tell.
`({ children, ...rest })`
`<StyledContainer isRowSelected={isRowSelected} {...rest}>`

This would allow for other attributes to be passed down too if we want to.

lib/src/components/data-table/bulk-actions/components/BulkActionsDefaultActions.tsx
```diff
@@ -0,0 +1,16 @@
...
+export const BulkActionsDefaultActions: React.FC = ({
+  children
+}: {
+  children: React.ReactNode
+}) => {
+  const { rowSelection } = useDataTable()
+
+  if (Object.keys(rowSelection).length > 0) return null
+
+  return <Box>{children}</Box>
```
LimeWub, line 15: Q: Do you need this `<Box />` ?

lib/src/components/data-table/DataTableHead.tsx
```diff
@@ -1,21 +1,23 @@
...
 import { Table } from '../table'
+import { DataTableSelectAllRowsCheckbox } from './DataTableSelectAllRowsCheckbox'
 import { DataTable, useDataTable } from './index'
 
 type DataTableHeadProps = Omit<
   React.ComponentProps<typeof Table.Header>,
   'children'
 > & {
   sortable?: boolean
+  rowSelection?: boolean
```
LimeWub, line 12: Is this an actual prop? Cause it doesn't seem to be used?

lib/src/components/data-table/DataTableMetaData.tsx
```diff
@@ -26,12 +26,14 @@ export const DataTableMetaData: React.FC<IDataTableMetaDataProps> = ({
...
   css
 }) => {
-  const { getState, columns, getRowModel } = useDataTable()
+  const { getState, columns, getRowModel, rowSelection } = useDataTable()
   const { sorting } = getState()
   const isSorted = sorting.length > 0
 
   const totalRows = getRowModel()?.rows?.length
 
+  const selectedRows = Object.keys(rowSelection || {}).length
```
LimeWub, line 35: For my understanding:
I keep seeing you need to do `Object.keys()` for the rowSelection.
What exactly is the format of the rowSelection? Does it have all row names with a boolean assigned for whether they are selected or not?
Or just the selected ones?

lib/src/components/data-table/bulk-actions/DataTableBulkActions.tsx
```diff
@@ -0,0 +1,62 @@
...
+    React.ReactElement<React.ComponentProps<typeof BulkActionsDefaultActions>>,
+    React.ReactElement<
+      React.ComponentProps<typeof BulkActionsSelectedRowActions>
+    >
+  ]
+}
+
+const StyledContainer = styled(Flex, {
+  p: '$2',
+  bg: 'transparent',
```
LimeWub, line 25: Do you need `bg: 'transparent'` ?

LimeWub (commented):
First pass ^^ Some of this is a bit complex to get my head around so leaving some comments for now :o

lib/src/components/data-table/bulk-actions/components/index.ts
```diff
@@ -0,0 +1,2 @@
+export { BulkActionsDefaultActions } from './BulkActionsDefaultActions'
```
LimeWub, line 1: One more thing. This folder structure is a bit weird. `bulk-actions/components`. Can everything live on the `bulk-actions` level?
If `DataTableBulkActions` needs to be moved one level up to live next to eg. `DataTableMetaData` this would make sense too.

lib/src/components/data-table/DataTable.types.ts
alexandrunemes: Huh 🤔 that is strange. No, I'm not. I'll take it out. Not sure how/why I added it.

lib/src/components/data-table/bulk-actions/components/index.ts
LimeWub: I would also not be opposed for everything bulk actions related to live in the same file: `DataTableBulkActions` since both the files in `components` are pretty small.

lib/src/components/data-table/DataTableHead.tsx
alexandrunemes: it used to be but i updated the implementation and i must've forgotten to remove it. nice catch!

lib/src/components/data-table/DataTableMetaData.tsx
alexandrunemes: just the selected ones. it uses the index of the row as the key, and then `true` as value:
```
{0: true, 1: true}
```

lib/src/components/data-table/DataTableRowSelectionCheckbox.tsx
alexandrunemes: cause I tend to use the `is` prefix for any variable/prop that's a boolean. if this isn't what we're doing here on the DS, I'll switch it, but i just find it easier to understand just by looking at it that it's a boolean.

lib/src/components/data-table/bulk-actions/DataTableBulkActions.tsx
alexandrunemes: Do we usually open up APIs this loosely? Wouldn't we want to keep some sort of control over how the components are used?

lib/src/components/data-table/bulk-actions/components/BulkActionsDefaultActions.tsx
alexandrunemes: No, I don't think so. TS complained at some point when I was just trying to return the children, but I think it was just some TS weirdness. It's not complaining anymore.

lib/src/components/data-table/bulk-actions/DataTableBulkActions.tsx
alexandrunemes: i think it should behave the same without it. i'll take it out

lib/src/components/data-table/DataTableMetaData.tsx
LimeWub: How weird they didn't use an array for that! `[0, 1]` would be the same thing if it is only storing the selected ones!
Okay thanks for explaining. Just wanted to clarify so we don't get bitten by a `{0: true, 1: false}` situation at any point in the case they were also keeping the unselected ones in the object. 👍

lib/src/components/data-table/DataTableRowSelectionCheckbox.tsx
alexandrunemes: TS complains about passing a `label` prop. should I use `aria-label` instead? That one shouldn't need any for of hiding

lib/src/components/data-table/bulk-actions/DataTableBulkActions.tsx
andoulla: we haven't when it comes to css overwrites as we don't want to have to build the occasional case as a variant.

lib/src/components/data-table/DataTableContext.tsx
alexandrunemes: We don't, really. It's just for the context to be able to pick up the row selection changes and assign them to the state variable, so we can pick that one up from the consumer side

lib/src/components/data-table/DataTableRowSelectionCheckbox.tsx
LimeWub: For native things like `checked`, `disabled` etc we defo are using the name directly. For other things it's debatable and there's no clear decision yet but checkbox itself is lit using `checked` so I'd keep it the same.

lib/src/components/data-table/DataTableContext.tsx
alexandrunemes: sure, i can change to `enable`

lib/src/components/data-table/DataTableRowSelectionCheckbox.tsx
alexandrunemes: 👍

lib/src/components/data-table/bulk-actions/DataTableBulkActions.tsx
LimeWub: AFAIK: We do have them quite open usually, yes. Most components do a `{...rest}` or similar.
There's no written down pattern for this but usually it's not a problem and in fact has been useful for testing or extending in implementation if needed. If something defo needs to not be overwritten and is initialised internally in the component and not coming from a prop we just add it after `{...rest}` to make sure it never is overwritten. 
Also, I think anything that doesn't match with what props this can `actually` have will flag in Typescript anyway?

alexandrunemes: Good point. I'll update the props to spread

lib/src/components/data-table/DataTableRowSelectionCheckbox.tsx
LimeWub: How are you passing a label?
`Checkbox` itself doesn't have a label inside of it so you'd need to use the `<Label>` component directly.
Have a look at `DataTableGlobalFilter` for how that's done - but your label is probably always hidden - unlike the one in `DataTableGlobalFilter` which depends on a prop ^^

lib/src/components/data-table/DataTableContext.tsx
LimeWub: Oh so from the consumer side - how do we access which rows are selected (if we need to?). Soz if there's docs somewhere for this and I've missed them.

lib/src/components/data-table/bulk-actions/components/index.ts
alexandrunemes: 🤔 I don't think the bulk actions should be moved one level up. it's closely related to the table. As for the `/components` folder, I think it also clarifies that those 2 components nested under that folder are part of the main component, but would not be used as siblings to it.

lib/src/components/data-table/DataTableContext.tsx
alexandrunemes: no worries 😄 . We use the `rowSelection` value, which gets set whenever the row selection changes, via the `onRowSelectionChange: setRowSelection` assignment.

lib/src/components/data-table/bulk-actions/components/index.ts
LimeWub: > I don't think the bulk actions should be moved one level up. it's closely related to the table.
It would still live in DataTable folder. Just directly in it rather than nested in `bulk-actions` - same level as the other components of DataTable

lib/src/components/data-table/DataTableContext.tsx
LimeWub: Ahhh so we can do `const { rowSelection } = useDataTable()` ? Fair enough!

alexandrunemes: exactly!

lib/src/components/data-table/DataTableRowSelectionCheckbox.tsx
alexandrunemes: Oh, I get it! Cool will add!

lib/src/components/data-table/bulk-actions/components/index.ts
alexandrunemes: And is it a common pattern to have multiple components exported from the same file?

LimeWub: Only one component is exported in that file: `DataTableBulkActions`
The other two are its `.Children` right?

Honestly my main issue is having the extra nesting of `bulk-actions/components`. It just doesn't feel right and nothing else is doing this sort of nesting in the DS.

Exporting multiple components in one file: there is no pattern for or against. I was just trying to get rid of the awkward nesting and make it more similar to the other components in the DataTable folder - which are all on the same level.

LimeWub: To clarify priority: This is not blocking but it doesn't feel right to me to have this sort of nesting. ^^

lib/src/components/data-table/DataTableRowSelectionCheckbox.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export const DataTableRowSelectionCheckbox = ({
+  rowIndex,
+  checked,
+  onCheckedChange
+}: DataTableRowSelectionCheckboxProps): React.ReactElement => {
+  return (
+    <>
+      <OptionallyVisuallyHiddenContainer hidden>
```
LimeWub, line 20: Oh Sorry I confused you! My bad!
Instead of using `OptionallyVisuallyHidden` use `VisuallyHidden` directly:
`import * as VisuallyHidden from '@radix-ui/react-visually-hidden'`
and then `<VisuallyHidden.Root><Label /></VisuallyHidden.Root>`


totally my bad! I didn't realise it wasn't using the VisuallyHidden.Root directly :o

lib/src/components/data-table/DataTableRowSelectionCheckbox.tsx
```diff
@@ -0,0 +1,32 @@
...
+        <Label htmlFor={`row${rowIndex}Selection`}>
+          {`Row ${rowIndex} selection`}
```
LimeWub, line 22: This should also come from translations if possible~

lib/src/components/data-table/DataTableRowSelectionCheckbox.tsx
```diff
@@ -0,0 +1,32 @@
...
+        </Label>
+      </OptionallyVisuallyHiddenContainer>
+      <Checkbox
+        checked={checked}
+        onCheckedChange={onCheckedChange}
+        name={`row${rowIndex}Selection`}
```
LimeWub, line 28: Any chance this name could be unique somehow? might cause bugs if there's multiples of a table on a page with the same named checkbox, I think? (I might be wrong and that's not the case for these checkboxes but unsure)
any kind of way of doing uniqueId or adding a table id or something to it would be A OK for me.

lib/src/components/data-table/DataTableBulkActions.tsx
```diff
@@ -0,0 +1,98 @@
...
+  mb: '$2',
+  justifyContent: 'space-between',
+  alignItems: 'center',
+  minHeight: '$6',
+  borderTopLeftRadius: '$0',
+  borderTopRightRadius: '$0',
+  variants: {
+    isRowSelected: {
+      true: {
+        bg: '$primaryLight'
```
LimeWub, line 38: V.Minor: instead of `$primaryLight` use `$blue100` :) We are trying to phase out use of "primary"

alexandrunemes: 👍

lib/src/components/data-table/DataTableRow.tsx
```diff
@@ -1,18 +1,50 @@
...
 
+const StyledRow = styled(Table.Row, {
+  bg: 'initial',
+  variants: {
+    isSelected: {
+      true: {
+        // this rule needs to be formatted this way in order to avoid using !important to set the bg.
+        // the bg property is set elsewhere and it's more specific than this one would be without targetting it this way.
+        // the :nth-child ones are needed to catch the rows in striped tables, for which the original bg rules are even more specific (due to the pseudo-selectors)
+        [`${StyledTableBody} &, ${StyledTableBody} &:nth-child(2n),
```
LimeWub, line 22: Firstly thanks for the comment - it's super useful!
However, I'd just use `!important` (+ keep part of this comment to explain why). I don't think it's such a big deal and the nesting here to make this a "heavier" css selector is a bit of a hack.

You could had probs done `* > &`  for the exact same result. I think `!important` is just simpler/cleaner and I can't think of any case it would be an issue/we'd want to overwrite it.

It'd just be so much simpler to read. imo :)

lib/src/components/data-table/DataTableSelectAllRowsCheckbox.tsx
```diff
@@ -0,0 +1,40 @@
...
+    if (getIsSomePageRowsSelected()) {
+      return toggleAllPageRowsSelected(false)
+    }
+    return toggleAllPageRowsSelected(!getIsAllPageRowsSelected())
+  }
+
+  return (
+    <>
+      <OptionallyVisuallyHiddenContainer hidden>
+        <Label htmlFor="allRowsSelection">All rows selection</Label>
```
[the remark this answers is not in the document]
LimeWub: ye exactly ^^

lib/src/components/data-table/DataTableRow.tsx
alexandrunemes: fair enough. if we're gonna need to override this one at some point (due to some other feature we'd be adding to the table), we can always go back to the non-!important implementation.

lib/src/components/data-table/DataTableSelectAllRowsCheckbox.tsx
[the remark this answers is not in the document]
alexandrunemes: nice!

lib/src/components/data-table/DataTableRowSelectionCheckbox.tsx
alexandrunemes: Pretty nifty components!

lib/src/components/data-table/DataTableRowSelectionCheckbox.tsx
alexandrunemes: It doesn't seem like there's any issue when adding 2 tables to the same page, but I've added a `uuid` to the tableprovider and use that as the tableId to build the unique names and labels

lib/src/components/table/TableBody.tsx
```diff
@@ -4,7 +4,7 @@ import { styled } from '~/stitches'
 
 import { StyledRow } from './TableRow'
 
-const StyledTableBody = styled('tbody', {
+export const StyledTableBody = styled('tbody', {
```
LimeWub, line 7: Minor: You no longer need the changes in table since you reverted the nested css hack.

LimeWub (approved):
Thanks for doing this work 👍

lib/src/components/table/TableBody.tsx
alexandrunemes: Haha, just committed the change to remove this.

lib/src/components/data-table/DataTableContext.tsx
```diff
@@ -27,6 +27,8 @@ import { getNewAsyncData } from './getNewAsyncData'
 import { usePagination } from './usePagination'
 import { useSortByColumn } from './useSorting'
 
+import { v4 as uuid } from '@lukeed/uuid'
```
LimeWub, line 30: I wonder if it's this giving you the error.
uuid is a dependency of `documentation` but not `lib`. 

Did you check if this is working? Maybe try to move the dependency one up to the common package.json and see if it unblocks this?

🤷

alexandrunemes: Weird. I def. tested to see what it outputs, and it works fine. Will do some more digging

LimeWub: The dependency should be included in the library's package.json anyway I'd assume - since it won't have it when the library is processed for npm ^^

lib/package.json
```diff
@@ -57,6 +57,9 @@
       "import": "{  useAlert, AlertProvider }"
     }
   ],
+  "dependency": {
```
LimeWub, line 60: this is not a meaningful property ^^
add `"@lukeed/uuid": "^2.0.0"` to 'dependencies' below :o

LimeWub: actually imma commit this change myself cause I'm curious if it'll fix the error 👀

alexandrunemes: jesus christ 🤦 , I just saw this in my editor and was like WTF?!

PR #527, opened 2023-05-30
Feat/sp 1474 add metadata component to datatable

chris-atomlearning:
https://atomlearningltd.atlassian.net/jira/software/projects/SP/boards/8/backlog?selectedIssue=SP-1474
"Add a label component to DataTable that shows the number of items and the column that the sorting is based on. This should be added as a compound component, i.e. DataTable.MetaData."

***DEFAULT SORTING***
image

***FILTERED WITH SEARCH***
image

***SORTED BY ANOTHER COLUMN***
image

***UNSORTED***
image

### DIFF
--- a/lib/src/components/data-table/DataTableMetaData.tsx
+++ b/lib/src/components/data-table/DataTableMetaData.tsx
@@ -1,0 +1,36 @@
+import { CSS } from '@stitches/react'
+import * as React from 'react'
+
+import { Text } from '../text'
+import { useDataTable } from './index'
+
+interface IDataTableMetaDataProps {
+  css?: CSS
+}
+
+export const DataTableMetaData: React.FC<IDataTableMetaDataProps> = ({
+  css
+}) => {
+  const { getState, columns, getRowModel } = useDataTable()
+  const { sorting } = getState()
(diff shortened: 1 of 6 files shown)


viniciusvolpe (approved):

lib/src/components/data-table/DataTableMetaData.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const isSorted = sorting.length > 0
+
+  const totalRows = getRowModel()?.rows?.length
+
+  const getColumnDisplayName = (id: string) => {
+    const sortedColumn = columns.find((col) => col.id === id)
+    return sortedColumn?.header || id
+  }
+
+  const getSortingString = (sorting) => {
```
LimeWub, line 25: Can this come from props? (it's okay to default it here but for translation reasons)

chris-atomlearning: Yeah absolutely, that's a good shout

chris-atomlearning: Even better suggestion, thanks!

I added the separator as well so everything is customisable :)

PR #528, opened 2023-06-01
docs: fix Slider examples

D7Torres:
Simply adding with to some examples to fix them

Before

image

After

image

Before

https://github.com/Atom-Learning/components/assets/4931116/73187419-08d1-4846-8293-77fa13ddb930

After

https://github.com/Atom-Learning/components/assets/4931116/96b4c420-8081-47eb-8c64-cc9d77dec3c4

### DIFF
--- a/documentation/content/components.form.fields.slider-field.md
+++ b/documentation/content/components.form.fields.slider-field.md
@@ -18,7 +18,7 @@ tabs:
 
 
       <CodeBlock live={true} preview={true} code={`<Form>
-        <SliderField name="slider" label="Select a value" defaultValue={[50]} />
+        <SliderField name="slider" label="Select a value" defaultValue={[50]} css={{ width: '260px' }} />
       </Form>`} language={"tsx"} />
 
 
--- a/documentation/content/components.form.primitives.slider.md
+++ b/documentation/content/components.form.primitives.slider.md
@@ -61,7 +61,7 @@ tabs:
       `Slider.Steps` work well with the built in `step` property, which defaults to 1 and changes the size of each movement. For example, this would limit the slider to three values only.
 
 
-      <CodeBlock live={true} preview={true} code={`<Slider defaultValue={[50]} min={10} max={20} step={5}>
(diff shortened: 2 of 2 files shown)


MiguelGDLR (approved):

documentation/content/components.form.fields.slider-field.md
```diff
@@ -18,7 +18,7 @@ tabs:
 
 
       <CodeBlock live={true} preview={true} code={`<Form>
-        <SliderField name="slider" label="Select a value" defaultValue={[50]} />
+        <SliderField name="slider" label="Select a value" defaultValue={[50]} css={{ width: '260px' }} />
```
andoulla, line 21: This is just adding some css to it? Why is this needed?

LimeWub: Have a look at the first recording in the description.
It's just fixing the example so it's not jumpy.

LimeWub (approved):

documentation/content/components.form.fields.slider-field.md
andoulla: I did watch them but I didn't notice it until you pointed it out. does this mean that we need to set the width every time we use it?

D7Torres: Yes, but this is not a problem with the component per se. If you have the width defined by the content, and the content is changing dynamically, it will be shaking, so you you should establish a fixed width

andoulla: couldn't we impose a min-width for it (which obviously people can override)?

D7Torres: And how much? you would have the same problem past that min-width, I'm not convinced, sorry.

andoulla: no worries it was just a thought. the only ask is that you document the need to set the CSS prop, as it is not really obvious at themoment

D7Torres: It' not needed if the container has a width, but I can make a note explaining it.

PR #529, opened 2023-06-06
Stu 1497 pagination improvements

mzeenkala23:
Jira ticket- https://atomlearningltd.atlassian.net/jira/software/projects/STU/boards/1?selectedIssue=STU-1497

In this task we implement a bug fix and add some small improvements to to the pagination component:

- Bugfix: add missing dependency to dependency array
- Improvements: make these props optional `indicatedPages disabledPages onSelectedPageChange visibleElementsCount` and fix issue where `The little dot for indicatedPages pushes the number up a few pixels, it looks a bit strange when some pages are indicated and some are not.`

### DIFF
--- a/lib/src/components/pagination/PaginationItem.tsx
+++ b/lib/src/components/pagination/PaginationItem.tsx
@@ -103,13 +103,13 @@ export const PaginationItem: React.FC<IPaginationItemProps> = ({
   const { currentPage, goToPage, indicatedPages, disabledPages, onItemHover } =
     usePagination()
 
-  const isIndicated = indicatedPages?.includes(pageNumber)
-  const isDisabled = disabledPages?.includes(pageNumber)
+  const isIndicated = indicatedPages.includes(pageNumber)
+  const isDisabled = disabledPages.includes(pageNumber)
 
   const isSelected = currentPage === pageNumber
 
   const handleOnHover = () => {
-    if (isSelected) return
+    if (isSelected || isDisabled) return
     onItemHover?.(pageNumber)
   }
 
--- a/lib/src/components/pagination/PaginationItem.tsx
+++ b/lib/src/components/pagination/PaginationItem.tsx
@@ -9,14 +9,15 @@ import { usePagination } from './usePagination'
 
 const Dot = styled(Box, {
   position: 'absolute',
-  bottom: '6px',
+  bottom: '-$1',
+  left: '0',
+  transform: 'translateX(50%)',
   borderRadius: '$round',
   size: '4px',
   bg: '$accent9'
 })
 
 const StyledButton = styled('button', {
-  position: 'relative',
   alignItems: 'center',
   border: 'unset',
   borderRadius: '$0',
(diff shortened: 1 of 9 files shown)


lib/src/components/pagination/types.ts
```diff
@@ -9,29 +8,25 @@ interface ILabels {
 }
 
 interface TBasePaginationType {
```
Mhoog, line 10:
```suggestion
interface IBasePaginationType {
```

mzeenkala23: good spot, updated

lib/src/components/pagination/types.ts
```diff
@@ -9,29 +8,25 @@ interface ILabels {
...
-  visibleElementsCount: TVisibleElementsCount
-  indicatedPages: number[]
-  disabledPages: number[]
+  pagesCount: number
+  onSelectedPageChange?: (pageNumber: number) => void
+  selectedPage?: number
+  visibleElementsCount?: TVisibleElementsCount
+  indicatedPages?: number[]
+  disabledPages?: number[]
```
Mhoog, line 16: I think we could make some changes here so that `indicatedPages` and `disabledPages` are optional props for `Pagination` and `PaginationProvider`, but are mandatory for the `PaginationContext`. That way you don't have to deal with optional chaining / null checks.

You can fall back to an empty array for those two when you set the context value.

```ts
interface IBasePagination {
    pagesCount: number
    visibleElementsCount: TVisibleElementsCount

    indicatedPages: number[]
    disabledPages: number[]

    labels: ILabels

    onItemHover: (pageNumber: number) => void
}

interface IPaginationContext extends IBasePagination {
  currentPage: number
  isMaxVisibleElementCount: boolean

  goToPage: (pagenumber: number) => void
  goToPreviousPage: () => void
  goToNextPage: () => void
}

type TPaginationProviderProps = Pick<IBasePagination, 'pagesCount'> & Partial<IBasePagination> & {
    selectedPage?: number
    onSelectedPageChange?: (pageNumber: number) => void
}

interface IPaginationProps extends TPaginationProviderProps {
  colorScheme?: TcolorScheme
  css?: CSS
}
```

lib/src/components/pagination/PaginationItem.tsx
```diff
@@ -57,6 +60,11 @@ const StyledButton = styled('button', {
         border: '1px solid $accent9',
         color: '$accent9',
         fontWeight: 600,
+        [`& ${Dot}`]: {
```
LimeWub, line 63: This is a bit hacky, I made a PR cleaning it up a bit and making it a bit more predictable: https://github.com/Atom-Learning/components/pull/530

Basically, put a Box around the number and then place the Dot relative to that number. Saves you having to care about any borders etc and it uses a DS token :)

Just merge it in this one if you're happy w the suggestion.

mzeenkala23: just tested it, works great thanks

lib/src/components/pagination/types.ts
mzeenkala23: looks much cleaner thanks, I also had to add default values for `onItemHover = () => null,
  labels = {},` in the provider to prevent typescript errors

LimeWub (approved):

Mhoog (approved):

PR #532, opened 2023-06-12
SP-1472 - New 'lg' size for Checkbox component

chris-atomlearning:
Added large size, and the prop to choose between `md` and `lg`, with `md` being the fallback.

image

image

Added test to make sure the checkbox toggles. Not necessarily related to this change, but thought it was useful.


lib/src/components/checkbox/Checkbox.tsx
```diff
@@ -50,21 +51,39 @@ const StyledCheckbox = styled(RadixCheckbox.Root, {
...
+        size: '$2'
+      }
     }
   }
 })
 
-type CheckboxProps = React.ComponentProps<typeof StyledCheckbox>
+type CheckboxSize = 'md' | 'lg'
+
+const getIconSize = (size: CheckboxSize) => {
```
LimeWub, line 68: Could you do this with `overrideStitchesVariantValue`? There should be other examples in the codebase with it too. And it would work responsive.

thomasdigby (approved):

lib/src/components/checkbox/Checkbox.test.tsx
```diff
@@ -32,4 +32,20 @@ describe(`Checkbox component`, () => {
 
     expect(await axe(container)).toHaveNoViolations()
   })
+
+  it('toggles the checkbox on click', async () => {
```
LimeWub, line 36: I'm not against this test. But, are we testing radix functionality basically?
Just wondering if it's needed. 
No big deal to keep it or not ^^

lib/src/components/checkbox/Checkbox.tsx
chris-atomlearning: I've added this in, thanks for the tip. I think I've used it correctly but would be good to get your view on it :)

chris-atomlearning:
> Are we doing Radio button separately?
I'll be honest, I somehow forgot that Radio was part of the ticket and focused just on Checkbox. I'll raise a separate PR for that. Should be pretty similar to this one.

lib/src/components/checkbox/Checkbox.tsx
```diff
@@ -50,26 +52,58 @@ const StyledCheckbox = styled(RadixCheckbox.Root, {
...
-          css={{
-            strokeWidth: '3',
-            size: 14
-          }}
-        />
-      </StyledIndicator>
-    </StyledCheckbox>
-  )
+  ({ size, ...props }, ref) => {
+    const checkboxSize = React.useMemo(
```
LimeWub, line 83: You shouldn't need this one. Checkbox size is the one you already have from the props :)

`overrideStitchesVariantValue` is for when you need to pass down a prop to a child or context. So it's only relevant to the Icon size here ^^

lmk if this doesn't make sense, happy to jump on a call to explain ^^

lib/src/components/checkbox/Checkbox.tsx
```diff
@@ -50,26 +52,58 @@ const StyledCheckbox = styled(RadixCheckbox.Root, {
...
+
+    return (
+      <StyledCheckbox {...props} size={checkboxSize} ref={ref}>
+        <StyledIndicator asChild>
+          <Icon
+            is={props.checked === 'indeterminate' ? Minus : Ok}
+            css={{
+              strokeWidth: '3'
+            }}
+            size={iconSize || 'sm'}
```
LimeWub, line 101: The logic for the iconSize is good but can we default on the entire checkbox for the size? Rather than just on the icon?

Then we shouldn't need this fallback here :)

lib/src/components/checkbox/Checkbox.tsx
chris-atomlearning: Ahh, gotcha. That makes sense. Is the Icon one ok tho?

LimeWub: Yes it looks correct to me :) You can always test if it works okay by padding the checkbox a breakpoint size variant and seeing if it works fine.

ie. `size={{'@initial': 'lg', '@md': 'md'}}`

lib/src/components/checkbox/Checkbox.tsx
```diff
@@ -2,9 +2,11 @@ import { Minus, Ok } from '@atom-learning/icons'
 import * as RadixCheckbox from '@radix-ui/react-checkbox'
 import * as React from 'react'
 
+import { Override } from '~/utilities/types'
```
LimeWub, line 5: This import is not needed

lib/src/components/radio-button/RadioButton.tsx
```diff
@@ -37,22 +49,35 @@ const StyledIndicator = styled(RadioGroup.Indicator, {
...
+      }
+    }
+  }
 })
 
 type RadioButtonProps = Override<
   React.ComponentPropsWithoutRef<typeof StyledRadioButton>,
   {
     as?: never
+    size?: 'md' | 'lg'
```
LimeWub, line 69: Why do you need this override? size coming from the StyledRadioButton should be good enough and more correct :o

lib/src/components/radio-button/RadioButton.tsx
```diff
@@ -16,6 +16,18 @@ const StyledRadioButton = styled(RadioGroup.Item, {
...
   p: 0,
   size: '$1',
+  variants: {
+    size: {
+      md: {
+        size: '$1'
+      },
+      lg: {
+        size: '$2',
+        mt: 'calc($1 * -1)',
```
LimeWub, line 26: Can you just do `-$1` ?

Why does lg have margins anyway?

chris-atomlearning: For some reason I thought that didn't work, but it obviously does!
The top margin was just to bring the larger checkbox/radio into line with the label.

LimeWub (approved):

MiguelGDLR:
Hey, I saw that you added the new size for Radio button too but the prop is not showing up in the docs?

MiguelGDLR:
also, is this a problem with the gap between items? 
Screenshot 2023-06-21 at 16 34 41

PR #533, opened 2023-06-13
fix: change all uses of absolute urls to relative

LimeWub:
JIRA: https://atomlearningltd.atlassian.net/browse/DS-421?atlOrigin=eyJpIjoiMzFhODJlZTUzOWY3NDRiOWE2YWI3ODdhOWI3OWRhZGYiLCJwIjoiaiJ9 

This PR changes all absolute internal links to relative, which means they work as expected in localhost use, as well as if we even change the url of the design system.
No more being sent to the live website accidentally and wondering why the changes aren't there :D


Mhoog (approved):

MiguelGDLR:
can/should I do the same with the card links in the homepage?

PR #534, opened 2023-06-13
Ds 425 list improvements use em unit for ordered list (+ some slight other tweaks)

LimeWub:
JIRA: https://atomlearningltd.atlassian.net/browse/DS-425

Fixes 3 things (yes I know, but they are quite small):

1. for ol number markers: replaces use of `$sm` fontSize with `.875em` (looks exactly the same in the demo example)
2. for ul dot markers: adds a single space next to the dot so it doesn't touch the text
3. for the list itself: adds list style position of 'inside' so the dots etc are affected by parent's test alignment

## Screenshots

Number fontSize:
**before**
Screenshot 2023-06-13 at 11 29 52

**after**
Screenshot 2023-06-13 at 11 37 25

UL dot spacing:
**before**
Screenshot 2023-06-13 at 11 37 04

**after**
Screenshot 2023-06-13 at 11 37 19

List style position:
**before**
Screenshot 2023-06-13 at 11 38 44

**after**
Screenshot 2023-06-13 at 11 40 39


lib/src/components/list/List.tsx
```diff
@@ -19,17 +20,17 @@ export const StyledList = styled('ul', {
...
-          '&::marker': { fontSize: '$sm', fontWeight: 'bold' }
+          '&::marker': { fontSize: '.875em', fontWeight: 'bold' }
         }
       },
       ul: {
         pl: '$3',
         [`& > ${StyledLi}`]: {
           pl: '$2',
-          '&::marker': { content: '•', fontWeight: 'bold' }
+          '&::marker': { content: '• ', fontWeight: 'bold' }
```
thomasdigby, line 33: Isn't this space supposed to be handled by the padding above?

LimeWub: I'm not sure what the padding above is for, but it doesn't seem it is handling it (based on what it ends up looking like :o )

thomasdigby: Yeah I _think_ that's what it's for, but if it ain't working I'm not against using whitespace, it will scale with the size of the font as well so probably more appropriate.

thomasdigby (approved):

Mhoog (approved):

PR #535, opened 2023-06-13
Feature: Flex gap stack

thomasdigby:


lib/src/components/stack/Stack.tsx
```diff
@@ -1,100 +1,40 @@
...
-        flexWrap: 'no-wrap'
-      },
-      'wrap-reverse': {
-        flexWrap: 'wrap-reverse'
-      }
+      wrap: { flexWrap: 'wrap' },
+      'no-wrap': { flexWrap: 'no-wrap' },
+      'wrap-reverse': { flexWrap: 'wrap-reverse' }
     },
     justify: {
```
LimeWub, line 22: I think now that stack is just a flex with a gap we can delete these justify/align/etc props as they can now be added directly with CSS.
Previously we needed them because the CSS was going on the CSS wrapper so there was no ez way to target the Stack but we can purge them now :3

thomasdigby: I actually quite like them, I think they're a nicer hook into flex layout than `css` directly. It does introduce slightly fuzzy "why support this property and not that one" vibes, _but_, I think for the majority of use cases it's a nice enhancement.

```tsx
<Stack justify="end">
```

```tsx
<Stack css={{ justifyContent: 'flex-end' }}>
```

LimeWub: Ye... tbh, as you said, I think they are a bit arbitrary as to what is supported, which has not been advised by design so it is missing things - it's more like: we've needed this before so there it is.
I also think the naming can be confusing as it doesn't match exactly the css property, I'm sure people will use css directly much more than trying to figure out the props. 
Putting them on the css to me is clearer as you don't have to have the extra abstraction and don't have to remember eg that in our DS `justify` is called `justifyContent`.
...just my 2c

On another note, now that we don't have to do CSS magic to achieve a Stack. What really is the benefit of the Stack as a component as opposed to Flex? Just wondering...

thomasdigby: I agree that we should align with design on this, Figma has a flex-based layout system I think anyway.

I think we ought to look at it from the other angle, from the developer's point of view, trying to lay things out on a page. We could just get them to use `Box` and add `display: flex` and all the other bits to it, but we provide a minimal abstraction to slightly improve their experience using flex layout. I think `Stack` is just the same sort of thing but another level up, helping developers add layout to a page, without having to write a bunch of custom CSS. The `Stack` abstraction is simple, but like most abstractions, I think it just helps remove the minor bits of complexity or incongruity with the underlying CSS properties.

I really think that page/component layout generally should be abstracted to provide a *consistent* way of placing UI on the page. There is a lot more we can do in the DS to allow devs to use components to layout and construct pages, but having a simple API to make it easy to use is pretty important. I don’t think that you shouldn’t have to write CSS to achieve basic, often-repeated layouts.

You can look at a bunch of other design systems, MUI, Chakra, etc. They all provide layout primitives to help developers put stuff on a page.

On the naming stuff, I agree that we could align with the actual CSS properties, so `justifyContent` and `alignItems`, but I also don’t think it’s a particularly big deal. There is 1 use of `justifyItems` and 4 of `alignContent` in `atom-core` compared to over 1000 of `justifyContent` and `alignItems`, so it’s really clear what people use to align things on either axis. I’ve also seen examples in other design systems of using `justify` and `align`, and it’s similar(ish) to `wrap` and `direction` for `flexWrap` and `flexDirection`.

Ultimately I just think this is a nicer API for laying things out on a page

```tsx
<Stack gap="3" direction="row" justify="end">
  <Button>Hello</Button>
  <Button>Hello</Button>
</Stack>
```

PR #537, opened 2023-06-20
fix: tweak NumberInputField so it connects to Form correctly + releva…

LimeWub:
…nt NumberInput primitive changes

JIRA: https://atomlearningltd.atlassian.net/jira/software/projects/DS/boards/27?selectedIssue=DS-435

## Description
`value` of the `NumberInputFiled` when used in a Form was seemingly updated only once (on the first click) when using watch. It was correct `onSubmit` but something was clearly out of sync.
Meanwhile `onChange` was blocked due to `trigger()` being used for validation purposes on the Field, taking up the `onChange` event - so that was not available.

## Implementation
Me and Michal were trying to debug the above issue for about 1hr, we got very close, getting `onChange` to work, however, validation and updating in the form's watch was still buggy.
I looked at how other custom form fields work; Like the RadioButtonField, and I copied the style of how that one connects to the form (via `useController`) and it just worked. Slight clean up after that.

FWIW this has been tested via the sandbox but it is a quick "get it out" change so if you see anything suspect in the code I may have missed lmk. :)

**_testing with this sandbox:_**
```

import { Form, NumberInputField, globalCss } from '../src'

const App = () => (
  <TooltipProvider>
    <Form
      onSubmit={(data) => { console.log({ data }) }}
      render={({ watch }) => {
        const extraTimeValue = watch('extraTimePercent')
        console.log('extraTimeValue', extraTimeValue)

        return (
          <NumberInputField
            label="A number field"
            name="extraTimePercent"
            min={0}
            step={5}
            validation={{
              min: {
                value: 4,
                message: 'You must have at least 3 cats!'
              }
            }}
            onValueChange={(newValue) => console.log('newValue', newValue)}
          />
        )
      }}
    />
  </TooltipProvider >
)

```


lib/src/components/number-input-field/NumberInputField.tsx
```diff
@@ -19,16 +19,31 @@ export interface NumberInputFieldProps extends NumberInputProps {
...
+    rules: validation,
+    defaultValue
+  })
   const { error } = useFieldError(name)
-  const ref = validation ? register(validation) : register
+
+  React.useEffect(() => {
+    // Update the react-hook-form inner value to match what is passed in.
+    if (typeof value !== 'undefined') onChange(value)
+  }, [value])
```
Mhoog, line 46: Not sure if I fully understand all details, but could we not compare the two values and change if they differ? Outside a `useEffect`, so that we don't do this after a render?

```js
if (value && value !== innerValue)
  onChange(value)
```

LimeWub: I'm not sure but I've copied this out of RadioButton and there must be a reason why I had done it this way originally? maybe it causes infinite loops? 

I'll give it a roll and lyk.

Soz the description started out pretty bare - just trying to do this ASAP so I don't get too distracted from my actual pod work tbf.

Mhoog (approved):

lib/src/components/number-input-field/NumberInputField.tsx
```diff
@@ -19,16 +19,31 @@ export interface NumberInputFieldProps extends NumberInputProps {
...
   validation,
-  prompt,
-  description,
+  onValueChange,
   ...remainingProps
 }) => {
-  const { register, trigger } = useFormContext()
+  const { control } = useFormContext()
+  const {
+    field: { ref, onChange, value: innerValue, name: innerName }
```
michalrozenek, line 34: is inner value different than name?

michalrozenek (approved):

michalrozenek:
Looking a bit complex, but as you said - it's done accordingly to other components so I hope it's correct. 🙏

lib/src/components/number-input-field/NumberInputField.tsx
LimeWub: I think it's actually the same. But just copying it from another component - maybe we can simplify and drop it in the future.

PR #539, opened 2023-06-30
fix: Pagination bugs

Mhoog:
## Description
This PR fixes two small bugs in the `Navigation` component:
1. The 'indicated page' dot was sometimes not centered horizontally in the button.
2. There was additional spacing on the right side of the popover

## Changes
- Changed the way the indicated-page-dot is centered (now using `left` and `marginLeft`.
- Removed unnecessary elements around the indicated-page-dot.
- Changed the list of page numbers to use `Flex` with `gap` instead of `Stack`.
- Center aligned the page numbers, so that no extra space on the right appears when items move to the next line.

### Before
image

### After
image

### DIFF
--- a/lib/src/components/pagination/PaginationItem.tsx
+++ b/lib/src/components/pagination/PaginationItem.tsx
@@ -3,23 +3,12 @@ import * as React from 'react'
 import { focusVisibleStyleBlock } from '~/utilities'
 
 import { styled } from '../../stitches'
-import { Box } from '..'
 import type { IPaginationItemProps } from './types'
 import { usePagination } from './usePagination'
 
-const Dot = styled(Box, {
-  position: 'absolute',
-  bottom: '-$1',
-  left: '0',
-  transform: 'translateX(50%)',
-  borderRadius: '$round',
-  size: '4px',
-  bg: '$accent9'
-})
-
 const StyledButton = styled('button', {
   alignItems: 'center',
-  border: 'unset',
+  border: '1px solid transparent',
   borderRadius: '$0',
   cursor: 'pointer',
   fontFamily: '$body',
@@ -32,6 +21,7 @@ const StyledButton = styled('button', {
   fontWeight: 400,
   color: '$grey800',
   bg: '$base1',
+  position: 'relative',
   '&:not(:disabled)': {
     '&:hover': {
       color: '$accent10',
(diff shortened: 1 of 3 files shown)


andoulla (approved):

lib/src/components/pagination/PaginationPopover.tsx
```diff
@@ -26,7 +26,15 @@ export const PaginationPopover = () => {
         </ActionIcon>
       </Popover.Trigger>
       <Popover.Content size="md" showCloseButton={false} css={{ p: 0 }}>
-        <Stack css={{ p: '$4', display: 'flex', flexWrap: 'wrap' }} gap={1}>
+        <Flex
```
LimeWub, line 29: What's the issue with using Stack?

lib/src/components/pagination/PaginationItem.tsx
```diff
@@ -77,16 +67,26 @@ const StyledButton = styled('button', {
       true: {
         fontWeight: 600,
         color: '$accent9',
+        '&:after': {
+          content: '',
+          position: 'absolute',
+          bottom: '$1',
+          left: '50%',
+          ml: '-2px',
```
LimeWub, line 75: instead of `ml: '-2px',` which seems a bit arbitrary (I know why it's 2 but still) - can u just use `transform: 'translateX(-50%)',` which will work for every size of the dot? Should do the exact same thing ^^

lib/src/components/pagination/PaginationItem.tsx
```diff
@@ -125,14 +125,7 @@ export const PaginationItem: React.FC<IPaginationItemProps> = ({
       aria-disabled={isDisabled}
       onMouseOver={handleOnHover}
     >
-      {isIndicated ? (
```
LimeWub, line 128: you've removed the isIndicated check but I don't see where you are checking it in the css above.

Where are you checking it? Am I blind ;-;

lib/src/components/pagination/PaginationPopover.tsx
LimeWub: Oh I see it. Ugh - hopefully flex-gap `Stack` soon.

LimeWub: @thomasdigby are we already supporting using flex for spacing now or not yet? Can we make the changes to Stack soon?

lib/src/components/pagination/PaginationItem.tsx
LimeWub: I found it! Ignore me! - great change.

lib/src/components/pagination/PaginationPopover.tsx
Mhoog: We're using it in the `core` app already, so I'd assume so.

In this case, it wouldn't look strange if `gap` isn't supported either, the button size makes up for most of the spacing between the numbers.

lib/src/components/pagination/PaginationItem.tsx
Mhoog: Yup, good one. Changed it.

avirati (approved):

mzeenkala23 (approved):

PR #540, opened 2023-07-05
feat: add create password field component

Marcdup1:
### Description

Add CreatePasswordField which replaces the version in atom-core and uses inline messages rather than a Popover to display the validation messages.

### Screenshots

Shows validation when trying to submit

https://github.com/Atom-Learning/components/assets/11061661/dbf0112b-f216-4fee-99fd-efee15dadba5

Validation messages remain neutral until the input is unfocused

https://github.com/Atom-Learning/components/assets/11061661/bb657915-99f2-468b-9101-2b5462915f1a

Validation 'row' direction
image


Marcdup1:
> [in the video examples] Can you make the field be in error state, marked with red colour border, like with regular validations?

@MiguelGDLR I updated it now, the error state wasn't being applied to the `PasswordField` either if there was a validation  error. (updated the videos in the PR description)

lib/src/components/create-password-field/CreatePasswordField.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+  })
+
+  it('displays validation messages when trying to submit the form', async () => {
+    customRender()
+
+    userEvent.click(screen.getByRole('button', { name: 'Submit' }))
+
+    expect(await screen.findByText('7+ characters')).toBeVisible()
+  })
+})
```
D7Torres, line 83: I wonder if we should check for the validation to become red after loosing focus via a snapshot or something 🤔 
Thoughts?

Marcdup1: I think its a good shout, just sucks that it has to be with a snapshot

D7Torres: Yeah... that's the part of me that is hesitant about doing it.

I think we introduced visual testing in core... but I don't think is available in `components` 🤔

D7Torres: So just to be clear, not a blocker

lib/src/components/create-password-field/CreatePasswordField.tsx
```diff
@@ -0,0 +1,103 @@
...
+      const result = await validate(password)
+
+      if (result) {
+        invariant(
+          typeof result === 'object',
+          'The validate function must return an object'
+        )
+
+        setValidationResult(result)
+        return Object.entries(result).every(([_, isValid]) => isValid)
```
D7Torres, line 56: I _think_ you can use `Object.values(...)` instead of entried so you don't need to write `_`

lib/src/components/create-password-field/CreatePasswordField.tsx
```diff
@@ -0,0 +1,103 @@
...
+            mt: '$2',
+            mb: '$4',
+            gap: '$2',
+            flexDirection: messageDirection,
+          }}
+        >
+          {Object.entries(validationResult).map(([message, result]) => (
+            <InlineMessage
+              key={message}
+              theme={result ? 'success' : isFocused ? 'neutral' : 'error'}
```
D7Torres, line 92: I was always advised against nested ternary conditionals.

I don't know what everyone else thinks, it's fine by me.

If you want to avoid it, you could extract it into a function.

D7Torres (approved):
Some optional comments left, but approved.

lib/src/components/create-password-field/CreatePasswordField.tsx
andoulla: shorter ternaries like this one might be ok, longer ones where each condition is elaborate are very difficult to read, especially when it involves rendering lots of html. 

Turning into a function sounds like a good idea if we want to be consistent

lib/src/components/create-password-field/CreatePasswordField.test.tsx
andoulla: to have VR testing we need Cypress set up which this repo doesn't. might be worth to discuss it with the rest of the team, maybe it is worth the investment of time as it will make testing sooo much easier

andoulla (approved):
Once this is merged, please let me know and I can add a ticket to migrate the old one from _design-system to this one

lib/src/components/create-password-field/CreatePasswordField.tsx
Marcdup1: Ah much better thanks

lib/src/components/create-password-field/CreatePasswordField.tsx
Marcdup1: For consistency I refactored it

lib/src/components/create-password-field/CreatePasswordField.test.tsx
Marcdup1: Also, maybe when we move away from stitches this sort of test will be easy

Marcdup1:
> Hey, I've finally launched the code to see the examples. Only 1 thing I would like to ask is: is it possible to make the messages wrap not internally (the text) but the component itself? Not sure how to explain it so adding an image.

@MiguelGDLR I update it to wrap the validation messages:

image


image

kerryambt (approved):

PR #541, opened 2023-07-11
Fix: limit commit type to feat, fix or perf

Marcdup1:
Adds a commitlint rule to limit the commit message types to `feat`, `fix` or `perf` based on our [semantic versioning](https://atomlearning.design/components/overview/versioning)

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,10 @@
       "body-max-line-length": [
         0,
         "always"
+      ],
+      "type-enum": [
+        2,
+        "always", ["feat", "fix", "perf"]
       ]
     }
   },
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
       ],
       "type-enum": [
         2,
-        "always", ["feat", "fix", "perf"]
+        "always", ["build", "ci", "docs", "feat", "fix", "perf", "refactor", "test"]
       ]
     }
   },


D7Torres:
@andoulla, I read you suggesting this limit. Could you help me understand where the limit is coming from and why we cannot use all the semantic verbs [here](https://www.conventionalcommits.org/en/v1.0.0/) in point `#4`?

andoulla:
> @andoulla, I read you suggesting this limit. Could you help me understand where the limit is coming from and why we cannot use all the semantic verbs [here](https://www.conventionalcommits.org/en/v1.0.0/) in point `#4`?

We are using `semantic release` to automatically release a PR when merged. This package, when using other than those 3 verbs seems to break that flow as from what I gather it can't calculate if it is a patch, minor or major increase in the semver pattern. https://atomlearning.design/components/overview/versioning
image

When this was set up `commit lint` used to throw an error if you used something outside the scope but it seems like that doesn't work and this is what Marc is kindly fixing here.

andoulla (approved):

LimeWub:
> chore: is for updating build tasks etc; no production code change

It makes sense it's not deploying with chore. Why limit it?

Marcdup1:
@LimeWub @D7Torres 
Does the release break if any of the commits contain a commit type other than feat/fix/perf?

LimeWub:
@Marcdup1 No it doesn't break. I've used refactor/chore many times bundled with other commits and the release has gone through. I think it picks up the heaviest level one from feat/fix/perf and uses that for the bump of the DS.
However it won't release if the only commit is say - a chore. (I'm not sure what it does with single refactor commits)

Marcdup1:
> @Marcdup1 No it doesn't break. I've used refactor/chore many times bundled with other commits and the release has gone through. I think it picks up the heaviest level one from feat/fix/perf and uses that for the bump of the DS. However it won't release if the only commit is say - a chore. (I'm not sure what it does with single refactor commits)

Oh I was under the impression it would break the release, this this PR feels pretty redundant then.

D7Torres:
@andoulla in the Semmantic Commits GitHub page it says that by default uses [this](https://github.com/angular/angular/blob/main/CONTRIBUTING.md#-commit-message-format)

So all these should be supported:
image

Having said that, I agree that maybe is not worth the effort

Marcdup1:
> @andoulla in the [`semmantic-release` GitHub page](https://github.com/semantic-release/semantic-release) it says that by default uses [this](https://github.com/angular/angular/blob/main/CONTRIBUTING.md#-commit-message-format)
> 
> So all these should be supported: image
> 
> Having said that, I agree that maybe is not worth the effort

@D7Torres @LimeWub @andoulla I updated the commit type to support the commit types as linked above 👆

D7Torres (approved):
This would leave `chore` out, are we happy with it @LimeWub ?

LimeWub:
In principle yes this logic makes sense. Actually though, this is all unnecessary. 
The commit lint config extends from this: https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional

Which basically already has all these rules in there; and is checking if our commit matches them. To add these few lines of code just to exclude `chore` doesn't make much sense to me personally.

Marcdup1:
> In principle yes this logic makes sense. Actually though, this is all unnecessary. The commit lint config extends from this: https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional
> 
> Which basically already has all these rules in there; and is checking if our commit matches them. To add these few lines of code just to exclude `chore` doesn't make much sense to me personally.

Well it'll excude `style` and `revert` as well - idk if anyone uses those. It does feel a bit overkill. Im very content with killing this PR - if using those commit types dont break the release then I dont see the need.

@andoulla what do you think?

andoulla:
> > In principle yes this logic makes sense. Actually though, this is all unnecessary. The commit lint config extends from this: https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional
> > Which basically already has all these rules in there; and is checking if our commit matches them. To add these few lines of code just to exclude `chore` doesn't make much sense to me personally.
> 
> Well it'll excude `style` and `revert` as well - idk if anyone uses those. It does feel a bit overkill. Im very content with killing this PR - if using those commit types dont break the release then I dont see the need.
> 
> @andoulla what do you think?

I think it is ok to exclude chore and style. Mainly because chore is a bit to vague of what it is and style we can use feat (ie we are adding new styles as a new feature) or fix (ie we are fixing a broken style)

revert is something that might be helpful to have incase we need to remove changes from the main branch

andoulla:
> In principle yes this logic makes sense. Actually though, this is all unnecessary. The commit lint config extends from this: https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional
> 
> Which basically already has all these rules in there; and is checking if our commit matches them. To add these few lines of code just to exclude `chore` doesn't make much sense to me personally.

in the last few months, anyone that had a chore: prefix broke the automatic release flow. so either we exclude chore or we extend the right package to work with chore. I don't think it is worth the effort to extend for what we will be getting

LimeWub:
> in the last few months, anyone that had a chore: prefix broke the automatic release flow. so either we exclude chore or we extend the right package to work with chore. I don't think it is worth the effort to extend for what we will be getting

chore doesn't break the release though - it just doesn't do a release by itself

LimeWub:
also fwiw here's the commit name style guidelines: https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type

`chore` is actually deprecated (replaced by `build`) but it's not biggie to use it imo.
But ye - none of the commit keys 'break' release. The ones that aren't urgent/supposed to release instantly just don't trigger a release. I really don't think we need this change.

thomasdigby:
Hey @Marcdup1, I'm gonna close this PR for now, it feels a touch redundant as it doesn't really change anything about the release process. Happy to continue to discuss if we learn that restricting PRs to a subset of commit styles can help our versioning/release process.

PR #542, opened 2023-07-12
fix: remove bottom margin from validation messages

Marcdup1:
### Description

This removes the bottom margin validation messages. I dont think it should have been added in the first place.
 - It might add unwanted space below the form field
 - The user should rather control how much margin (if any) should be applied. It can be applied though the css prop rather

### DIFF
--- a/lib/src/components/create-password-field/CreatePasswordField.tsx
+++ b/lib/src/components/create-password-field/CreatePasswordField.tsx
@@ -87,7 +87,6 @@ export const CreatePasswordField = ({
         <Flex
           css={{
             mt: '$2',
-            mb: '$4',
             gap: '$2',
             flexWrap: 'wrap',
             flexDirection: messageDirection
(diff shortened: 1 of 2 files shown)


kerryambt (approved):

lib/src/components/create-password-field/CreatePasswordField.tsx
```diff
@@ -87,7 +87,6 @@ export const CreatePasswordField = ({
         <Flex
           css={{
             mt: '$2',
-            mb: '$4',
```
D7Torres, line 90: what it's even crazier in my opinion is that we use a margin top.
Following the same logic, should we remove it?

Marcdup1: That margin top is used for space between the input and the validation messages. It sort of feels part of the component

D7Torres: 👍 
I don't know if we still follow the rule, but we had the rule to avoid margin tops in Notion.
However, it looks like your case could be ok for them as it's quite controlled (won't be placed wherever, but only under the field)

Marcdup1: Yep I also think in this case it'll be fine, if we reverse it and add mb to the input then we have to conditionally apply it when there are errors and that feels a bit over complicated.

D7Torres (approved):

PR #545, opened 2023-07-18
Badge: New emphasis, themes and overflow truncation features

LimeWub:
JIRA: https://atomlearningltd.atlassian.net/browse/DS-406 and subsequently https://atomlearningltd.atlassian.net/browse/DS-401

## Description

1. Adds new variants that allow a new emphasis style: `subtle` (default) and `bold` (new)
2. Adds new theme variants that allow for non-semantic colours to the badges (new, currently the purple one)
3. Implements similar logic to ColorScheme for all colours including semantic.
4. Adds new overflow truncation features: `wrap` (default - better performance - breaking change potentially) and `ellipsis` Will need to cautiously update core due to the `wrap` default.
5. Additionally the component now exposes `Badge.Icon` and `Badge.Text` for more flexibility when using.

## Screenshots
Screenshot 2023-07-26 at 21 26 33

Screenshot 2023-07-26 at 21 16 58

Screenshot 2023-07-26 at 21 16 12

Screenshot 2023-07-26 at 21 16 37


mikebeat:
great job!! :)

I don't always read the code so can you guys always share with me when you create new design tokens? so that I can start documenting them and use them in Figma too, possibly reusing in other components (referring to '$backgroundBold' etc)

On the emphasis example, can you make the text say Subtle and Bold. That would be easier to read for non-devs.

lib/src/components/badge/Badge.tsx
```diff
@@ -17,90 +23,110 @@ const StyledBadge = styled(Flex, {
...
+  Text: typeof BadgeText
 }
 
+type TBadgeInnerProps = Omit<Omit<TBadgeProps, 'size'>, 'overflow'>
+
+const BadgeInner = React.forwardRef<HTMLDivElement, TBadgeInnerProps>(
+  ({ theme = 'non-semantic', emphasis = 'subtle', children, ...rest }, ref) => {
+    const { size, overflow, isOverflowing } = React.useContext(BadgeContext)
+    const [badgeElRef, setBadgeElRef] = useCallbackRefState()
+    React.useImperativeHandle(ref, () => badgeElRef as HTMLInputElement)
```
Mhoog, line 69: Why is it an `HTMLInputElement`?

LimeWub: oops

lib/src/utilities/optional-tooltip-wrapper/OptionalTooltipWrapper.tsx
```diff
@@ -0,0 +1,32 @@
...
+    )
+  }
+
+  // children could be multiple elements/components,
+  // so we need a fragment here.
+  // eslint-disable-next-line react/jsx-no-useless-fragment
+  return <>{children}</>
+}
+
+OptionalTooltipWrapper.displayName = 'OptionalTooltipWrapper'
```
D7Torres, line 32: I wonder if the component couldn't simply be called `OptionalTooltip`

documentation/content/components.feedback.badge.md
```diff
@@ -14,19 +14,77 @@ tabs:
...
+        <Badge emphasis="bold">Badge</Badge>
+      </Stack>`} language={"tsx"} />
+
+
+      ## Theme
+
+      The following `theme`s semantic themes are available: `success`, `warning`, `danger`, `neutral`, `info`.
+
+
+      Moreover, any colour that has been set up as a ColorScheme accent is also available to use.  Currently the available non-semantic colours are `grey`, `blue`, `purple`, `cyan`,  `green`, `magenta`, `red`, `teal`, `orange`, `yellow`, `lime`.
```
D7Torres, line 38: Can we link somewhere that is the source of truth instead of manually listing so this doesn't get out of date?

lib/src/components/badge/Badge.tsx
```diff
@@ -17,90 +23,110 @@ const StyledBadge = styled(Flex, {
...
+          >
+            {React.Children.map(children, (child) => {
+              if (typeof child === 'string' || typeof child === 'number') {
+                return <BadgeText>{child}</BadgeText>
+              }
+              if (React.isValidElement(child) && child.type === Icon) {
+                return <BadgeIcon {...child.props} />
+              }
+              return child
+            })}
```
D7Torres, line 112: Interesting pattern, so far what I've seen is we exposing things like `<Badge.Icon>` and `<Badge.Text>` for the consumers to use.

I see we do it here too, so: am I right to assume they can use both?

```
<Badge>
  <BadgeText>Hello</BadgeText>
</Badge>
```

and also

```
<Badge>Hello</Badge>
```

?

lib/src/components/badge/BadgeText.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { useCallbackRefState } from '~/utilities/hooks/useCallbackRef'
+import { useResizeObserver } from '~/utilities/hooks/useResizeObserver'
+
+/*
+ * Instead of sticking a resize observer on every `BadgeText`
+ * regardless of overflow type, split the resize logic into a component.
+ * Basically call the hook conditionally, only for the version with
+ * the `overflow === 'ellipsis'` which needs it.
+ * This saves us from initialising a resize observer for any badge which doesn't need it.
+ */
```
D7Torres, line 16: Excellent comment

lib/src/utilities/optional-tooltip-wrapper/OptionalTooltipWrapper.tsx
LimeWub: You aren't wrong. It probably could.
I'm copying this out `ActionIcon` from where it was named `ConditionallyWrapWithTooltip`, then mirroring the naming of `OptionalVisuallyHiddenContainer` -> but which I changed the name of to `OptionalVisuallyHidden**Wrapper**` to match the use of wrapper we tend to use, rather than `Container` (ie. `CSSWrapper`, `FieldWrapper`, `NoOverflowWrapper`).
Both of these could be called `OptionalThing` rather that `OptionalThingWrapper` and it would make as much sense to me - to be fair all the components which have `Wrapper` in the name could drop that and they would still make sense to me. Maybe `Wrapper` indicates how simple and basic the component is (?)

Given this background - do you thing it should be renamed to remove `Wrapper` from the name?

D7Torres: It's a mild suggestion, and I'll let you choose as I don't feel strongly about it.

But I think "wrapper" suggests is doing something more, and in this case is not doing anything else than rendering it optionally. "wrapper" or "container" also can suggest a piece o html wrapping the tooltip, but in this case there's none.
So I'd be ok with removing "wrapper" from it in this case, but it's your call

lib/src/components/badge/Badge.tsx
LimeWub: Ye so the Badge used to work with this format: `<Badge><Icon/>Hello</Badge>` but while I was reworking I took out the manipulations it was doing to the Icon and the Text and I moved them to their own components and exported as well.

So now both the previous and: 

```
<Badge>
  <Badge.Icon /><Badge.Text>Hello</Badge.Text>
</Badge>
```
syntaxes will work.

I chose to export the Icon and Text directly in case they ever need to be nested in a Fragment of similar. 


The reason why I didn't change the previous way of doing it is mainly backward compatibility + maintaining ease of use.

documentation/content/components.feedback.badge.md
D7Torres: Won't we have a page listing all the colours that have been set up as a ColorScheme accent?

If several components are going to refer to them, it would be a good idea to have them in a single place, even if we have to maintain it mthemanually.

lib/src/components/badge/Badge.tsx
D7Torres: > The reason why I didn't change the previous way of doing it is mainly backward compatibility + maintaining ease of use.

Then "no objections, your honor"

D7Torres (approved):
Think about centralising the list of colours, but it's good to go :D
Great job

lib/src/utilities/optional-tooltip-wrapper/OptionalTooltipWrapper.tsx
LimeWub: It is a bit arbitrary what is named a `Wrapper` and what is not. I think imma leave them in here as they are and maybe we can discuss this naming in the next DS meeting.
Because it's a bit arbitrary, I am leaning to agree on removing `Wrapper` from their name; but I'd also be similarly tempted to remove it from `NoOverflow{Wrapper}` - which is used outside the DS I think.
So ye better discuss this with the others actually and just decide on what gets that naming and what doesn't and I can rename separated.

documentation/content/components.feedback.badge.md
LimeWub: Yeeeee I'll put in a ticket to make a ColorScheme docs page actually.

lib/src/utilities/optional-visually-hidden-wrapper/OptionalVisuallyHiddenWrapper.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+  })
+
+  it('renders visible content', async () => {
+    const { container } = render(
+      <OptionalVisuallyHiddenWrapperImplementation />
+    )
+
+    expect(container).toMatchSnapshot()
+  })
+})
```
Mhoog, line 30: Might be easier to just query for `hello` instead of snapshots here. Safer, and the snapshots won't add much value here, as there's not a lot of HTML in there.

lib/src/utilities/optional-tooltip-wrapper/OptionalTooltipWrapper.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+    expect(container).toMatchSnapshot()
+  })
+
+  it('renders without tooltip', async () => {
+    const { container } = render(<OptionalTooltipWrapperImplementation />)
+
+    expect(container).toMatchSnapshot()
+  })
+})
```
Mhoog, line 35: Same comment as: https://github.com/Atom-Learning/components/pull/545/files?w=1#r1294709471

lib/src/components/badge/BadgeText.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+type TBadgeTextProps = React.ComponentProps<typeof StyledBadgeText>
+
+const toTextSize = {
+  xs: 'sm',
+  sm: 'md',
+  md: 'md'
+}
+
+export const BadgeText: React.FC<TBadgeTextProps> = ({ children, ...rest }) => {
```
Mhoog, line 58:
```suggestion
export const BadgeText = ({ children, ...rest }: TBadgeTextProps) => {
```

lib/src/components/badge/BadgeText.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const { size: badgeSize, overflow } = React.useContext(BadgeContext)
+
+  const size = React.useMemo(
+    () => overrideStitchesVariantValue(badgeSize, (s) => toTextSize[s]),
```
Mhoog, line 62: Why do we need `toTextSize` here, instead of:

```ts
() => overrideStitchesVariantValue(badgeSize, (size) => size),
```

lib/src/components/badge/BadgeIcon.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react'
+
+import { Icon } from '~/components/icon'
+import { styled } from '~/stitches'
+
+const StyledBadgeIcon = styled(Icon, { flexShrink: 0 })
+export const BadgeIcon: React.FC<
+  React.ComponentProps<typeof StyledBadgeIcon>
+> = (props) => <StyledBadgeIcon size="sm" {...props} />
```
Mhoog, line 9:
```suggestion
export const BadgeIcon = (props: React.ComponentProps<typeof StyledBadgeIcon>) => <StyledBadgeIcon size="sm" {...props} />
```

lib/src/components/badge/Badge.tsx
```diff
@@ -17,90 +23,110 @@ const StyledBadge = styled(Flex, {
...
+
+export const Badge = React.forwardRef<HTMLDivElement, TBadgeProps>(
+  ({ size = 'sm', overflow = 'wrap', ...rest }, ref) => {
+    return (
+      <BadgeProvider size={size} overflow={overflow}>
+        <BadgeInner {...rest} ref={ref} />
+      </BadgeProvider>
+    )
+  }
+) as TBadge
```
Mhoog, line 128: You can get around typecasting:

```tsx
const BadgeComponent = React.forwardRef<HTMLDivElement, TBadgeProps>(
  ...
)

export const Badge = Object.assign(BadgeComponent, {
  Icon: BadgeIcon,
  Text: BadgeText
})
```

lib/src/components/badge/Badge.context.tsx
```diff
@@ -0,0 +1,29 @@
...
+export const BadgeContext = React.createContext<TBadgeContext>({})
+
+export const BadgeProvider: React.FC<TBadgeProviderProps> = ({
+  size,
+  overflow,
+  children
+}) => {
+  const [isOverflowing, setIsOverflowing] = React.useState(false)
+
+  const value = React.useMemo<TBadgeContext>(
```
Mhoog, line 24: You can leave out the type here. Typescript will complain if this is not correct anyways when you pass it into the `BadgeContext.Provider` below.

lib/src/components/badge/BadgeText.tsx
LimeWub: Because the text size is not overriding 1 to 1.
```
const toTextSize = {
  xs: 'sm',
  sm: 'md',
  md: 'md'
}
```

But even if it was I'd keep it using the same pattern for consistency/ease of updating it in the future.

LimeWub: Actually if it was we would not even need to override. `<Text size={size} />` would be just fine.

lib/src/components/badge/Badge.tsx
LimeWub: The typecasting version feels cleaner to me as I do not have to define a new variable for it to work. Does it cause any actual issue?

lib/src/components/badge/Badge.context.tsx
LimeWub: Is it a problem to leave it in? It's more readable to me with it rather than without it as it indicates that this is the bit that's building the context.

lib/src/utilities/optional-visually-hidden-wrapper/OptionalVisuallyHiddenWrapper.test.tsx
Mhoog: What decides whether it's visible or not? Is that the `data-state` attribute? If that's the case you could use `expect().toHaveAttribute()`

lib/src/components/badge/BadgeText.tsx
Mhoog: Oh lol, did not see that it was mapping to different sizes.

All good then 👌

lib/src/utilities/optional-visually-hidden-wrapper/OptionalVisuallyHiddenWrapper.test.tsx
LimeWub: it's literally just inline css it adds - nothing to target. That's why the snapshot kinda makes sense too.

Screenshot 2023-08-17 at 11 20 08

Mhoog: Should it then have an aria attribute to indicate that it's hidden?

LimeWub: No because it's not hidden to screen readers. That's the point of it :o 
https://www.radix-ui.com/primitives/docs/utilities/visually-hidden

LimeWub:
I believe I've addressed all comments which makes sense to me to address here :)

lib/src/components/badge/Badge.tsx
Mhoog: I think the `Object.assign` option is more solid.

- You don't have to manually specify `React.ForwardRefExoticComponent<TBadgeProps>`
- If something changes about the props / ref, the inferred type will pick that up. If you typecast, the type might become invalid.

Mhoog (approved):

PR #546, opened 2023-07-18
feat: added lg size for input fields

Jdrazong:
Added lg size for the following components:

Combobox, Date input, input, password input, search input, select, number input

Zrzut ekranu 2023-07-18 o 22 25 06

[Jira ticket](https://atomlearningltd.atlassian.net/jira/software/projects/PAR/boards/11/backlog?selectedIssue=PAR-671)


lib/src/components/date-input/DateInput.tsx
```diff
@@ -72,10 +75,14 @@ export const DateInput = React.forwardRef<HTMLInputElement, DateInputProps>(
         <Popover modal open={calendarOpen} onOpenChange={setCalendarOpen}>
           <Popover.Trigger asChild>
             <ActionIcon
-              css={{ position: 'absolute', top: '0', right: '0' }}
+              css={{
+                position: 'absolute',
+                top: size === 'lg' ? '4px' : '0',
```
LimeWub, line 80: top: 4px? is this for centering? can we do it with a transform instead so we don't have random px values like this? ie. `top: '50%', translate: 'transformY(-50%)'`

LimeWub (approved):
Good work - one comment about positioning :) Otherwise LGTM

D7Torres:
@Jdrazong It would be great to have a link to the ticket in the PR description

lib/src/components/combobox/ComboboxInput.tsx
```diff
@@ -1,9 +1,12 @@
...
 import { ComboboxInput as BaseComboboxInput } from '@reach/combobox'
 
 import { styled, theme } from '~/stitches'
+import { Override } from '~/utilities/types'
 import { encodeBackgroundIcon } from '~/utilities'
 
-export const ComboboxInput = styled(BaseComboboxInput, {
+export const StyledComboboxInput = styled(BaseComboboxInput, {
   boxShadow: 'none', // prevent default iOS default styling
   fontSize: '$md', // prevent iOS zooming on focus
```
D7Torres, line 11: As you have the `fontSize` defined in the variant, does it make sense to remove this one?

lib/src/components/combobox/ComboboxInput.tsx
```diff
@@ -37,10 +40,40 @@ export const ComboboxInput = styled(BaseComboboxInput, {
...
     }
   }
 })
+
+export type ComboboxInputProps = Override<
+  React.ComponentProps<typeof StyledComboboxInput>,
+  {
+    name: string
+  }
+>
```
D7Torres, line 73: Just for my undertanding, why is this override of `name` needed?

lib/src/components/date-input/DateInput.tsx
D7Torres: And in that case... maybe we can do it for all sizes instead of checking if we are in the `lg` case, right?

lib/src/utilities/style/get-field-icon-size.ts
```diff
@@ -0,0 +1,10 @@
+import { overrideStitchesVariantValue } from '~/utilities/override-stitches-variant-value/overrideStitchesVariantValue'
+
+const toIconSize = {
+  sm: 'sm',
+  md: 'md',
+  lg: 'md'
+}
+
+export const getFieldIconSize = (size: any) =>
+  overrideStitchesVariantValue(size, (s) => toIconSize[s])
```
D7Torres, line 10: I wonder if we could call the file name something more generic like `get-icon-size`, so if we need different `getXXXIconSize()` utils in the future, we can add them here without changing the file name.

D7Torres:
@Jdrazong The documentation is missing. From the ticket:
```
In all components

* Add a test that shows the use of a non-default size. If this test already exist, you don’t need to add another one just to test the lg size

* Add the sizes with examples to the documentation (easier to add it using the CMS locally, which will make changes in the .md file ensuring the format is good).
```

please, let me know if you have any trouble using the CMS locally and we can pair on it if needed 😄

lib/src/utilities/style/get-field-icon-size.ts
LimeWub: I think we can rename it if/when we need it :o Doesn't have to be very generically named when it's a very specific content rn. ^^

D7Torres: It's fine, pretty minor. But probably what will happen is that when someone needs another `getXXXIconSize()` another file will be created

D7Torres (approved):

PR #547, opened 2023-07-19
Ds 447 build ds date field component input not updating correctly during on change validation

LimeWub:
**Edit:** There was an additional problem with this field not updating in Forms correctly which Michal brought up (JIRA 2 link). I've reworked this PR to fix this problem and remove the original remounting hack introduced.

JIRA: https://atomlearningltd.atlassian.net/browse/DS-447
JIRA 2: https://atomlearningltd.atlassian.net/browse/DS-469

Basically the bug is that when the DateField is used in combination with validation, once we submit the form and we get to a state of validating `onChange` - typing in the Input field glitches out and eats the letters we are typing as well as backspaces etc. 
From a day's worth of research/work it looks to me like this might be a bug within `react-hook-form`'s validation logic - I have not exactly pinpointed where it is in their code though. 

All I can tell is our `onChange` does not run when validation happens `onChange` (post submit) and the message is about to change. I did not find any github issues reporting on this - [the closest I got was this stackoverflow ticket.](https://stackoverflow.com/questions/65928525/react-hook-form-cant-correctly-validation-with-value-and-onchange-after-submit)

This is an attempt to fix the reported bug. It is an unorthodox method and kind of a hack - but I could not figure out the proper way to solve this. I'm assuming this is a bug or even expected functionality from `react-hook-form`. This sort of bug I'm also assuming affects any other field which relies on controlled value but is used with validation - perhaps particularly with a validation function.

Comments and alternative solutions are welcome.

**Edit:** 
I changed the solution to trigger onChange on the input instead. Which solves the issue with the component not marking as `dirty` when observed via a `Form`'s `dirtyFields`

## Tested with sanbox
```

import { Form, DateField, globalCss, InputField } from '../src'
import { DEFAULT_DATE_FORMAT } from '../src/components/date-input/constants'

import dayjs from 'dayjs'
import customParseFormat from 'dayjs/plugin/customParseFormat'
dayjs.extend(customParseFormat)

const formatDateToString = (date?: Date, dateFormat = DEFAULT_DATE_FORMAT) => date ? dayjs(date).format(dateFormat) : ''

const App = () => {

  return (
    <TooltipProvider>
      <Form
        onSubmit={(data) => { console.log({ data }) }}
        validationMode={undefined}
        defaultValues={{
          'name': '',
          'exam-date': formatDateToString(new Date()),
        }}
        render={({ watch, formState }) => {
          console.log('dirtyFields', formState.dirtyFields)

          return (
            <>
              <InputField
                name={'name'}
                defaultValue=""
                validation={{
                  validate: (value) => {
                    if (value.length >= 2) {
                      return 'too long';
                    }
                    return true
                  }
                }}
              />
              <DateField name="exam-date" label="Exam date" validation={{
                validate: (value) => {
                  var now = new Date;
                  const d = value.slice(0, 2),
                    m = value.slice(3, 5),
                    y = value.slice(6, 10);
                  var target = new Date(y, m - 1, d);
                  console.log({ now, target })

                  if (target >= now) {
                    console.log('Date in the future!')
                    return 'in future';
                  }
                  return true
                }
              }} initialDate={new Date()} />
            </>
          )
        }}
      />
    </TooltipProvider >
  )
}

```

## Videos

**before**

https://github.com/Atom-Learning/components/assets/6905473/2a474504-480b-4f76-b840-36ad19043a56

**after**

https://github.com/Atom-Learning/components/assets/6905473/a116b7f0-1153-475f-8cf6-e26e6f540474


lib/src/components/date-input/DateInput.tsx
```diff
@@ -43,7 +46,41 @@ export const DateInput = React.forwardRef<HTMLInputElement, DateInputProps>(
...
+      [dateFormat]
+    )
+
+    const [inputRefreshKey, setInputRefreshKey] = React.useState(0)
+    const handleCalendarChange = React.useCallback(
+      (newDate) => {
+        const parsedInputDate = dayjs(newDate)
+        setDate(newDate)
+        setDateString(parsedInputDate.format(dateFormat))
+        setInputRefreshKey(Math.random()) // Remount the Input
```
D7Torres, line 79: Could we increase `inputRefreshKey` in 1 instead?

D7Torres (approved):

lib/src/components/date-input/DateInput.tsx
LimeWub: Done 👍

lib/src/components/date-input/DateInput.tsx
```diff
@@ -44,7 +47,41 @@ export const DateInput = React.forwardRef<HTMLInputElement, DateInputProps>(
     },
     ref
   ) => {
-    const { date, dateString, setDate } = useDate(initialDate, dateFormat)
+    const [date, setDate] = React.useState(
+      initialDate ? dayjs(initialDate).toDate() : undefined
+    )
+
+    const [dateString, setDateString] = React.useState(
```
michalrozenek, line 54: I can just suggest to remove `dateString` and  `setDateString` from state, because anyway we set it always together with `setDate`.
Instead i propose to just declare it below as a variable that is something kind of

```
const formatDateToString = (date: Date) => {
  date ? dayjs(date).format(dateFormat) : ''
}

const dateString = formatDateToString(date)
```

michalrozenek: ☝️ ofc just a proposal + you'd need to double check if that's working as expected

michalrozenek (approved):
Honestly I am not 💯 to go this way because it looks hacky, but I am ok if you tested it properly if that works.

So, on the one hand - if it solves the problem, its' better than now. On the other hand it's hacky :slightly_smiling_face:
Maybe we can merge it now and do more research + refactor working on isDirty issue.
Maybe it's more than small changes here, maybe complex refactor from scratch, dunno

D7Torres (approved):

lib/src/components/date-input/DateInput.tsx
```diff
@@ -44,7 +51,52 @@ export const DateInput = React.forwardRef<HTMLInputElement, DateInputProps>(
...
+      [dateFormat]
+    )
+
+    const handleCalendarChange = React.useCallback(
+      (newDate) => {
+        setDate(newDate)
+
+        const mirrorChangeToInputElement = () => {
+          if (!inputElRef) return
+          const nativeInputValueSetter = Object.getOwnPropertyDescriptor(
```
michalrozenek, line 84: Does it works fine with other form fields (meaning the form with few inputs)?

lib/src/components/date-input/DateInput.tsx
```diff
@@ -44,7 +51,52 @@ export const DateInput = React.forwardRef<HTMLInputElement, DateInputProps>(
...
+            window.HTMLInputElement.prototype,
+            'value'
+          )?.set
+          nativeInputValueSetter?.call(
+            inputElRef,
+            formatDateToString(newDate, dateFormat)
+          )
+          const event = new Event('input', { bubbles: true })
```
michalrozenek, line 92: I think that would be great to add comments to this fn to describe what happens in particular line.

It's not something really common and people might be confused.

lib/src/components/date-input/use-date.test.ts
```diff
@@ -1,51 +0,0 @@
-import { act, renderHook } from '@testing-library/react-hooks'
```
michalrozenek, line 1: What about tests?
Don't we need these or it's broken now?

lib/src/components/date-input/DateInput.tsx
LimeWub: I'll add a link to the StackOverflow I got this from :o

lib/src/components/date-input/use-date.test.ts
LimeWub: I deleted the utility and doing everything in the file. Will check if I need to add any tests to `DateInput` to cover these.

lib/src/components/date-input/DateInput.tsx
LimeWub: I don't understand this question

PR #548, opened 2023-07-24
fix: move left arrow in TabsTriggerList so always over list, negate t…

LimeWub:
…abIndex as not useful to tab on.

JIRA: https://atomlearningltd.atlassian.net/browse/DS-448

## Description

Fixes bug with left TabsTriggerList arrow.
Adds negative TabIndex to the arrows as they are useless/obstructive for navigation (Tabbing actually selects tabs and moves scroll on the TabsTriggerList so much better)

## Screenshots

**before**
Screenshot 2023-07-24 at 11 44 35

**after**
Screenshot 2023-07-24 at 11 44 02


lib/src/components/tabs/TabsTriggerList.tsx
```diff
@@ -96,13 +98,12 @@ export const TabsTriggerList: React.FC<
             left: 0
           }}
           onClick={() => scrollList(-SCROLL_STEP)}
+          tabIndex={-1}
```
thomasdigby, line 101: Surely these action icons are available in the tab order already as they're buttons, what does this `tabIndex` change?

LimeWub: it makes the arrow scroll buttons unfocusable via tabbing. 
They don't need to be focusable as it's not useful to someone navigating with tabbing, which can instead just focuses the tab buttons and does the scroll based on the button u are on.

D7Torres: Do you, perhaps, have the PR screenshots swapped? (before <-> after)

D7Torres (approved):

lib/src/components/tabs/TabsTriggerList.tsx
LimeWub: Nope - I think the one in the second snapshot is outlined because I had clicked it before trying to take a snapshot...which switched me over to keyboard nav and it was my currently focused button, so focus visible style appeared.

PR #549, opened 2023-07-25
feat: remove uses of asWorkaround in favour of stitches withConfig

LimeWub:
Found out during the Flex/Stack work that `asWorkaround` is not needed. Instead there is an actual workaround in stitches to ask it to not use the `as` prop and pass it through. Which works perfect for our use cases.

No visual changes.


thomasdigby (approved):
🙏🙏

D7Torres (approved):
👏

PR #552, opened 2023-08-21
Feature: `Button asChild` / `Link asChild`

thomasdigby:
This PR changes how the `Link` and `Button` components handle polymorphism. Instead of using `as="a"`/`as="button"` internally or externally, we can now use `asChild` and render the required element as children. This pattern helps with a bunch of typescript issues and retains the composability of styling and logic that we had previously. We use the Radix `Slot` component internally to merge any `style`, `className`, `ref`, `onClick` props, so we can compose `Link` with any child component.

I've also added logic for some useful (necessary?) attributes in the `Link` component if the `href` is not relative. This logic is currently handled by our "middleware" design-system components in `atom-core`, but they will eventually be removed so I moved it here. We lose the ability to do this with our `Button` components however, as we delegate responsibility for rendering the element to the user, so we can't repeat the same logic in `Button`. I think this is an ok compromise as the logic is not complex at all and the number of non-relative href "button"s we render is minimal.

There are ~100 instances of `Link` and `Button` components that will need to be updated in `atom-core`, but it should be a fairly simple migration. I have avoided updating `ActionIcon` with this logic as we're looking at deprecating the component soon, but I could look to update `Tile` if we like this approach.

```jsx
<Button asChild css={{ mb: '$2' }}>
  <a href="https://google.com">Click here</a>
</Button>

// will render

<a href="https://google.com" class="c-fkUJsw c-fkUJsw-dXsdwv-size-md c-fkUJsw-gsycxF-cv c-fkUJsw-ijkLxnS-css">
  Click here
</a>
```

```jsx
<Link asChild>
  <button onClick={() => console.log('yolo')}>Click here</button>
</Link>

// will render

<button class="c-fucOQC c-fucOQC-gvmVBy-size-md c-fucOQC-fmPQHp-noCapsize-true">
  Click here
</button>
```

With this pattern we get nicely merged `css` props, valid types and autocomplete on the child and parent, and composed `onClick` methods as expected

```jsx
<Button
  asChild
  css={{ bg: 'red' }}
  onClick={() => console.log('bonjour')}
>
  <Box
    as="a"
    href="/yolo"
    css={{ bg: 'yellow' }}
    onClick={() => console.log('hello')}
  >
    Click me
  </Box>
</Button>
```


lib/src/components/button/Button.tsx
```diff
@@ -189,108 +193,72 @@ export const StyledButton = styled('button', {
...
-      return 16
-  }
-}
-
-const getChildren = (children, size) =>
-  React.Children.map(children, (child, i) => {
-    if (child?.type === Icon) {
-      return React.cloneElement(child, {
-        css: {
-          [i === 0 ? 'mr' : 'ml']: size === 'sm' ? '$2' : '$3',
```
thomasdigby, line 227: I'm super happy that we can remove this logic thanks to... `gap` support! woohoo!

Mhoog: https://www.youtube.com/watch?v=sscfMV5rd6s

lib/src/components/stepper/types.ts
```diff
@@ -17,9 +17,9 @@ export type Direction = 'vertical' | 'horizontal'
 
 export type Context = {
   steps: Step[]
-  goToPreviousStep?: () => void
-  goToNextStep?: () => void
-  goToStep?: (index: number) => void
+  goToPreviousStep: () => void | undefined
+  goToNextStep: () => void | undefined
+  goToStep: (index: number) => void | undefined
```
Mhoog, line 22: Is this change needed? Seems like the exact same type.

Mhoog: Never mind, read that wrong.

Mhoog (approved):

lib/src/components/stepper/types.ts
thomasdigby: To be clear I'm not entirely sure why the `undefined` is necessary, but I think I know why they needed to be changed.

So I had some types failing after implementing the `asChild` change, which must related to either `Link` or `Button` dropping the `NavigatorActions` type, as it defined a narrower `onClick` type. We're now reliant on the default Stitches types to define `onClick` which required this change.

Marcdup1 (commented):
Very cool 😎

elliotlaws (approved):

Marcdup1 (approved):

lib/src/components/button/Button.tsx
```diff
@@ -189,108 +193,80 @@ export const StyledButton = styled('button', {
...
+          isLoading={isLoading || false}
+          onClick={!isLoading ? onClick : undefined}
+          {...props}
+          as={Slot}
+        />
+      )
+    }
 
-    // Note: button is not disabled when loading for accessibility purposes.
-    // Instead the click action is not fired and the button looks faded
```
D7Torres, line 271: Isn't the comment still true?
If so, I think it's a useful coment

lib/src/components/button/Button.test.tsx
```diff
@@ -183,14 +183,5 @@ describe(`Button component`, () => {
...
     })
-
-    it('renders an anchor if provided a link', async () => {
-      render(<Button href="https://atomlearning.co.uk">ATOM</Button>)
-
-      expect(await screen.findByRole('link')).toHaveAttribute(
-        'href',
-        'https://atomlearning.co.uk'
-      )
-    })
```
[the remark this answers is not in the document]
thomasdigby: We test `asChild` in the test directly above!

lib/src/components/button/Button.tsx
thomasdigby: Yeah true, I'll add it back in!

lib/src/components/button/Button.tsx
```diff
@@ -189,108 +193,81 @@ export const StyledButton = styled('button', {
...
+      children,
+      ref,
+      // Note: button is not disabled when loading for accessibility purposes.
+      // Instead the click action is not fired and the button looks faded
+      isLoading: isLoading || false,
+      onClick: !isLoading ? onClick : undefined
+    }
 
-    const buttonSpecificProps = href ? {} : { type }
+    if (asChild) {
```
andoulla, line 261: I don't understand how the loader will work if we are using the asChild option?

thomasdigby:
Closing this PR while we look into alternative options for improving our polymorphism

PR #557, opened 2023-08-25
fix: use hashes for classes created via createTheme

LimeWub:
Spotted mistake, after a great question by @thomasdigby 
Screenshot 2023-08-25 at 11 52 13

I have been naming all of the classes in all my `createTheme` uses 🤦  That's what I get for copy/pasting and not reading the docs **thoroughly**... This is obviously not great, as we end up with generic classnames like `danger`!

In this PR, I'm undoing this mistake and letting `createTheme` classes just be hashes. As they should.

**No visual changes** but defo important css changes :)


thomasdigby (approved):
🎉

D7Torres (approved):

PR #558, opened 2023-08-29
fix: add correct polymorphic types for Button component

Mhoog:
We had some issues with the types of `Button`:
- If you used the `as` prop (or rendered it as `a` by adding a `href`), Typescript would still only check for attributes on a `button`.
- If you forwarded a ref, this had to be of type `HTMLButtonElement`, even if you rendered as another element.

This PR fixes those types.

Screenshot 2023-08-29 at 11 00 09


lib/src/components/button/Button.tsx
```diff
@@ -219,60 +227,59 @@ const WithLoader = ({
   </>
 )
 
-type ButtonProps = Override<
-  React.ComponentProps<typeof StyledButton>,
+type ButtonProps<
+  H extends string | undefined,
+  C extends React.ElementType
```
avirati, line 232: Optional parameter should be at the right.

```suggestion
  C extends React.ElementType,
  H extends string | undefined,
```

Otherwise, we might see types like - `ButtonProps<undefined, ...>`

avirati (approved):

lib/src/components/button/Button.tsx
Mhoog: These should always be taken from the `href` and `as` props and never manually defined.

PR #560, opened 2023-08-30
Fix: `Link` polymorphic types

thomasdigby:
Applied @Mhoog's polymorphic type fixes to the `Link` component (which has the same (but opposite) `a || button` style polymorphism)

### DIFF
--- a/lib/src/components/link/Link.tsx
+++ b/lib/src/components/link/Link.tsx
@@ -1,14 +1,14 @@
 import * as React from 'react'
 
 import { styled } from '~/stitches'
-import { NavigatorActions } from '~/types'
-import { Override } from '~/utilities'
+import { PolymorphicComponentPropWithRef } from '~/types'
 import { isExternalLink } from '~/utilities/uri'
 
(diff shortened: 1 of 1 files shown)


Mhoog (approved):

lib/src/components/link/Link.tsx
```diff
@@ -33,29 +33,51 @@ export const StyledLink = styled('a', {
...
+  ) => {
+    const externalLinkProps = isExternalLink(href)
+      ? { target: '_blank', rel: 'noopener noreferrer' }
+      : {}
 
-Link.displayName = 'Link'
+    return (
+      <StyledLink
+        as={as || (!href ? 'button' : undefined)}
+        noCapsize={!href ? true : undefined}
```
andoulla, line 75: very small but can we update the documentation to mention the capsize change? it will probably trip up a couple people

thomasdigby: This existed previously but in a different format `{...(!href && { as: 'button', noCapsize: true })}`, so there's no change here

PR #561, opened 2023-08-30
Feature: `Icon` spacing/positioning within `Link`

thomasdigby:
Similar to how `Button` handles `Icon` as a child, this PR adds positioning and sizing to an `Icon` when rendered within a `Link`. e.g.

```jsx
<Link>
  <Icon is={ChevronLeft} />
  Go back
</Link>
```

Unfortunately we can't use the same `flex`/`gap` convention from `Button` as we're having to fight against capsize a bit, but I think the solution is still fairly simple.

Screenshot 2023-08-29 at 22 24 35


D7Torres (approved):

andoulla (approved):
Finally, no more debating what margin it should be!

lib/src/components/link/Link.tsx
```diff
@@ -1,15 +1,36 @@
...
 import { StyledLi } from '../list/List'
 import { StyledMarkdownEmphasis } from '../markdown-content/components'
 import { StyledText, textVariants } from '../text/Text'
 
+const iconSizeVariants = {
+  xs: { size: 14, mt: '-2px' },
+  sm: { size: 16, mt: '-2px' },
+  md: { size: 20, mt: '-4px' },
+  lg: { size: 24, mt: '-4px' },
+  xl: { size: 30, mt: '-4px' }
```
thomasdigby, line 20: I initially put these sizes in to broadly match the x-height of the sibling text, but I have realised that the icons have specific sizes across the rest of the UI that we might want to match. @MiguelGDLR What would you prefer to do about this? Would you rather custom icon sizes that match the link text? Or stricter, consistent sizes that may not align nicely with the text?

MiguelGDLR: I think we should have strict DS sizes. That's how it is in Figma and we didn't notice any inconvenient (besides of course it probably looking a bit nicer the way you did it for some icons). See example.

But DS is all about using our tokens and scales systematically.
We might also change the library during rebranding so having something standard and easy might be better than just matching visually to the current icons/text sizes. 
Screenshot 2023-11-20 at 16 56 28

What do u think?

thomasdigby: Yeah this sounds good, I'll update and when discuss/review this PR again when it's ready

PR #562, opened 2023-08-30
Fix: Pagination UX

thomasdigby:
### Description

An attempt to improve the UX of the pagination component without drastically changing how it currently works.

https://atomlearningltd.atlassian.net/browse/DS-419

https://www.figma.com/file/KjgQh7eTwEfJmH6e56k9mX/DS-Documentation?type=design&node-id=5093-38072&mode=design&t=rnqD2dDVPwvs65Ov-0

### Implementation

- Move the popover button to the start of the item list when appropriate, avoiding the dreaded `5 6 ... 7`
- Attempt to simplify and improve the `getPaginationItemsToRender()` function, hopefully making it clearer what gets returned and _why_

Screenshot 2023-08-30 at 17 41 49

(ignore missing arrows in first pagination)

https://github.com/Atom-Learning/components/assets/3226804/d7beda23-5f6d-446b-b09c-f20621cba9d5

### DIFF
--- a/lib/src/components/pagination/pagination.constants.ts
+++ b/lib/src/components/pagination/pagination.constants.ts
@@ -4,3 +4,7 @@ export enum VisibleElementsAmount {
   LESS = 6,
   MORE = 8
 }
+
+export const MAX_PAGINATION_ITEMS = 4
+
+export const MIN_PAGINATION_ITEMS = 2
--- a/lib/src/components/pagination/pagination.constants.ts
+++ b/lib/src/components/pagination/pagination.constants.ts
@@ -4,7 +4,3 @@ export enum VisibleElementsAmount {
   LESS = 6,
   MORE = 8
 }
-
-export const MAX_PAGINATION_ITEMS = 4
-
-export const MIN_PAGINATION_ITEMS = 2
--- a/documentation/content/components.navigation.pagination.md
+++ b/documentation/content/components.navigation.pagination.md
@@ -4,7 +4,7 @@ title: Pagination
 tabs:
   - title: Code
     content: >-
-      
+
       NOTE: Currently, only the `md` variant has been implemented.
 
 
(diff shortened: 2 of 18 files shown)


MiguelGDLR:
Sorry, my bad, I told that to Tom without checking designs. 
I thought without the arrows would make sense for small amount of pages because it removes the need for ellipsis and it makes it more simple but Myrto is right, we agreed on leaving the arrows for consistency, since it's just navigating.

thomasdigby:
@MiguelGDLR @LimeWub Thanks both! I'll revert 😄

lib/src/components/pagination/pagination.constants.ts
```diff
@@ -4,3 +4,7 @@ export enum VisibleElementsAmount {
   LESS = 6,
   MORE = 8
 }
+
+export const MAX_PAGINATION_ITEMS = 4
```
andoulla, line 8: how come these arent 5 and 3 respectively?

thomasdigby: Because the 1st and last pagination items are rendered separately from this list, so we only want to return 4 or 2 as the 5th and 3rd items are separate.

andoulla (approved):

Mhoog (approved):

lib/src/components/pagination/pagination.constants.ts
LimeWub: This is one thing I also find confusing tbf.

andoulla: it could be just a comment explaining this. 
```suggestion
export const MAX_PAGINATION_ITEMS = 4 // excluding first and last pagination items that are always rendered
```

thomasdigby: Yeah that's fair, I didn't want to rock the boat too much by changing these values but a comment could help

PR #563, opened 2023-09-01
Pagination: further improvements

LimeWub:
## Description/ Implementation

This PR was originally written as a proposal at the same time as the most recent changes @thomasdigby had made to the pagination, at the time it was not reviewed as the changes to the pagination would be blocking but I believe it is worth considering due to the below merits:

### 1. Simplifies mental workload regarding what items are included in the pagination and how that is calculated based on the `visibleElementsCount`

Changes the code in `pagination.helper` to include ALL of the Pagination Items in its output. Including `Arrows` and the `ViewAllPopover`. 

This was done to bring all the logic in one place (**now:** the helper function) rather than having it split between various views (**previously:** `PaginationPages`, `PaginationItems` and the helper function and subtracting seemingly random numbers from `visibleElementsCount` to account for permanent elements in these views). 

The new code has everything in one place so it is easier to understand and all the mental workload focuses on one function rather than multiple.
Additionally as we now simplify the code by being explicit in what the helper returns: **the full array of every pagination element**, we no longer need the `paginationAlignment` context property which was previously a confusing was to indicate where the `ViewAllPopover` should appear in the pagination elements.

eg. for input `visibleElementsCount={6}` the `getPaginationElementsToRender` function returns as such...
**previous output:** `[1, 2]`
**new output:** `[GO_TO_PREVIOUS_PAGE, 1, 2, VIEW_ALL_POPOVER, 8, GO_TO_NEXT_PAGE]`

### 2. Adds support for any number as a prop for `visibleElementsCount` (but for now leaves the TS asking for either `6` or `8` in place)

The `getPaginationElementsToRender` function has been tweaked to work with simple math rather than returning specific output based on the value of `6` or `8`. This means it now works for any number of visible elements and even has contingencies for lower numbers where certain elements would need to not be rendered.

https://github.com/Atom-Learning/components/assets/6905473/59e2ae3d-19c4-4f73-a0e9-37cbd30bcea9

### 3. Fixes logic for disabled Previous/Next arrow

**before**

https://github.com/Atom-Learning/components/assets/6905473/b627c9a6-52bb-4a2e-a643-2b8abc6699e9

**after**

https://github.com/Atom-Learning/components/assets/6905473/658d1287-dadb-423c-b170-3abe8e6a421a


MiguelGDLR (approved):

lib/src/components/pagination/Pagination.tsx
```diff
@@ -2,32 +2,44 @@ import * as React from 'react'
...
-        </PaginationProvider>
-      </Flex>
-    </ColorScheme>
+    <PaginationProvider {...paginationProviderProps}>
+      <ColorScheme base="grey1" accent="blue1" {...colorScheme} asChild>
+        <Flex gap={1} {...rest}>
+          <PaginationItems />
+        </Flex>
+      </ColorScheme>
+    </PaginationProvider>
```
D7Torres, line 42: I'm sure you have a good reason to move the PaginationProvider outside of ColorScheme, but could you explain?

lib/src/components/pagination/pagination.helper.ts
```diff
@@ -1,145 +1,180 @@
...
+
+  if (canFitViewAllPopover) {
+    // Subtract one from visiblePagesCount to accomodate for us adding the `[…]` VIEW_ALL_POPOVER
+    visiblePagesCount -= 1
   }
 
+  if (canFitEdgePage) {
+    // Subtract one more from visiblePagesCount to accomodate for us adding the GO_TO_PREVIOUS/NEXT_PAGE arrows
+    visiblePagesCount -= 1
+  }
```
D7Torres, line 60: I don't get this, doesn't seem like being able to fit the edge pages should reduce the number. Call?

D7Torres (commented):
Sorry, I still need a bit more time to finish

lib/src/components/pagination/pagination.helper.ts
```diff
@@ -1,145 +1,180 @@
...
-   * render the entire page list.
+   * If fewer pages than threshold for truncating render them all.
    *
-   * We need to remove `2` from `visibleElementsCount` to account
-   * for the previous & next buttons
-   *
-   *  pagesCount: 4
-   *  visibleElementsCount: 6 // VisibleElementsAmount.LESS
-   *  returns [1, 2, 3, 4]
+   * pagesCount: 6
```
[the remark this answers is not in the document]
LimeWub: no this is about the `pages` aka how many pages are in this pagination; it's not the items/elements, that's `visibleElementsCount`

lib/src/components/pagination/pagination-context/PaginationContext.tsx
```diff
@@ -46,37 +44,27 @@ export const PaginationProvider: React.FC<TPaginationProviderProps> = ({
...
+    if (nextAvailablePage) goToPage(nextAvailablePage)
+  }, [goToPage, nextAvailablePage])
 
-  const paginationItems = getPaginationItemsToRender(
-    currentPage,
-    pagesCount,
-    visibleElementsCount
-  )
-  const paginationAlignment = getPaginationAlignment(
+  const paginationItems = getPaginationElementsToRender(
```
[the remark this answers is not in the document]
LimeWub: If anything it should always be called `Elements` because the prop is `visibleElementsCount` and not `visibleItemsCount`

lib/src/components/pagination/Pagination.tsx
LimeWub: Ok I checked this again and there's defo no effect to this change it's just a style change ^^

lib/src/components/pagination/pagination-context/PaginationContext.tsx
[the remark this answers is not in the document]
LimeWub: So I went to change this but it's a bunch of changes to keep consistency! 

For example:
* if we want to keep `Items` as the name of what this is - then the prop `visibleElementsCount` needs to be renamed to `visibleItemsCount` which is a change that affects the API.
* if we want to keep `Elements` the file `PaginationItem` needs to be renamed to `PaginationElement` and that's inconsistent with some of our other components ie. `AccordionItem`

I'll leave as is for now if it's okay....

lib/src/components/pagination/pagination.helper.ts
LimeWub: Did we end up having this call? I don't remember. It makes sense though as the `edgePage` is an additional space that will need to be accounted for - which in very few `visibleElementsCount` setting would not need to be accounted for as it wouldn't be able to fit.

lib/src/components/date-input/__snapshots__/DateInput.test.tsx.snap
```diff
@@ -450,14 +450,14 @@ exports[`DateInput component renders lg size 1`] = `
     display: table;
   }
```
LimeWub, line 452: Non-change change but otherwise snapshots were failing. Not sure what happened here at all but it's exactly the same so 🤷

D7Torres (approved):

PR #564, opened 2023-09-01
Fix: Pagination truncation issue

thomasdigby:
Remove hard-coded `TRUNCATED_THRESHOLD` and use `visibleElementsCount` instead

### Before

Screenshot 2023-09-01 at 09 43 46
Screenshot 2023-09-01 at 09 43 40

### After

Screenshot 2023-09-01 at 09 43 23

### DIFF
--- a/lib/src/components/pagination/pagination.helper.ts
+++ b/lib/src/components/pagination/pagination.helper.ts
@@ -40,9 +40,19 @@ export const getPaginationItemsToRender = (
    * We need to remove `2` from `visibleElementsCount` to account
    * for the previous & next buttons
    *
-   *  +---+  +---+  +---+  +---+
-   *  | 1 |  | 2 |  | 3 |  | 4 |
-   *  +---+  +---+  +---+  +---+
+   *  pagesCount: 4
+   *  visibleElementsCount: 6 // VisibleElementsAmount.LESS
+   *  returns [1, 2, 3, 4]
+   *  +---+  +---+  +---+  +---+  +---+  +---+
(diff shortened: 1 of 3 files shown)


andoulla (approved):

lib/src/components/pagination/pagination.helper.ts
```diff
@@ -38,13 +35,16 @@ export const getPaginationItemsToRender = (
...
    * If there are fewer pages than our threshold for truncating,
-   * render the entire page list
+   * render the entire page list.
+   *
+   * We need to remove `2` from `visibleElementsCount` to account
+   * for the previous & next buttons
    *
    *  +---+  +---+  +---+  +---+
    *  | 1 |  | 2 |  | 3 |  | 4 |
    *  +---+  +---+  +---+  +---+
```
D7Torres, line 45: What are these boxes with 4 numbers?

They are not saying anything to me, maybe we can remove them.

D7Torres (approved):

LimeWub (approved):

lib/src/components/pagination/pagination.helper.ts
thomasdigby: This is a visual representation of the pagation, based on the current page, total page count and visible elements constraint. I added arrows to make it clearer. It helped me to understand what shape the data was in with each conditional and what we were expecting to return/render.

D7Torres: yes, sorry, this is good. I didn't see the other examples and didn't have the full context. I was not aware, for example, than in some cases we would have only the page numbers without arrows.

Now that I've seen everything it's clearer

PR #570, opened 2023-09-07
chore: Replace deprecated Stack component with Flexbox and update tests for BannerRegularActions

avirati:
No visual changes expected

Before : 
Screenshot 2023-09-07 at 07 37 35

After :
Screenshot 2023-09-07 at 07 31 20

### DIFF
--- a/lib/src/components/banner/banner-regular/BannerRegularActions.tsx
+++ b/lib/src/components/banner/banner-regular/BannerRegularActions.tsx
@@ -3,24 +3,34 @@ import * as React from 'react'
 
 import { overrideStitchesVariantValue } from '~/utilities/override-stitches-variant-value/overrideStitchesVariantValue'
 
-import { Stack } from '../../stack'
+import { Flex } from '../../flex'
 import { useBannerContext } from '../BannerContext'
 import { BannerRegularButton } from './BannerRegularButton'
 
 const MAX_ALLOWED_CHILDREN = 2
 
 const toGap = {
-  sm: 2,
-  md: 4
+  sm: '$2',
+  md: '$4'
+}
+
+const toDirection = {
+  sm: 'column',
+  md: 'row'
 }
 
 export const BannerRegularActions: React.FC<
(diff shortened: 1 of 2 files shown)


andoulla (approved):

lib/src/components/banner/banner-regular/BannerRegularActions.tsx
```diff
@@ -3,34 +3,51 @@ import * as React from 'react'
...
   )
 
   return (
-    <Stack gap={gap} {...props}>
+    <Flex
+      css={{
+        gap,
+        flexDirection,
+        ...css
+      }}
```
thomasdigby, line 48: You can do similar things with `Flex` now that you used to be able to do only on `Stack` https://atomlearning.design/components/layout/flex. That will help to simplify the CSS stuff here

```jsx
<Flex gap={gap} direction={flexDirection} {...props}>
```

thomasdigby: You'll need to revert your `toGap` changes along with this

thomasdigby:
Worth keeping in mind that `Flex` doesn't provide a default `justifyContent`, `alignItems` or `flexWrap` like `Stack` used to, so you should test a few responsive layouts to confirm that it still works as expected

lib/src/components/banner/banner-regular/BannerRegularActions.tsx
avirati: Makes sense, thanks

avirati: Implemented the changes 👍

thomasdigby:
> While developing this component, we decided that this component will not depend on css breakpoints but instead, use `size` prop to decide how it would look. Do you think we still need to address these ?

Yeah that's fine, you just need to be aware that it will no longer align, justify, and wrap content like it used. It might be fine, but maybe more content might trigger a layout regression

thomasdigby (approved):

PR #571, opened 2023-09-12
Feature: add subRow support to DataTable SP-1745

alexandrunemes:
This PR introduces subRow support to the DataTable. If the data passed into the table is structured as
``` {
       ...data,
       subRows: [
           ...data
       ]
```
the table will display appropriate Chevron icons next to the expandable rows. The checkbox on each row now also supports the indeterminate state, so that when only some of the subRows are selected, it is displayed.

I've also added a pretty hefty unit test that also provides support w.r.t how one would perform bulk actions on the selected rows, without affecting the others. 

## Video recording

https://github.com/Atom-Learning/components/assets/9009155/06eb4146-35a8-4433-a962-9785bb9278ae


alexandrunemes:
@MiguelGDLR , do you think the indentation of the subRow checkboxes is enough?

MiguelGDLR:
> @MiguelGDLR , do you think the indentation of the subRow checkboxes is enough?

Idk, it could perfectly be 32 instead of 24 but then we are taking a lot of horizontal space between the chevron, checkbox... not good in mobile where space is very limited, but I see your concern.  🤷🏻‍♂️

MiguelGDLR:
Video at 0:31 feels weird. if the parent is selected shouldn't the child be selected?

lib/src/components/data-table/DataTableRow.tsx
```diff
@@ -25,16 +27,37 @@ const StyledRow = styled(Table.Row, {
...
+    if (row.getIsSomeSelected()) return 'indeterminate'
+    return row.getIsSelected()
+  }
 
   return (
     <StyledRow isSelected={row.getIsSelected()}>
+      {getCanSomeRowsExpand() && (
+        <Table.Cell
+          data-testid={`expand-icon-${row.id}`}
+          css={{ width: '$4', cursor: 'pointer' }}
```
viniciusvolpe, line 45: wouldn't this make an empty cell clickable if `row.getCanExpand()` returns false? isn't better to move the `cursor: pointer` to the icon?

alexandrunemes:
> Video at 0:31 feels weird. if the parent is selected shouldn't the child be selected?

i think in that scenario, it should be fine. you're selecting the parent, and all the subrows also get selected. then you deselect the subrows one by one, and when you're left with no subrows selected, it's just the parent that remains selected, so its icon changes from `indeterminate` to `checked`

viniciusvolpe (approved):
LGMT just added a small comment about the clickable column

thomasdigby (approved):

PR #572, opened 2023-09-13
Feature: sticky header on table header SP-1779

alexandrunemes:
This PR introduces a sticky header opt-in feature on the DataTable. To enable it, the user must add an `hasStickyHeader` flag to the DataTable provider.

### Video recording
https://github.com/Atom-Learning/components/assets/9009155/fe347cbd-ff87-4b41-9c15-fe21a7125256


lib/src/components/table/TableHeader.tsx
```diff
@@ -29,6 +29,13 @@ const StyledTableHeader = styled('thead', {
           color: '$tonal600'
         }
       }
+    },
+    hasStickyHeader: {
+      true: {
+        position: 'sticky',
+        top: 0,
+        zIndex: 1
```
alexandrunemes, line 37: i think this zIndex value should play nice with our modals 🤔

viniciusvolpe: I think it will be a problem with sticky columns, z-index there is set to 2 and I think we don't want the column to cover the header

lib/src/components/data-table/DataTable.types.ts
```diff
@@ -48,6 +48,7 @@ export type DataTableContextType<T = unknown> = Table<T> & {
   runAsyncData?: (options: Partial<TAsyncDataOptions>) => Promise<void>
   enableRowSelection?: boolean | ((row: Row<unknown>) => boolean)
   rowSelection: RowSelectionState
+  hasStickyHeader: boolean
```
viniciusvolpe, line 51: I think this prop is more suitable for the `DataTable.Table` component since it's specific to the table layout. You can check `numberOfStickyColumns` as an example

lib/src/components/table/TableHeader.tsx
alexandrunemes: s**t

chris-atomlearning:
Looks good. Will it work if it's not at the top of the page? Like on the stats page it already has a sticky tab bar and the header and stuff above it. Or maybe we need to add an offset value? I can't remember exactly how clever css sticky is but I remember having issues like that before.

lib/src/components/table/TableHeader.tsx
alexandrunemes: set it to 3, but also added an extra prop `cssHeader` to the table component, so you can more easily control the styling if need be, instead of having to break down the table into its root components.

alexandrunemes:
> Looks good. Will it work if it's not at the top of the page? Like on the stats page it already has a sticky tab bar and the header and stuff above it. Or maybe we need to add an offset value? I can't remember exactly how clever css sticky is but I remember having issues like that before.

I've put together a demo that showcases the table when added to a page that has a similar sticky container above it. I'm also adding a screenshot with the relevant CSS. In this scenario, you're going to have to use the `headerCss` prop to add an offset to the `top` property.

https://github.com/Atom-Learning/components/assets/9009155/724611c0-772d-4f4f-828d-fb398a43d4e2

Screenshot 2023-09-15 at 08 40 05

viniciusvolpe (approved):

chris-atomlearning (approved):

lib/src/components/table/TableHeader.tsx
```diff
@@ -37,7 +44,16 @@ type TableHeaderProps = React.ComponentProps<typeof StyledTableHeader>
 
 export const TableHeader: React.FC<TableHeaderProps> = ({
   theme = 'primaryDark',
+  hasStickyHeader = false,
```
LimeWub, line 47: why not just call the prop `sticky` since you're applying it to the Header only?
Header > hasStickyHeader is a bit of a weird name unless the header has a header in it?

alexandrunemes: Good point. I'll rename it along the way and get a quick patch out.

lib/src/components/table/TableStickyColumnsContainer.tsx
```diff
@@ -34,7 +34,6 @@ export const TableStickyColumnsContainer: React.FC<
       role="scrollbar"
       ref={scrollContainerRef}
       css={{
-        overflow: 'auto',
```
LimeWub, line 37: @alexandrunemes Hey I think this may have caused a bug in datatable. Can you lmk why you removed this `overflow: auto` from here?

LimeWub: Mzee reported this bug with it where it not longer overflow correctly on small screens so I thought with this being one of the latest changes to it it might be relevant: 
https://atomlearningltd.atlassian.net/browse/DS-489

alexandrunemes: I'll take a look

PR #574, opened 2023-09-13
Ds 465 👩🏻‍💻 build navigation sidedrawer

LimeWub:
JIRA: https://atomlearningltd.atlassian.net/browse/DS-465

* Splits [`Sidedrawer`](https://atomlearning.design/components/navigation/sidedrawer?tab=code) into 2 components: `Drawer` and `NavigationMenuVertical`
* Drawer now supports more positions although I suspect the top/bottom styles might be revisited when we actually use it 
* DELETES Sidedrawer (will need to replace in two cases). I will also need to wrap apps in a top level ColorScheme (noting here so I don't forget)
* Restyles `NavigationMenuVertical` as per [new designs ](https://www.figma.com/file/KjgQh7eTwEfJmH6e56k9mX/DS-Documentation?type=design&mode=design&t=8KXdfVviLgiSkIL1-0) and reworks its internals so that it uses `NavigationMenu` by radix.

## Implementation

* Unfortunately radix's `NavigationMenu` has some bugged out logic when it come to Accordions (Trigger/Content elements) so had to implore the use of `Collapsible` in combination with it to get these to work properly.
* Have moved some shared styling/utility functionality to the utilities folder
* Updated docs.

## Videos
**NavigationMenuVertical**

https://github.com/Atom-Learning/components/assets/6905473/2ac2cb54-4075-478d-9544-fd4169f96191

**Dialog**

https://github.com/Atom-Learning/components/assets/6905473/b7d2bc09-68ab-4ffe-a6dc-24a51ed2ccf3


MiguelGDLR (commented):
https://youtu.be/r13riaRKGo0?si=xBKzruJ0vex33M7S

lib/src/components/navigation-menu-vertical/NavigationMenuVerticalList.tsx
```diff
@@ -0,0 +1,24 @@
...
+  m: 0,
+  listStyle: 'none',
+  '--navigation-menu-vertical-item-pl': '$space$4',
+  '& &': {
+    '--navigation-menu-vertical-item-pl': '$space$5'
+  },
+  '& & &': {
+    '--navigation-menu-vertical-item-pl': '$space$7'
+  },
+  '& & & &': {
```
kerryambt, line 16: stupid question probably but what are these selectors doing?

lib/src/utilities/event/preventEvent.ts
```diff
@@ -2,7 +2,7 @@
 Used to block default radix hover to open menu behaviour
 Props: https://github.com/radix-ui/primitives/issues/1630
 */
-export const preventHover = (event: any) => {
+export const preventEvent = (event: unknown): void => {
```
kerryambt, line 5: can this maybe be `(event: MouseEvent | KeyboardEvent)`?

lib/src/utilities/style/backdrop-overlay.ts
```diff
@@ -0,0 +1,34 @@
...
+    '&[data-state="open"]': {
+      animation: string
+    },
+    '&[data-state="closed"]': {
+      animation: string
+    }
+  }
+}
+
+export const backdropOverlay = (): TbackdropOverlayReturn => {
```
kerryambt, line 19: why is this a function and not just an object? maybe typed `as const` if consistent typing is an issue where it's used?

lib/src/components/navigation-menu-vertical/NavigationMenuVertical.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+      {children}
+    </NavigationMenuVertical>
+  )
+}
+
+const mockOnOpenChange = jest.fn((_value) => null)
+
+describe('NavigationMenuVertical', () => {
+  it('renders', async () => {
+    const { findByRole } = render(<NavigationMenuVerticalImplementation />)
```
Marcdup1, line 26: Does it make a difference in this test to use `findByRole` from render instead of screen?

lib/src/components/navigation-menu-vertical/NavigationMenuVertical.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+    expect(container).toMatchSnapshot()
+
+    const accordionTrigger = await findByRole('button', {
+      name: 'Accordion Trigger'
+    })
+    expect(accordionTrigger).toHaveAttribute('data-state', 'open')
+    const accordionContent = getByTestId('accordion-content')
+    expect(accordionContent).toBeVisible()
+
+    fireEvent.click(accordionTrigger)
```
Marcdup1, line 61: Its perhaps better to use `userEvent`  since it simulates brower interactions more closely

https://testing-library.com/docs/user-event/intro/#differences-from-fireevent

lib/src/components/navigation-menu-vertical/NavigationMenuVertical.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { NavigationMenuVerticalItemContent } from './NavigationMenuVerticalItemContent'
+import { NavigationMenuVerticalLink } from './NavigationMenuVerticalLink'
+import { NavigationMenuVerticalList } from './NavigationMenuVerticalList'
+import { NavigationMenuVerticalText } from './NavigationMenuVerticalText'
+
+const StyledRoot = styled(Root, {
+  width: '100%'
+})
+
+type TNavigationVerticalProps = React.ComponentProps<typeof StyledRoot> & {
```
Marcdup1, line 20: You could use `Omit` here instead:
```ts
type TNavigationVerticalProps = Omit<React.ComponentProps<typeof StyledRoot>, 'delayDuration' | 'skipDelayDuration' | etc>
```

I think if you explicity have `delayDuration: never` its still going to bring it up as an option with autocomplete. It wont allow you to set a value since the type is `never` but it coming up as an option is a bit misleading

lib/src/components/navigation-menu-vertical/NavigationMenuVerticalAccordionContent.tsx
```diff
@@ -0,0 +1,42 @@
+import { Content } from '@radix-ui/react-collapsible'
+import React from 'react'
+
+import { styled } from '~/stitches'
+
+import { NavigationMenuVerticalAccordionContext } from './NavigationMenuVerticalAccordion'
+import { NavigationMenuVerticalList } from './NavigationMenuVerticalList'
+
+const StyledNavigationMenuVerticalAccordionContent = styled(Content, {})
```
Marcdup1, line 9: Does this need to be an empty styled component?

lib/src/components/navigation-menu-vertical/NavigationMenuVerticalItem.tsx
```diff
@@ -0,0 +1,22 @@
+import { Item } from '@radix-ui/react-navigation-menu'
+import React from 'react'
+
+import { styled } from '~/stitches'
+
+const StyledNavigationMenuVerticalItem = styled(Item, {})
```
Marcdup1, line 6: Same here, can we use the component directly?

lib/src/components/navigation-menu-vertical/NavigationMenuVerticalItem.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+type TNavigationMenuVerticalItemProps = React.ComponentProps<
```
Marcdup1, line 8: Can also use Omit here

lib/src/utilities/style/backdrop-overlay.ts
Marcdup1, line 19: Can this just be an object?

lib/src/components/drawer/DrawerContent.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { DrawerContext } from './Drawer.context'
+import { Flex } from '~/components/flex'
+
+import { ColorScheme } from '~/experiments/color-scheme'
+
+const animationTop = slideInOut({ outPosition: { x: 0, y: '-100%' } })
+const animationRight = slideInOut({ outPosition: { x: '100%', y: 0 } })
+const animationBottom = slideInOut({ outPosition: { x: 0, y: '100%' } })
+const animationLeft = slideInOut({ outPosition: { x: '-100%', y: 0 } })
+const setupAnimation = (animation) => {
```
Marcdup1, line 18: You could make it a bit more typesafe:
```suggestion
const setupAnimation = (animation: ReturnType<typeof slideInOut>) => {
```

lib/src/components/navigation-menu-vertical/NavigationMenuVerticalList.tsx
LimeWub: Not a stupid question at all lol.
They are selecting nested levels of `NMVList` and setting up a css variable for `pl` (paddingLeft) so that nested List Items can use that and indent correctly - further in for each level.
Goes up to a level 4 of nesting ^^

Does this answer the question?

lib/src/utilities/style/backdrop-overlay.ts
LimeWub: Yeeeeeeee it could be an object. I was just mimicing how we(I)'ve done it for focusVisibleStyleBlock which is a function which accepts parameters before it returns the styles. For this one there's no parameters currently - but it coooooould have some in the future maybe. 
I'm just following a _\pattern(?)\_ but you and @Marcdup1  have both commented on this and you are right. 
It could be just an object.
FWIW I don't think this would actually slow down the website as it resolves in the compile step - but yes it could b an object and it'd b as valid.

Shall I change it to an object you think?

lib/src/components/drawer/Drawer.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+          <Drawer.Footer>FOOTER</Drawer.Footer>
+        </Drawer.Content>
+      </Drawer>
+    </Tooltip.Provider>
+  )
+}
+
+describe('Drawer', () => {
+  let rendered: RenderResult
+  let trigger
```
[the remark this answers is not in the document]
LimeWub: I'll be honest - I just copied these over from `Dialog`'s tests lol

lib/src/components/drawer/DrawerContent.tsx
```diff
@@ -0,0 +1,109 @@
...
+})
+
+export const DrawerContent: React.FC<
+  React.ComponentProps<typeof StyledContent>
+> = ({ children, ...rest }) => {
+  const { position } = React.useContext(DrawerContext)
+
+  return (
+    <Portal>
+      <DrawerOverlay data-testid="drawer_overlay" />
```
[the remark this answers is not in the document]
LimeWub: We don't 👍

lib/src/components/navigation-menu-vertical/NavigationMenuVertical.test.tsx
LimeWub: I'll be honest, I just copied this from the `Sidedrawer` tests 🤣

lib/src/components/navigation-menu-vertical/NavigationMenuVertical.tsx
LimeWub: Great point and I think this will solve the TS errors I'm getting. Will defo try to do this with omits.

lib/src/components/navigation-menu-vertical/NavigationMenuVerticalAccordionContent.tsx
LimeWub: Yes because wrapping `styled` allows for use of `css` and `as` in implementation. Radix doesn't have these out of the box it's stitches (`styled`) which adds them.

lib/src/components/navigation-menu-vertical/NavigationMenuVerticalItem.tsx
LimeWub: Same as above.

lib/src/components/drawer/Drawer.context.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react'
+
+import { StyledContent } from './DrawerContent'
+
+type TDrawerProviderProps = {
```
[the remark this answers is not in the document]
Marcdup1: I remember a ton of back and forths in a few FE roundtables...buut we do have this typescript guide that @elliotlaws created that I try to rely on: https://www.notion.so/atomlearning/TS-StyleGuide-a26b6becddc04b24ac03387799ee57ac#46bb349d1cd24730931858772306995e

lib/src/components/navigation-menu-vertical/NavigationMenuVerticalAccordionContent.tsx
Marcdup1: ah yes of course! thanks for explaining

lib/src/components/drawer/Drawer.context.tsx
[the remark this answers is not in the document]
LimeWub: Ah I think this hasn't been agreed on 100%. I'm going with the `T`/`I` setup as this is what has come out of discussions in the FE guild before but I'd be happy to change if dropping the `T` actually gets agreed on :o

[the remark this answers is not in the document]
Marcdup1: Oof our pod has 100% dropped the prefixes, it'll be great to be consistent at least. I do think the reasons to drop it are compelling enough

lib/src/components/navigation-menu-vertical/NavigationMenuVerticalList.tsx
kerryambt: oooh I'd just never seen `&` used without anything else, certainly not repeated. I think I get it 😅

lib/src/utilities/style/backdrop-overlay.ts
kerryambt: I would say an object would be my preference just because it feels a bit superfluous as a function and we can remove the return type. that said I agree it likely would only be a code improvement and make no difference to performance.

lib/src/utilities/event/preventEvent.ts
LimeWub: Sadly this type doesn't work ):
I was having problems with typing this to begin with - that's why it is `unknown` ...

Screenshot 2023-09-26 at 14 53 52

lib/src/components/drawer/Drawer.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+    trigger = rendered.getByText('TRIGGER')
+  })
+
+  it('renders the trigger with the popover hidden by default', async () => {
+    expect(await screen.queryByText('BODY')).not.toBeInTheDocument()
+    expect(rendered.container).toMatchSnapshot()
+  })
+
+  it('opens the popover once trigger is clicked', async () => {
+    expect(await trigger).toBeInTheDocument()
```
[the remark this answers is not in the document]
LimeWub: I think I fixed these now? lmk

lib/src/utilities/style/backdrop-overlay.ts
LimeWub: changed to an object ^^

lib/src/utilities/event/preventEvent.ts
kerryambt: ah that's a shame. thanks for trying :)

kerryambt (approved):

lib/src/components/drawer/Drawer.context.tsx
```diff
@@ -0,0 +1,21 @@
...
+}
+
+type TDrawerContext = TDrawerProviderProps
+
+export const DrawerContext = React.createContext<TDrawerContext>({
+  position: 'left'
+})
+
+export const DrawerProvider = ({
+  position = 'left',
```
mzeenkala23, line 16: correct me if i am wrong but is left the only position we will have?

mzeenkala23 (approved):
Looks great, add one question though

lib/src/components/drawer/Drawer.context.tsx
LimeWub: I've added in `top`, `right` and `bottom` as well ^^
Currently `left` is the only one in use but with this becoming a separate Drawer component I've added the extra positions so they can be used if needed for menus other than the navigation, ie. filters, user menus etc. 
Basically any type of side panel ^^ This has been cleared/agreed on with @MiguelGDLR

lib/src/components/drawer/Drawer.test.tsx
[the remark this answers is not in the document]
Marcdup1: Oh I mean since trigger is getting assigned like this
`trigger = rendered.getByText('TRIGGER')` // which isnt a promise 
you dont need the `await trigger`

```suggestion
    expect(trigger).toBeInTheDocument()
```



(one could argue if you even need this assertion since you calling trigger.focus() which would fail anyways if it wasnt there & it is technically captured in the snapshot in the test above - no harm in leaving it though)

lib/src/components/drawer/Drawer.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+    expect(screen.queryByText('BODY')).not.toBeInTheDocument()
+
+    trigger.focus()
+    userEvent.click(trigger)
+
+    expect(screen.queryByText('BODY')).toBeInTheDocument()
+
+    const dialog = await screen.getByRole('dialog')
```
Marcdup1, line 51: same here the with the await 
```suggestion
    const dialog = screen.getByRole('dialog')
```

lib/src/utilities/event/preventEvent.ts
Marcdup1: A bit late here maybe this would work 💡 🤔 

```suggestion
export const preventEvent = (event: React.PointerEvent | Event): void => {
```

Marcdup1 (approved):

lib/src/utilities/event/preventEvent.ts
LimeWub: This worked! Cheers @Marcdup1

lib/src/components/drawer/Drawer.test.tsx
LimeWub: thanks for clarifying! done!

PR #576, opened 2023-09-18
Feat: Create Slim variant for the Promo Banner

avirati:
Note to reviewers : 

`BannerSlim` slightly differs in sub component structure. I did not want to use `absolute` positioning of elements to achieve desired design, instead I used a different composition structure.

Documentation : 
image

Preview:
https://github.com/Atom-Learning/components/assets/1936119/a195c456-f664-4a9e-9e2c-3b0e7d693709


avirati:
> What's the reason for not adding the variant into `Banner`?

@thomasdigby `Banner` acts as a container for both `BannerSlim` and `BannerRegular`.
If you look at `lib/src/components/banner/index.ts`, you'll see that we don't expose `Banner` to end users and hence it would be considered an internal component.

lib/src/components/banner/Banner.tsx
```diff
@@ -25,12 +26,19 @@ export const Banner: React.FC<IBannerProps> & {
   colorScheme,
   emphasis = 'lowContrast',
   value = 'dismiss-banner-regular',
-  onDismiss
+  onDismiss,
+  fullWidth = true,
```
LimeWub, line 30: Why is this needed? Can't someone just `width: 100%` if they need to?

lib/src/components/banner/banner-slim/BannerSlimActions.tsx
```diff
@@ -0,0 +1,30 @@
...
+  css,
+  ...props
+}) => {
+  const { size } = useBannerContext()
+  const width = React.useMemo(
+    () => overrideStitchesVariantValue(size, (s) => toWidth[s]),
+    [size]
+  )
+  return (
+    <Flex gap="4" css={{ p: '$4', width, ...css }} {...props}>
```
LimeWub, line 24: doesn't this break if `css` is undefined?

For the DS we(I) usually use `styled()` to generate the styled components rather than directly css so you don;t have to worry about `...css` and splitting css from `...rest`

lib/src/components/banner/banner-slim/BannerSlimActions.tsx
```diff
@@ -0,0 +1,30 @@
...
+  md: 'auto'
+}
+
+export const BannerSlimActions: React.FC<React.ComponentProps<typeof Flex>> = ({
+  children,
+  css,
+  ...props
+}) => {
+  const { size } = useBannerContext()
+  const width = React.useMemo(
```
LimeWub, line 19: I wouldn't use this here. I'd use a `styled` variant directly.

`overrideStitchesVariantValue` is for passing down variant props to child components (ie. Text component size or Icon component size) and not meant for simple css.

lib/src/components/banner/banner-slim/BannerSlimContent.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import { Flex } from '../../flex'
+import { useBannerContext } from '../BannerContext'
+
+const Container = styled(Flex, {
+  p: '$4',
+  gap: '$4',
+  variants: {
+    size: {
+      sm: { maxWidth: '736px', pb: 0 },
```
LimeWub, line 13: `maxWidth: '736px'` is this number just the breakpoint this is supposed to scale in?
If so then this shouldn't be needed in the variant.

lib/src/components/banner/banner-slim/BannerSlimImage.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { Image } from '../../image'
+import { useBannerContext } from '../BannerContext'
+
+const Container = styled(Box, {
+  overflow: 'hidden',
+  borderRadius: '$round',
+  variants: {
+    size: {
+      sm: {
+        minHeight: '$4',
```
LimeWub, line 15: ummm.... why not just `size: '$4'`?
you're setting both min and max values to `$4` might as well set `size/width/height` directly (?)

lib/src/components/banner/banner-slim/BannerSlimImage.tsx
```diff
@@ -0,0 +1,49 @@
...
+        minHeight: '$5',
+        maxHeight: '$5',
+        minWidth: '$5',
+        maxWidth: '$5'
+      }
+    }
+  }
+})
+
+const StyledImage = styled(Image, {
```
LimeWub, line 30: Minor: 
this css could be just
```
{
  size: '100%',
  objectFit: 'cover'
}
```
I think

lib/src/components/banner/banner-slim/BannerSlimText.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { styled } from '~/stitches'
+
+import { Text } from '../../text'
+import { useBannerContext } from '../BannerContext'
+
+const StyledText = styled(Text, {
+  color: '$grey900',
+  fontWeight: '600',
+  variants: {
+    containerSize: {
```
LimeWub, line 12: What is container size?

lib/src/components/banner/banner-slim/BannerSlimText.tsx
```diff
@@ -0,0 +1,55 @@
...
+      sm: {},
+      md: {}
+    },
+    emphasis: {
```
LimeWub, line 16: You don't need to set colour here, I think as it just uses what's been set above it (in the theme level) - unless I'm missing something.

lib/src/components/banner/banner-slim/BannerSlimText.tsx
```diff
@@ -0,0 +1,55 @@
...
+        color: '$grey100'
+      },
+      midContrast: {},
+      lowContrast: {}
+    },
+    hasDismiss: {
+      true: {}
+    }
+  },
+  compoundVariants: [
```
LimeWub, line 27: okay I just got it lol. I was very confused why you had all of this setup here but it's just for this margin-right :o

I wonder if this fairly long setting object could be bypassed somehow. Maybe with css variables or by using flex order or something?
I might pull this down and try something out with this at some point - but yes it's a bit awkward how much setup is needed for one margin-right :o

lib/src/components/banner/Banner.tsx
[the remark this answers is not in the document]
LimeWub: Actually what was wrong with the `width: 100%` this originally had?

lib/src/components/banner/banner-slim/BannerSlimText.tsx
```diff
@@ -0,0 +1,55 @@
+import * as React from 'react'
+
+import { styled } from '~/stitches'
+
+import { Text } from '../../text'
+import { useBannerContext } from '../BannerContext'
+
+const StyledText = styled(Text, {
+  color: '$grey900',
```
LimeWub, line 9: u shouldn't need this color here either cause it is set up in `BannerContainer`

thomasdigby:
> > What's the reason for not adding the variant into `Banner`?
> 
> @thomasdigby `Banner` acts as a container for both `BannerSlim` and `BannerRegular`. If you look at `lib/src/components/banner/index.ts`, you'll see that we don't expose `Banner` to end users and hence it would be considered an internal component.

Wait, we have a component called `BannerRegular`?? Why isn't that just `Banner`? Surely the dev in `atom-core` isn't importing `BannerRegular`?

If that is the case then I'd really recommend renaming it and figuring out how to combine these two components into something that allows `Banner type="slim"` (or `compact`, `reduced`, etc.).

lib/src/components/banner/banner-slim/BannerSlimText.tsx
thomasdigby: A parent level flex gap?

avirati:
> > > What's the reason for not adding the variant into `Banner`?
> > 
> > 
> > @thomasdigby `Banner` acts as a container for both `BannerSlim` and `BannerRegular`. If you look at `lib/src/components/banner/index.ts`, you'll see that we don't expose `Banner` to end users and hence it would be considered an internal component.
> 
> Wait, we have a component called `BannerRegular`?? Why isn't that just `Banner`? Surely the dev in `atom-core` isn't importing `BannerRegular`?
> 
> If that is the case then I'd really recommend renaming it and figuring out how to combine these two components into something that allows `Banner type="slim"` (or `compact`, `reduced`, etc.).

Yes, we import `BannerRegular` in `atom-core`. Agreed that `Banner type="slim"` would be cleaner. Would you want this to happen in this PR ? Also, this would be considered a breaking change.

thomasdigby:
> Yes, we import `BannerRegular` in `atom-core`. Agreed that `Banner type="slim"` would be cleaner. Would you want this to happen in this PR ? Also, this would be considered a breaking change.

I'll let you decide where to make that change with the DS team, but as we have two examples in the codebase it should be a quick migration.

LimeWub:
We would need an additional component to be making the choice based on `type` to use `BannerRegular` or `BannerSlim` components.

Is it worth the effort? 🤷 
Personally I don't mind the separation of this as 2 components (after all it was my suggestion to split them out as they were different enough) but if it's that shocking we can do this work (urgency is ???).

LimeWub:
Unless you somehow merge the two components props+css to achieve this. If it's easy enough. But I thought the 2 different variations as 2 different components was cleaner when I was planning this.

lib/src/components/banner/banner-slim/BannerSlimText.tsx
LimeWub: Screenshot 2023-09-18 at 12 05 42


it's done to leave margin-right for absolute positioning of the X to not overlap the text.

That's why I'm mentioning flex order as a potential solution.

lib/src/components/banner/Banner.tsx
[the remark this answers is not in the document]
avirati: You're right. This was not needed. Width can be controlled from outside. Reverting this change 👍

lib/src/components/banner/banner-slim/BannerSlimActions.tsx
avirati: It doesn't break if `css` is `undefined`. The `styled()` approach is definitely cleaner. Will make the change 👍

lib/src/components/banner/banner-slim/BannerSlimActions.tsx
avirati: Done

lib/src/components/banner/banner-slim/BannerSlimContent.tsx
avirati: `sm` variant is not supposed to scale to the entire container's width. It is capped at `737px` as per design. Where should I move it then ?

Screenshot 2023-09-20 at 05 32 02

lib/src/components/banner/banner-slim/BannerSlimImage.tsx
avirati: The image does not retain its expected dimensions when the screen size is smaller

Screenshot 2023-09-20 at 05 34 00

lib/src/components/banner/banner-slim/BannerSlimText.tsx
avirati: I copied the approach from `lib/src/components/banner/banner-regular/BannerRegularText.tsx`

@thomasdigby parent level flex gap won't work as the `BannerRegularText` and `BannerDismiss` are not siblings. Moreover, `BannerDismiss` is `absolute` positioned in `sm` variant

@LimeWub would `order` work for both `sm` and `md` variants ?

lib/src/components/banner/banner-slim/BannerSlimText.tsx
avirati: Right, it's handled already 👍

lib/src/components/banner/banner-slim/BannerSlim.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { BannerSlimImage } from './BannerSlimImage'
+import { BannerSlimText } from './BannerSlimText'
+
+const toDirection = {
+  sm: 'column',
+  md: 'row'
+}
+
+export const BannerSlim: React.FC<React.ComponentProps<typeof Banner>> & {
+  Content: typeof BannerSlimContent
```
elliotlaws, line 19: The component types should be inferred so declaring each component and using `typeof Component` shouldn't be necessary, is there a reason to do this?

avirati: We have been following this approach while building `composable` components.

elliotlaws (approved):

andoulla (approved):

lib/src/components/banner/banner-slim/BannerSlimImage.tsx
LimeWub: fixed 👍

lib/src/components/banner/banner-slim/BannerSlimText.tsx
LimeWub: this approach is fine 👍  flex-order wouldn't work for multiple paragraphs.

lib/src/components/banner/banner-slim/BannerSlimContent.tsx
LimeWub: that is addressed via the responsive size default variant (and support for responsive variants) so no need for the maxWidth here ^^

LimeWub:
Just waiting for @MiguelGDLR to check to pull and check/confirm this component is okay and then lets get it merged ^^

avirati:
> Just waiting for @MiguelGDLR to pull and check/confirm this component is okay and then lets get it merged ^^

The component is Acceptance Tested by @AtanasKovachev. Merging

PR #577, opened 2023-09-20
feat: update Toast to allow for greater UI customisation 

elliotlaws:
Ticket: https://atomlearningltd.atlassian.net/browse/STU-1722

Also includes the this small ticket https://atomlearningltd.atlassian.net/browse/DS-120

- Allow toast content to expand as per the contents. Currently, it is fixed at 400px.
- Add onDismiss callback
- Allow toast(s) to be offset from the top
- Look into dismiss button’s position. At present, it is absolute positioned, which makes it difficult for us to align them vertically.

## Implementation: 
- ToastProvider now accepts css prop to allow us to set the top offset from outside
- `Toast`, `Toast.Icon` and `Toast.Close` are now exported for easy reuse with custom toast UIs.
- Removed absolute positioning from Dismiss button
- Added Spacer component

### Sandbox code
```tsx
import { Info } from '@atom-learning/icons'
import {
  Text,
  Flex,
  globalCss,
  Icon,
  Button,
  toast,
  ToastProvider,
  Grid,
  Spacer
} from '../src'
import { Toast } from '../src/components/toast/Toast'

const TOAST_DURATION = 9999999

export const MockTestToast = () => {
  return (
    <Toast css={{ width: 600 }}>
      <Toast.Icon is={Info} />
      <Text>You have a mock test in progress.</Text>
      <Spacer />
      <Text css={{ p: '$2', cursor: 'pointer' }}>Resume mock test</Text>
      <Toast.Close />
    </Toast>
  )
}

const App = () => (
  <Flex
    css={{
      minHeight: '100vh',
      justifyContent: 'center',
      alignItems: 'center',
      flexDirection: 'column'
    }}
  >
    <TooltipProvider>
      <ToastProvider />
      <Grid gap={5}>
        <Button
          onClick={() =>
            toast.success('Something went right!', { duration: TOAST_DURATION })
          }
        >
          <Flex>
            <Text>Click for toast!</Text>
          </Flex>
        </Button>
        <Button
          onClick={() => toast(<MockTestToast />, { duration: TOAST_DURATION })}
        >
          <Flex>
            <Text>Click for a custom toast</Text>
          </Flex>
        </Button>
      </Grid>
    </TooltipProvider>
  </Flex>
)

```

https://github.com/Atom-Learning/components/assets/44065175/20ccec9a-bd32-4cf1-bfd8-f0b47e8e908b


lib/src/components/toast/ToastDismissButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+  onDismiss
+}: ToastDismissButtonProps) => {
+  return (
+    <ActionIcon
+      css={{
+        color: 'white',
+        mr: '-$2',
+        '&:hover,&:focus': { color: 'white', opacity: 0.9 }
+      }}
+      label="Close alert"
```
andoulla, line 25: Can we please make this label a prop to be able to overwrite it with a translated string?

andoulla (approved):

lib/src/components/toast/ToastContent.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export const TOAST_WIDTH = 400
+
+export const ToastContent = ({
+  id,
+  message,
+  type
+}: Pick<Toast, 'id' | 'message' | 'type'>) => {
+  return (
+    <Flex css={{ alignItems: 'center', '@md': { width: TOAST_WIDTH } }}>
```
avirati, line 19: Is the toast content able to expand more than `TOAST_WIDTH` if needed ?

lib/src/components/toast/ToastDismissButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+    <ActionIcon
+      css={{
+        color: 'white',
+        mr: '-$2',
+        '&:hover,&:focus': { color: 'white', opacity: 0.9 }
+      }}
+      label={label}
+      onClick={() => {
+        toast.dismiss(id)
+        onDismiss && onDismiss()
```
avirati, line 32:
```suggestion
        onDismiss?.()
```

lib/src/components/toast/ToastContent.tsx
elliotlaws: The ToastContent component is used only for when a custom UI is not required. For custom UI we'd pass in a function (rendering a component), that way the width is not restricted.

avirati: Right, makes sense

avirati (approved):

lib/src/components/toast/ToastDismissButton.tsx
elliotlaws: Updated

lib/src/components/toast/ToastProvider.tsx
```diff
@@ -3,31 +3,34 @@ import { useToaster } from 'react-hot-toast'
...
+
+const DEFAULT_OFFSET = '$2'
 
 const Container = styled('div', {
-  left: '$2',
   position: 'fixed',
-  top: '$2',
-  right: '$2',
   zIndex: MAX_Z_INDEX,
+  top: DEFAULT_OFFSET,
```
LimeWub, line 17: Minor: did you know when you want to set all the position properties at the same time [you can you `inset`](https://developer.mozilla.org/en-US/docs/Web/CSS/inset)

lib/src/components/toast/ToastProvider.tsx
```diff
@@ -3,31 +3,34 @@ import { useToaster } from 'react-hot-toast'
...
+  left: DEFAULT_OFFSET,
+  right: DEFAULT_OFFSET,
+  bottom: DEFAULT_OFFSET,
+  pointerEvents: 'none',
```
LimeWub, line 21: Why `pointerEvents: 'none'`?

lib/src/components/toast/index.ts
```diff
@@ -1 +1,3 @@
+export type { ToastProps } from './Toast'
 export { ToastProvider, toast } from './ToastProvider'
+export { ToastDismissButton } from './ToastDismissButton'
```
LimeWub, line 3: It is a bit weird that unlike our other components `Toast` doesn't actually export... well.. `Toast` and thus its `DismissButton` isn't `Toast.DismissButton` but `ToastDismissButton` :S 

I wonder - is there nothing useful in `ToastContainer` or just `Toast`,  to form a `Toast` out of so that it matches more with our usual pattern? .-.

lib/src/components/toast/ToastContent.tsx
LimeWub, line 19: Do you not need this to be available in your ToastContent overrides at all? Is this not useful to be exported (even if partially) - ie for the Icon or the Loader? 👀

lib/src/components/toast/Toast.tsx
```diff
@@ -50,15 +46,10 @@ const StyledToast = styled('div', {
   color: 'white',
   display: 'flex',
   minHeight: '$5',
-  pl: '$4',
   position: 'relative',
-  pr: '$6',
-  py: '$4',
+  px: '$3',
+  py: '$1',
```
LimeWub, line 51: How comes the paddings changes? Are these just safety gaps? (if yes I was expecting to see a minHeight or something) - is this coming from design?

lib/src/components/toast/ToastProvider.tsx
elliotlaws: Without it, you're unable to click on the button to trigger the toast. This is also done in the `Toaster` component from 'react-hot-toast' - see https://github.com/timolins/react-hot-toast/blob/main/src/components/toaster.tsx#L105

lib/src/components/toast/Toast.tsx
elliotlaws: It was changed to accomodate the change in component composition but, the horizontal padding does look different onscreen though, so I'll look into that.

lib/src/components/toast/index.ts
elliotlaws: Yeah this crossed my mind too, but I think if we ultimately end up going with Radix Toast, then I thought we may as well wait and not overthink it too much at this point. Thoughts?

lib/src/components/toast/ToastContent.tsx
elliotlaws: Same as above, maybe this is something to consider when we move to using Radix Toast?

lib/src/components/toast/Toast.tsx
elliotlaws: Padding has been updated to match more closely with what we had before

lib/src/components/toast/index.ts
elliotlaws: Updated so that we now have `Toast`, `Toast.Icon` and `Toast.Close`

lib/src/components/toast/ToastProvider.tsx
```diff
@@ -1,45 +1,138 @@
...
     </>
   )
 }
 
+export const useToastContext = () => {
+  const context = React.useContext(ToastContext)
+
+  if (!context) {
+    throw new Error(
+      'useSectionMessageContext must be used within a SectionMessageProvider'
```
LimeWub, line 131: copy/pasta :)

documentation/content/components.layout.spacer.md
```diff
@@ -0,0 +1,32 @@
+---
```
LimeWub, line 1: noice

lib/src/components/toast/ToastProvider.tsx
```diff
@@ -1,45 +1,138 @@
...
   '@sm': {
-    top: '$3',
-    right: 'auto',
-    left: `calc(50% - ${TOAST_WIDTH / 2}px)`
+    top: '$3'
+  }
+})
+
+const slideIn = keyframes({
+  '0%': { transform: `translate3d(0,-100%,0)`, opacity: 0 },
```
LimeWub, line 25: Curious (and more so because I'm tweaking `slideIn`/`Out` in another PR):
Are you animating on the zAxis? why `translate3d`?

lib/src/components/toast/ToastProvider.tsx
elliotlaws: oh my

lib/src/components/toast/ToastProvider.tsx
```diff
@@ -1,45 +1,138 @@
...
+                aria-live={toast.ariaLive}
+                css={{
+                  transform: `translateY(${offset}px)`,
+                  pointerEvents: 'auto',
+                  alignItems: 'center',
+                  borderRadius: '$0',
+                  boxSizing: 'border-box',
+                  minHeight: '$5',
+                  position: 'relative',
+                  transition: 'background-color 50ms ease-out',
```
LimeWub, line 100: Minor: was it animating `background-color` before? (why?)

lib/src/components/toast/ToastProvider.tsx
elliotlaws: updated

lib/src/components/toast/ToastProvider.tsx
```diff
@@ -1,45 +1,136 @@
...
+                  }
+                }}
+              >
+                <ToastContext.Provider value={toast}>
+                  {typeof children === 'function' ? (
+                    children(toast)
+                  ) : React.isValidElement(children) ? (
+                    children
+                  ) : (
+                    <ToastBadge />
```
LimeWub, line 113: Minor: 
Rather than splitting `ToastBadge` in its own component - could you just paste what it renders in here? Then we don't need to find a name that makes sense for it ie. `ToastBadge`, `ToastComposed`, `ToastPrebuilt`, `ToastFallback` :p 
Just cause someone who is viewing the folder might think the Toast has a badge (like a child component badge type of thing - the same way it has an Icon) but if it's in here as a nameless fallback it'll be less confusing imo. (Generally it's a great idea to split things out tho so ye, mixed on this one!)

lib/src/components/toast/ToastBadge.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export const TOAST_WIDTH = 400
+
+export const ToastBadge = () => {
+  const { type, message } = useToastContext()
+
+  return (
+    <Toast css={{ width: TOAST_WIDTH }}>
+      {type === 'error' && <Toast.Icon is={Error} />}
+      <Text>{message}</Text>
```
LimeWub, line 18: Do you need message in the context? or could you just do `<Text>{children}</Text>` for this case?

lib/src/components/toast/ToastProvider.tsx
elliotlaws: To be perfectly honest, I'm not sure. I just copied this directly from the previous iteration of Toast.

lib/src/components/toast/ToastProvider.tsx
elliotlaws: It was animating on the Toast itself (which it still is) so this isn't required here. Will remove

lib/src/components/toast/ToastProvider.tsx
elliotlaws: Yeah I was in two minds about this too, will move back to ToastProvider

lib/src/components/toast/ToastProvider.tsx
elliotlaws: I've updated this to `translateY`

LimeWub (approved):

PR #584, opened 2023-10-10
B2B-165: fix horizontal scrolling when sticky columns are used

alexandrunemes:
This PR fixes the broken horizontal scroll when using sticky columns and also makes the feature compatible with sticky headers. This will, however, add a vertical scrollbar to the container which wraps the table when using sticky columns, as shown in the video below:

https://github.com/Atom-Learning/components/assets/9009155/e762276f-6457-4ef0-83b6-3e9862a174f6

### DIFF
--- a/lib/src/components/table/TableStickyColumnsContainer.tsx
+++ b/lib/src/components/table/TableStickyColumnsContainer.tsx
@@ -34,6 +34,7 @@ export const TableStickyColumnsContainer: React.FC<
       role="scrollbar"
       ref={scrollContainerRef}
       css={{
+        overflow: 'auto',
         maxWidth: '100%',
         ...columnsCss,
         [`& td:nth-child(${numberOfStickyColumns}), th:nth-child(${numberOfStickyColumns})`]:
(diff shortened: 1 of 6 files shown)


lib/src/components/table/TableStickyColumnsContainer.tsx
```diff
@@ -34,6 +34,7 @@ export const TableStickyColumnsContainer: React.FC<
       role="scrollbar"
       ref={scrollContainerRef}
       css={{
+        overflow: 'auto',
```
alexandrunemes, line 37: This change will indeed break the sticky header behaviour when `hasStickyHeader` is used in combination with `numberOfStickyColumns` but i think this is a limitation we're going to have to live with right now. I couldn't figure out a clean and non-janky way of having the header both sticky and maintain horizontal scroll position in sync with the rest of the table when the table gets wrapped by the sticky columns container.

chris-atomlearning (approved):

viniciusvolpe (approved):

LimeWub (approved):
It's good for now thanks for fixing 👍 
I think at some point in the future it might be worth looking at the Table/DataTable API again to see if we can make it a bit simpler to use but until then this seems good enough ^^

PR #585, opened 2023-10-18
Feat: showAlert dialog improvements

Marcdup1:
### Description 

This PR makes the `showAlert` dialog a bit more customisable.

### Changes
- Adds a `theme` option: `'info' | 'warning' | 'danger' | 'success'`, adding a theme will display an appropriate icon along side the title
- Adds a `size` prop, defaults to `sm`
- Makes `onAction` optional; avoids having to do `onAction: () => null,` just to fix ts errors.
- Adds `showCloseButton` options, defaults to `true`
- Adds `confirmElement` & `cancelElement` to allow flexibility over the buttons being rendered.
- Set dialog max height to 90vh and scroll content
- Adds `noInline` to code blocks so we can render functional component with hooks

### Screenshots

image

image

image

https://github.com/Atom-Learning/components/assets/11061661/3211a27a-54d9-4d86-88b6-31ac2d1f53c9


lib/src/components/alert-dialog/alert-context/AlertDialog.tsx
```diff
@@ -16,46 +18,65 @@ type AlertDialogContentProps = React.ComponentProps<typeof AlertDialog> & {
...
 export const Alert: React.FC<AlertDialogContentProps> = ({
   title,
+  size,
+  theme,
   description,
   onAction,
   cancelActionText,
   confirmActionText,
   onClose,
+  showCloseButton = true,
```
thomasdigby, line 28: We shouldn't include this prop, the `AlertDialog` is there specifically to get user confirmation on an action. There's a reason that we're currently `preventDefault()`ing on ESC key.

Can we remove this and rely on the current `Cancel` button action to close?

Marcdup1: aaah 😅 

>A modal dialog that interrupts the user with important content and expects a response.

// @MiguelGDLR

lib/src/components/alert-dialog/alert-context/types.ts
```diff
@@ -1,10 +1,24 @@
...
   description?: string
-  onAction: (result: boolean) => void
+  onAction?: (result: boolean) => void
   confirmActionText: string
   cancelActionText?: string
+  showCloseButton?: boolean
+  confirmButtonTheme?: ButtonTheme
+  confirmButtonAppearance?: ButtonAppearance
+  cancelButtonTheme?: ButtonTheme
+  cancelButtonAppearance?: ButtonAppearance
```
thomasdigby, line 23: I'm not _loving_ this set of options for `showAlert()` now. Feels similar to something we want to avoid in component land, with a prop for every variation of the component.

Could we have a few "theme"s or variants of the two buttons? e.g. one where cancel is outline and they're both primary, one where cancel and confirm and both solid, cancel red, confirm green, etc. Can we align that with the `theme` we're using for the icon?

I'd be wary of opening up too much flexibility with a key like this.

lib/src/components/alert-dialog/alert-context/AlertDialog.tsx
MiguelGDLR: Yeah very controversial topic. It depends on the use case. I think it might be safer to exclude it from Alert dialog for now then.

thomasdigby: Thanks both, there's little difference between this and the standard `Dialog` if we add the ability to close it.

lib/src/components/alert-dialog/alert-context/types.ts
MiguelGDLR: I don't understand these props and if they are needed. But we can't provide a set of themes for each button as it all depends on the use case. We might want all ghost-neutral. It's what I tried to explain here, and I thought it wouldn't require any logic, just the flexibility to choose how many actions and what theme/appearance on each, and the order. 

I remember seeing examples in the admin panel with alert dialogs that had a 'Save' button in red 🥵 So I wanted to avoid that

Screenshot 2023-10-18 at 17 46 53

thomasdigby: I wonder if we can provide an entry point for components to be passed in, e.g.

```tsx
showAlert({
  title: 'Delete student',
  description: 'You are going to delete this student. Are you sure?'
  confirmElement: <Button theme="danger">Yes, delete them please</Button>,
  cancelElement: <Button appearance="outline">Nah, don't do that</Button>
})
```

We would have to do a bit of event composition, ensuring that any additional `onClick`s get combined with the default. That would make `onAction` redundant, as well as `confirmActionText`/`cancelActionText`.

Not 100% on it, but I feel it could allow flexibility without requiring alll the additional props.

Marcdup1: @thomasdigby Could you do something like this with those props:
```tsx
<AlertDialog.Cancel asChild>{cancelElement}</AlertDialog.Cancel>
<AlertDialog.Action asChild>{confirmElement}</AlertDialog.Action>
```
This does open a lot more flexibility though.

D7Torres: We do need the flexibility. Each case is completely different and the colours and button themes depend completely on the action.

We discussed this ages ago and we decided to go this way, because using `showAlert` is so much better a dev experience than composing the component in this specific case, as it manages the open/close state for you so you don't have to pollute the code with that logic

MiguelGDLR (commented):
All good from my side. Great work.
The only issue I found is that somehow when I change the props (size+text) the buttons don't behave responsively. 
However they do when I resize the screen 🤔
Screenshot 2023-10-19 at 10 15 33

lib/src/components/alert-dialog/alert-context/types.ts
Marcdup1: @D7Torres @thomasdigby are there any strong opinions on the two approaches: Passing in Elements vs  Button props?

It feels like passing in elements requires a bit more dev work and deteriorates the dev experience David touched on, but looks slightly cleaner. We could have a default confirmElement (eg `<Button theme="primary" appearance="solid" size="sm">Confirm</Button>`) but I feel like _most_ of the time we will want to update the copy - we would rarely/never want to update `size` though?

D7Torres: @Tom @MiguelGDLR 
> Not 100% on it, but I feel it could allow flexibility without requiring alll the additional props.

Do we want the flexibility? sometimes the lack of flexibility is a good thing. Miguel, can you confirm will always have a primary button and an outline button? if that's the case, isn't it better to restrict it?

Regarding the "too many props hell". This hell happens when the props are conditionals, or excluding each other, and the components start to do too many things that should actually be different components.
However, in this case I feel that the props are legit: each one setting one thing that actually needs to be set.

thomasdigby: @Marcdup1 Can I ask how you think it deteriorates the dev experience? I don't think I've suggested anything more complex than a list of very specific single-purpose keys.

If it wasn't clear in my previous message (and it probably wasn't tbh), I wasn't suggesting removing the previous props, `confirmActionText`, `onAction`, etc., just adding an additional way to define your confirm/cancel actions. So you can _either_ add a `confirmElement` to tweak specific button props, content (maybe you want an icon in this _one_ specific case), and action.

```tsx
showAlert({
  title: 'Delete student',
  description: 'You are going to delete this student. Are you sure?',
  confirmElement: <Button theme="danger" onClick={doThing}>Yes, delete them please</Button>,
  cancelElement: <Button appearance="outline">Nah, don't do that</Button>
})
```

or

```tsx
showAlert({
  title: 'Delete student',
  description: 'You are going to delete this student. Are you sure?',
  confirmActionText: 'Yes, delete them please',
  cancelActionText: 'Nah, don't do that',
  onAction: (confirmed) => {
    if (confirmed) doThing()
  }
})
```

If we don't try to compose our components, then I think we'll continue adding specific props for the buttons, content, until we have a long list for single use-cases across the codebase. And obviously that's something we want to avoid.

D7Torres: It's true, today we might be told that buttons will never have icons, and then tomorrow they might be needed.
So maybe composing is a better idea, but then I'm actually in favour of having only one way of doing it (your piece of code #1, and kill `confrimActionText`, etc. 

Having multiple ways of doing the same thing can seem like a good dev UX, but is going to make the component messier, imo

Marcdup1: @thomasdigby I was mainly thinking about having to declare a whole button element again when you just want a copy change - I wasn't thinking about supporting both ways.  I agree with the composition for sure 💯  So I'll defs kill the long list of params

I'm torn, how bad could supporting both really look 🤔 
```tsx
{cancelActionText && (
  <Button
    as={AlertDialog.Cancel}
    onClick={() => onAction?.(false)}
    appearance="outline"
  >
    {cancelActionText}
  </Button>
)}
{confirmActionText && 
  <Button as={AlertDialog.Action} onClick={() => onAction?.(true)}>
    {confirmActionText}
  </Button>
}
{cancelElement && (<AlertDialog.Cancel asChild>{cancelElement}</AlertDialog.Cancel>)}
{confirmElement && (<AlertDialog.Action asChild>{confirmElement}</AlertDialog.Action>)}
```
We can also make it more type-safe so you can't pass in both `confirmActionText` & `confirmElement`?  should enforce the conditional checks as well? `{confirmActionText && !confirmElement && (...)` 😕

documentation/components/markdown/editor-component/CodeBlock.tsx
```diff
@@ -12,6 +12,7 @@ type CodeBlockProps = {
   language?: string
   live?: boolean
   preview?: boolean
+  noInline?: boolean
```
LimeWub, line 15: What does`noInline` mean?

lib/src/components/alert-dialog/alert-context/types.ts
thomasdigby: This might be a touch simpler?

```tsx
  <AlertDialog.Cancel asChild>
    {cancelElement || (
      <Button onClick={() => onAction?.(false)} appearance="outline">
        {cancelActionText}
      </Button>
    )}
  </AlertDialog.Cancel>
  <AlertDialog.Action asChild>
    {confirmElement || (
      <Button onClick={() => onAction?.(true)}>{confirmActionText}</Button>
    )}
  </AlertDialog.Action>
```

thomasdigby: As I said before, I'm not 100% on this approach over all others, I think it should primarily be used when you need total control over the element rather than a way of just passing in content. So if we retain the current props, `confirmActionText`, `onAction`, etc., but allow future flexbility, we get the best of both worlds, while adding a bit of an API headache for some engineers that don't know what to choose.

I think it really depends on the long-term usage of this component. How much control are we giving the engineer, and what do we want to keep consistent.

lib/src/components/alert-dialog/alert-context/AlertDialogIcon.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from 'react'
+import { Info, Danger, OkCircle } from '@atom-learning/icons'
+import { styled } from '~/stitches'
+
+import { Icon } from '../../icon'
+
+export type AlertDialogTheme = 'info' | 'warning' | 'danger' | 'success'
+
+const StyledIcon = styled(Icon, {
+  mr: '$2',
```
thomasdigby, line 10: Can we use `gap` instead of this margin?

thomasdigby (approved):
Just a few minor changes related to layout/spacing. Otherwise I think this looks good 👍

lib/src/components/alert-dialog/alert-context/AlertDialog.tsx
```diff
@@ -16,46 +17,60 @@ type AlertDialogContentProps = React.ComponentProps<typeof AlertDialog> & {
...
+      size={size}
       onEscapeKeyDown={(e) => e.preventDefault()}
       onCloseAutoFocus={onClose}
+      css={{ display: 'flex', flexDirection: 'column', maxHeight: '90vh' }}
       {...remainingProps}
     >
-      <Heading as={AlertDialog.Title} size="sm" css={{ mb: '$5' }}>
-        {title}
-      </Heading>
+      <Flex css={{ alignItems: 'center', mb: '$5' }}>
```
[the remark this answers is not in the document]
Marcdup1: Nice! it did thank you 👍

MiguelGDLR (approved):

documentation/components/markdown/editor-component/CodeBlock.tsx
```diff
@@ -55,6 +56,7 @@ export const CodeBlock: React.FC<CodeBlockProps> = React.memo(
         code={code}
         scope={scope}
         theme={theme}
+        noInline={noInline}
```
D7Torres, line 59: I don't know how this works, but well done!

lib/src/components/alert-dialog/alert-context/AlertContext.test.tsx
```diff
@@ -35,48 +35,49 @@ const ComponentTest = ({ showSecondAlert, onAction = () => null }) => {
...
+
 const AlertContextTest = (props) => (
   <AlertProvider>
     <ComponentTest {...props} />
   </AlertProvider>
 )
 
 describe('Alert context', () => {
-  it('renders the trigger with the dialog hidden by default', async () => {
+  it('renders the trigger with the dialog hidden by default', () => {
```
D7Torres, line 48: Well done removing all unnecessary async/awaits

documentation/content/components.feedback.alert-dialog.md
```diff
@@ -14,7 +14,7 @@ tabs:
       `AlertDialog` exports a number of components that can be composed together to create a modal pop up that expects a response from the user.
 
 
-      `AlertDialog` also exports a custom hook `useAlert`, that can be used to dynamically render an alert based on some content and a callback. `showAlert` accepts a few props allowing the `AlertDialog` to be customised in a number of ways; you can see an example of this below.
+      `AlertDialog` also exports a custom hook `useAlert`, that can be used to dynamically render an alert based on some content and a callback. `showAlert` accepts a `theme` and `size` prop which can be used to customise the `AlertDialog`
```
D7Torres, line 17: I wonder if we should write something about this being the preferred way of using it.

lib/src/components/alert-dialog/alert-context/types.ts
```diff
@@ -1,23 +1,31 @@
...
+      confirmElement?: never
+      cancelElement?: never
+    }
+  | {
+      onAction?: never
+      confirmActionText?: never
+      cancelActionText?: never
+      confirmElement: React.ReactNode
+      cancelElement?: React.ReactNode
+    }
```
D7Torres, line 29: I might understand little about types but I think here you could not have
`confirmElement` and `onAction` at the same time, why is that not a valid combination?

D7Torres (approved):

lib/src/components/alert-dialog/alert-context/types.ts
Marcdup1: Yup, you are correct! the reason is because if you are passing in `confirmElement` you would specify your own `onClick`:

```tsx
  confirmElement: <Button theme="danger" size="sm" onClick={deleteStudent}>Yes, delete them please</Button>,
```

`onAction` wont be used, it only gets added to the button if `confirmElement` is omitted:
```tsx
          {confirmElement || (
            <Button size="sm" onClick={() => onAction?.(true)}>
              {confirmActionText}
            </Button>
          )}
```

documentation/components/markdown/editor-component/CodeBlock.tsx
Marcdup1: I came across it in the documentation [here](https://formidable.com/open-source/react-live/#render-function-demo)
I guess this means we can add some live examples for other components that use hooks etc

lib/src/components/alert-dialog/alert-context/AlertDialogIcon.tsx
```diff
@@ -0,0 +1,36 @@
...
+  flexShrink: 0,
+  alignSelf: 'flex-start',
+  color: '$blue1100',
```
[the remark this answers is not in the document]
Marcdup1: Oops! will remove  👍

documentation/content/components.feedback.alert-dialog.md
Marcdup1: "In order to keep `AlertDialogs` consistent this would be the preferred way of using it"? Im not too sure tbh 🤔

Isn't this type of thing handled in the `usage` tab?

D7Torres: I think `usage` is for designers. It has things like when to use, when not to use, and how to use (from a designer point of view).

I think some people might miss that there's the `showAlert` way, and go directly to composition.
But I'll leave it up to you

lib/src/components/alert-dialog/alert-context/types.ts
D7Torres: oh! right.

I didn't see the `onClick` on the documentation example, though. Maybe it's worth adding it.
image

D7Torres: I already approved, but just to be clear: ✅

documentation/content/components.feedback.alert-dialog.md
Marcdup1: Ah I see, I added a short paragraph now to suggest the recommended approach

PR #586, opened 2023-10-18
Fix: Add external link props to links

thomasdigby:
A minor fix to ensure that any component that can render a link also contains the logic to add props specific to external hrefs. e.g. `<a href="http://google.com" />` now renders as `<a href="http://google.com" target="_blank" rel="noopener noreferrer" />`

Please let me know if I've missed anything.

### DIFF
--- a/lib/src/components/action-icon/ActionIcon.tsx
+++ b/lib/src/components/action-icon/ActionIcon.tsx
@@ -11,6 +11,7 @@ import { OptionalTooltipWrapper } from '~/utilities/optional-tooltip-wrapper'
 
 import { Icon } from '../icon/Icon'
 import { ActionIconSizeMap } from './ActionIcon.constants'
+import { isExternalLink } from '~/utilities/uri'
 
 const getSimpleVariant = (base: string, interact: string, active: string) => ({
   bg: 'transparent',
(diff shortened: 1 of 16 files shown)


andoulla (approved):

lib/src/components/action-icon/ActionIcon.tsx
```diff
@@ -237,6 +238,9 @@ export const ActionIcon = React.forwardRef<HTMLButtonElement, ActionIconProps>(
           'aria-disabled': !!disabled
         } as const)
       : ({ type: 'button' } as const)
+    const externalLinkProps = isExternalLink(href)
+      ? { target: '_blank', rel: 'noopener noreferrer' }
```
D7Torres, line 242: As we are using this in a few places now
`{ target: '_blank', rel: 'noopener noreferrer' }`,

I wonder if it would be worth it to have it centralised in a util or something, so if we need to add something we add it to one place

D7Torres (approved):

lib/src/components/action-icon/ActionIcon.tsx
thomasdigby: Yep that's a good shout, I'll pop it in with our URI utilities

PR #590, opened 2023-10-31
Feat/add stitches theme to navigation and topbar

mzeenkala23:
The goal of this PR is to allow the ability to add custom theming to the `NavigationMenu` , `NavigationMenuVertical`  and the `Topbar` components, so that in the marketing site we can add our own "dark version" of the above components, without having to resort to the messy method of overriding every state e.g focus, hover, active etc.

### DIFF
--- a/lib/src/components/navigation/stitches.navigationMenu.colorscheme.config.ts
+++ b/lib/src/components/navigation/stitches.navigationMenu.colorscheme.config.ts
@@ -2,23 +2,23 @@ import { createTheme } from '~/stitches'
 
 const light = createTheme({
   colors: {
-    text: '$tonal400',
+    text: '$textSubtle',
     background: 'white',
-    backgroundHover: '$tonal50',
-    textHover: '$tonal600',
-    backgroundActive: '$tonal100',
-    textActive: '$tonal600',
-    backgroundDataActive: '$primaryLight',
-    dataActiveText: '$primary',
-    dataActiveTextHover: '$primary',
-    dataActivePressedText: '$primary',
-    dataActiveFocus: '$primary',
-    dataActiveBackgroundHover: '$tonal50',
-    dataActiveBackground: '$tonal100',
-    activeItemColor: '$tonal500',
-    activeItemBackgroundColor: '$tonal500',
-    triggerOpenBackground: '$tonal100',
-    navigationDropdown: 'white'
+    backgroundHover: '$grey100',
+    textHover: '$textForeground',
+    backgroundActive: '$grey200',
+    textActive: '$grey1000',
+    backgroundSelected: '$blue100',
+    textSelected: '$blue800',
+    textSelectedHover: '$blue800',
+    textSelectedPressed: '$blue800',
+    textSelectedFocus: '$blue800',
+    backgroundSelectedHover: '$grey100',
+    backgroundSelectedPressed: '$grey200',
+    itemTextSelected: '$grey900',
+    itemBackgroundSelected: '$grey900',
+    triggerOpenBackground: '$grey200',
+    navigationDropdownBackground: 'white'
   }
 })
 
--- a/lib/src/components/navigation/stitches.navigationMenu.colorscheme.config.ts
+++ b/lib/src/components/navigation/stitches.navigationMenu.colorscheme.config.ts
@@ -1,0 +1,27 @@
+import { createTheme } from '~/stitches'
+
+const light = createTheme({
+  colors: {
+    text: '$tonal400',
+    background: 'white',
+    backgroundHover: '$tonal50',
+    textHover: '$tonal600',
+    backgroundActive: '$tonal100',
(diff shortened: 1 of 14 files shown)


lib/src/components/navigation/stitches.navigationMenu.colorscheme.config.ts
```diff
@@ -0,0 +1,27 @@
...
+    dataActiveTextHover: '$primary',
+    dataActivePressedText: '$primary',
+    dataActiveFocus: '$primary',
+    dataActiveBackgroundHover: '$tonal50',
+    dataActiveBackground: '$tonal100',
+    activeItemColor: '$tonal500',
+    activeItemBackgroundColor: '$tonal500',
+    triggerOpenBackground: '$tonal100',
+    navigationDropdown: 'white'
+  }
```
LimeWub, line 22: can you use the blue/grey rather than primary/tonal?

All colours available here: https://atomlearning.design/theme/overview/colours

lib/src/components/top-bar/stitches.topBar.colorscheme.config.ts
```diff
@@ -0,0 +1,13 @@
+import { createTheme } from '~/stitches'
+
+const light = createTheme({
+  colors: {
+    background: 'white',
+    dividerBackground: '$tonal100',
+    borderBottomBackground: '$tonal100'
```
LimeWub, line 7: I think the last 2 here are both "divider"/"separator" colours, could both be a single variable and don't need "Background" in their name (as they are not a background/text section)

lib/src/components/top-bar/TopBar.tsx
```diff
@@ -73,16 +77,22 @@ type StyledRootProps = React.ComponentProps<typeof StyledRoot>
 
 interface TopBarProps extends StyledRootProps {
   css?: CSS
+  className?: string
```
LimeWub, line 80: you shouldn't need to add classname in the type as StyledRootProps will have it (?)

lib/src/components/navigation/stitches.navigationMenu.colorscheme.config.ts
```diff
@@ -0,0 +1,27 @@
...
+const light = createTheme({
+  colors: {
+    text: '$tonal400',
+    background: 'white',
+    backgroundHover: '$tonal50',
+    textHover: '$tonal600',
+    backgroundActive: '$tonal100',
+    textActive: '$tonal600',
+    backgroundDataActive: '$primaryLight',
+    dataActiveText: '$primary',
```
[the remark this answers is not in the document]
LimeWub: Not sure what `activeItemColor` is - is it for dropdown triggers? (Maybe we do need a pressed after all...)

lib/src/components/navigation/stitches.navigationMenu.colorscheme.config.ts
mzeenkala23: sure thing I will update the colours token

lib/src/components/top-bar/stitches.topBar.colorscheme.config.ts
mzeenkala23: as discussed we will just remove background from those two properties, as we need them to be two different variables for more specific styling

lib/src/components/top-bar/TopBar.tsx
mzeenkala23: hmm I just tried removing it from the interface, but then started to get type Errors.

```
Type '{ children: Element; className: string; }' is not assignable to type 'IntrinsicAttributes & TopBarProps & { children?: ReactNode; }'.
  Property 'className' does not exist on type 'IntrinsicAttributes & TopBarProps & { children?: ReactNode; }'
  
  ```

lib/src/components/navigation/stitches.navigationMenu.colorscheme.config.ts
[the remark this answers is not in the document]
mzeenkala23: so `activeItemColor` was using for when the a `NavigationMenuLInk` or `NavigationMenuTrigger` as the active prop so I updated the token name too `$itemBackgroundSelected` and `$itemTextSelected`

lib/src/components/top-bar/TopBar.tsx
LimeWub: ok fine to leave it 👍

LimeWub (approved):
one little thing~ thanks for doing these changes hope they work well for the dark mode theming on marketing 🤞

lib/src/components/navigation/stitches.navigationMenu.colorscheme.config.ts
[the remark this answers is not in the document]
LimeWub: nice 👍

Marcdup1 (approved):

PR #593, opened 2023-11-08
Feature: Side bar

thomasdigby:
### Description

Created a `SideBar` component that allows us to have improved and more visible navigation for users between mobile and desktop. The aim is for this component to be paired with `NavigationMenuVertical` and for type to be toggled to `static` at a breakpoint that's big enough to display the whole nav.

[Figma docs here](https://www.figma.com/file/KjgQh7eTwEfJmH6e56k9mX/DS-Documentation?type=design&node-id=14049%3A24793&mode=design&t=CL9MXpXpFVsR2HL9-1)

### Implementation

- Create new `SideBar` component with relevant sub-components.
- Implement "expandable" functionality to expand the side-bar container if:
  - A user hovers over the side-bar
  - A user focuses any element within the side-bar
  - A user touches or clicks the side-bar
- Additionally, ensured that the initial user touch does not trigger any link or button behaviour, once opened the user can complete their action
- Created and exported `useExpandableSidebar()` hook so any child component within the `SideBar` can access the current collapsed/expanded state. This will be useful for custom components or implementations that require 2 variants dependent on the parent `SideBar` width.

#### `type="static"`
Screen Shot 2023-11-08 at 16 31 06

#### `type="expandable"`

https://github.com/Atom-Learning/components/assets/3226804/a69c0229-5f7c-41ea-acdf-8379fa7344d5


lib/src/components/side-bar/SideBar.tsx
```diff
@@ -0,0 +1,151 @@
...
+      : {}
+
+  return (
+    <SideBarExpandableContext.Provider value={{ isExpanded }}>
+      <Root {...props} className={className} type={type}>
+        <Wrapper {...expandableProps} isExpanded={isExpanded} type={type}>
+          <Content>
+            {children}
+            {type === 'expandable' && (
+              <PointerBlocker isVisible={!isHovered && !isExpanded} />
```
thomasdigby, line 135: When the user hovers, clicks or touches the collapsed side bar (or anything within it), the `PointerBlocker` will remain in place for 50ms while it's expanding. It is here to solve a problem for touch devices specifically, where the user can click a link before the sidebar expands.

This is a bit of a blunt-force approach, but I think it's preferable to searchin for child focusable elements and preventing default, especially when we don't (or shouldn't) have control over any child elements.

lib/src/components/side-bar/SideBar.tsx
```diff
@@ -0,0 +1,151 @@
...
+    }
+  }
+})
+
+const Wrapper = styled('div', {
+  borderRight: '1px solid $border',
+  boxShadow: '4px 0 4px -2px rgba(31, 31, 31, 0)',
+  height: '100%',
+  overflow: 'hidden',
+  willChange: 'width',
```
LimeWub, line 48: dayum :3 this is for perf right? noice

lib/src/components/side-bar/SideBar.tsx
```diff
@@ -0,0 +1,151 @@
...
+  variants: {
+    type: {
+      expandable: { width: SIZE_COLLAPSED }
+    }
+  }
+})
+
+const Wrapper = styled('div', {
+  borderRight: '1px solid $border',
+  boxShadow: '4px 0 4px -2px rgba(31, 31, 31, 0)',
```
LimeWub, line 45: What is this boxShadow and how comes it's not a token? :o

LimeWub (commented):
Thanks for doing this work :D

I've just done a very very very very rough/rushed review/first thoughts before my standup; just posting to start the convo basically.
Haven't followed the actual hover/pointer code too deeply so far.

lib/src/components/navigation-menu-vertical/NavigationMenuVerticalText.tsx
```diff
@@ -16,5 +36,13 @@ type TStyledNavigationMenuVerticalTextProps = React.ComponentProps<
 export const NavigationMenuVerticalText = (
   props: TStyledNavigationMenuVerticalTextProps
 ): JSX.Element => {
-  return <StyledNavigationMenuVerticalText size="md" as="span" {...props} />
+  const { isExpanded } = useExpandableSidebar()
```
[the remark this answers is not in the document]
LimeWub: Update: It does makes sense after I read the rest of the PR just felt a bit odd when I was reading this initially. I wonder if it can be done any other way.... (?) iunno (as I said in the comment this is a very rushed review for now soz :3 )

lib/src/components/side-bar/SideBar.tsx
thomasdigby: This came from `TopBar`, I tried to keep it consistent. Agree that it should be a token though! Is there one you'd suggest for both?

lib/src/components/side-bar/SideBar.tsx
thomasdigby: Yeah, it _may_ not be necessary tbh, but as we are only animating this one property we can cheaply/safely prep the browser for animating it.

thomasdigby:
> So does this mean the collapsed version doesn't allow directly clicking the icons to go somewhere or am I misreading this?

@LimeWub We ideally want the side bar to expand before selecting an item. The problem we had on touch devices previously was that you would click on an icon, it would navigate to the new page, but you'd still have no idea what you actually clicked on.

How it works currently is:
- Mouse users: Hover on the side bar, it opens, you can click a link/button within. There is a period of only 50ms where users cannot click on the icons, so by the time they've hovered on the icon it will already be available (even if the side bar isn't fully expanded) 
- Keyboard users: Focus anything within the sidebar, it opens, you can tab to other options and select something. If focus moves to something outside of the side bar, it closes
- Touch users: Touch anywhere on the side bar to open it, including the links/icons, then select an option after it is open.

lib/src/components/side-bar/SideBar.tsx
LimeWub: I like it! It'd be best if we weren't animating width but I don't see how it would be possible here so this property is good if it works!

lib/src/components/navigation-menu-vertical/NavigationMenuVerticalText.tsx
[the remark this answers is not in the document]
thomasdigby: I think this approach is a good compromise between composition/flexibility/decoupling components, and opt-in behaviour for components that are likely to be used with this component.

The side bar may not always be used with a nav, we may also want to tweak some other UI within the side bar dependent on the expanded/collapsed state.

So it makes sense to me not to bind the expand/collapse state with the nav, but allow the nav to access it when necessary.

lib/src/components/side-bar/SideBarComponents.tsx
```diff
@@ -0,0 +1,40 @@
+import { styled } from '~/stitches'
+
+import { Image } from '../image'
+import { Text } from '../text'
+
+export const SideBarHeader = styled('div', {
```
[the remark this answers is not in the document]
thomasdigby: I think they're pretty much the same? I'd definitely be keen to reuse some stuff here, where would you want to store them? I'm not super on board with components using other components sub-components, if that makes sense

lib/src/components/side-bar/SideBarExpandableContext.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react'
```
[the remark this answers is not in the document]
thomasdigby: Yeah, fair. I was just trying to be a touch clearer about its use. `useSideBar()` doesn't make as much sense in the context... 🤔

lib/src/components/side-bar/SideBar.tsx
thomasdigby: Yeah it's unavoidable unfortunately

lib/src/components/side-bar/SideBarExpandableContext.tsx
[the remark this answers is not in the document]
thomasdigby: Updated to `useSideBar()`

lib/src/components/side-bar/SideBar.tsx
```diff
@@ -0,0 +1,155 @@
...
+      >
+        <Content {...expandableProps} isExpanded={isExpanded} type={type}>
+          {children}
+          {/**
+           * When the SideBar is collapsed, ensure that the initial tap event
+           * is used to expand the nav first before making the items available
+           */}
+          {type === 'expandable' && (
+            <PointerBlocker isVisible={!isHovered && !isExpanded} />
+          )}
```
D7Torres, line 141: wasn't it possible to render stuff that is actually not clickable and then when is expanded to render the clickable elements?
You probably had a good reason to do it this way

D7Torres (approved):

lib/src/components/navigation-menu-vertical/NavigationMenuVerticalText.tsx
```diff
@@ -3,10 +3,37 @@ import React from 'react'
 import { Text } from '~/components/text'
 import { styled } from '~/stitches'
 
+import { useSidebar } from '../side-bar'
+
+const SIZE_EXPANDED_MAX = '160px'
```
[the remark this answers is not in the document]
thomasdigby: Yeah potentially, if it takes into account the internal padding of the link/button as well as the icon, and the padding of the sidebar... 🤔 For this initial PR I think I'll leave the hard-coded value as is, but there are a few things I want to tidy up as improvements in a patch next week, so I'll add this to the list.

Agree that the all the widths in this PR could be set as `rem` as everything within will be scaling using `rem`.

lib/src/components/navigation-menu-vertical/NavigationMenuVerticalText.tsx
```diff
@@ -3,10 +3,37 @@ import React from 'react'
...
 })(Text, {
-  fontWeight: 'var(--navigation-menu-vertical-item-font-weight)'
+  fontWeight: 'var(--navigation-menu-vertical-item-font-weight)',
+  lineHeight: 1.2,
+  variants: {
+    isExpanded: {
+      true: {
+        width: 'max-content',
+        maxWidth: SIZE_EXPANDED_MAX,
+        '@allowMotion': {
```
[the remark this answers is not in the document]
thomasdigby: Did you search the components codebar? 😜 It's an alias for `(prefers-reduced-motion: no-preference)`

lib/src/components/side-bar/SideBar.tsx
```diff
@@ -0,0 +1,155 @@
...
+  flexDirection: 'column',
+  height: '100%',
+  overflow: 'hidden',
+  willChange: 'width',
+  '@allowMotion': {
+    transition: 'width 125ms ease-out, box-shadow 125ms ease-out'
+  },
+  variants: {
+    type: {
+      static: {},
```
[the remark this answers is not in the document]
thomasdigby: This is just for the types, ensuring that we have both `static` and `expandable` as the two options for that prop. That type is then used in `Content` so they need to match

lib/src/components/side-bar/SideBar.tsx
thomasdigby: Yeah I just wanted a separation of concerns. We do have a slight coupling with `useSideBarState()` in `NavigationMenuVerticalText`, but other than that they're completely separate. We can be certain that the sidebar is expanded before interacting with _any_ content, whether its a nav, or individual link in the footer, logo in the head, etc.

lib/src/components/side-bar/SideBarComponents.tsx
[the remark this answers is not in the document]
thomasdigby: Will tackle this in a follow-up PR

lib/src/components/side-bar/SideBar.tsx
thomasdigby: Will tackle this in a follow-up PR

lib/src/components/side-bar/SideBar.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+      <SideBar.Brand>
+        <SideBar.BrandLogo alt="Logo" src="" />
+      </SideBar.Brand>
+    </SideBar.Header>
+    <SideBar.Body>Main content</SideBar.Body>
+    <SideBar.Footer>Footer content</SideBar.Footer>
+  </SideBar>
+)
+
+describe('SideBar component', () => {
```
Mhoog, line 20: Is it worth it to add the scenarios for expand/collapsing with focus/blur (and a scenario for closing with `hoverEnd`)?

lib/src/components/side-bar/SideBarContext.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react'
+
+export const SideBarContext = React.createContext<{
+  isExpanded?: boolean
+}>({
+  isExpanded: undefined
+})
```
Mhoog, line 7: You could make `isExpanded` mandatory and set the initial value to `false`, so that the type is `boolean` only.

thomasdigby: `isExpanded` may not be a boolean though, as the `SideBar` type could be `static`. I guess that could still be an `isExpanded` value of `true` though... 🤔 Yeah I think I can default to `false` 👍

lib/src/components/side-bar/SideBar.test.tsx
thomasdigby: Yep for sure, I'll add them 👍

lib/src/components/side-bar/SideBarContext.tsx
Mhoog: Ah fair, if it's a hassle don't bother. It's very minor

Mhoog (approved):

lib/src/components/side-bar/SideBarComponents.tsx
```diff
@@ -0,0 +1,40 @@
+import { styled } from '~/stitches'
```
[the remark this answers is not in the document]
thomasdigby: It feels a bit unnecessary, especially as they're all small `styled` components, but either way we can discuss for a future release 👍

PR #596, opened 2023-11-17
b2b-313-add-disabled-state-to-DS-components

swapnil1807:
DS - New consistent Disabled state style for all components [Fix]
https://atomlearningltd.atlassian.net/browse/B2B-313
Implementation 👍:

- Added file `disabledStyle` under `lib/src/utilities/style/ ` to apply the same to all the components
- Changed all the List of components:

Tabs
Action Icon
Button
Link
Checkbox
Date Input
Date field
Select
Slider
Slider field
Switch
Dropdown Menu

- Updated all the sanpshots
- Below is the screenshot for disabled states of all the requested components
Screenshot 2023-11-17 at 15 28 45
Screenshot 2023-11-17 at 12 11 17


lib/src/utilities/style/disabledStyle.ts
```diff
@@ -0,0 +1,15 @@
...
+  opacity: number
+  cursor: 'not-allowed'
+}
+
+export const disabledStyle = (): disabledStyleReturnType => {
+  return {
+    opacity: 0.3,
+    cursor: 'not-allowed'
+  }
+}
```
thomasdigby, line 15: We don't have to use a function for this, we can just return a plain object. We could probably also simplify the type to be `CSS`, e.g.

```ts
import type { CSS } from '~/stitches'

export const disabledStyle = {
  opacity: 0.3,
  cursor: 'not-allowed'
} as CSS
```

(You may not even need the type assertion, the type validate script will let you know if it's necessary)

lib/src/components/switch/Switch.tsx
```diff
@@ -26,6 +27,7 @@ const StyledSwitch = styled(RadixSwitch.Root, {
   '&[data-state="checked"]:hover': {
     backgroundColor: '$primaryMid'
   },
+  '&[disabled]': { ...disabledStyle() },
```
thomasdigby, line 30:
```suggestion
  '&[disabled]': disabledStyle,
```

There's no need to spread the result here, we can just reference it directly. (treating it as an object as per my other comment)

chris-atomlearning (approved):
Looks great Swapnil, nice one. Good suggestion from Tom to simplify the style but apart from that it looks great :)

MiguelGDLR (approved):
all components look great Swapnil! 👏 thanks a lot

lib/src/components/dropdown-menu/DropdownMenuItem.tsx
```diff
@@ -12,10 +13,7 @@ export const itemStyles = {
   position: 'relative',
   px: '$3',
   py: '$2',
-  '&[data-disabled]': {
-    color: '$tonal200',
-    pointerEvents: 'none'
-  },
+  '&[disabled]': { disabledStyle, pointerEvents: 'none' },
```
thomasdigby, line 16: This will need to be:

```suggestion
  '&[disabled]': { ...disabledStyle, pointerEvents: 'none' },
```

lib/src/components/link/Link.tsx
```diff
@@ -25,6 +25,11 @@ export const StyledLink = styled('a', {
   '&:active': {
     color: '$primaryDark'
   },
+  '&[disabled]': {
+    disabledStyle,
```
thomasdigby, line 29:
```suggestion
    ...disabledStyle,
```

lib/src/components/markdown-content/__snapshots__/MarkdownContent.test.tsx.snap
```diff
@@ -116,32 +116,41 @@ exports[`MarkdownContent component renders 1`] = `
...
-  .c-fucOQC:active {
+  .c-gyuptP:active {
     color: var(--colors-primaryDark);
   }
 
-  .c-dyvMgW > .c-fucOQC,
-  .c-jryFOR > .c-fucOQC,
-  .c-PJLV > .c-fucOQC,
-  .c-bYNGXt > .c-fucOQC {
+  .c-gyuptP[disabled] disabledStyle {
```
thomasdigby, line 129: @swapnil1807 FYI this is the output by doing `'selector[disabled]': { disabledStyled, 'pointer-events': 'none' }`

thomasdigby (approved):

PR #600, opened 2023-11-22
fix: remove unnessesary prop on BannerRegular Text

LimeWub:
Component had been tweaked but prop forgotten and thus showing in output.


thomasdigby (approved):
Nice! Thanks for the quick turnaround! 😄

andoulla (commented):
while we are at it can we please make value as optional

```<BannerRegular
  colorScheme={{ base: 'blue1' }}
  emphasis="subtle"
  size="md"
  value=""
>```

In most of the cases in atom-core, we are force to pass an empty string to silence TS.

D7Torres (approved):

LimeWub:
@andoulla 
Done this, but on `Dismissible` itself here:
https://github.com/Atom-Learning/components/pull/603

PR #601, opened 2023-11-27
Fix: Replace margins with gap

thomasdigby:
Converting all of our "internal" margin usage over to flex gap. I've tested each of these components before/after in the sandbox and everythings looking good.


andoulla (approved):
sorry for the delay took a while to get through these. Approving pending one question if thats ok

LimeWub:
Why is this change being made?
Are you removing all internal margins for gaps? Is this better somehow?
(Just trying to understand before I review)

PR #602, opened 2023-11-27
Fix: Add `flex-shrink: 0` to `Icon` & `ActionIcon`

thomasdigby:
We use `flexShrink: 0` a lot on our Icons both within the design system and in core, we can add this default to the `Icon` component.


andoulla (approved):

LimeWub (approved):
yessssss

PR #604, opened 2023-12-08
New Component: Tree

LimeWub:
Figma: https://www.figma.com/file/KjgQh7eTwEfJmH6e56k9mX/DS-Documentation?type=design&node-id=13555-33291&mode=design&t=aG7rjRPnKLXBHra4-0

Introduces the `Tree` component which I need as part of the `CheckboxTree` work.

>Tree is a component that represents data in a hierarchical or tree-like structure, where items are organised into parent nodes and child nodes.
>
>Users can expand or collapse nodes to reveal or hide their child nodes, allowing them to navigate through the hierarchy efficiently.

Having some issues with TS due to stitches bugs and not sure how to fix. Help and comments are more than welcome.

## Video

https://github.com/Atom-Learning/components/assets/6905473/2670e82f-6496-4986-9c39-eaa75fc514dc


documentation/content/components.navigation.tree.md
```diff
@@ -0,0 +1,91 @@
...
+        <Tree.Item>Two</Tree.Item>
+        <Tree.Collapsible>
+          <Tree.CollapsibleTrigger label="Expand: Trigger 1"><Icon is={Sun} /> Trigger 1</Tree.CollapsibleTrigger>
+          <Tree.CollapsibleContent>
+            <Tree.Item>Nested: One</Tree.Item>
+            <Tree.Item>Nested: Two</Tree.Item>
+          </Tree.CollapsibleContent>
+        </Tree.Collapsible>
+        <Tree.Collapsible defaultOpen={true}>
+          <Tree.CollapsibleTrigger label="Expand: Trigger 2">Trigger 2<Icon is={Bell} /></Tree.CollapsibleTrigger>
```
LimeWub, line 31: Not loving the repetition of `label` and the actual text content of the trigger.
Could give the trigger a ref and use `.innerText` for the label but it introduces a bit more code. What do we think?

lib/src/components/tree/TreeListItem.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react'
+
+import { Flex } from '~/components/flex'
+
+type TreeListItemProps = React.ComponentProps<typeof Flex>
+
+export const TreeListItem = (props: TreeListItemProps): JSX.Element => {
+  return <Flex as="li" gap={2} align="center" {...props} />
```
LimeWub, line 8: For when this flags in TS: https://github.com/stitchesjs/stitches/issues/1105 :(

lib/src/components/tree/TreeCollapsibleTrigger.tsx
```diff
@@ -0,0 +1,65 @@
...
+  children,
+  label,
+  ...rest
+}: TNavigationMenuCollapsibleTriggerProps) => {
+  const { setTriggerRef, triggerRef } = React.useContext(TreeCollapsibleContext)
+
+  return (
+    <StyledTreeCollapsibleTrigger
+      {...rest}
+      onClick={() => triggerRef?.current?.click()}
```
Mhoog, line 48: What is the reason that we need to do this? Is it not possible to render the `children` inside the `Trigger` component?

lib/src/components/tree/TreeCollapsibleTrigger.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+import { TreeCollapsibleContext } from './TreeCollapsible'
+import { TreeItemContent } from './TreeItemContent'
+
+const StyledTreeCollapsibleTrigger = styled(TreeItemContent, {
+  cursor: 'pointer'
+})
+
+const StyledActionIcon = styled(ActionIcon, {
+  position: 'absolute',
```
Mhoog, line 17: Is it possible to use flexbox, instead of absolute positioning?

lib/src/components/tree/TreeCollapsibleTrigger.tsx
LimeWub: This is the best combination for accessibility of having both the ActionIcon clickable and with the full aria stuff and ActionIcon styles on it and the full trigger area clickable.

lib/src/components/tree/TreeCollapsibleTrigger.tsx
LimeWub: What's the concern with absolute positioning?

It messes up the spacing if it's not absolutely positioned.  We want to be aligning on text - so we don't want the arrow to be taking up space. 
absolute positioning is just the correct way to do this, unless we want to be employing optional negative margins etc (which is not worth it imo)

Screenshot 2023-12-08 at 16 09 27

documentation/content/components.navigation.tree.md
LimeWub: I tried the `.innerText` and it was not getting generated/available in time for tests which was a PITA so I removed it...

MiguelGDLR (approved):

lib/src/components/tree/Tree.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+    expect(mockOnOpenChange).toBeCalledWith(true)
+    expect(collapsibleTriggerActionIcon).toHaveAttribute('data-state', 'open')
+    const collapsibleContent = getByTestId('collapsible-content')
+    expect(collapsibleContent).toBeVisible()
+
+    userEvent.click(collapsibleTriggerActionIcon)
+
+    expect(mockOnOpenChange).toBeCalledWith(false)
+    expect(collapsibleTriggerActionIcon).toHaveAttribute('data-state', 'closed')
```
andoulla, line 66: you could also assert the label here (but that's just extra)

lib/src/components/tree/TreeCollapsibleTrigger.tsx
andoulla, line 48: it would be good to have a way of passing more stuff into the onClick. the most common thing we miss in the DS when adding new components is this, it could be from triggering an event to triggering a different action. just having a onClick prop that gets destructured and called within onclick will fix this

lib/src/components/tree/Tree.tsx
```diff
@@ -0,0 +1,39 @@
...
+)
+
+export const Tree = Object.assign(TreeRoot, {
+  Collapsible: TreeCollapsible,
+  CollapsibleContent: TreeCollapsibleContent,
+  CollapsibleTrigger: TreeCollapsibleTrigger,
+  Item: TreeItem,
+  ItemContent: TreeItemContent,
+  Icon: TreeIcon,
+  Text: TreeText
```
thomasdigby, line 38: Are all of these components necessary for the dev? As far as I can see from the test and the docs, we use `Tree`, `Tree.Item`, `Tree.Collapsible`, `Tree.CollapsibleContent`, and `Tree.CollapsibleTrigger` when using the component.

It would be good to remove any that aren't essential for the dev for now.

lib/src/components/tree/TreeCollapsibleTrigger.tsx
Mhoog: Ah, I see. I thought the arrow was aligned with the start of the text of the parent item, but seems like it isn't.

More of a design question, but should it not be aligned with the parent item? In your screenshot I find it a bit confusing to see the structure (especially with `Nested: #4`).

More like:

```
  Item 1
⌄ Item 2
    Item 3
  > Item 4
```

Instead of:

```
  Item 1
⌄ Item 2
   Item 3
 > Item 4
```

^ note the `>` for Item 4.

lib/src/components/tree/TreeCollapsibleTrigger.tsx
Mhoog: Why do the `aria-` attributes only need to be on the `ActionIcon`?

I think this HTML structure would be fine.
```html
<button aria-...>
  <img src="icon.svg" />
  <span>Some label</span>
</button>
```

(I'm probably missing a requirement somewhere, just trying to understand)

LimeWub: I can't make the full thing a Trigger for a couple of reasons:
* in the `CheckboxTree` case it can have nested checkboxes in the trigger - so this is nesting interactive in interactive and a11y check complains as it's invalid
* I wouldn't be able to use the ActionIcon and styles nested inside of it as once again it's button in button which is not good and it would cause me having to duplicate Icon styles for this use case

So the ActionIcon is used directly wrapped in the trigger - so it's the only button technically and it is the bit which is focusable with the keyboard controls, and as it gets all the aria correctly from radix, makes sense for keyboard users.

But, because it was preferable in terms of design to have the whole label open/close the collapsable; I'm adding an onClick forwarding basically, on the full trigger as well. 

But the bit which is important/more correct is the ActionIcon itself to be the actual trigger button so we can nest focusable/interactable elements in the general trigger area.

lib/src/components/tree/TreeCollapsibleTrigger.tsx
LimeWub: That's just the design. 
Nested elements are all in by x amount of left padding depending on nesting and then we are using the ActionIcon component for the arrow down which has a bit of extra padding on it. 

Screenshot from figma:

Screenshot 2024-01-02 at 15 23 01

Mhoog: Yeah fair, @MiguelGDLR what do you think?

lib/src/components/tree/TreeList.tsx
```diff
@@ -0,0 +1,42 @@
...
+>
+
+export const TreeList = React.forwardRef(
+  (
+    props: TreeListProps,
+    ref: React.ForwardedRef<HTMLUListElement>
+  ): JSX.Element => (
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore: Stitches polymorphic components issue due to `as="ul"`
+    <StyledList as="ul" ref={ref} {...props} direction="column" />
```
Mhoog, line 40: Could you add the `tree` role on this?
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tree_role

lib/src/components/tree/TreeListItem.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export const TreeListItem = React.forwardRef(
+  (
+    props: TreeListItemProps,
+    ref: React.ForwardedRef<HTMLLIElement>
+  ): JSX.Element => {
+    return (
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore: Stitches polymorphic components issue due to `as="li"`
+      <StyledTreeListItem as="li" gap={2} align="center" {...props} ref={ref} />
```
Mhoog, line 22: Could you add the `treeitem` role on this?
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/treeitem_role

Mhoog:
@LimeWub,

For a separate PR, but it would be nice to have the recommended keyboard interactions implemented once all PRs are combined:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tree_role#multi-select_keyboard_interactions

lib/src/components/tree/TreeCollapsibleTrigger.tsx
LimeWub: General info: The specific onClick is the trigger container onClick which forwards the onClick even basically to the actual button trigger. This specific use case is to make the full trigger clickable while maintaining a11y and allowing the contents of the trigger to be interact elements.

But in terms of functionality I did have to check how the onClick/onOpenChange works cause I couldn't remember.

in terms of **onOpenChange**
Open/Close state callback can be passed in through the `Collapsible.Root` ie. `Tree.Collapsible` as such:

Screenshot 2024-01-02 at 16 44 08

I think this would cover most use cases.

in terms of **onClick**
I also tested adding an `onClick` event to the `.Trigger` specifically and ye it does get overwritten by this forwarding click hack.
I've added this line to support it - in case it's ever needed as you suggested ^^

Screenshot 2024-01-02 at 16 52 24

lib/src/components/tree/TreeCollapsibleTrigger.tsx
MiguelGDLR: true, we discussed this before but I guess my solution doesn't work well. 
I wanted to use an action icon 32x32px  to have a good minimum clickable target area, but including the margins it sums up a lot of distance from every level down if you align with the parent. I'm not sure how else to solve this, will think about it. Any ideas welcome
Screenshot 2024-01-02 at 17 54 15

lib/src/components/tree/Tree.tsx
LimeWub: `TreeIcon` and `TreeText` are useful to export as fallback for the styling, in case we want to nest text and icons in fragments or other elements as the map in `TreeItemContent` which replaces the `Icon` with `TreeIcon` etc won't be able to catch via the children map to replace them. 

but I think `ItemContent` can stay internal so I'll remove that ones export.

lib/src/components/tree/TreeList.tsx
LimeWub: Yep added the 3 relevant roles - I think I've done it alright although not 100% sure cause there was no example in MDN ;-; 
This is the commit: https://github.com/Atom-Learning/components/pull/604/commits/578a2d31b2e243dff3d7803f5128d9e22a44c5dc

lib/src/components/tree/TreeCollapsibleTrigger.tsx
LimeWub: Oh @Mhoog just to clarify my screenshot in this comment thread isn't how it has been built. It is to demonstrate that it doesn't work out correctly if not using position absolute.
I think that maybe has been confusing with reviewing this 2 weeks ago and coming bk to it now :)
It's actually formatted as in the video in the description :) Hope that makes sense?

Mhoog (approved):

PR #605, opened 2023-12-12
Feat/checkbox group

LimeWub:
Adds the `CheckboxGroup` primitive component.
This component can be controlled or uncontrolled.
Keeps track of mounted checkboxes so only they are affected when clicking the `AllItem` checkbox.
Component will be used in `CheckboxTree` as well as future dropdown selects.

https://github.com/Atom-Learning/components/assets/6905473/59dcc8ba-ddc1-4756-abbb-ed49e4ac9f26


MiguelGDLR (approved):

lib/src/components/checkbox-group/CheckboxGroup.tsx
```diff
@@ -0,0 +1,50 @@
...
+      <CheckboxGroupMountedProvider>
+        <Component {...rest} />
+      </CheckboxGroupMountedProvider>
+    </CheckboxGroupCheckedProvider>
+  )
+}
+
+export const CheckboxGroup = Object.assign(CheckboxGroupRoot, {
+  Item: CheckboxGroupItem,
+  AllItem: CheckboxGroupAllItem,
```
andoulla, line 48: This should be AllItems in my opinion, I know it is small but it is easier to address it now than later. Same with the CheckboxGroupAllItems

lib/src/components/checkbox-group/CheckboxGroupAllItem.tsx
```diff
@@ -0,0 +1,52 @@
...
+      return 'indeterminate'
+
+    return false
+  })()
+
+  return (
+    <Checkbox
+      onCheckedChange={handleItemCheckedChange}
+      checked={isAllChecked}
+      title={title}
```
andoulla, line 48: checkbox doesn't have a title, neither the Html one, does it? if we need this of accessibility why not pass an aria-label instead?

lib/src/components/checkbox/Checkbox.tsx
```diff
@@ -68,17 +69,17 @@ const toIconSize = {
...
       [size]
     )
 
     return (
-      <StyledCheckbox {...props} size={size} ref={ref}>
+      <StyledCheckbox ref={ref} checked={checked} size={size} {...rest}>
         <StyledIndicator asChild>
           <Icon
-            is={props.checked === 'indeterminate' ? Minus : Ok}
+            is={checked === 'indeterminate' ? Minus : Ok}
```
andoulla, line 82: in accessibility documentation,  they tend to use mixed instead of indeterminate, and I think it makes more sense than indeterminate because it is actually mixed (we have both states). Matching it to that terminology I think just makes it easier to understand due to consistency

lib/src/components/checkbox-group/CheckboxGroup.tsx
LimeWub: I disagree on the suggestion for this one. 
The `AllItem` is the one which selects all the checkboxes nested or equal level under it. 
So it's the `CheckAll`. 
So it's not a collection of `Items` but instead it's a singular "select all the checkboxes item" which is why I names it `AllItem`.

Naming could be improved I agree for sure but not by making `Item` into `Items` as it's still just a singular checkbox...

lib/src/components/checkbox-group/CheckboxGroupAllItem.tsx
LimeWub: [The HTMLElement has a title property.](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/title)

The A11y test is failing without the title provided, so I had to provide a title on the checkboxes. The `CheckboxItem` defaults to the value of the item passed in (and can be overwritten if needed) but the `AllItem` has no value so I've made it an explicit prop...

lib/src/components/checkbox/Checkbox.tsx
LimeWub: Hey the image hasn't attached correctly - but this is a[ radix naming in the checkboxes](https://www.radix-ui.com/primitives/docs/components/checkbox) not our own choice - and we usually trust radix's choice of a11y.
 
Screenshot 2024-01-02 at 11 57 38

MiguelGDLR: if it helps I've studied so many DSs and everyone calls it "indeterminate" state, at least in the user-facing UI terminology. Not sure about accessibility

documentation/content/components.form.primitives.checkbox-group.md
```diff
@@ -0,0 +1,63 @@
...
+      (!) Note that `Item`s need to be mounted to affect the `AllItem` state and be targetted by it. If filtered out or collapsed without them being in the DOM they do not count. This is by design!
+
+
+      ## Examples
+
+
+      <CodeBlock live={true} preview={true} code={`  <CheckboxGroup>
+          <Flex direction="column" gap={1}>
+            <CheckboxGroup.AllItem />
+            <CheckboxGroup.Item value={1} />
```
thomasdigby, line 29: Can we use an example that includes label text and a semi-realistic scenario? e.g. a series of checkbox terms/things that the user has to agree and one `AllItem` at the top to say "Agree to all"

thomasdigby:
Why do the checkboxes have to be mounted? You mentioned in the docs that it was by design but I can't see anything that describes why

LimeWub:
@thomasdigby The idea is that the CheckboxGroup will also be used in Dropdown Selects (when they have the checkbox at least) and when that was designed, a while back there was this conversation on the example which provides filtering.


Screenshot 2024-01-02 at 14 13 35
Screenshot 2024-01-02 at 14 14 55

So I'm keeping track of whether the checkboxes are mounted or not to show checked/unchecked/indeterminate and select only the checkboxes that are actually visible/available to the user.

thomasdigby:
> @thomasdigby The idea is that the CheckboxGroup will also be used in [Dropdown Selects](https://www.figma.com/file/KjgQh7eTwEfJmH6e56k9mX/DS-Documentation?type=design&node-id=5249%3A39751&mode=design&t=mgZIUWWrHiowxCzS-1) (when they have the checkbox at least) and when that was designed, a while back there was this conversation on the example which provides filtering.
> So I'm keeping track of whether the checkboxes are mounted or not to show checked/unchecked/indeterminate and select only the checkboxes that are actually visible/available to the user.

Ok this makes a lot of sense, thanks. 👍

lib/src/components/checkbox-group/CheckboxGroupItem.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  React.useEffect(() => {
+    handleItemMountedChange(true, value)
+    return () => {
+      handleItemMountedChange(false, value)
+    }
+  }, [handleItemMountedChange, value])
+
+  return (
+    <Checkbox
```
Mhoog, line 43: Would it be possible to 'force' users of the component to create a label for the checkbox, so that we don't need to specify the `title`?

Right now it would be possible to use the checkbox without a label. Do we need that?

LimeWub: The `Checkbox` component itself can be used without the `InlineFieldWrapper` - my logic is if that component can get away with not having a label (a11y will be broken for Checkbox without the title as well) then the checkbox group should as well 🤷

Mhoog (approved):

documentation/content/components.form.primitives.checkbox-group.md
LimeWub: I've added the title so I'm demoing with something actually accessible but will not make other changes to this example as the `Checkbox` primitive doesn't include label or further components and I'm mirroring that one.

If we want the full label etc added we should make a `CheckboxGroupField` component and add these via that one to maintain the pattern ^^

PR #608, opened 2024-01-08
Feature: Update documentation nav, integrate new Quest theme

thomasdigby:
### Implementation

- Update documentation nav to use `SideDrawer`/`Sidebar` (depending on viewport size) & `NavigationMenuVertical`
- Implement a theme switcher in the nav to switch between "Atom" and "Quest" themes
- Remove leading numbers in titles and navigation items
- Update the reference to `@atom-learning/theme` to point to correct file
- Add reference to new Quest font

https://github.com/Atom-Learning/components/pull/609 is now merged into this PR, so there are far more changes here than I intended. The changes to the documentation site are all that needs to be reviewed.

### Screenshots

Screen Shot 2024-01-08 at 10 30 33
Screen Shot 2024-01-08 at 10 30 37
Screen Shot 2024-01-08 at 10 30 43


MiguelGDLR (approved):

lib/src/components/section-message/SectionMessage.tsx
```diff
@@ -44,8 +44,8 @@ const StyledSectionMessage = styled(Dismissible, {
         color: '$grey1000'
       },
       info: {
-        bg: '$blue100',
-        color: '$blue1000'
+        bg: '$infoLight',
```
andoulla, line 47: why keep the infoLight/Dark shorthands if we are removing the primaryLight/Dark ones?

andoulla (approved):

lib/src/components/section-message/SectionMessage.tsx
thomasdigby: This naming scheme is consistent across the other "semantic" values, e.g. warning, danger, success. I'm not sure if we're looking to change that any time soon, but it made more sense to be consistent with the others than to introduce something new just for this.

PR #609, opened 2024-01-08
Fix: New primary theme access

thomasdigby:
### Implementation

- Update all primary colour tokens to their new respective value on the primary scale:
  - `$primaryLight` = `$primary100`
  - `$primary` = `$primary800`
  - `$primaryMid` = `$primary900`
  - `$primaryDark` = `$primary1000`
- Create new `primary1` & `primary2` colorscheme configurations that access either `$primary` scale values
- Update some components to use the new `$info` colour palette where blue as a semantic "information" colour (alongside orange, green, red) is necessary
  - `AlertDialog` (icon in header)
  - `Badge`
  - `InlineMessage`
  - `SectionMessage`
  - `Toast`
- Simplify `Button` & `ActionIcon` variant implementation
- Update documentation to use the correct `$primary` scale values, retain `blue1`/`blue2` if used alongside other colorscheme palettes


MiguelGDLR:
I think in the case of button it should stay as primary, not info. They are 2 different concepts

thomasdigby:
> I think in the case of button it should stay as primary, not info. They are 2 different concepts

Sorry if I wasn't clear, I meant to add an _additional_ theme of `info`. Alongside `primary`. That would mean you could display a button with the theme of `info`, `success`, `warning` or `error`, regardless of the brand,

thomasdigby:
> > > I think in the case of button it should stay as primary, not info. They are 2 different concepts
> > 
> > 
> > Sorry if I wasn't clear, I meant to add an _additional_ theme of `info`. Alongside `primary`. That would mean you could display a button with the theme of `info`, `success`, `warning` or `error`, regardless of the brand,
> 
> oh ok, We would VERY rarely use it, I think. I haven't seen other DSs using that either. I can ask around and try to find use cases. In the future we might also consider removing 'success' as is not properly used and adds complexity.

Ok no problem then, thought it might be useful but if not then no worries. I'll leave it as is 🙂

lib/src/components/action-icon/ActionIcon.tsx
```diff
@@ -88,104 +104,26 @@ const StyledButton = styled('button', {
     }
   },
 
+  // prettier-ignore
```
andoulla, line 107: just curious with what's wrong with prettier here, it looks fine to me

andoulla (approved):

MiguelGDLR (approved):

lib/src/components/action-icon/ActionIcon.tsx
andoulla: yes I was wondering which rule it was trying to pause, good to know it was the single line bit

PR #611, opened 2024-01-10
fix: need to omit defaultChecked from Box so HTMLElement boolean type…

LimeWub:
… is not picked up

I love typescript and nothing suspicious is happening with types at all. Anyway this fixes this wrong type:

testing via:
`type Test = CheckboxGroupRootProps['defaultChecked']`

**before:**
Screenshot 2024-01-10 at 13 40 03

**after:**
Screenshot 2024-01-10 at 13 39 43


thomasdigby (approved):
Weird

PR #624, opened 2024-02-21
fix: remove usage of React.FC type

Mhoog:
To prevent conflicts when we upgrade the main repo to React 18, where `React.FC` does not include the `children` prop anymore.

This PR removes usage of the type and instead types the props directly (with `React.PropsWithChildren` where needed).

### DIFF
--- a/lib/src/components/icon/Icon.tsx
+++ b/lib/src/components/icon/Icon.tsx
@@ -28,7 +28,7 @@ type IconProps = Override<
   }
 >
 
-export const Icon: React.FC<IconProps> = React.forwardRef(
+export const Icon = React.forwardRef<SVGSVGElement, IconProps>(
   ({ is: SVG, size = 'md', ...remainingProps }, ref) => (
     <StyledIcon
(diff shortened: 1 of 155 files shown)


lib/src/components/icon/Icon.tsx
```diff
@@ -28,7 +28,7 @@ type IconProps = Override<
   }
 >
 
-export const Icon: React.FC<IconProps> = React.forwardRef(
+export const Icon = React.forwardRef<SVGSVGElement, IconProps>(
```
thomasdigby, line 31: Is this a typo or is `SVGSVGElement` actually valid?

Mhoog: https://developer.mozilla.org/en-US/docs/Web/API/SVGSVGElement 🤷

thomasdigby: Weird

thomasdigby (approved):
👍 😬 👍

PR #634, opened 2024-02-29
fix: added disabled prop to type to allow disabling Link

chris-atomlearning:
The Link component wasn't allowing a disabled prop to be added as it isn't included in `React.ComponentProps<typeof 'a'>`
This has lead to the styling being overwritten in the apps.

The styling was already there so it just needed to allow the disabled prop explicitly.


lib/src/components/link/Link.tsx
```diff
@@ -48,15 +48,16 @@ type LinkProps = Override<
   React.ComponentProps<typeof StyledLink>,
   {
     as?: React.ComponentType | React.ElementType
-  } & NavigatorActions
+  } & NavigatorActions & { disabled?: boolean }
```
LimeWub, line 51: that's really weird that this is needed - I'd expect `StyledLink` to be covering the disabled prop as it's based on an anchor :o ...

lib/src/components/link/Link.tsx
```diff
@@ -48,15 +48,16 @@ type LinkProps = Override<
...
 >
 
 export const Link: React.ForwardRefExoticComponent<LinkProps> =
-  React.forwardRef(({ as, href, ...rest }, ref) => (
+  React.forwardRef(({ as, disabled, href, ...rest }, ref) => (
     <StyledLink
       as={as || (!href ? 'button' : undefined)}
       noCapsize={!href ? true : undefined}
       href={href}
+      {...(disabled && { disabled: true })}
```
LimeWub, line 60: why do you need to do this? It's part of `rest` so you shouldn't need to be adding this property optionally.

chris-atomlearning: I guess we don't need to really. I saw it as a custom property so wanted to give it more visibility rather than a generic prop that we're passing through to the link.

lib/src/components/link/Link.tsx
LimeWub: I hadn't realised that - I thought it was a legit anchor attribute! Fair dos then!

LimeWub (approved):

thomasdigby:
@chris-atomlearning Is there anything blocking this PR?

PR #638, opened 2024-03-07
B2B-651 - Custom popover view for pagination

chris-atomlearning:
This ticket is part of this B2B ticket:
https://atomlearningltd.atlassian.net/browse/B2B-651
AS A Quest student
I WANT to find the navigation easy and intuitive, even if I am a 5-year-old
SO THAT I can do my best in the assessment without being confused or distracted

The pagination popover is exactly what we needed to display, just with a different trigger, and no horizontal numbered list of pages.

I considered pulling the <PaginationPopover> component out so it could be added directly, but it would need the Provider and ColorScheme around it, and that quickly seemed impractical.

So I have added a couple of new props, to turn the "popover only" view on, and specify what the trigger looks like. 
I'm open to suggestions on naming and other alternatives but this seems like a really neat way to achieve what we want.

image

### DIFF
--- a/lib/src/components/pagination/Pagination.test.tsx
+++ b/lib/src/components/pagination/Pagination.test.tsx
@@ -12,6 +12,8 @@ import { Pagination } from './index'
 import { Flex } from '../flex'
 import { Text } from '../text'
 import userEvent from '@testing-library/user-event'
+import { Icon } from '../icon'
+import { ChevronDown } from '@atom-learning/icons'
 
 const pageChangeSpy = jest.fn()
 
@@ -226,6 +228,7 @@ describe('Pagination component', () => {
         <Pagination.Popover>
           <Flex gap={1} align="center" css={{ cursor: 'pointer' }}>
             <Text>{`Question 1 / 10`}</Text>
+            <Icon is={ChevronDown} size="sm" />
           </Flex>
         </Pagination.Popover>
       </Pagination>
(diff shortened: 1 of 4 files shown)


lib/src/components/pagination/Pagination.tsx
```diff
@@ -17,11 +18,13 @@ export const Pagination = ({
   disabledPages = [],
   onItemHover = () => null,
   labels = {},
+  showPopoverTrigger = false,
+  popoverTriggerLabel = null,
```
chris-atomlearning, line 22: Do these props make sense? Is there a neater way?

chris-atomlearning: Good shout, I like that, works really nice :)

lib/src/components/pagination/Pagination.test.tsx
```diff
@@ -216,4 +219,25 @@ describe('Pagination component', () => {
       await waitFor(() => axe(container), { timeout: 5000 })
     ).toHaveNoViolations()
   })
+
+  it('renders with custom popover trigger if turned on', async () => {
+    await render(
+      <Pagination pagesCount={10} selectedPage={1}>
+        <Pagination.Popover>
+          <Flex gap={1} align="center" css={{ cursor: 'pointer' }}>
```
LimeWub, line 227: Is the Flex here needed for `asChild` to work?

LimeWub (approved):

lib/src/components/pagination/Pagination.test.tsx
chris-atomlearning: Ah no, in the sandbox version I had an icon as well and I just copied it without the icon for some reason. I'll add the icon back in so it's the same as how we're going to use it.

PR #639, opened 2024-03-14
B2B-651 - Indicated page style on popover

chris-atomlearning:
We need the ability to define custom indicated styles for the indicated pages in the popover, primarily a solid coloured background, but we want to keep the circle indicator as a default for the normal list of pages.

I added a prop to handle the extra css to the page and the popover.

image

image

### DIFF
--- a/lib/src/components/pagination/PaginationPage.tsx
+++ b/lib/src/components/pagination/PaginationPage.tsx
@@ -96,7 +96,11 @@ const StyledButton = styled('button', {
   }
 })
 
-export const PaginationPage = ({ pageNumber, css }: PaginationPageProps) => {
+export const PaginationPage = ({
+  pageNumber,
+  css,
+  indicatedCSS
+}: PaginationPageProps) => {
   const { currentPage, goToPage, indicatedPages, disabledPages, onItemHover } =
     usePagination()
 
@@ -115,7 +119,7 @@ export const PaginationPage = ({ pageNumber, css }: PaginationPageProps) => {
       selected={isSelected}
       size="md"
       onClick={() => goToPage(pageNumber)}
-      css={css}
+      css={{ ...css, ...(isIndicated ? { ...indicatedCSS } : {}) }}
       indicated={isIndicated}
       disabled={isDisabled}
       aria-current={isSelected && 'page'}
--- a/lib/src/components/pagination/PaginationPopover.tsx
+++ b/lib/src/components/pagination/PaginationPopover.tsx
@@ -4,10 +4,19 @@ import React from 'react'
 import { ActionIcon, Flex, Icon, Popover } from '..'
 import { PaginationPage } from './PaginationPage'
 import { usePagination } from './usePagination'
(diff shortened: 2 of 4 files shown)


LimeWub:
Why is this change needed? 
It feels like we are adding complexity to the DS component for a single specific use case. 
Why is the indicator being the dot a problem?
How does this affect interaction styles which use background colours? 
Has it been reviewed by design?

lib/src/components/pagination/PaginationPopover.tsx
```diff
@@ -4,10 +4,19 @@ import React from 'react'
...
+
+interface PaginationPopoverProps {
+  pageCSS?: CSS
+  indicatedPageCSS?: CSS
+  children: React.ReactNode
+}
 
 export const PaginationPopover = ({
+  pageCSS,
+  indicatedPageCSS,
```
LimeWub, line 17: The Popover does not need to know about this property as it can be added to `PaginationPage` directly.

lib/src/components/pagination/PaginationPage.tsx
```diff
@@ -115,7 +119,7 @@ export const PaginationPage = ({ pageNumber, css }: PaginationPageProps) => {
       selected={isSelected}
       size="md"
       onClick={() => goToPage(pageNumber)}
-      css={css}
+      css={{ ...css, ...(isIndicated ? { ...indicatedCSS } : {}) }}
```
LimeWub, line 122: You shouldn't even need to add this here - just override the CSS as you need in your use case (?) It's already accepting the css prop.

chris-atomlearning:
> Why is this change needed? It feels like we are adding complexity to the DS component for a single specific use case. Why is the indicator being the dot a problem? How does this affect interaction styles which use background colours? Has it been reviewed by design?

These are all great questions 😅 It does feel like a bit of an over complication but I couldn't see another way...
Basically I'm trying to achieve this design for the popover, and thought this approach would solve the problem while keeping the dots styling as a default: 
image

It would be great if we could just use this filled in style everywhere instead of the dots, which would mean that we don't need to make these changes to have it customisable. But I think we need the dots to match the ISEB design 🫤 Miguel is going to check that.

Do you think it would be better to force the filled in style only when a custom popover is used? Would that be more confusing?

lib/src/components/pagination/PaginationPopover.tsx
chris-atomlearning: Yeah I think this is where it's getting a little unclear. I needed to specify it in the popover because I'm not using PaginationPage directly in the app.

image

LimeWub: oh sorry I missed that... having a think... (I have refinement now but might be worth a meeting?)

...
Maybe something about passing theme for the dot version or a filled out theme for the quest version would be a shout? Kind of like we do with the button with solid/outline themes? That can go in the context and pulled through to PaginationPage (both for the popover ones and the regular ones if u ever go back to that)

PR #642, opened 2024-03-22
devops - actions to use Atom CI Actions app for privileged operations

euanhunteratom:


LimeWub (approved):
Thanks for this!

Does this also need to happen in the other workflows which didn't break?
ie. `sync-from-main-to-documentation`?

Does anything change for how these are ran? (on merge and on push respectively) - there's nothing manual that will need doing - correct?

euanhunteratom:
> Thanks for this!
> 
> Does this also need to happen in the other workflows which didn't break? ie. `sync-from-main-to-documentation`?
> 
> Does anything change for how these are ran? (on merge and on push respectively) - there's nothing manual that will need doing - correct?

Yeah, should be no changed needed for the other workflows, just the ones that used `BOT_ACCESS_TOKEN`. And yeah, should be no changes to the UX

PR #649, opened 2024-04-05
Feat/new pagination design for data table

Jdrazong:
-Updated data table pagination to use current DS pagination
-Updated test cases and removed obsolte ones
-Indexing is different for the context of data table vs DS pagination- in change handler and initial index it had to be incremented.

Zrzut ekranu 2024-04-5 o 11 58 13
Zrzut ekranu 2024-04-5 o 11 58 08


avirati (approved):

lib/src/components/data-table/pagination/Pagination.tsx
```diff
@@ -2,32 +2,38 @@ import * as React from 'react'
...
+  gap: '$4',
+  mt: '$4',
+  '@md': {
+    flexDirection: 'row'
+  }
 })
 
-type PaginationProps = React.ComponentProps<typeof StyledNav>
+type PaginationProps = React.ComponentProps<typeof StyledNav> & {
+  colorScheme: TcolorScheme
```
andoulla, line 27: shouldn't this be optional to avoid TS errors on core?

andoulla (approved):

PR #657, opened 2024-05-08
chore: updated theme version to include new colours

chris-atomlearning:
We've updated the colours to include new palettes for Marsh, Maroon and Lapis, plus updated colours for Quest.

This just bumps the theme version from 3.0.1 to 3.1.0 and updates the snapshots


thomasdigby (approved):

thomasdigby:
@chris-atomlearning This won't automatically release because you aren't using `feat:` or `fix` in your commit. `chore` skips the release. However I have a PR that I want to merge right after this to bump the icons lib so that should be fine

PR #670, opened 2024-07-15
Aq 1405 create phone number input in design system

LimeWub:
## Description

Introduces new form elements `PhoneNumberInput` and `PhoneNumberField` to the design system.

These components will hopefully help solve ongoing issues with wrongly inputted phone numbers on the marketing site, where loose validation has been implemented but is still allowing for errors.

> We need to build a new input in the DS that uses libphonenumber to fully validate phone numbers using a double input field with country selection.

**Jira:** 

- https://atomlearningltd.atlassian.net/browse/AQ-1405  
- https://atomlearningltd.atlassian.net/browse/STRAP-29

## Implementation

* Split `Input` into `InputBackground` and `InputText`, moved to separate PR: https://github.com/Atom-Learning/components/pull/671
* Working from @miaambt 's [SPIKE outcome code](https://atomlearningltd.atlassian.net/browse/STRAP-29?focusedCommentId=27440), cleaned up and built `PhoneNumberInput` component. You may notice, this component has some seemingly hacky internal code; That was required, as while `react-phone-number-input` does quite a lot of the logic for us, it did not respect `defaultValue` and `value` with the same pattern as our other components did and also did not style `placeholder` correctly. I've gone through quite a bit of trial and error to iron out this functionality, which is why the components code might not look as straightforward. Have added relevant jest tests to cover the custom `value` and `defaultValue` manipulations.
* Added the `PhoneNumberField` component, a simple hook-in of the `PhoneNumberInput` to work with 'react-hook-form'
* Tweaked rollup to handle importing CSS from `react-phone-number-input` in `node-modules`
* Updated node-version for our Github actions to match the one in `atom-core`: This was also required to install `rollup-plugin-import-css` correctly
* Updated `size-limit` file path, which is now further nested as the `preserveModules` output of bundling now also has a `node_modules` folder since we're importing css from node_modules.

## Videos

**Primitive**

https://github.com/user-attachments/assets/8842047c-bdaa-4c96-98f2-7c7f4b9002a5

**Form Field**
(Note: Every time a console log appears in the video below is because I have pressed `Enter` thus submitting the demo form)

https://github.com/user-attachments/assets/7662559e-d3b6-48ed-9873-93412a24738a


lib/src/components/phone-number-input/PhoneNumberInput.tsx
```diff
@@ -0,0 +1,121 @@
...
+        if (typeof valueProps === 'undefined') return
+        setValue(parsedValue?.number || valueProps || '')
+      }, [valueProps, parsedValue?.number])
+
+      React.useEffect(() => {
+        const phoneInputWithCountrySelectStyleTagId =
+          'phone-number-input-country-select-styles'
+        if (document.getElementById(phoneInputWithCountrySelectStyleTagId))
+          return // styles' already injected
+        document.head.insertAdjacentHTML(
```
thomasdigby, line 87: I'd really like to avoid applying styles after the initial render, we don't want a flash of unstyled content. If there's no way to import the 3rd party styles without that, then we should just keep it simple and copy the styles into a `styled()` component.

lib/package.json
```diff
@@ -43,17 +43,17 @@
   "size-limit": [
     {
       "limit": "10kb",
-      "path": "dist/index.js",
+      "path": "dist/lib/src/index.js",
```
thomasdigby, line 46: I'd like to avoid changing the output structure if we don't get a big benefit from the change. There are multiple imports in core that rely on this structure, as well as the main/module/typings exports above.

As we're only doing this to support a sub-optimal application of styles anyway, I just think it's worth keeping it simple, reverting these changes, and just copy the CSS.

lib/package.json
```diff
@@ -167,6 +168,7 @@
     "react-hook-form": "^6.15.4",
     "react-hook-form-persist": "^2.0.0",
     "react-hot-toast": "^1.0.2",
+    "react-phone-number-input": "^3.4.3",
```
thomasdigby, line 171: Are we doing anything with this package to avoid importing 190kb of JS? https://bundlephobia.com/package/react-phone-number-input@3.4.3

lib/src/components/phone-number-input/PhoneNumberInput.tsx
LimeWub: There is no flash of unstyled content - it happens so fast that it doesn't flash at all. I tested it with slow network.

lib/package.json
thomasdigby: Does it tree-shake? Are we importing the entire 190kb? Or only a sub-set of it? Will it be included with every bundle if the component is used anywhere in the tree?

lib/package.json
LimeWub: Doesn't core use the bundled index file the same way the documentation site does? The documentation site is not affected so core shouldn't be either(?)
Which imports are you concerned about?

In my opinion it is much better to import the adjoining css from `node_modules` as, if we update the component we definitely have the latest CSS there instead of relying on individual's knowledge to remember to fetch and manually replace CSS in the component.

lib/src/components/phone-number-input/PhoneNumberInput.tsx
thomasdigby: I think we should set a rule to never do this, even if it is super fast. Styles have to be available for an element before we render it into the DOM. Otherwise there's always a chance that certain network or browser conditions mean we're waiting on styles to be applied.

lib/package.json
thomasdigby: Ah yeah you're right, I'm thinking of `@atom-learning/theme`.

Either way, I just don't think it's worth it? The last updates to non-internal packages in the components lib were 2 months ago with Radix, and before then, 8 months ago. And more importantly, the last time they updated the styles in `react-phone-number-input` was almost 2 years ago, and that was a minor addition to support read-only inputs, the update before then was 4 years ago when the package was created.

It's just not worth all of these changes

lib/package.json
thomasdigby: Do we need all of the supported phone number formats in `libphonenumber-js`? We're surely only going to need a small subset of what they support. Could we simplify and use a "international prefix" dropdown, and then a number field? That would greatly reduce the amount of JS we need to support this.

Also, tangentially, why do we need to validate phone numbers? Much like with emails, do we not rely on the user typing in the correct data?

lib/src/components/phone-number-input/PhoneNumberInput.tsx
thomasdigby: Honestly, I would just copy the styles, there's very, very little to gain from trying to import from the 3rd party. It would just work.

LimeWub:
(added to marketing repo for now - closing this PR)

PR #671, opened 2024-07-22
feat: split and export InputBackground and InputText to allow for separate use

LimeWub:
_Required for: https://github.com/Atom-Learning/components/pull/670 
... was originally part of the above PR but splitting out so both the PRs are easier to review._

It is becoming more and more a need that we split out the `InputBackground` from the actual `<input />` element: `InputText`. There are multiple components built or planned to be built, which nest elements other than or alongside the usual text inside the input.

The card input/combobox is nesting a card icon next to the selected option:

Screenshot 2024-07-22 at 13 03 30

The planned/WIP multi-select is nesting `ChipDismissible` and a `Dropdown/SelectTrigger`:

Screenshot 2024-07-22 at 13 04 38

The [currently in PR](https://github.com/Atom-Learning/components/pull/670) `PhoneNumberInput` is nesting a full country `Select` alongside the text:

Screenshot 2024-07-22 at 13 05 10

Some of these designs can be achieved by positioning the adjacent elements absolutely, but this is a more hacky solution as we are using magic numbers for the spacing and placement of the text and as it limits the amount of elements nested and their placement.

It is therefore cleaner and more reusable to split the `InputBackground` as a container from the `InputText` (the actual `<input />`) and allow them to be used separately for those use-cases, rather than relying on hackier positioning - which will not support all our design use-cases anyway.

This change is not expect to affect anything visually, but will require quite a few snapshot changes in core, as the Input component is widely used.


lib/src/components/input/Input.tsx
```diff
@@ -2,88 +2,129 @@ import * as React from 'react'
...
+import { overrideStitchesVariantValue } from '~/utilities/override-stitches-variant-value/overrideStitchesVariantValue'
 import { Override } from '~/utilities/types'
 
-const StyledInput = styled('input', {
-  appearance: 'none',
-  border: '1px solid $tonal400',
+import { Flex } from '../flex'
+import { Text } from '../text'
+
+export const InputBackground = styled(Flex, {
```
Mhoog, line 11: Could we make them subcomponents of `Input`? `Input.Background` and `Input.Text`?

LimeWub: Yes we could! Before I do this though: 
The reason why I didn't make them subcomponents is because I **don't** expect this to be used like so 

```
<Input>
<Input.Background>
<AnotherComponent />
<Input.Text/>
</Input.Background>
</Input>
```

instead it'd be used directly like so:
```
<Input.Background>
<AnotherComponent />
<Input.Text/>
</Input.Background>
```

or even just 
```
<Input.Background>
<AnotherComponent />
</Input.Background>
```

So, in my opinion, this is not matching the subcomponent structure, as the top "level" of that component is just the composed `<Input />`. Rather it is just two elements which combined make an `<Input />`.

Do you still want these out as subcomponents? 

Would the implementation without a top level `<Input />` to nest them under not be confusing?  I could probably also export `<Input />` as a wrapper, but that wouldn't be the composed Input which is in this PR, and which maintains backwards compat. 
(We'd be exporting additionally a `<ComposedInput />` of sorts for all the actual uses of input currently in our codebase)

Mhoog: Yeah that's a fair point. Let's go with this :)

Mhoog (approved):

miaambt (approved):

PR #678, opened 2024-08-23
fix: add patch to fix memory usage issues with Stitches type

Mhoog:
Our machines were running out of memory when running `tsc`, turns out this was caused by this issue:
https://github.com/stitchesjs/stitches/issues/1038

This PR applies the patch described here:
https://github.com/stitchesjs/stitches/issues/1038#issuecomment-1516638929

Note that this patch will only be applied when installing packages at root level, we should upgrade `yarn` (or move to `pnpm`) so that we don't need `patch-package` and instead can just use the built-in patch functionality of `yarn` / `pnpm`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,11 +16,14 @@
     "dev:sandbox": "cd lib && yarn start:sandbox",
     "build:types": "cd lib && NODE_ENV=production tsc --emitDeclarationOnly && tsc-alias",
     "lint-staged": "cd lib && yarn run precommit",
-    "build-all": "run-s build:*"
+    "build-all": "run-s build:*",
+    "postinstall": "patch-package"
   },
   "devDependencies": {
     "husky": "^4.3.8",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -16,11 +16,14 @@
...
     "lint-staged": "cd lib && yarn run precommit",
-    "build-all": "run-s build:*"
+    "build-all": "run-s build:*",
+    "postinstall": "patch-package"
   },
   "devDependencies": {
     "husky": "^4.3.8",
-    "npm-run-all": "^4.1.5"
+    "npm-run-all": "^4.1.5",
+    "patch-package": "^8.0.0",
```
thomasdigby, line 25: Can't we use `yarn patch` instead?

Mhoog: Nope, we're still on yarn v1 😞

elliotlaws (approved):

thomasdigby (approved):

PR #689, opened 2024-08-28
fix: replace the @stitches/react patch with a forked version of the repo

Mhoog:
Having the Yarn patch version of `@stitches/react` caused issues with installing it in other repositories that did not use Yarn v4.

This replaces the patch with a forked version of the repo (including the patch). It also reverts back to using `@semantic-release/npm` instead of `semantic-release-yarn`, as we only needed that to handle the `patch:` protocol.


elliotlaws (approved):

lib/src/components/sortable/SortableItem.tsx
```diff
@@ -1,7 +1,7 @@
 import type { UniqueIdentifier } from '@dnd-kit/core'
 import { useSortable } from '@dnd-kit/sortable'
 import { Slot } from '@radix-ui/react-slot'
-import { CSS, styled } from '@stitches/react'
+import { CSS, styled } from '@atom-learning/stitches-react'
```
thomasdigby, line 4: These should come from `~/stitches` right?

thomasdigby: One for another PR

PR #697, opened 2024-08-28
fix: rename prepublishOnly to prepublish for yarn

Mhoog:


lib/package.json
```diff
@@ -210,7 +210,6 @@
       "@semantic-release/changelog",
       "semantic-release-yarn",
       [
-        "semantic-release-slack-bot",
```
thomasdigby, line 213: Do we need to remove this?

Mhoog: Nope, forgot to re-add after testing locally

PR #699, opened 2024-08-29
feat: upgrade NextJS, Typescript and React to the latest versions

Mhoog:


package.json
```diff
@@ -44,8 +44,8 @@
     }
   },
   "resolutions": {
-    "react": "17.0.2",
-    "react-dom": "17.0.2",
+    "react": "^18.3.1",
+    "react-dom": "^18.3.1",
```
thomasdigby, line 48: These need to be actual version numbers, not a range (I think)

PR #704, opened 2024-09-18
Feat/keyboard shortcut

LimeWub:
## Description

Adds the `KeyboardShortcut` utility component.

Basically a KeyboardEvent spy to be wrapped around an area or be used with `asChild` as an area and a pre-styled `<Text as="kbd" />` element to maintain consistency of look across uses.

This was developed as a need for the Question container regular and particularly the paper tests input view and @MiguelGDLR  has confirmed he'd like this functionality reusable to make use of in other areas of the site.

## Screenshots

Indicator use:
Screenshot 2024-09-18 at 09 10 33

General use:

https://github.com/user-attachments/assets/7c07bfff-5876-463e-965c-db248b22bf5f


MiguelGDLR (approved):

lib/src/components/keyboard-shortcut/KeyboardShortcut.tsx
```diff
@@ -0,0 +1,78 @@
...
+type KeyboardShortcutProps = React.ComponentProps<typeof Box> & {
+  asChild?: boolean
+  config: {
+    shortcut: Partial<React.SyntheticEvent<KeyboardEvent>>, action: ({ event, shortcut }: {
+      event: KeyboardEvent | React.KeyboardEvent<HTMLDivElement>,
+      shortcut: Partial<React.SyntheticEvent<KeyboardEvent>>
+    }) => void
+  }[],
+  targetWindow?: boolean
+  onKeyDown?: (e: KeyboardEventWindowOrElement) => void
```
Mhoog, line 23: Why do we need the additional `onKeyDown` prop? If you needed to execute something on the event, you could add it in the `action` function, right?

lib/src/components/keyboard-shortcut/KeyboardShortcut.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+      return () => {
+        window.removeEventListener('keydown', handleOnKeydown)
+      }
+
+    }, [targetWindow, handleOnKeydown])
+
+    const Component = asChild ? StyledSlot : Box
+
+    return (<Component onKeyDown={targetWindow ? () => null : handleOnKeydown} ref={setTargetElRef} {...rest} />)
```
Mhoog, line 49: Would this work?

```suggestion
    return (<Component onKeyDown={targetWindow ? undefined : handleOnKeydown} ref={setTargetElRef} {...rest} />)
```

lib/src/components/keyboard-shortcut/KeyboardShortcut.tsx
```diff
@@ -0,0 +1,78 @@
...
+  alignItems: 'center',
+  borderRadius: '$1',
+  flexShrink: 0
+})
+
+type KeyboardShortcutIndicatorProps = React.ComponentProps<typeof Text>
+
+export const KeyboardShortcutIndicator = (props: KeyboardShortcutIndicatorProps): JSX.Element => {
+  return (<StyledKeyboardShortcutIndicator size="sm" as="kbd" {...props} />)
+}
```
Mhoog, line 75: If you take the props from the styled version instead of the `Text` it will also include variants if we ever add any.

You could also omit the return type, it should infer it.

```suggestion
export const KeyboardShortcutIndicator = (
  props: React.ComponentProps<typeof StyledKeyboardShortcutIndicator>
) => <StyledKeyboardShortcutIndicator size="sm" as="kbd" {...props} />
```

lib/src/components/keyboard-shortcut/KeyboardShortcut.tsx
LimeWub: Ngl I added it in a bit automatically because I didn't want to completely "steal" the event.

The actions will run if the combination matches. Thinking about it, I suppose this would be useful if you wanted an action to happen regardless of what has been typed into the event. So not matching a shortcut.

Mhoog (approved):

Marcdup1 (approved):

PR #710, opened 2024-10-14
Feat: Allow CSS overrides for TopBar wrapper element

avirati:
https://atomlearningltd.atlassian.net/browse/HMP-1178

We encountered a bug last week where TopBar was not rendering well with other elements in the page
image

Upon investigation, we found out that overriding z-index of the root of TopBar component fixed the issue. However, it was done so by forcing the override from outside. This PR enables css overrides for root / wrapper element of TopBar. After this is released, we will remove the forced override and pass the required css as a prop instead.

More information - https://github.com/Atom-Learning/atom-core/pull/10161

### DIFF
--- a/lib/src/components/top-bar/TopBar.tsx
+++ b/lib/src/components/top-bar/TopBar.tsx
@@ -16,7 +16,6 @@ const TopBarDivider = () => (
 const StyledRoot = styled(Flex, {
   bg: '$background',
   position: 'sticky',
-  alignItems: 'center',
   width: '100vw',
   top: '0',
   zIndex: 1,
@@ -53,6 +52,7 @@ const TopBarComponent = ({
 
   return (
     <StyledRoot
+      align="center"
       className={className}
       hasScrolled={!!scrollPositionY}
       {...props}
--- a/lib/src/components/top-bar/TopBar.tsx
+++ b/lib/src/components/top-bar/TopBar.tsx
@@ -48,12 +48,14 @@ type StyledRootProps = React.ComponentProps<typeof StyledRoot>
 
 interface TopBarProps extends StyledRootProps {
   css?: CSS
+  wrapperCss?: CSS
   className?: string
 }
 
 const TopBarComponent = ({
   size = 'md',
   className = topBarColorSchemes['light'],
+  wrapperCss,
   ...props
 }: TopBarProps) => {
   const { y: scrollPositionY } = useWindowScrollPosition()
@@ -63,6 +65,7 @@ const TopBarComponent = ({
       className={className}
       hasScrolled={!!scrollPositionY}
       size={size}
+      css={wrapperCss}
     >
       <Container {...props} />
     </StyledRoot>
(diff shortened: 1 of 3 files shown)


andoulla (approved):

lib/src/components/top-bar/TopBar.tsx
```diff
@@ -48,12 +48,14 @@ type StyledRootProps = React.ComponentProps<typeof StyledRoot>
 
 interface TopBarProps extends StyledRootProps {
   css?: CSS
+  wrapperCss?: CSS
```
LimeWub, line 51: While this is fine, we don't usually do this sort or `thingCss` thing in the DS...

**First Idea:** 
What do you think about exporting `Container` out as `TopBar.Container` so it's available to use separately and can be properly targeted with CSS?

**Second Idea:**
I don't think we actually need the `Container` to exist at all...
I think if StyledRoot changes like so: 

```
const StyledRoot = styled(Flex, {
  bg: '$background',
  position: 'sticky',
  display: 'flex',
  alignItems: 'center',
  width: '100vw',
  top: '0',
  zIndex: 1,
  borderBottom: '1px solid $borderBottom',
  transition: 'box-shadow .2s ease-out',
  px: '$4',
  py: '$2',
  '@md': {
    px: '$5'
  },
  variants: {
    hasScrolled: {
      true: { boxShadow: '0px 4px 4px -2px rgba(31, 31, 31, 0.1);' }
    },
    size: {
      md: { height: '$6' },
      lg: { height: '$7' }
    }
  }
})
```

and we add an `align="center"` on `StyledRoot` use below, we basically get the exact same result (?)

@elliotlaws Was there a reason which I'm not picking up for the nested `Container` use here? (I think you originally wrote this component? A while back tbf). From the CSS only I don't see the need for it.
I think if it can be removed it'll be the cleanest solution (and we're saving one div).

avirati: I will look into it, thanks @LimeWub

avirati: I went with option 2. I tested the changes in atom-core by copying the component into shared and importing it in parent, student app. Seemed to work fine without any visible changes.

lib/src/components/top-bar/TopBar.tsx
```diff
@@ -62,10 +55,8 @@ const TopBarComponent = ({
     <StyledRoot
       className={className}
       hasScrolled={!!scrollPositionY}
```
LimeWub, line 57: To match behaviour 100% you'd want an `align="center"` here.

LimeWub (approved):
I think you're missing one line of code but pre-emptively approving :) ^^

lib/src/components/top-bar/TopBar.tsx
avirati: I did not add that as `StyledRoot` already has a css style for it. Do we still need `align="center"` on this ?
` alignItems: 'center',`

LimeWub: Oh I think it's fine then - in my code I had removed the css and replaced with a prop as using the props is 'preferred' when using a Flex, so i thought it was still missing ^^ 
It's not really a big deal imo, as it works the same, but if it's easy to change it it'd be closer to guidelines. 🤷

avirati: Understood, thanks !

PR #713, opened 2024-11-05
Feat/visual regression

avirati:
The following was done as part of this PR -

1. Setup Playwright
2. Add Visual Regression spec to crawl documentation site and take screenshot
3. Add Github workflow file for running visual regression tests on push to `main` and on PRs with label `visual-regression`
4. Add missing concurrency for `test` workflow


playwright.config.ts
```diff
@@ -0,0 +1,87 @@
...
+      name: 'firefox',
+      use: { ...devices['Desktop Firefox'] }
+    },
+
+    {
+      name: 'webkit',
+      use: { ...devices['Desktop Safari'] }
+    }
+
+    /* Test against mobile viewports. */
```
andoulla, line 60: Are these follow-ups?

andoulla (approved):

playwright.config.ts
avirati: No, these are additional configs, in case we want to test against different viewports / browsers

andoulla: ok, lets add a comment explaining this.

Also, question: for the branded versions is there a way to set the version it is running against, or is it because CI uses some images for setting up the environment to run against the ones we support we need to update the image

avirati: It will always be compared against `main`, unfortunately we can't set a custom branch / base for a specific build. The baseline will be updated automatically once `main` build is completed (i.e. post a new release)

playwright.config.ts
```diff
@@ -0,0 +1,87 @@
...
+      '@argos-ci/playwright/reporter',
+      {
+        uploadToArgos: Boolean(process.env.ARGOS_TOKEN),
+        token: Boolean(process.env.ARGOS_TOKEN)
+      }
+    ]
+  ],
+  /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
+  use: {
+    trace: 'off'
```
chmielsen, line 38: When a test fail, it should be 100% retryable locally, right? That's why `trace: 'off'` makes sense for us? Might be good to add comment on this

playwright.config.ts
```diff
@@ -0,0 +1,87 @@
...
+export default defineConfig({
+  testDir: './e2e',
+  /* Run tests in files in parallel */
+  fullyParallel: true,
+  /* Fail the build on CI if you accidentally left test.only in the source code. */
+  forbidOnly: !!process.env.CI,
+  /* Retry on CI only */
+  retries: process.env.CI ? 2 : 0,
+  /* Opt out of parallel tests on CI. */
+  workers: process.env.CI ? 1 : undefined,
```
chmielsen, line 23: Might be worth mentioning that we can only run one test in parallel because `ubuntu-latest` in private repositories has 2vCPUs:
https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for--private-repositories

and the proportion of workers is # vCPU / 2 (because one per browser and test runner)

chmielsen (approved):

.github/workflows/visual-regression.yml
```diff
@@ -1,21 +1,35 @@
...
   contents: read
   pull-requests: write
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
-  test:
+  debug-stuff:
```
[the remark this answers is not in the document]
avirati: Oops, this needs to be removed :)

playwright.config.ts
avirati: I set it to `off` as it was eating up space on CI. Maybe I can disable it for `CI` only ? Traces would be helpful when running these locally.

playwright.config.ts
avirati: Done 👍

playwright.config.ts
avirati: Updated to `trace: process.env.CI ? 'off' : 'on'`

PR #714, opened 2024-11-05
Feat: Add github workflow file for visual regression

avirati:


package.json
```diff
@@ -16,7 +16,8 @@
     "dev:sandbox": "cd lib && yarn start:sandbox",
     "build:types": "cd lib && NODE_ENV=production tsc --emitDeclarationOnly && tsc-alias",
     "lint-staged": "cd lib && yarn run precommit",
-    "build-all": "run-s build:*"
+    "build-all": "run-s build:*",
+    "visual-regression": "echo 'vr:chromium'"
```
avirati, line 20: Note to reviewers - this will be replaced with playwright command in PR https://github.com/Atom-Learning/components/pull/713

.github/workflows/visual-regression.yml
```diff
@@ -0,0 +1,52 @@
+name: Visual Regression
+
+on:
+  workflow_call:
+    inputs:
+      dsDocumentationUrl:
+        description: 'URL of environment to run VR tests against. Should include protocol (e.g. https://)'
+        required: false
+        default: ''
+        type: string
```
andoulla, line 10: sorry but my yaml is week , does the section below mean that it will run on every PR?

avirati: No, this is triggered by workflow calls. In other words, it is triggered by manual invocation, either by going to https://github.com/Atom-Learning/components/actions or by calling it from another workflow file.

andoulla (approved):

package.json
avirati: I plan to do the following - 

1. PR level builds - spin up a local server, run VR tests against that. No cloudflare access token should be required. But let me know if that is not the case
2. Baseline Builds triggered on push to main branch, after atomlearning.design deployment completes - Target https://atomlearning.design and run VR tests on it. Cloudflare access tokens should not be required as it is a public url.

euanhunteratom: Ah, that makes sense. No, that shouldn't need a CF token then

.github/workflows/visual-regression.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Build Library
+        if: ${{ inputs.dsDocumentationUrl != '' }}
+        run: yarn build:lib
+
+      - name: Build Documentation
+        if: ${{ inputs.dsDocumentationUrl != '' }}
+        run: yarn build:docs
+
+      - name: Start Web Server as a Daemon
+        if: ${{ inputs.dsDocumentationUrl != '' }}
```
euanhunteratom, line 42: Should these ifs instead be `==` so these run if the URL _isn't_ specified?

.github/workflows/visual-regression.yml
```diff
@@ -0,0 +1,52 @@
...
+        run: yarn start:docs &
```
euanhunteratom, line 43: This might need the `disown` command after this to make sure the process stays running after the step finishes.

euanhunteratom (approved):

.github/workflows/visual-regression.yml
avirati: If we provide `dsDocumentationUrl`, we need to run VR tests on that URL only. In case we don't provide it, we would build the docs locally and run VR tests on `localhost`.

.github/workflows/visual-regression.yml
avirati:
```
run: yarn start:docs &
     disown -h %1
```

@euanhunteratom will this do ☝️ ?

